PR #4, opened 2017-05-30
Add functional test for meshcop

bukepo:
This PR,
* Adds functional test for meshcop(involves OpenThread, wpantund)
* Formalize the state machine of DTLS session.
* Fix timeout of DTLS.
* Support piggy-back response of CoAP API.

### Details

This test is based on the POSIX simulation of OpenThread and wpantund. The scenario is simple. Firstly, create an one node Thread network, thus the only node acts as Thread leader as well as Thread border router. Secondly, authenticate an external commissioner. At last, authenticate the joiner.

#### tests/meshcop/commissioner.cpp
Implements a simple commissioner app with fixed PSKd `123456`, steering data(for joiner whose EUI64 is 18b4300000000002) and commissioner id `OpenThread`. It connects to local border agent and petition to be the commissioner. Once success, it will authenticate the joiner.

#### OpenThread
Latest OpenThread is built to provide simulated Thread devices. One is in NCP mode and controlled by wpantund and act as the border router, the other is in CLI mode and act as the joiner whose EUI64 is 18b4300000000002.

#### wpantund
Latest wpantund is built and installed into `$(top_src)/_stage` to act as the gateway of thread network. Since d-bus policy file must be present in /etc/dbus-1/system.d, it will be deployed during the test and deleted after the test.

This tests the overall inter-operation of OpenThread, wpantund and Border Router. And it should be improved continuously.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ matrix:
   include:
     - env: BUILD_TARGET="pretty-check"
       os: linux
-    - env: BUILD_TARGET="posix-check"
+    - env: BUILD_TARGET="posix-check" VERBOSE=1
       os: linux
       compiler: gcc
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -42,17 +42,14 @@ linux)
     # This is for configuring wpantund
     sudo apt-get install autoconf-archive
 
-    # mDNS
-    sudo apt-get install libavahi-core-dev
-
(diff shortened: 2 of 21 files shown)


bukepo:
@jwhui any comments?

jwhui:
@bukepo, this is a great start to functional testing!  Thanks for this.

PR #7, opened 2017-05-31
Scripts to rule them all

bukepo:
This PR adds several scripts according to [Scripts to rule them all](https://github.com/github/scripts-to-rule-them-all)

- *bootstrap* install dependencies.
- *setup* build and install border router, wpantund.
- *update* rebuild border router and update wpantund.
- *server* start wpantund and border router.
- *console* start border router and wpantund.
- *cibuild* used only on travis.
- *test* run test locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,41 @@ OTBR also provides the following services:
 
 Third-party components for Border Router Services include Simple Web Server and Material Design Lite for the framework of the web UI.
 
+# Get started
+
+We follow philosophy of [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all). Validated platforms include:
+
+* Raspberry Pi 3B (running recent [Raspbian Jessie Lite](https://www.raspberrypi.org/downloads/raspbian/))
+
+## Build and Install
+
+```sh
+# Install dependencies
+./script/bootstrap
+
+# Build and install border router and wpantund
+./script/setup
+```
+
+## Configure
+
+Edit `/etc/wpantund.conf` according to the comments. Basically only path to NCP's serial port is required. Plug in the NCP mode OpenThread
+device and check the corresponding device file of it. Here's an example when the device's serial port is `/dev/ttyUSB0`.
+
+```
+Config:NCP:SocketPath "/dev/ttyUSB0"
+```
+
+## Start services
+
+All border router services will be automatically started on next boot. Or you can start immediately by:
+
+```sh
+./script/server
+```
+
+Now you can access border router's Web UI through browsers and explore its features.
+
 # Need help?
 
 ## Interact
--- a/script/_ipforward
+++ b/script/_ipforward
@@ -1,0 +1,52 @@
+#!/bin/sh
+#
+#  Copyright (c) 2017, The OpenThread Authors.
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#  3. Neither the name of the copyright holder nor the
+#     names of its contributors may be used to endorse or promote products
+#     derived from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+#  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+#  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+#  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+#   Description:
+#       This script manipulates ip forward configuration.
+#
+
+SYSCTL_IP_FORWARD=/etc/sysctl.d/60-otbr-ip-forward.conf
+
+ipforward_install()
+{
+    sudo tee $SYSCTL_IP_FORWARD <<EOF
+net.ipv6.conf.all.forwarding = 1
+net.ipv4.ip_forward = 1
+EOF
+}
+
+ipforward_uninstall()
+{
+    test ! -f $SYSCTL_IP_FORWARD || sudo rm $SYSCTL_IP_FORWARD
+}
+
+ipforward_enable()
+{
+    echo 1 | sudo tee /proc/sys/net/ipv6/conf/all/forwarding
+    echo 1 | sudo tee /proc/sys/net/ipv4/ip_forward
+}
--- a/script/_otbr
+++ b/script/_otbr
@@ -1,0 +1,70 @@
+#!/bin/sh
+#
+#  Copyright (c) 2017, The OpenThread Authors.
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#  3. Neither the name of the copyright holder nor the
+#     names of its contributors may be used to endorse or promote products
+#     derived from this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+#  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+#  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+#  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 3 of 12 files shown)


script/_wpantund
```diff
@@ -0,0 +1,67 @@
...
+    test -d $BUILD_DIR/wpantund || return 0
+
+    (cd $BUILD_DIR/wpantund &&
+        (test ! -f config.status || sudo make uninstall))
+}
+
+wpantund_install()
+{
+    test -d $BUILD_DIR/wpantund || (cd $BUILD_DIR &&
+        git clone --depth 1 https://github.com/openthread/wpantund.git)
```
beriberikix: Assumes git is installed on the machine - I think we should apt-get install git in `bootstrap`

script/bootstrap
```diff
@@ -0,0 +1,140 @@
...
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+. "$(dirname "$0")"/_initrc
+
+install_packages_apt()
+{
+    sudo apt-get install -y build-essential
+    # For wpantund
```
beriberikix: apt-get install git here?

script/bootstrap
```diff
@@ -0,0 +1,140 @@
...
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+. "$(dirname "$0")"/_initrc
+
+install_packages_apt()
+{
```
beriberikix: sudo apt-get update?

script/bootstrap
```diff
@@ -0,0 +1,140 @@
...
+
+    # For libcoap to generate .sym and .map
+    sudo apt-get install -y ctags
+
+    # Boost
+    sudo apt-get install -y libboost-dev libboost-filesystem-dev libboost-system-dev
+
+    # npm bower
+    if ! which bower; then
+        curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
```
beriberikix: I don't think this install NPM. I recieved 'sudo: npm: command not found`

beriberikix: Also curl wasn't installed on my VM. apt-get install curl

script/_wpantund
```diff
@@ -0,0 +1,67 @@
...
+
+    (cd $BUILD_DIR/wpantund &&
+        (test -f configure || ./bootstrap.sh) &&
```
beriberikix: After running `./setup`, I received the output below. Maybe a missing dep or config issue?:

vagrant@precise64:~/borderrouter/script$ ./setup     
otbr-web: no process found
otbr-agent: no process found
wpantund: no process found
Cloning into 'wpantund'...
remote: Counting objects: 1759, done.
remote: Compressing objects: 100% (1065/1065), done.
remote: Total 1759 (delta 757), reused 1536 (delta 630), pack-reused 0
Receiving objects: 100% (1759/1759), 2.15 MiB | 912 KiB/s, done.
Resolving deltas: 100% (757/757), done.
 ***************************** 
./bootstrap.sh: 33: ./bootstrap.sh: [[: not found

./bootstrap.sh: 36: ./bootstrap.sh: [[: not found
 *** autoreconf failed with error code 1

beriberikix: Please review the deps @ https://github.com/openthread/wpantund/blob/master/INSTALL.md#1-install-dependencies

script/_wpantund
jwhui: Just so everyone is aware, this is a proposal to adopt [GitHub's Scripts to Rule Them All](https://githubengineering.com/scripts-to-rule-them-all/) convention.  The [GitHub boilerplate scripts](https://github.com/github/scripts-to-rule-them-all) do not include installing git itself.  I have not dug in to see if that was a conscious decision by GitHub.

beriberikix: That also works, most users will have git but something to add for completeness. Alternatively we can call out requirements in the README or guide that walks the user through the setup process.

script/_wpantund
bukepo: I think your environment lack bash, which is not listed as dependencies in the link you provided either. Could you please share with me the distribution of your VM？

beriberikix: I'm using the default image for Vagrant of [precise64](https://atlas.hashicorp.com/ubuntu/boxes/precise64), which is Ubuntu Server 12.04 LTS (Precise Pangolin.)

beriberikix:
Update: I think the original `wpantund` wiki doc is out of date. I used the deps from the Vagrant file I created found [here](https://github.com/openthread/openthread/blob/master/etc/vagrant/Vagrantfile#L54).

After I apt-get'd NPM and cURL, I was able to *almost* run setup. Last line output is `sudo: systemctl: command not found.` Searching a bit seems to be related to systemd (see comment above.)

script/_otbr
```diff
@@ -0,0 +1,56 @@
...
+
+    test ! -f config.status || sudo make uninstall
+}
+
+otbr_install()
+{
+    test -f config.status || ./configure --prefix=/usr --sysconfdir=/etc
+    make -j$(getconf _NPROCESSORS_ONLN)
+    sudo make install
+    sudo systemctl enable otbr-web otbr-agent
```
beriberikix: Handle when systemctl is not installed or disabled by [default](https://wiki.ubuntu.com/SystemdForUpstartUsers).

bukepo: These scripts are tested on raspberry pi 2 and ubuntu 16.04.2 LTS. And I also planned to add sysV support when systemd is not available, but I don't get that much time to get it done for the moment. Do you think we should support sysV in this PR or later?

script/_wpantund
bukepo: OK, I will test in this environment. Thanks.

lanyuwen:
I don't think it's a high priority to support various of platforms/distributions at this time. Trying to accommodate a wide range of different Linux distributions will unavoidably slow down the speed to provide a comprehensive support on our target platforms such as Raspberry Pi.

Can we start from focusing on a single platform for now, and then expand to others after that? By delivering clear scripts to build a workable OTBR on a specific platform, we would also enable people to support more platforms/distributions as their wish.

beriberikix:
Agreed, focus on the pi and maybe the latest Ubuntu (for CI.) The one thing
I'd like to consider is to throw a more helpful error, like "only platforms
with systemd are currently supported," here and elsewhere the focused
platform support becomes relevant.

script/_wpantund
bukepo: It seems the VM's default shell is dash.
```sh
ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Jan 21  2014 /bin/sh -> dash
```
However, wpantund's bootstrap.sh actually requires `bash`. I will submit a PR to fix this issue in wpantund.

script/bootstrap
bukepo: Thanks, added.

script/bootstrap
bukepo: Thanks, added curl. And downgraded to nodejs 6.x, which is the stable version and is more widely supported.

script/_otbr
bukepo: Added error information.

bukepo:
I've updated and adjusted, Thank you all for your suggestions.

beriberikix:
Tested on the latest nightly of Ubuntu 16.04.2 LTS and I think I got as far as I can test with a VM! `./console` fails because I don't have a USB device (wpantund[4381]: Unable to open socket. "No such file or directory" (2) (filename = /dev/ttyUSB0, type = 6)
) - so I feel comfortable with the current state. One improvement for a future release would be to better handle USB addresses. For example, if another device is USB0 this would also fail.

With this upgrade do we still need `bootstrap` in the root directory?

README.md
```diff
@@ -42,6 +42,38 @@ OTBR also provides the following services:
 
 Third-party components for Border Router Services include Simple Web Server and Material Design Lite for the framework of the web UI.
 
+# Get started
+
+We follow philosophy of [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all). Scripts are tested under,
+
```
lanyuwen: Suggestion: "Scripts are tested under," -> "Validated platforms include:"

README.md
```diff
@@ -42,6 +42,38 @@ OTBR also provides the following services:
...
+* Raspberry PI2
```
lanyuwen: this should be "Raspberry Pi 3B (running [Raspbian Jessie Lite)](https://www.raspberrypi.org/downloads/raspbian/)"

README.md
```diff
@@ -42,6 +42,38 @@ OTBR also provides the following services:
...
+./script/setup
+```
+
+## Configure
+
+Edit `/etc/wpantund.conf` according to the comments. Basically only path to serial port is required. Below is the example when NCP is recognized as `/dev/ttyUSB0`.
+
+```
+Config:NCP:SocketPath "/dev/ttyUSB0"
+```
```
lanyuwen: remind readers to get NCP device ready here before start services?

README.md
```diff
@@ -42,6 +42,38 @@ OTBR also provides the following services:
...
+
+## Start services
+
+```sh
+./script/server
+```
+
+Now you may configure the NCP throught web gui port 80.
```
lanyuwen: Suggestion: "Now you can access border router's Web UI through browsers to explore its features."

script/_otbr
```diff
@@ -0,0 +1,70 @@
...
+        sudo systemctl daemon-reload
+    fi
+}
+
+otbr_install()
+{
+    test -f config.status || ./configure --prefix=/usr --sysconfdir=/etc
+    make -j$(getconf _NPROCESSORS_ONLN)
+    sudo make install
+    if which systemctl; then
```
lanyuwen: border router won't function well without enabling Linux's IP forwarding which is disabled by default on most Linux distributions. we should figure out where to enable it before enable border router services

bukepo: To make PR smaller, how about submitting a separate PR for this and other system specified configuration?

bukepo:
@beriberikix Thanks. The `bootstrap` in the root directory is generated by nlbuild-autotools and used to generate automake related files. The same or equivalent file exists in each dependent projects.

bootstrap.sh - wpantund
autogen.sh - libcoap
bootstrap - OpenThread

README.md
bukepo: updated.

README.md
bukepo: updated.

README.md
bukepo: updated.

README.md
```diff
@@ -42,6 +42,39 @@ OTBR also provides the following services:
 
 Third-party components for Border Router Services include Simple Web Server and Material Design Lite for the framework of the web UI.
 
+# Get started
+
+We follow philosophy of [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all). Validated platforms include:
+
+* Raspberry Pi 3B (running recent Raspbian Jessie Lite)
```
lanyuwen: add the link of "raspbian jessie lite"?

script/_otbr
lanyuwen: I think this PR is a good fit to include system specified configurations, but I'm fine if you'd rather like to get that done in another PR timely. I would also recommend removing new added "Get started" content in README.md until then.

README.md
bukepo: Thanks, updated.

script/_otbr
bukepo: Thanks. Added script to enable ip forward in rc.local.

script/_ipforward
```diff
@@ -0,0 +1,49 @@
...
+sysctl -w net.ipv6.conf.all.forwarding=1\
+sysctl -w net.ipv4.ip_forward=1\
+#</ip-forward>' $RC_LOCAL
+    fi
+}
+
+ipforward_uninstall()
+{
+    sudo sed -i '/#<ip-forward>/,/#<\/ip-forward>/d' $RC_LOCAL
+}
```
lanyuwen: Thanks for working on this. This could work, but may not be easy for future maintenance. How about consolidating all such kinds of configurations in a dedicated script, and only insert that script into rc.local?

PR #31, opened 2017-06-06
Declare the method 'GetPskc()' to const type

LeZhang2016:


bukepo:
I'm wondering if we should simplify this class to a single function such as
```cpp
/**
 * This function computes PSKc.
 *
 * @returns PSKc in bytes.
 *
 */
const uint8_t *ComputePSKc(const char *aPassphrase, const char *aNetworkName, const uint8_t *aExtPanId);
```

PR #34, opened 2017-06-07
Make checks more strict

bukepo:
This PR,
1. adds `scan-build` check
2. adds `-Werror` to `CFLAGS`
3. fixes a mistake in libcoap patch
3. fixes some warnings in libcoap

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,3 +31,20 @@ matrix:
     - env: BUILD_TARGET="posix-check" VERBOSE=1
       compiler: gcc
       os: linux
+    - env: BUILD_TARGET="scan-build"
+      os: linux
+      compiler: clang
+      addons:
+        apt:
+          packages:
+          - libdbus-1-dev
+          - autoconf-archive
+          - doxygen
+          - ctags
+          - libboost-dev
+          - libboost-filesystem-dev
+          - libboost-system-dev
+          - libavahi-core-dev
+          - libavahi-client-dev
+          - expect
+          - clang
--- a/configure.ac
+++ b/configure.ac
@@ -200,7 +200,7 @@ AC_PATH_PROG(PERL, perl)
 #   -Wall                        CC, CXX
 #
 
-PROSPECTIVE_CFLAGS="-Wall"
+PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Werror"
 PROSPECTIVE_CXXFLAGS=""
 
 AX_CHECK_COMPILER_OPTIONS([C],   ${PROSPECTIVE_CFLAGS})
--- a/configure.ac
+++ b/configure.ac
@@ -200,11 +200,30 @@ AC_PATH_PROG(PERL, perl)
 #   -Wall                        CC, CXX
 #
 
-PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Werror"
-PROSPECTIVE_CXXFLAGS=""
+PROSPECTIVE_CFLAGS="-Wall -Wextra -Wshadow -Werror -std=c99 -pedantic-errors"
+PROSPECTIVE_CXXFLAGS="-Wall -Wextra -Wshadow -Werror -std=gnu++98 -Wno-c++14-compat"
+
+AC_CACHE_CHECK([whether $CC is Clang],
+    [nl_cv_clang],
+    [nl_cv_clang=no
+    if test "x${GCC}" = "xyes"; then
+        AC_EGREP_CPP([NL_CC_IS_CLANG],
+            [/* Note: Clang 2.7 lacks __clang_[a-z]+__ */
+#            if defined(__clang__) && defined(__llvm__)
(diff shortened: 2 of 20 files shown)


.travis.yml
```diff
@@ -3,6 +3,9 @@ language: cpp
 sudo: required
 dist: trusty
 
+git:
+  depth: 1
```
jwhui: Are we sure we want to do this?  [Travis docs](https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth) has this warning:

> Please note that if you use a depth of 1 and have a queue of jobs, Travis CI won’t build commits that are in the queue when you push a new commit.

bukepo: I also noticed it, I set it to 1 to,
* speeding up failing old commit in the queue
* possibly saving some time when doing `git clone`

I think it's OK to fail old if there's a new commit pushed to an queued branch. Because the resource is limited, quickly fail building an outdated commit can make room for the latest commit. I think that's why [Building only the latest commit](https://docs.travis-ci.com/user/customizing-the-build#Building-only-the-latest-commit) is introduced.

As far as I know, `depth 1` and `Building only the latest commit` is different. The former may cause the running job fail for being unable to find the commit id, while the latter will not fail if the commit id is within 50 depth.

As [Git Clone Depth](https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth) suggested,
> Travis CI clones repositories to a depth of 50 commits, which is only really useful if you are performing git operations.

I also did an experiment, which shows that this will not affect two branches.

jwhui: The issue with setting git depth to 1 is if multiple PRs are merged in succession to master, then any queued master builds will fail.  That's why we only enable building last commit on pull requests and not branches.

bukepo: thanks for the explanation, I will revert the change.

bukepo: thanks for the explanation, I will revert the change.

configure.ac
```diff
@@ -200,7 +200,7 @@ AC_PATH_PROG(PERL, perl)
 #   -Wall                        CC, CXX
 #
 
-PROSPECTIVE_CFLAGS="-Wall"
+PROSPECTIVE_CFLAGS="-Werror"
```
jwhui: Do we want to keep `-Wall`, or is that too strict?

bukepo: Thanks. If this causes problem, I will revert and let `scan-build` do this.

lanyuwen: Noticing we have excluded "third_party" code from the check, I think it may be worth to try stricter check on our own code. Any concerns on the potential issues? @jwhui

jwhui: No concerns - I'm proposing stricter checks here by having both `-Wall` and `-Werror`.

lanyuwen: Ah, i think it's a good idea then. :)

bukepo: Thanks for all your suggestions. Now both `-Wall` and `-Werror` are added.

configure.ac
```diff
@@ -200,7 +200,7 @@ AC_PATH_PROG(PERL, perl)
 #   -Wall                        CC, CXX
 #
 
-PROSPECTIVE_CFLAGS="-Wall"
+PROSPECTIVE_CFLAGS="-Wall -Werror"
```
jwhui: While we're at it, what about the following flags:
- `-Wextra`
- `-Wshadow`

We should also apply the same flags to `PROSPECTIVE_CXXFLAGS`.

Thoughts?

bukepo: Great, I thought `-Wall` would enable all possible warnings. Adding these two found more warnings. It will take some time to get them fixed.

And It seems we don't need to add same flags to `PROSPECTIVE_CXXFLAGS`. It's combined with `PROSPECTIVE_CFLAGS ` at last.

> PROSPECTIVE_CFLAGS="-Werror"
PROSPECTIVE_CXXFLAGS=""
AX_CHECK_COMPILER_OPTIONS([C],   ${PROSPECTIVE_CFLAGS})
AX_CHECK_COMPILER_OPTIONS([C++], ${PROSPECTIVE_CFLAGS} ${PROSPECTIVE_CXXFLAGS})

third_party/mbedtls/Makefile.am
```diff
@@ -62,6 +62,7 @@ libmbedtls_la_SOURCES                   = \
 libmbedtls_la_CPPFLAGS                        = \
     -I$(srcdir)/repo/include                    \
     -I$(srcdir)/repo/configs                    \
+    -D_GNU_SOURCE                               \
```
jwhui: Just curious, what portion of mbedtls requires this flag?

bukepo: I didn't override warning flags for mbedtls.

`third_party/mbedtls/repo/library/entropy_poll.c:105` uses `syscall()`, which from [man page](http://man7.org/linux/man-pages/man2/syscall.2.html), requires this flag.

PR #36, opened 2017-06-08
Switch the JSON lib from boost::property_tree to jsoncpp

LeZhang2016:


LeZhang2016:
I will count the resource consumption and then reopen this PR.

PR #40, opened 2017-06-12
Make OTBR services robust

bukepo:
Currently, OTBR mainly runs three services provided by OpenThread(wpantund, otbr-web, otbr-agent). Although they are already managed by systemd, but the configuration files are not properly set to make sure these services be restarted if some unexpected error happened. Especially for the case that NCP is unplugged, wpantund dies and will not recover.

This PR utilizes systemd's restart on-failure feature to make sure these services be restarted once exited unexpectedly. With this PR,  when unplugging NCP, systemd will try starting wpantund every 5s. Thus when the device is plugged, wpantund and otbr-agent can recover.

More information
* [1] https://www.freedesktop.org/software/systemd/man/systemd.service.html
* [2] https://www.freedesktop.org/software/systemd/man/systemd.unit.html

### DIFF
--- a/src/agent/otbr-agent.service.in
+++ b/src/agent/otbr-agent.service.in
@@ -8,7 +8,7 @@ EnvironmentFile=-@sysconfdir@/default/otbr-agent
 ExecStart=@sbindir@/otbr-agent $BA_OPTS
 Restart=on-failure
 RestartSec=5
-RestartPreventExitStatus=255
+RestartPreventExitStatus=SIGKILL
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


DuaneEllis-TI:
Hmm - another problem occurs on BeagleBone...   and TI-LaunchPads

TI-LaunchPads appear as /dev/ttyACMn  where n=some number.

At boot - and per per the boot logs, the /dev/ttyACMn does not appear until a few seconds *after* wpantund is started.  

granted, this 'restart' fix will help - but there will still be a red [FAILED] message during boot which causes noob confusion

Options:

1) At boot make the service wait for the device to appear (with a timeout=5 seconds?) 

2) Put a hard coded 5 second wait.... for anything which is not really optimal

3) Change Wpantund to do the wait (would this be valid? I don't think so)

I also think the exact sequence, and timing depends on the number of things that are happening on boot.

bukepo:
Thanks for reporting this issue. Since the actual time about how long the USB device will appear is not fixed for all possible platforms. I think we have to involve some other mechanism to make the solution general for all platforms. Could you please submit a separate issue?

PR #61, opened 2017-06-20
disable the ipv6 mDNS via setting the 'use-ipv6' and 'publish-aaaa-on-ipv4' in 'avahi_daemon.conf'

LeZhang2016:


LeZhang2016:
@bukepo , Thanks a lot for your suggestion. i have modified and tested the code according to your suggestion. The mDNS client can advertise service on different types of network by setting the parameter `protocol` in function `avahi_entry_group_add_service`. But it can not block the service entries with IPv6 address shown on the Commissioner App, because AAAA type record on IPv4 is enabled in default, Avahi can not provide corresponding API  to disable publishing AAAA record on IPv4. So we have to construct the new `host` with A type record  and change the target of SRV record to the new A record to solve this problem, in addition, we should find and enumerate all IPv4 addresses of available network interfaces on the OS to support A type record. The whole task is a little complex, and i need some time to achieve it.

jwhui:
@LeZhang2016, label this PR as "work-in-progress"?

LeZhang2016:
@jwhui , label "work-in-progress" is better than "P4", thanks!

jwhui:
Any updates on this PR?

LeZhang2016:
@jwhui , I will update and recommit this PR tomorrow.

PR #64, opened 2017-06-21
Add otbrDump() in the logging library.

bukepo:
This PR adds otbrDump to dump memory as hex string for debugging.

### DIFF
--- a/src/common/logging.cpp
+++ b/src/common/logging.cpp
@@ -28,32 +28,61 @@
 
 #include "logging.hpp"
 
+#include <assert.h>
 #include <stdarg.h>
-
+#include <stdint.h>
 #include <syslog.h>
 
-static int sLevel = LOG_INFO;
+static int        sLevel = LOG_INFO;
+static const char HEX_CHARS[] = "0123456789abcdef";
 
 void otbrLogInit(const char *aIdent, int aLevel)
 {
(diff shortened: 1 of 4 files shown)


src/common/logging.cpp
```diff
@@ -28,34 +28,63 @@
...
         return;
     }
+
     va_list args;
     va_start(args, aFormat);
     vsyslog(aLevel, aFormat, args);
     va_end(args);
 }
 
+void otbrDump(int aLevel, const char *aPrefix, const void *aMemory, size_t aSize)
```
jwhui: Is it useful to use the same `otDump` call from [src/core/common/logging.cpp](https://github.com/openthread/openthread/blob/master/src/core/common/logging.cpp#L114), which provides some additional formatting?

bukepo: Thanks, I used the output of `otDump`. It provides more formatting. However, I didn't benefit much from the formatting.

1. Most dumping happens on binary data, it not that useful to provide ascii conversion.
2. When I check the dumps, I wanna simply copy and parse it with tools, those formatting is annoying because I have to concat those segments into the original hex string.

Thoughts?

jwhui: Sounds good, just wanted to make sure you were aware of some existing code.

PR #78, opened 2017-07-13
Fix issue that server socket fails.

bukepo:
This PR mainly tries to recover DTLS from server socket failure.

I encounter this issue once. This issue results in a lot of logs in syslog.

Other changes,
* Added a function to convert error into string.


bukepo:
any comments? @jwhui

PR #86, opened 2017-08-13
[coap] support new message with id.

### LINKED ISSUE #81: Commissioner Proxy Issues
I believe there are 2 issues here.

Issue 1 is in the Agent, described here.

Issue 2 is in the embedded side, where the Commissioning Petition is actually handled (there is no provision for duplicate message detection) I believe there should be.
Described here: https://github.com/openthread/openthread/issues/2037

Short Description: The Agent does not seem to handle PROXY requests correctly.
By correctly, I specifically mean with the ability to detect duplicate COAP requests.

Example: 
   https://github.com/openthread/borderrouter/blob/master/src/agent/border_agent.cpp#L102

When a commissioner request is created (example: by the android app) the message is received with both a OPAQUE tag, and a MESSAGE ID - per RFC7252 - Section 2.1 - the Message ID is used to detect duplicates.

However, when the message is forwarded - a call is made to mCoap->NewMessage(), which results in a *NEW* message id for each forwarded message, See:

https://github.com/openthread/borderrouter/blob/master/src/agent/coap_libcoap.cpp#L150 

If the commissioner (ie: The Android App) sends a duplicate CoAP petition, each forwarded message has a "unique message id" - thus when the leader receives the petition - each comes with a unique message id and thus there is *NO* means for the Leader to detect a duplicate request.

I'm not sure of the correct solution here, 

option (1) would be to duplicate (pass through) the message original message ID, 

or 

Option (2) in the agent, mapping of the "android-app-message-id" vrs "proxy-message-id"

Separately, there's an issue with the embedded leader where it does not handles duplicate commissioning  CoAP requests. More generally - in the embedded side I think some common scheme to handle duplicate message detection is an important feature that is missing, while this problem was noticed in the commissioning process at the application layer there is a strong need for duplicate message detection.

bukepo:
This PR fixes #81 by setting message id when forwarding CoAP messages from external commissioner.

Since a request from external commissioner is responded with separate response, the response message id is not overridden with the message id from Thread leader.

### DIFF
--- a/src/agent/border_agent.cpp
+++ b/src/agent/border_agent.cpp
@@ -106,7 +106,7 @@ void BorderAgent::ForwardCommissionerRequest(const Coap::Resource &aResource, co
     const uint8_t *token = aMessage.GetToken(tokenLength);
     const char    *path = aResource.mPath;
 
-    Coap::Message *message = mCoap->NewMessage(Coap::kTypeConfirmable, Coap::kCodePost,
+    Coap::Message *message = mCoap->NewMessage(Coap::kTypeConfirmable, Coap::kCodePost, aMessage.GetMessageId(),
                                                token, tokenLength);
     Ip6Address     addr(kAloc16Leader);
     uint16_t       length = 0;
@@ -185,9 +185,8 @@ void BorderAgent::HandleRelayTransmit(const Coap::Message &aMessage, const uint8
         Ip6Address     addr(rloc);
         uint8_t        tokenLength = 0;
         const uint8_t *token = aMessage.GetToken(tokenLength);
-        Coap::Message *message = mCoap->NewMessage(Coap::kTypeNonConfirmable,
-                                                   Coap::kCodePost,
-                                                   token, tokenLength);
+        Coap::Message *message = mCoap->NewMessage(Coap::kTypeNonConfirmable, Coap::kCodePost,
+                                                   aMessage.GetMessageId(), token, tokenLength);
 
         message->SetPath(OT_URI_PATH_RELAY_TX);
         message->SetPayload(payload, length);
--- a/src/agent/coap.hpp
+++ b/src/agent/coap.hpp
@@ -95,6 +95,14 @@ public:
     virtual ~Message(void) {};
 
     /**
+     * This method returns the CoAP message id of this message.
+     *
+     * @returns the CoAP message id of the message.
+     *
+     */
+    virtual uint16_t GetMessageId(void) const = 0;
+
+    /**
      * This method returns the CoAP code of this message.
      *
      * @returns the CoAP code of the message.
@@ -273,6 +281,20 @@ public:
      *
      */
     virtual Message *NewMessage(Type aType, Code aCode, const uint8_t *aToken,
+                                uint8_t aTokenLength) = 0;
+    /**
+     * This method creates a CoAP message with the given arguments.
+     *
+     * @param[in]   aType           The CoAP type.
+     * @param[in]   aCode           The CoAP code.
+     * @param[in]   aMessageId      The CoAP message id.
+     * @param[in]   aToken          The CoAP token.
+     * @param[in]   aTokenLength    Number of bytes in @p aToken.
+     *
+     * @returns The newly CoAP message.
+     *
+     */
+    virtual Message *NewMessage(Type aType, Code aCode, uint16_t aMessageId, const uint8_t *aToken,
                                 uint8_t aTokenLength) = 0;
 
     /**
--- a/src/agent/coap_libcoap.cpp
+++ b/src/agent/coap_libcoap.cpp
@@ -75,6 +75,11 @@ const uint8_t *MessageLibcoap::GetToken(uint8_t &aLength) const
     return mPdu->hdr->token;
 }
 
+uint16_t MessageLibcoap::GetMessageId(void) const
+{
+    return mPdu->hdr->id;
+}
+
 Code MessageLibcoap::GetCode(void) const
 {
     return static_cast<Code>(mPdu->hdr->code);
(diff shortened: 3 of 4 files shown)


jwhui:
I'm fine with merging this for now if it helps things.  However, does this completely solve the problem?  In particular, if we are copying the Message ID from the Commissioner, shouldn't we also forward/proxy the message via a UDP socket rather than via the CoAP agent?

bukepo:
I'm not clear how to forward with UDP socket. The connection between border agent and external commissioner is DTLS, simply forwarding UDP packets cannot work. Besides, the URI of CoAP message may change for some requests. This only forward message ID from external commissioner

jwhui:
As we discussed with @DuaneEllis-TI, the ideal case is that CoAP communication between the Commissioner and Leader is end-to-end - the Border Agent does not maintain any CoAP-related state.  For example, the Border Agent should not perform duplicate detection nor should it perform retransmissions on its own.  Communication between the Border Agent and Leader happen via UDP.

bukepo:
@jwhui Thanks for the explanation. It's a great idea to involving a UDP proxy on NCP side. I will submit a PR later.

This PR doesn't involving duplication detection nor retransmission into border agent. However, it keeps the its original behavior to respond separately as defined by the spec. Here's the current workflow on host-side border agent:

1. Commissioner sends a COMM_PET/COMM_KA to border agent.
2. Border agent responds a empty ACK message.
3. Border agent generates a new CoAP message LEAD_PET/LEAD_KA with the same message id of the original request and delivers it to to NCP.
4. If commissioner retransmits the request(for not receiving the empty message or bad implemented), border agent redo from step 2.
5. When border agent receives the response from leader it converts the response to non-confirmable message with a new message id and sends it to commissioner.

The fault of this implementation is that commissioner will not retransmit if it successfully handled the empty message. Since there's no retransmission in border agent, no retransmission will happen between border agent and leader in this case.

To save the retransmission performed on border agent, border agent should not reply empty ACK message on its own, besides, we may have to make a choice between:

1. Leader always responds LEAD_PET/LEAD_KA separately.
2. Border agent only converts the response from leader to non-confirmable if its type is ACK, thus pretending the empty ACK message is lost(CoAP agent should process the actual response as an ACK according to RFC7252).

Or shall we ignore the separate response requirement in the spec?

jwhui:
@bukepo I believe the right approach is to keep the CoAP layer at the Border Agent stateless.  That means that CoAP Responses from the Leader to Border Agent are translated into CoAP Responses to the Border Agent to Commissioner.

Can you point me to the location in the spec that states the Border Agent and Leader need to respond separately?

bukepo:
@jwhui Only border agent needs respond separately. As defined in section 8.6.1, transaction pattern of COMM_PET and COMM_KA are both defined as `Req+Rsp_Separate (request part)`, which is different from LEAD_PET and LEAD_KA defined in section 8.8.1. I also wanna keep border agent stateless, so I asked in my last comment if we could ignore the separate response requirement for COMM_PET and COMM_KA.

jwhui:
@bukepo, yes I see the intended design with Section 8.6.1 now.  I think Section 8.6.1 could be made to work, but it is not completely stateless.  Rather than copying the Message ID between the two sides of the Border Agent, we actually copy the Token instead.  The process then looks like this:

1. Commissioner sends a COMM_PET.req/COMM_KA.req to border agent.
2. Border Agent responds a empty ACK message.
3. Border Agent sends a LEAD_PET.req/LEAD_KA.req with the same Token as the received COMM_PET.req/COMM_KA.req.
4. When Border Agent receives the LEAD_PET.rsp/LEAD_KA.rsp from the Leader, it sends the COMM_PET.rsp/COMM_KA.rsp as confirmable with the same Token as above.

Note that if the Commissioner does not receive a COMM_PET.rsp/COMM_KA.rsp, it should send a new COMM_PET.req/COMM_KA.req with a new Token.  That should cause the Border Agent to send a new LEAD_PET.req/LEAD_KA.req to the Leader.

I agree that this is not as ideal as the fully stateless approach.  The Border Agent still needs to maintain CoAP Request state for the LEAD_PET.req/LEAD_KA.req messages.  But it does allow the Commissioner to perform the necessary retransmissions.

At this point, I'm inclined to follow the existing Thread 1.1.1 Specification.  Thoughts?

bukepo:
The new process seems to be similar with the implementation before this PR from step 1 to step 3. The different of step 4 is that it sent non-confirmable response before this PR.

Does this solution still involves the UDP proxy on NCP side? If so, that means the border agent must do retransmission for requests to leader. Because the UDP proxy won't do that.
And if we respond confirmable response at step 4, we'll also have to implement retransmission too.

Besides, even responding an Empty ACK at step 2, there could still be retransmissions from commissioner, because the Empty ACK could be lost or delayed, that means, the duplication cannot be saved either.

As a result, it seems we'll have to enable full functionality of CoAP in border agent with this solution.

I'm considering the following solution, which only requires a duplication detection based on message id.
```
               detect duplication
commissioner <-------------------> border agent
                                        ^
                                        | ensure no duplicated message
                retransmission          v 
 Leader      <------------------->     NCP
```
Border agent still response Empty ACK. NCP still use the CoAP agent to send messages, which does the retransmission. Once NCP got the response, it delivers to host side border agent. And border agent respond commissioner with non-confirmable message. Thoughts?

jwhui:
@bukepo, yes, because the Thread Spec requires maintaining CoAP state on the Border Agent, the UDP proxy approach no longer applies.  I should have made that more clear in my prior comment.

With my current understanding of the Thread Spec:
- The Border Agent implements CoAP request duplicate detection for COMM_PET.req/COMM_KA.req message.
- The Border Agent performs its own retransmissions of LEAD_PET.req/LEAD_KA.req.
- The Border Agent generates COMM_PET.rsp/COMM_KA.rsp messages after receiving the LEAD_PET.rsp/LEAD_KA.rsp.
- The Border Agent sends COMM_PET.rsp/COMM_KA.rsp as **confirmable** messages.

With the exception of my last point above, I believe this is inline with your current thinking as well.

bukepo:
@jwhui OK, I will create another PR for the final solution.

DuaneEllis-TI:
@jwhui  - Note that if the Commissioner does not receive a COMM_PET.rsp/COMM_KA.rsp, it should send a new COMM_PET.req/COMM_KA.req with a new Token. That should cause the Border Agent to send a new LEAD_PET.req/LEAD_KA.req to the Leader.

I'm no CoAP expert here - but I don't think there is a CoAP requirement to have a token, let alone make the token unique per transmission.

PR #88, opened 2017-08-14
[agent] disable exceptions.

bukepo:
This PR simply removes exceptions from *otbr-agent*, as suggested in [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Exceptions).

### DIFF
--- a/src/agent/ncp.hpp
+++ b/src/agent/ncp.hpp
@@ -69,6 +69,15 @@ class Controller : public EventEmitter
 {
 public:
     /**
+     * This method initalize the NCP controller.
+     *
+     * @retval  OTBR_ERROR_NONE     Successfully initialized NCP controller.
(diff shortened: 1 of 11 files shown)


src/agent/ncp.hpp
```diff
@@ -69,6 +69,15 @@ class Controller : public EventEmitter
 {
 public:
     /**
+     * This method initalize the NCP controller.
+     *
+     * @retval  OTBR_ERROR_NONE     Successfully initialized NCP controller.
+     * @retval  *                   Failed to initialize NCP controller.
```
jwhui: Do we know the set of possible return values?

bukepo: Thanks, I thought this is just the interface. I will update this PR.

PR #91, opened 2017-08-17
[code] fix warnings.

bukepo:
This PR fixes,
* signed conversion warnings when enabled `-Wsign-conversion`, which is forced under some environment.
* `va_start` warning for now starting from promoted type argument.
* changed return type of `GetNow()` to `unsigned long`.


bukepo:
It seems the google cla check is stuck.

PR #92, opened 2017-08-17
[tests] remove hardcoding for commissioner test app

DuaneEllis-TI:
Current test app: ${borderrouter}/tests/meshcop/commissioner.cpp - and the associated test script meshcop are very hard coded for a very specific test scenario.

This pull request:

* Removes hard coding of the test app
* Adds features to the test app to calculate for example the PSKc
* And can be used with any target device by adjusting various command line parameters.

General intent is that the commissioner test app can be used from a script to commission *actual* physical hardware via a script in a test farm scenario.

NOTE: A number of bugs/issues where found with the test scripts during this process, they are noted here: https://github.com/openthread/borderrouter/issues/94

### DIFF
--- a/src/agent/dtls_mbedtls.cpp
+++ b/src/agent/dtls_mbedtls.cpp
@@ -77,8 +77,31 @@ static void MbedtlsDebug(void *aContext, int aLevel, const char *aFile, int aLin
         aLevel = OTBR_LOG_DEBUG;
         break;
     }
+    char buf[100];
+    /* mbed inserts a EOL
+     * And so does the otbrLog()
+     * remove that here.
+     */
 
-    otbrLog(aLevel, "%s:%04d: %s", aFile, aLine, aMessage);
+    /* make a copy because we modify it */
+    strncpy(buf, aMessage, sizeof(buf) - 1);
+    /* force termination */
+    buf[sizeof(buf) - 1] = 0;
+
+    /* now remove the extra cr/lf */
+    char *cp;
+    cp = strchr(buf, '\n');
+    if (cp)
+    {
+        *cp = 0;
+    }
+    cp = strchr(buf, '\r');
+    if (cp)
+    {
+        *cp = 0;
+    }
+
+    otbrLog(aLevel, "%s:%04d: %s", aFile, aLine, buf);
     (void)aContext;
 }
 
--- a/src/common/logging.cpp
+++ b/src/common/logging.cpp
@@ -40,13 +40,13 @@
 #include <sys/time.h>
 
 static int        sLevel = LOG_INFO;
-static const char HEX_CHARS[] = "0123456789abcdef";
+static const char kHEX_CHARS[] = "0123456789abcdef";
 
-static unsigned msecs_start;
-static bool     log_col0 = true; /* we start at col0 */
-static FILE    *log_fp;
-static bool     syslog_enabled = true;
-static bool     syslog_opened = false;
+static unsigned sMsecsStart;
+static bool     sLogCol0 = true; /* we start at col0 */
+static FILE    *sLogFp;
+static bool     sSyslogEnabled = true;
+static bool     sSyslogOpened = false;
 
 #define LOGFLAG_syslog 1
 #define LOGFLAG_file   2
@@ -54,14 +54,14 @@ static bool     syslog_opened = false;
 /** Set/Clear syslog enable flag */
 void otbrLogEnableSyslog(bool b)
 {
-    syslog_enabled = b;
+    sSyslogEnabled = b;
 }
 
 /** Enable logging to a specific file */
 void otbrLogSetFilename(const char *filename)
 {
-    log_fp = fopen(filename, "w");
-    if (log_fp == NULL)
+    sLogFp = fopen(filename, "w");
+    if (sLogFp == NULL)
     {
         fprintf(stderr, "Cannot open log file: %s\n", filename);
         perror(filename);
@@ -83,7 +83,7 @@ void otbrLogSetLevel(int aLevel)
 }
 
 /** Determine if we should not or not log, and if so where to */
-static int do_log(int aLevel)
+static int LogCheck(int aLevel)
 {
     int r;
 
@@ -91,7 +91,7 @@ static int do_log(int aLevel)
 
     r = 0;
 
-    if (syslog_opened && syslog_enabled && (aLevel <= sLevel))
+    if (sSyslogOpened && sSyslogEnabled && (aLevel <= sLevel))
     {
         r = r | LOGFLAG_syslog;
     }
@@ -100,7 +100,7 @@ static int do_log(int aLevel)
      * situation is they are debugging a problem, or need a extra
      * information. In this case, we do not test the log level.
      */
-    if (log_fp != NULL)
+    if (sLogFp != NULL)
     {
         r = r | LOGFLAG_file;
     }
@@ -120,64 +120,64 @@ static unsigned msecs_now(void)
     v = v * 1000;
     v = v + (tv.tv_usec / 1000);
 
-    if (msecs_start == 0)
+    if (sMsecsStart == 0)
     {
-        msecs_start = (unsigned)(v);
+        sMsecsStart = (unsigned)(v);
     }
-    v = v - msecs_start;
+    v = v - sMsecsStart;
     return v;
 }
 
 
 /** Write this string to the private log file, inserting the timestamp at column 0 */
-static void log_str(const char *cp)
+static void LogString(const char *cp)
 {
     int ch;
 
     while ((ch = *cp++) != 0)
     {
-        if (log_col0)
+        if (sLogCol0)
         {
-            log_col0 = false;
+            sLogCol0 = false;
             unsigned n;
             n = msecs_now();
-            fprintf(log_fp, "%4d.%03d | ", (n / 1000), (n % 1000));
+            fprintf(sLogFp, "%4d.%03d | ", (n / 1000), (n % 1000));
         }
         if (ch == '\n')
         {
-            log_col0 = true;
+            sLogCol0 = true;
         }
-        fputc(ch, log_fp);
+        fputc(ch, sLogFp);
         if (ch == '\n')
         {
             /* force flush (in case something crashes) */
-            fflush(log_fp);
+            fflush(sLogFp);
         }
     }
 }
 
 /** Print to the private log file */
-static void log_vprintf(const char *fmt, va_list ap)
+static void LogVprintf(const char *fmt, va_list ap)
 {
     char buf[1024];
 
     /* if not enabled ... leave */
-    if (log_fp == NULL)
+    if (sLogFp == NULL)
     {
         return;
     }
     vsnprintf(buf, sizeof(buf), fmt, ap);
 
-    log_str(buf);
+    LogString(buf);
 }
 
 /** Print to the private log file */
-static void log_printf(const char *fmt, ...)
+static void LogPrintf(const char *fmt, ...)
 {
     va_list ap;
 
     va_start(ap, fmt);
-    log_vprintf(fmt, ap);
+    LogVprintf(fmt, ap);
     va_end(ap);
 }
 
@@ -189,9 +189,9 @@ void otbrLogInit(const char *aIdent, int aLevel)
     assert(aLevel >= LOG_EMERG && aLevel <= LOG_DEBUG);
 
     /* only open the syslog once... */
-    if (!syslog_opened)
+    if (!sSyslogOpened)
     {
-        syslog_opened = true;
+        sSyslogOpened = true;
         openlog(aIdent, LOG_CONS | LOG_PID | LOG_PERROR, LOG_USER);
     }
     sLevel = aLevel;
@@ -214,18 +214,18 @@ void otbrLogv(int aLevel, const char *aFormat, va_list ap)
 
     assert(aFormat);
 
-    r = do_log(aLevel);
+    r = LogCheck(aLevel);
 
 
     if (r & LOGFLAG_file)
     {
         va_list cpy;
         va_copy(cpy, ap);
-        log_vprintf(aFormat, cpy);
+        LogVprintf(aFormat, cpy);
         va_end(cpy);
 
         /* logs do not end with a NEWLINE, we add one here */
-        log_str("\n");
+        LogString("\n");
     }
 
     if (r & LOGFLAG_syslog)
@@ -243,7 +243,7 @@ void otbrDump(int aLevel, const char *aPrefix, const void *aMemory, size_t aSize
     int            r;
     int            addr;
 
-    r = do_log(aLevel);
+    r = LogCheck(aLevel);
     if (r == 0)
     {
         return;
@@ -277,8 +277,8 @@ void otbrDump(int aLevel, const char *aPrefix, const void *aMemory, size_t aSize
 
         for (pEnd = p8 + this_size; p8 < pEnd; p8++)
         {
-            *++ch = HEX_CHARS[(*p8) >> 4];
-            *++ch = HEX_CHARS[(*p8) & 0x0f];
+            *++ch = kHEX_CHARS[(*p8) >> 4];
+            *++ch = kHEX_CHARS[(*p8) & 0x0f];
             *++ch = ' ';
         }
         *ch = 0;
@@ -289,7 +289,7 @@ void otbrDump(int aLevel, const char *aPrefix, const void *aMemory, size_t aSize
         }
         if (r & LOGFLAG_file)
         {
-            log_printf("%s: %04x: %s\n", aPrefix, addr, hex);
+            LogPrintf("%s: %04x: %s\n", aPrefix, addr, hex);
         }
     }
 }
@@ -329,6 +329,6 @@ const char *otbrErrorString(otbrError aError)
 
 void otbrLogDeinit(void)
 {
-    syslog_opened = false;
+    sSyslogOpened = false;
     closelog();
 }
--- a/src/common/logging.cpp
+++ b/src/common/logging.cpp
@@ -148,6 +148,11 @@ static void log_str(const char *cp)
             log_col0 = true;
         }
         fputc(ch, log_fp);
+        if (ch == '\n')
+        {
+	    /* force flush (in case something crashes) */
+            fflush(log_fp);
+        }
     }
 }
 
--- a/src/common/logging.cpp
+++ b/src/common/logging.cpp
@@ -28,64 +28,267 @@
 
 #include "logging.hpp"
 
+#include <stdlib.h>
+#include <stdio.h>
 #include <assert.h>
 #include <errno.h>
 #include <stdarg.h>
 #include <stdint.h>
 #include <string.h>
 #include <syslog.h>
+#include <sys/time.h>
 
 static int        sLevel = LOG_INFO;
 static const char HEX_CHARS[] = "0123456789abcdef";
 
+static unsigned msecs_start;
+static bool     log_col0 = true; /* we start at col0 */
+static FILE    *log_fp;
+static bool     syslog_enabled = true;
+static bool     syslog_opened = false;
+
+#define LOGFLAG_syslog 1
+#define LOGFLAG_file   2
+
+/** Set/Clear syslog enable flag */
+void otbrLogEnableSyslog(bool b)
+{
+    syslog_enabled = b;
+}
+
+/** Enable logging to a specific file */
+void otbrLogSetFilename(const char *filename)
+{
+    log_fp = fopen(filename, "w");
+    if (log_fp == NULL)
+    {
+        fprintf(stderr, "Cannot open log file: %s\n", filename);
+        perror(filename);
+        exit(EXIT_FAILURE);
+    }
+}
+
+/** Get the current debug log level */
+int  otbrLogGetLevel(void)
+{
+    return sLevel;
+}
+
+/** Set the debug log level */
+void otbrLogSetLevel(int aLevel)
+{
+    assert(aLevel >= LOG_EMERG && aLevel <= LOG_DEBUG);
+    sLevel = aLevel;
+}
+
+/** Determine if we should not or not log, and if so where to */
+static int do_log(int aLevel)
+{
+    int r;
+
+    assert(aLevel >= LOG_EMERG && aLevel <= LOG_DEBUG);
+
+    r = 0;
+
+    if (syslog_opened && syslog_enabled && (aLevel >= sLevel))
+    {
+        r = r | LOGFLAG_syslog;
+    }
+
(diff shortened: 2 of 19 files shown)


DuaneEllis-TI:
Note: I expect the pull request to fail the first time or two as I work through the automated test adjustments...

tests/meshcop/meshcop
```diff
@@ -27,99 +27,637 @@
...
+    if [ ! -x $NCP_EXE ]
+    then
+	EXIT_MESSAGE="Missing ncp: $NCP_EXE"
+	echo $EXIT_MESSAGE
+	exit 1
+    fi
+
+    if [ ! -x $CLI_EXE ]
+    then
+	EXIT_MESSAGE="MISSING cli: $CLI_EXE"
```
bukepo: MISSING or Missing?

src/agent/dtls_mbedtls.cpp
```diff
@@ -77,8 +77,31 @@ static void MbedtlsDebug(void *aContext, int aLevel, const char *aFile, int aLin
         aLevel = OTBR_LOG_DEBUG;
         break;
     }
+    char buf[100];
+    /* mbed inserts a EOL
+     * And so does the otbrLog()
```
DuaneEllis-TI: In the end, there are blank lines between logs - that is why they are removed.
I believe syslog - quietly removes them, they are not removed when this goes to the seperate file

src/common/logging.cpp
```diff
@@ -28,64 +28,270 @@
...
 #include <stdarg.h>
 #include <stdint.h>
 #include <string.h>
 #include <syslog.h>
+#include <sys/time.h>
 
 static int        sLevel = LOG_INFO;
 static const char HEX_CHARS[] = "0123456789abcdef";
 
+static unsigned msecs_start;
```
DuaneEllis-TI: I thought this was for public functions only, not private functions - that said I can  change this.

src/common/logging.cpp
```diff
@@ -28,64 +28,270 @@
...
+    struct timeval tv;
+    uint64_t       v;
+
+    gettimeofday(&tv, NULL);
+
+    v = tv.tv_sec;
+    v = v * 1000;
+    v = v + (tv.tv_usec / 1000);
+
+    if (msecs_start == 0)
```
DuaneEllis-TI: Could do that ...

src/common/logging.hpp
```diff
@@ -50,6 +50,34 @@ enum
...
+ * @param[in] enable true to log to/via syslog
+ *
+ */
+void otbrLogEnableSyslog(bool enabled);
+
+/**
+ * This function causes logs to be written to a specific file
+ * Note: Logs are still written to the syslog.
+ */
+void otbrLogSetFilename(const char *filename);
```
DuaneEllis-TI: does not always work - especially when you are running the tool under gdb.

tests/meshcop/commissioner.cpp
```diff
@@ -201,32 +103,54 @@ void HandleRelayReceive(const Coap::Resource &aResource, const Coap::Message &aM
...
             memset(&addr, 0, sizeof(addr));
             addr.sin_family = AF_INET;
             addr.sin_port = htons(kPortJoinerSession);
 
-            ret = sendto(context.mSocket, requestTlv->GetValue(), requestTlv->GetLength(),
-                         0, reinterpret_cast<const struct sockaddr *>(&addr), sizeof(addr));
+            otbrLog(OTBR_LOG_INFO, "Encapsulation: %d bytes for port: %d",
+                    requestTlv->GetLength(), kPortJoinerSession);
+            {
+                char buf[50];
```
DuaneEllis-TI: ok - did know of that one

tests/meshcop/commissioner.cpp
```diff
@@ -241,47 +165,90 @@ void HandleRelayReceive(const Coap::Resource &aResource, const Coap::Message &aM
...
+        recvfrom(aContext.mSocket, dtlsEncapsulation, sizeof(dtlsEncapsulation), 0, (struct sockaddr *)(&from_addr),
+                 &addrlen);
 
     VerifyOrExit(ret > 0);
+    {
+        // Print this, because in some environments...
+        // other things on the network use the same port
+        // as open thread is using here :-(
+        char buf[50];
+        get_ip_str((struct sockaddr *)(&from_addr), buf, sizeof(buf));
```
DuaneEllis-TI: No - we use ipv4 and we use ipv6 - the android app does both, hence it does both.

tests/meshcop/commissioner.cpp
```diff
@@ -318,18 +289,40 @@ static ssize_t SendCoap(const uint8_t *aBuffer, uint16_t aLength, const uint8_t
...
 
+/** callback to print debug from mbed */
 static void my_debug(void *ctx, int level,
                      const char *file, int line,
                      const char *str)
 {
     ((void) level);
-
+    (void)(ctx);
+#if 0
```
DuaneEllis-TI: ok - i did not see a #define type thing to use here.

tests/meshcop/commissioner.cpp
```diff
@@ -340,47 +333,61 @@ int export_keys(void *aContext, const unsigned char *aMasterSecret, const unsign
...
+            }
             break;
 
         case Meshcop::kCommissionerSessionId:
             context.mCommissionerSessionId = tlv->GetValueUInt16();
+            otbrLog(OTBR_LOG_INFO, "COMM_PET.rsp: session-id=%d", context.mCommissionerSessionId);
             break;
 
         default:
+            otbrLog(OTBR_LOG_INFO, "COMM_PET.rsp: ignore-tilv: %d", tlv_type);
```
DuaneEllis-TI: typo ok

tests/meshcop/commissioner.cpp
bukepo: Thanks for pointing out this. On border agent side, we created a single `AF_INET6` socket to handle both IPv4 and IPv6 requests. We could use this technique here too, thus simplify the code. Thoughts?

tests/meshcop/commissioner.cpp
```diff
@@ -705,11 +864,30 @@ int run(Context &context)
...
+
     context.mSsl = &ssl;
     context.mNet = &client_fd;
-#if defined(MBEDTLS_DEBUG_C)
-    mbedtls_debug_set_threshold(DEBUG_LEVEL);
-#endif
+    {
+        int n;
+        n = otbrLogGetLevel();
+        mbedtls_debug_set_threshold(n);
```
DuaneEllis-TI: could be but why should it?

Of the places I have worked over the last 30+ years writing C code I have seen multiple coding standards across multiple companies.  What you do greatly depends on what your measuring stick is, and how you judge code.

if it is "code coverage" - you write code that is very hard to debug.

If it is "cost to maintain" - you have look at what some coverage rules do to the debug process. Remember: many others will step through this code at some point in the future - and that cost is higher then you think.

thus there are 3 reasons one writes code this way:

1) In a debugger you can stop/halt - and inspect a variable, or modify it during debugging.

This is very important if what looks like a function is actually a macro... 

2) In a debugger a statement like  `foo( bar() )`  is really hard to step through. Debuggers only step *LINES* - not columns. This example is simple: there is a *SINGLE* nested function but there are many cases where there are more complex statements that make it super hard.  

Another is that of practicality - often various GUI wrappers for GDB can't always step correctly - then you have lost your debug state, this includes Eclipse, Emacs-GUD mode, and countless others.

3) While it *looks* more efficient, let the optimizer make it better.  it can always do a better job then you.

tests/meshcop/commissioner_compute.cpp
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * @file
+ *   The file computes various values used during commissioning.
+ */
+
+#include "commissioner.hpp"
+
+
+/** the hashmac is used in the steering data */
+bool compute_hashmac(void)
```
DuaneEllis-TI: No - because the intent is to use the global context to perform the work.

tests/meshcop/commissioner_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * @file
+ *   Misc functions used by commissioning test app
+ */
+
+#include "commissioner.hpp"
+
+/** return a hex string to print for logging */
+const char *hex_string(const uint8_t *pBytes, int n)
+{
+    static char buf[80 + 1];
```
DuaneEllis-TI: No - because it is a limited string that is purposely clipped to only show the first portion of the hex dump in ascii form.

Also - this is the *SINGLE* location where this "magic number" is used - where would the define be placed?

tests/meshcop/meshcop
```diff
@@ -27,99 +27,637 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
+IS_SUCCESS=false
+
+# Get our starting directory and remember it
+THIS_DIR=`pwd`
```
DuaneEllis-TI: good point.

tests/meshcop/meshcop
```diff
@@ -27,99 +27,637 @@
...
+# The agent is on the local host.
+OT_AGENT_IPADDR=127.0.0.1
+OT_AGENT_IPPORT=49191
+
+# Optionally we can write the commisiner log to a file
+OT_COMMISSIONER_LOG_FILE=./otbr-commissioner-log.txt
+
+# Using the commissioner app - Compute the PSKc from the above settings.
+set -x
+_x=`$OTBR_COMMISSIONER_EXE --disable-syslog --network-name $OT_NETWORK_NAME --xpanid $OT_XPANID --agent-passphrase $OT_AGENT_PASSPHRASE --compute-pskc`
```
DuaneEllis-TI: Trying to keep this 100% self contained and did not know that the other existed

tests/meshcop/meshcop
DuaneEllis-TI: typo

src/agent/dtls_mbedtls.cpp
bukepo: Thanks for pointing out this. Is it possible to remove them inside `otbrLog()` just like how `syslog()` does? I think this could avoid duplicated code in the commissioner side.

tests/meshcop/commissioner.cpp
bukepo: Thanks for sharing your considerations here. I learned a lot.

src/common/logging.cpp
```diff
@@ -28,64 +28,270 @@
...
+    if (log_fp != NULL)
+    {
+        r = r | LOGFLAG_file;
+    }
+    return r;
+}
+
+
+/** return the time, in milliseconds since application start */
+static unsigned msecs_now(void)
```
DuaneEllis-TI: The only example I find like this - of a static function is:
```
static void MbedtlsDebug(void *aContext, int aLevel, const char *aFile, int aLine, const char *aMessage)
```
In the src/agent/dtls_mbed.cpp file

So I followed that style,

src/common/logging.cpp
DuaneEllis-TI: don't think this is needed, fine the way it is.

src/common/logging.cpp
```diff
@@ -28,64 +28,270 @@
...
+{
+    int r;
+
     assert(aFormat);
-    assert(aLevel >= LOG_EMERG && aLevel <= LOG_DEBUG);
 
-    if (aLevel > sLevel)
+    r = do_log(aLevel);
+
+
```
DuaneEllis-TI: done

tests/meshcop/commissioner.cpp
```diff
@@ -160,6 +52,10 @@ void HandleJoinerFinalize(const Coap::Resource &aResource, const Coap::Message &
                           Coap::Message &aResponse,
                           const uint8_t *aIp6, uint16_t aPort, void *aContext);
 
+
+
```
DuaneEllis-TI: done

tests/meshcop/commissioner.cpp
```diff
@@ -188,11 +86,15 @@ void HandleJoinerFinalize(const Coap::Resource &aResource, const Coap::Message &
...
+void HandleRelayReceive(const Coap::Resource &aResource,
+                        const Coap::Message &aMessage,
                         Coap::Message &aResponse,
-                        const uint8_t *aIp6, uint16_t aPort, void *aContext)
+                        const uint8_t *aIp6,
+                        uint16_t aPort,
+                        void *aContext)
 {
     int            ret = 0;
+    int            tlvtype;
```
DuaneEllis-TI: ok

tests/meshcop/commissioner.cpp
DuaneEllis-TI: ok, actually more complicated because you need to pick the larger of INET6 or INET4 length.

tests/meshcop/commissioner.cpp
```diff
@@ -201,32 +103,54 @@ void HandleRelayReceive(const Coap::Resource &aResource, const Coap::Message &aM
...
+            }
+            ret = sendto(context.mSocket,
+                         requestTlv->GetValue(),
+                         requestTlv->GetLength(),
+                         0,
+                         reinterpret_cast<const struct sockaddr *>(&addr),
+                         sizeof(addr));
+            if (ret < 0)
+            {
+                otbrLog(OTBR_LOG_ERR, "relay receive, sendto() fails with %d", ret);
```
DuaneEllis-TI: sendto() returns the number of byte sent, or -1 on error, the negative test is sufficent because you cannot send a negative number of bytes.

print errno - good catch.

tests/meshcop/commissioner.cpp
DuaneEllis-TI: so I made both use the get_ip_str() function - and created a common #define for the bufsize here

tests/meshcop/commissioner.cpp
DuaneEllis-TI: FYI - this was here previously, I changed it

tests/meshcop/commissioner.cpp
```diff
@@ -416,64 +424,94 @@ int CommissionerPetition(Context &aContext)
...
 
+    otbrLog(OTBR_LOG_INFO, "COMM_PET.req: complete");
     return ret;
 }
 
+/** This handles the commissioner data set response, ie: response to the steering data etc */
 void HandleCommissionerSetResponse(const Coap::Message &aMessage, void *aContext)
 {
     uint16_t       length;
+    int            tlv_type;
```
DuaneEllis-TI: globally  tlv_type -> tlvType  in this file, done.

src/common/logging.cpp
bukepo: With the current implementation, the returned time actually depends on the first call to `msecs_now()`.  That is, if `msecs_now()` is firstly called `10s` after the program runs, it will return `0` instead of expected `10 * 1000`. I think this is different from what the comment indicates: `/** return the time, in milliseconds since application start */`. On the other hand, `otbrLogInit()` is supposed to be called at the beginning of the application. Initializing `msecs_start ` inside `otbrLogInit()` would provide more stable output.

tests/meshcop/meshcop
DuaneEllis-TI: other places too fixed

src/common/logging.cpp
bukepo: I meant the naming should be in upper camel case, as suggested in the OpenThread Style Guide. As you see, most OpenThread code follows the naming convention. Besides, there's already a function called `GetNow()` in `common/timer.hpp`. We might finally merge this function with that, either in this PR or in future PR.
> All OpenThread class, namespace, structure, method, function, enumeration, and type names in the C++ language shall be in upper camel case. Exception: the top level OpenThread namespace 'ot'.
> All OpenThread instantiated names of instances of classes, namespaces, structures, methods, functions, enumerations, and types as well as method and function parameters in the C++ language shall be in lower camel case.

tests/meshcop/commissioner.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+    /* this is the commissioner session id */
+    uint16_t mCommissionerSessionId;
+
+    /** all things for the border agent */
+    struct br_agent
+    {
+
+        /** network port, in form mbed likes it (ascii) */
+        char mPort_ascii[ 7 ];
```
bukepo: `mPort_ascii -> mPort`?
naming convention:
> Names shall not give any idea of type, such as is done with System Hungarian notation.

tests/meshcop/commissioner.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+        /** network address, in form mbed likes it (ascii) */
+        char mAddress_ascii[64];
+
+        /** xpanid from command line & binary form */
+        /* Note: xpanid is used to calculate the PSKc */
+        struct
+        {
+            char    ascii[  (kXpanidLength * 2) + 1 ];
```
bukepo: is it possible to check and to convert the `xpanid` to binary at the phase of parsing argument(in 
 `handle_xpanid` in commissioner_argcargv.cpp) so that we could simplify here and fail earlier if argument is not valid?

tests/meshcop/commissioner.hpp
```diff
@@ -0,0 +1,309 @@
...
+        /* Note: passphrase is used to calculate the PSKc */
+        char mPassPhrase[ kBorderRouterPassPhraseLen + 1 ];
+
+
+        /** the PSKc used with the agent */
+        struct
+        {
+
+            /* this class does the calculation */
+            Psk::Pskc mTool;
```
bukepo: Is it better to keep `Psk::Pskc` in local scope?

tests/meshcop/commissioner.hpp
```diff
@@ -0,0 +1,309 @@
...
+    /** All things about the joiner */
+    struct joiner
+    {
+
+        /** the EUI64 of the Joiner or HASHMAC, either from cmdline or computed */
+        struct
+        {
+            char    ascii[ (kEui64Len * 2) + 1 ];
+            uint8_t bin[ kEui64Len ];
+        } mEui64, mHashMac;
```
bukepo: same concern as `xpanid`, would it be better to handle parsing its argument handler?

tests/meshcop/commissioner_argcargv.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+using namespace ot;
+using namespace ot::Utils;
+using namespace ot::BorderRouter;
+
+/** see: commissioner_argcargv.hpp, processes 1 command line parameter */
+int argcargv::parse_args(void)
+{
+    const char                *arg;
+    const struct argcargv_opt *pOPT;
```
bukepo: name as `opt`?

tests/meshcop/commissioner.hpp
DuaneEllis-TI: not so simple here, the _bin and _ascii parts serve two uses.

The ascii component, ie: xpanid._ascii[0] - if non-zero means the value has been provided on the command line (ie: set), and/or it has been calculated. 

For example, the  HASHMAC can be given on the command line or the EUI64 can be given on the command line and the value calculated. 

When the values are calculated, the "ascii form" is created effectively for two purposes - 1) as a flag that this is already completed and 2) so that it can be immediately printed by code that should log the result.  

In addition, the sizeof( _ascii ) for each of these is used during command line parsing to validate the ascii hex encoded data the user might provide, for example what if the user provided 1 to few ascii bytes on for the PSKc on the command line?  Do we insert 4 bits of zeros?

I can't use the _bin[0] component, because the data element *MIGHT* start with a zero byte.

I understand what you are saying, but what you are asking for is a bunch of flags and small ascii buffers with magic values everywhere which is worse.

Space/memory here is not the concern - this is test/validation code not embedded code.

tests/meshcop/commissioner.hpp
DuaneEllis-TI: don't understand your suggestion

tests/meshcop/commissioner.hpp
DuaneEllis-TI: This, and the address  - see comment about the xpandid._ascii below
it's named this way for consistency with the others.

tests/meshcop/commissioner.hpp
bukepo: Thanks for the explanation to let me know more about the details.

Although the commissioner code is just test code for now, it could be improved or referenced to implement a fully functional commissioner in future, that's why I'd suggest those changes, which might improve reusability.

For the flag purpose, is it possible to use a invalid value as the not set flag? For example, 0xffff is not a valid value as panid, initializing it to 0xffff could possibly save a flag variable. For extended pan id, since no special value is defined(correct me if I'm wrong), I think involving a flag variable in the parser(commissioner_argcargv.cpp) is reasonable. No matter how the flag is implemented, restricting it in the parsing phrase seems to help leave the core logic of commissioner(commissioner.cpp) cleaner, so that easier to reuse.

For logging, I found these values would not be in log frequently. I only see `mPSKd_ascii` and `mAddress_ascii` are actually used in logging. So do you think we could call a bin2hex conversion function when logging?

DuaneEllis-TI: Saving 1 flag - but making it inconsistent with others - is not an improvement.

If/when the commissioner code is improved - why don't we address these things then.

At this point I am running out of time - I leave for vacation tomorrow and this will not be addressed for the next 3 weeks.

src/common/logging.cpp
DuaneEllis-TI: ok

src/common/logging.cpp
```diff
@@ -28,64 +28,270 @@
...
+/** Set/Clear syslog enable flag */
+void otbrLogEnableSyslog(bool b)
+{
+    sSyslogEnabled = b;
+}
+
+/** Enable logging to a specific file */
+void otbrLogSetFilename(const char *filename)
+{
+    sLogFp = fopen(filename, "w");
```
DuaneEllis-TI: done

tests/meshcop/commissioner.hpp
DuaneEllis-TI: as above

tests/meshcop/commissioner_argcargv.cpp
DuaneEllis-TI: done and other places

tests/meshcop/commissioner.hpp
bukepo: As OpenThread style guide suggested
>Data objects shall be declared at the smallest possible level of scope.

This `mTool` was used only once at a function. I think it should be the function(i.e.`compute_pskc()`)'s local scope variable.

DuaneEllis-TI:
@bukepo & @jwhui  - I am in the office tomorrow, then leave for a few weeks (vacation)
I would really like to get this merged in before I leave, otherwise it will sit until I return.

Thanks.

PR #98, opened 2017-08-25
[agent] add dataset uris.

bukepo:
This PR adds more resources for commissioner.

This PR also allows null pointer passed to `otbrDump()`. This still needs to be improved, but I don't have enough time to clean up this part right now.

### DIFF
--- a/src/agent/uris.hpp
+++ b/src/agent/uris.hpp
@@ -1,33 +1,150 @@
 /*
- *    Copyright (c) 2017, The OpenThread Authors.
- *    All rights reserved.
+ *  Copyright (c) 2016-2017, The OpenThread Authors.
+ *  All rights reserved.
  *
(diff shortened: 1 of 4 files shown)


src/agent/uris.hpp
```diff
@@ -125,4 +242,38 @@
...
 #define OT_URI_PATH_COMMISSIONER_KEEP_ALIVE "c/ca"
 
-#endif // URIS_HPP_
+/**
+ * @def OT_URI_PATH_DIAGNOSTIC_GET_REQUEST
+ *
+ * The URI Path for Network Diagnostic Get Request.
+ *
+ */
+#define OT_URI_PATH_DIAGNOSTIC_GET_REQUEST "d/dg"
```
jwhui: nit: value alignment seems off-by-one from `#define` above.

PR #140, opened 2018-04-26
Publish MDNS in agent

bukepo:


bukepo:
The travis error is caused by https://github.com/openthread/openthread/pull/2684

PR #143, opened 2018-05-11
fix warnings on strict environment

bukepo:


bukepo:
failure caused by travis network error.

bukepo:
@jwhui fixed travis error.

PR #156, opened 2018-07-20
Fix join thread network from web gui

gjc13:
There are both frontend and backend problems to use otbr web gui to join a thread network.
- fix frontend bug that join button doesn't even post to server
- update deprecated property to set network key


gjc13:
@bukepo Please kindly review

PR #159, opened 2018-07-25
[wpan-service] fix pskc buffer overflow

gjc13:
Fix overflow of pskcStr since it is manipulated by strlen() in dbus_set.cpp

### DIFF
--- a/src/web/web-service/wpan_service.cpp
+++ b/src/web/web-service/wpan_service.cpp
@@ -89,13 +89,12 @@ exit:
 
 std::string WpanService::HandleFormNetworkRequest(const std::string &aFormRequest)
 {
-    Json::Value      root;
-    Json::FastWriter jsonWriter;
-    Json::Reader     reader;
-    std::string      response;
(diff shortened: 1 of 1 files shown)


src/web/web-service/wpan_service.cpp
```diff
@@ -89,12 +89,13 @@ std::string WpanService::HandleJoinNetworkRequest(const std::string &aJoinReques
...
-    std::string              response;
-    ot::Psk::Pskc            psk;
-    char                     pskcStr[OT_PSKC_MAX_LENGTH * 2];
+    Json::Value      root;
+    Json::FastWriter jsonWriter;
+    Json::Reader     reader;
+    std::string      response;
+    ot::Psk::Pskc    psk;
+    char             pskcStr[OT_PSKC_MAX_LENGTH * 2 + 1]; // for manipulating with strlen
+    pskcStr[OT_PSKC_MAX_LENGTH * 2] = 0;
```
bukepo: move this line after declaration block? and prefer use `'\0'` to `0`;

PR #161, opened 2018-08-02
[fix-commission-petition] fix commission test utility function

gjc13:
fix reject handling and commissioner id comparing

### DIFF
--- a/src/common/tlv.hpp
+++ b/src/common/tlv.hpp
@@ -198,9 +198,9 @@ enum
 
 enum
 {
-    kPetitionAccepted = 1,
-    kPetitionPending  = 0,
-    kPetitionRejected = -1,
+    kStateAccepted = 1,
+    kStatePending  = 0,
+    kStateRejected = -1,
 };
 
 } // namespace Meshcop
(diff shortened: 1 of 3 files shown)


src/common/tlv.hpp
```diff
@@ -196,6 +196,13 @@ enum
     kJoinerRouterKek         = 21,
 };
 
+enum
+{
+    kPetitionAccepted = 1,
```
bukepo: name as `kState*`?

tests/meshcop/commissioner.cpp
```diff
@@ -838,7 +832,7 @@ int CommissionerServe(Context &aContext)
 exit:
     otbrLog(OTBR_LOG_INFO, "CommissionerServe: result=%d", ret);
 
-    return ret;
+    return aContext.mState == kStateDone ? 0 : -1;
```
librasungirl: why not return `ret` directly?

gjc13: Returning ret directly will need additional flags for the double loop

librasungirl: A little confused about the function of the variable `ret` in your interpretation.
How about initializing `ret` as `-1 `and exit without reassign `ret` with `errno` when fails to create the socket in Line 746?

gjc13:
No need to merge now since this part of code is undergoing a full rewrite.

Issue #157, opened 2018-07-23
Vulnerability to XSS


gjc13:
```
GET /no_such_file.xxx?"><script>alert(73541);</script> HTTP/1.1
Host: 127.0.0.1
Connection: Keep-Alive
```
Causes the web page to alert

gjc13:
Temporarily limiting access to localhost in this commit:https://github.com/gjc13/borderrouter/commit/cc1a253773007a42fdae0ec556c109eb4b5b79b0
The vulnerability, however, comes from the http framework we are using now. To totally fix this issue, maybe we should wait for the api definition of openthread and implement the web server with modern frameworks.

bukepo:
Is it reasonable to provide a match all resource responding with escaped information as a quick fix for this issue?

jwhui:
Resolved by #162.

PR #162, opened 2018-08-07
Fix xss

gjc13:
Rosolve issue #157

### DIFF
--- a/src/web/main.cpp
+++ b/src/web/main.cpp
@@ -123,7 +123,7 @@ int main(int argc, char **argv)
     otbrLog(OTBR_LOG_INFO, "border router web started on %s", interfaceName);
 
     server = new ot::Web::WebServer();
-    server->StartWebServer(interfaceName, kDefaultListenAddr, port);
+    server->StartWebServer(interfaceName, httpListenAddr, port);
 
     otbrLogDeinit();
 
--- a/src/web/main.cpp
+++ b/src/web/main.cpp
@@ -50,7 +50,7 @@
 
 static const char kSyslogIdent[]          = "otWeb";
 static const char kDefaultInterfaceName[] = "wpan0";
-static const char kDefaultListenAddr[]    = "127.0.0.1";
+static const char kDefaultListenAddr[]    = "0.0.0.0";
 
 void PrintVersion(void)
 {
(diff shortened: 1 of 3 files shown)


src/web/main.cpp
```diff
@@ -50,6 +50,7 @@
 
 static const char kSyslogIdent[]          = "otWeb";
 static const char kDefaultInterfaceName[] = "wpan0";
+static const char kDefaultListenAddr[]    = "127.0.0.1";
```
bukepo: We should allow request from remote address by default.

src/web/main.cpp
```diff
@@ -103,6 +108,12 @@ int main(int argc, char **argv)
         printf("interfaceName not specified, using default %s\n", interfaceName);
     }
 
+    if (httpListenAddr == NULL)
+    {
+        httpListenAddr = kDefaultListenAddr;
+        printf("litenAddress not specified, using default %s\n", httpListenAddr);
```
bukepo: use `stderr` for error message. typo: `litenAddress` -> `Bind address`.

src/web/web-service/web_server.hpp
```diff
@@ -80,10 +80,11 @@ class WebServer
      * This method starts the Web Server.
      *
      * @param[in]  aIfName  The pointer to the interface name of wpantund.
+     * @param[in]  aListenAddr The http server listen address, can be NULL for any address.
```
bukepo: align the description with above and below lines.

src/web/web-service/web_server.cpp
```diff
@@ -195,7 +199,7 @@ void WebServer::DefaultHttpResponse(void)
 
         } catch (const std::exception &e)
         {
-            std::string content = "Could not open path " + request->path + ": " + e.what();
+            std::string content = "Could not open path";
```
bukepo: how about include the escaped path?

gjc13: Escaping url turns out to be tricky and is not really that necessary

src/web/main.cpp
gjc13: It's just a info, so I think stdout suffices.

src/web/main.cpp
```diff
@@ -112,7 +123,7 @@ int main(int argc, char **argv)
     otbrLog(OTBR_LOG_INFO, "border router web started on %s", interfaceName);
 
     server = new ot::Web::WebServer();
-    server->StartWebServer(interfaceName, port);
+    server->StartWebServer(interfaceName, kDefaultListenAddr, port);
```
bukepo: `kDefaultListenAddr` -> `httpListenAddr`?

PR #177, opened 2018-08-22
[nat64] fix tayga start

gjc13:
tayga inner ipv6 address should be specified for default ipv6 translation prefix


jwhui:
Thanks for this fix. Anything we can do in continuous integration testing to avoid this kind of regression in the future?

gjc13:
Trying to add nat64 travis test but encountered weird ipv6 problem with travis.
See [this issue](https://github.com/travis-ci/travis-ci/issues/8711).
The travis build log can be found [here](https://travis-ci.org/openthread/borderrouter/jobs/419522509)

PR #180, opened 2018-08-27
[otbr-agent] should always use system dbus

gjc13:


bukepo:
@gjc13 could you please also fix this issue in otbr-web?

PR #186, opened 2018-09-05
[meshcop] allow commissioner to set joiner filter in separate method 

gjc13:


tests/meshcop/commissioner.cpp
```diff
@@ -398,6 +400,11 @@ bool Commissioner::IsValid(void) const
     return mCommissionState != kStateInvalid;
 }
 
+bool Commissioner::IsCommissionerAccepted()
```
jwhui: "All void functions or methods shall explicitly declare and specify the void type keyword."

tests/meshcop/commissioner.hpp
```diff
@@ -111,6 +115,14 @@ class Commissioner
      */
     bool IsValid(void) const;
 
+    /**
+     * This method returns whether the commissioner petition succeeded
+     *
+     * @returns inner commisioner state is kStateAccepted
+     *
+     */
+    bool IsCommissionerAccepted();
```
jwhui: All void functions or methods shall explicitly declare and specify the void type keyword.

PR #195, opened 2018-09-18
[commissioner] move commissioner to src

bukepo:
This PR just moves commissioner from `tests` into `src`.


src/commissioner/commissioner.cpp
```diff
@@ -613,7 +613,7 @@ int Commissioner::SendRelayTransmit(uint8_t *aBuf, size_t aLength)
     responseTlv->SetValue(mJoinerRouterLocator);
     responseTlv = responseTlv->GetNext();
 
-    if (mJoinerSession->NeedAppendKek())
+    if (mJoinerSession != NULL && mJoinerSession->NeedAppendKek())
```
gjc13: Maybe we can add assert instead of condition, since it will be very unreasonable to have a NULL joiner session since we are forwarding data from it.

Issue #134, opened 2018-02-06
Dnsmasq service fails to start - Address already in use


ubx-caloca:
I downloaded the OTHBR code from the latest commit (07ea4a54a5ee4105475627fa23d65d2cdda9ec14) and followed all the steps exactly as described in the OTBR setup guide (https://openthread.io/guides/border_router) in a Raspberry Pi 3 with Raspbian Lite 2017-11-29 . However I'am getting an error when booting Raspbian :

.....
0;32m  OK  0m Started OpenBSD Secure Shell server.
0;1;31mFAILED 0m Failed to start dnsmasq - A lightweight DHCP and caching DNS server.
See 'systemctl status dnsmasq.service' for details.
0;32m  OK  0m Reached target Host and Network Name Lookups.
.....

The status of the dnsmasq service (got from executing "sudo systemctl status dnsmasq") is the following 
....
dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS server
   Loaded: loaded (/lib/systemd/system/dnsmasq.service; enabled; vendor preset: enabled)
   Active: failed Result: exit-code since Fri 2018-02-02 22:17:15 UTC; 2min 52s ago
  Process: 497 ExecStart=/etc/init.d/dnsmasq systemd-exec (code=exited, status=2)
  Process: 467 ExecStartPre=/usr/sbin/dnsmasq --test (code=exited, status=0/SUCCESS)

Feb 02 22:17:14 raspberrypi systemd 1: Starting dnsmasq - A lightweight DHCP and caching DNS server...
Feb 02 22:17:14 raspberrypi dnsmasq 467: dnsmasq: syntax check OK.
Feb 02 22:17:15 raspberrypi systemd 1: dnsmasq.service: Control process exited, code=exited status=2
Feb 02 22:17:15 raspberrypi dnsmasq 497: dnsmasq: failed to create listening socket for 192.168.1.2: Address already in use
Feb 02 22:17:15 raspberrypi systemd 1: Failed to start dnsmasq - A lightweight DHCP and caching DNS server.
Feb 02 22:17:15 raspberrypi systemd 1: dnsmasq.service: Unit entered failed state.
Feb 02 22:17:15 raspberrypi systemd 1: dnsmasq.service: Failed with result 'exit-code'
.....

[...]

PR #198, opened 2018-09-18
[script] disable DNS of dnsmasq on port failure

bukepo:
This PR disables DNS service of dnsmasq if failed. Possible fix for #134


Vyrastas:
Does this change the dnsmasq configuration in the guide?  https://openthread.io/guides/border-router/access-point#configure-dnsmasq

Or should we just leave it as is to be safe (because it works).

bukepo:
This issue #134 is about the setup guide in openthread.io. However, the script doesn't manage dnsmasq through that way. So I close this PR. And since the guide has been updated, I think we can also close #134. @jwhui

PR #203, opened 2018-10-02
Adds an access control list to dns config

### LINKED ISSUE #181: Recursive DNS on public network allows DDOS attack
The current configuration of BIND is for a recursive DNS server, which has the potential to be used for amplification attacks if the device is on a public network. Is there any reason why the border router cannot simply act as a forwarder?

Additionally, the web gui, if public facing, provides a vector for outside parties to modify the thread network without any kind of authorization.

nealjack:
Adds the `thread` acl to restrict queries (recursive and not) to the
local thread network. Fixes #181

### DIFF
--- a/script/_dns64
+++ b/script/_dns64
@@ -34,7 +34,7 @@ BIND_CONF_OPTIONS=/etc/bind/named.conf.options
 NAT64_PREFIX=64:ff9b::/96
 
 DNS64_NAMESERVER_ADDR=127.0.0.1
-DNS64_CONF="dns64 `echo $NAT64_PREFIX | tr \"/\" \"\/\"` { clients { any; }; recursive-only yes; };"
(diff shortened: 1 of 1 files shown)


nealjack:
I am unable to reproduce an invalid `/etc/bind/name.conf.options`. I tested by running both `script/setup` and `script/update` repeatedly. Restarting the bind9 service after performing these actions results in no failures. Can you supply the before and after `name/conf.options`? Are you starting from a fresh install of bind?

nealjack:
What platform are you running the script on? I've tested on Raspbian and Ubuntu and cannot replicate your results.

nealjack:
Additionally, which version of sed are you using?

nealjack:
Ok, I spun up a fresh Ubuntu 18.04.1 virtual machine with sed 4.4 and cannot replicate your problems. Are you sure you are cleanly checked out and on the correct commit (45166870cb)?

PR #210, opened 2018-10-12
[libc] do not use glibc-internal header sys/cdefs.h

sjlongland:
The `sys/cdefs.h` is an internal header of `glibc` and is not meant to
be used by user code.  The macros needed from it can be trivially
implemented where required, relying on this header being present
excludes the OpenThread Border Router from being used on systems where
`glibc` is not the used C library (e.g. OpenWRT, where `uclibc` or
`musl` is standard).

https://wiki.musl-libc.org/faq.html#Q:-When-compiling-something-against-musl,-I-get-error-messages-about-%3Ccode%3Esys/cdefs.h%3C/code%3E

### DIFF
--- a/third_party/wpantund/repo/src/util/string-utils.h
+++ b/third_party/wpantund/repo/src/util/string-utils.h
@@ -28,14 +28,16 @@
 #include <stdint.h>
 #include <string.h>
 #include <stdbool.h>
-#include <sys/cdefs.h>
 
 #define strcaseequal(x, y)   (strcasecmp(x, y) == 0)
 #define strncaseequal(x, y, n)   (strncasecmp(x, y, n) == 0)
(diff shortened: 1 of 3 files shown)


jwhui:
These changes are specific to `wpantund`. Rather than patching third_party code directly here, would it be better to merge https://github.com/openthread/wpantund/pull/362 first and then sync with the latest wpantund?

sjlongland:
This is an option.  In the meantime though, people who run into this at least have a patch they can apply should they need it.

I'll do a re-base when the other branch is merged and see what's left. :-)

jwhui:
@bukepo, should we merge this PR as is? Or should we sync the changes in from openthread and wpantund repos?

bukepo:
I think it's OK to merge this PR. I will sync wpantund later.

PR #215, opened 2018-10-23
[commissioner] use getopt

bukepo:
This PR mainly utilizes `getopt_long()` to parse arguments for commissioner. 

* fixed initializing otbrLog()
* removed unused envelope_time
* cleaned code
* rename `--agent-addr` to `--agent-host` to support hostname


src/commissioner/commissioner_argcargv.cpp
```diff
@@ -297,244 +74,207 @@ static void handle_pskd(ArgcArgv *pThis, CommissionerArgs *args)
...
-    n = pThis->NumParam();
-    /* sanity... */
+    if (!allowAllJoiners)
+    {
+        VerifyOrExit(aArgs.mPSKd != NULL, fprintf(stderr, "Missing PSKd!"));
+        VerifyOrExit(isEui64Set, fprintf(stderr, "Missing EUI64!"));
 
-    /* note: 86400 = 1 day in seconds */
-    if ((n < 3) || (n > 86400))
+        aArgs.mSteeringData.ComputeJoinerId(joinerEui64, joinerEui64);
```
gjc13: Previous code puts steering data computation(and any other non-trivial task) outside argument parsing to make it easier to identify main procedures in commissioning. Is it a preferred way of organizing code?

bukepo: I think computing steering data is trivial and should be done when parsing arguments. We may support multiple joiners later, if we don't compute steering data as we parsing argument, we have to store these arguments, which is not clean.
Besides, some command line arguments are not used after parsing, we don't need to save them all the time.

Issue #243, opened 2019-02-06
strncpy used unsafely throughout the code base


doanac:
**Describe the bug**
The code base has a lot of uses of strncpy that aren't accounting for the space the nul-byte requires. Additionally, in the places it does, there are no checks that strncpy had to truncate the src buffer.

**To Reproduce**
~~~
$ git grep strncpy | grep -v 1
src/agent/mdns_avahi.cpp:        strncpy(service.mName, aName, sizeof(service.mName));
src/agent/mdns_avahi.cpp:        strncpy(service.mType, aType, sizeof(service.mType));
src/agent/mdns_mdnssd.cpp:        strncpy(service.mName, aName, sizeof(service.mName));
src/agent/mdns_mdnssd.cpp:        strncpy(service.mType, aType, sizeof(service.mType));
src/commissioner/addr_utils.cpp:        strncpy(aOutBuf, "Unknown AF", aLength);
src/commissioner/commissioner.cpp:    strncpy(buf, aStr, sizeof(buf));
src/web/web-service/wpan_service.hpp:    void SetInterfaceName(const char *aIfName) { strncpy(mIfName, aIfName, sizeof(mIfName)); }
src/web/wpan-controller/dbus_base.cpp:    strncpy(mDestination, aDestination, sizeof(mDestination));
src/web/wpan-controller/dbus_base.cpp:    strncpy(mIface, aIface, sizeof(mIface));
src/web/wpan-controller/dbus_base.cpp:    strncpy(mInterfaceName, aInterfaceName, sizeof(mInterfaceName));
src/web/wpan-controller/dbus_base.cpp:    strncpy(mPath, aPath, sizeof(mPath));
src/web/wpan-controller/dbus_base.cpp:    strncpy(mDBusName, aDBusName, sizeof(mDBusName));
src/web/wpan-controller/dbus_gateway.hpp:        strncpy(mAddressString, aAddressString, strlen(aAddressString));
src/web/wpan-controller/dbus_get.cpp:        strncpy(mPropertyList[propCnt].name, pName, sizeof(mPropertyList[propCnt].name));
src/web/wpan-controller/dbus_get.cpp:        strncpy(mPropertyList[i].value, GetPropertyValue(mPropertyList[i].name), sizeof(mPropertyList[i].value));
src/web/wpan-controller/dbus_ifname.cpp:            strncpy(mDBusName, aItemDBusName, DBUS_MAXIMUM_NAME_LENGTH);
[...]

PR #245, opened 2019-02-14
Fix improper uses of strncpy

doanac:
This addresses bug #243. In many places stncpy was used unsafely in
way that wouldn't null terminate buffers. strlcpy is a safer
replacement. Additionally, I've added error logging to note there's
a real issue with the code.

Signed-off-by: Andy Doan <andy@foundries.io>

### DIFF
--- a/third_party/wpantund/repo/src/wpanctl/wpanctl-utils.c
+++ b/third_party/wpantund/repo/src/wpanctl/wpanctl-utils.c
@@ -471,7 +471,9 @@ lookup_dbus_name_from_interface(char* dbus_bus_name, const char* interface_name)
 				&& (NULL != item_dbus_name)
 				&& (strcmp(item_interface_name, interface_name) == 0)
 			) {
-				strncpy(dbus_bus_name, item_dbus_name, DBUS_MAXIMUM_NAME_LENGTH);
+				if(strlcpy(dbus_bus_name, item_dbus_name, DBUS_MAXIMUM_NAME_LENGTH) >= DBUS_MAXIMUM_NAME_LENGTH)
+                                    fprintf(stderr,
+				            "Buffer truncation detected for dbus_bus_name\n");
 				ret = 0;
 				break;
 			}
--- a/src/agent/mdns_avahi.cpp
+++ b/src/agent/mdns_avahi.cpp
@@ -550,8 +550,10 @@ otbrError PublisherAvahi::PublishService(uint16_t aPort, const char *aName, cons
 
     {
         Service service;
-        strncpy(service.mName, aName, sizeof(service.mName));
-        strncpy(service.mType, aType, sizeof(service.mType));
+        if(strlcpy(service.mName, aName, sizeof(service.mName)) >= sizeof(service.mName))
+            otbrLog(OTBR_LOG_ERR, "Buffer truncation detected for service.mName");
+        if(strlcpy(service.mType, aType, sizeof(service.mType)) >= sizeof(service.mType))
+            otbrLog(OTBR_LOG_ERR, "Buffer truncation detected for service.mType");
         service.mPort = aPort;
         mServices.push_back(service);
     }
--- a/src/agent/mdns_mdnssd.cpp
+++ b/src/agent/mdns_mdnssd.cpp
@@ -337,8 +337,10 @@ void PublisherMDnsSd::RecordService(const char *aName, const char *aType, DNSSer
     {
         Service service;
 
-        strncpy(service.mName, aName, sizeof(service.mName));
-        strncpy(service.mType, aType, sizeof(service.mType));
+        if(strlcpy(service.mName, aName, sizeof(service.mName)) >= sizeof(service.mName))
+            otbrLog(OTBR_LOG_ERR, "Buffer truncation detected for service.mName");
+        if(strlcpy(service.mType, aType, sizeof(service.mType)) >= sizeof(service.mType))
+            otbrLog(OTBR_LOG_ERR, "Buffer truncation detected for service.mType");
         service.mService = aServiceRef;
         mServices.push_back(service);
     }
--- a/src/commissioner/addr_utils.cpp
+++ b/src/commissioner/addr_utils.cpp
@@ -32,6 +32,7 @@
  */
 
 #include "addr_utils.hpp"
+#include "common/logging.hpp"
 
 #include <string.h>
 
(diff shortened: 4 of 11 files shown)


doanac:
hmm - it doens't look like all your build configurations allow for strlcpy. let me know how you'd like to proceed.

third_party/wpantund/repo/src/wpanctl/wpanctl-utils.c
```diff
@@ -471,7 +471,9 @@ lookup_dbus_name_from_interface(char* dbus_bus_name, const char* interface_name)
 				&& (NULL != item_dbus_name)
 				&& (strcmp(item_interface_name, interface_name) == 0)
 			) {
-				strncpy(dbus_bus_name, item_dbus_name, DBUS_MAXIMUM_NAME_LENGTH);
+				if(strlcpy(dbus_bus_name, item_dbus_name, DBUS_MAXIMUM_NAME_LENGTH) >= DBUS_MAXIMUM_NAME_LENGTH)
```
bukepo: Could you please submit this fix to https://github.com/openthread/wpantund

doanac:
`sizeof-1` is always going to be needed for strncpy. However, I think the most portable fix for this code base might be something slight different than what I've suggested so far. In general, you guys have probably chosen these destination buffer sizes for specific reasons (eg IFNAMESZ). So if we are ever getting truncation, then an assertion is somehow being broken.

With that in mind, I'm thinking the easiest thing might be the following:
1) define a macro (not sure the best header file to use in your code base): 
~~~
#define safe_strncpy(dst, src, dstsiz))\
    do {\
        strncpy(dst, src, dstsiz); \
        if(dst[dstsiz-1] != '\0') abort(); \
    } while(0)
~~~

2) Go through and fix callers of strncpy to pass the proper buffer size *and* use this new macro to give some error detection.

bukepo:
@doanac Sounds good to me. And we must be careful because some buffers length definition includes the null terminator(IFNAMSIZ), some not.

doanac:
@bukepo - thanks. I'll look at that.

One other thing to consider before we proceed too far with this. I haven't looked closely at the usage pattern for these objects, but have you considered using std::string or strdup + freeing memory in destructor for these buffers? ie - we are doing some heavy lifting here that in many cases C++ is okay at handling.

doanac:
An example of a more C++ friendly approach could be:

~~~
diff --git a/src/agent/mdns_avahi.cpp b/src/agent/mdns_avahi.cpp
index 90a6419..1798fa5 100644
--- a/src/agent/mdns_avahi.cpp
+++ b/src/agent/mdns_avahi.cpp
@@ -549,10 +549,7 @@ otbrError PublisherAvahi::PublishService(uint16_t aPort, const char *aName, cons
     SuccessOrExit(error);
 
     {
-        Service service;
-        strncpy(service.mName, aName, sizeof(service.mName));
-        strncpy(service.mType, aType, sizeof(service.mType));
-        service.mPort = aPort;
+        Service service(aName, aType, aPort);
         mServices.push_back(service);
     }
 
diff --git a/src/agent/mdns_avahi.hpp b/src/agent/mdns_avahi.hpp
index 8e47416..eb113fa 100644
--- a/src/agent/mdns_avahi.hpp
+++ b/src/agent/mdns_avahi.hpp
@@ -34,6 +34,9 @@
 #ifndef MDNS_AVAHI_HPP_
 #define MDNS_AVAHI_HPP_
 
+#include <string.h>
+#include <cassert>
+#include <cstdlib>
 #include <vector>
 
 #include <avahi-client/client.h>
@@ -282,9 +285,20 @@ private:
 
     struct Service
     {
-        char     mName[kMaxSizeOfServiceName];
-        char     mType[kMaxSizeOfServiceType];
-        uint16_t mPort;
+        char *mName;
+        char *mType;
+        uint16_t   mPort;
+
+       Service(const char *name, const char *type, uint16_t port) {
+               assert(name != NULL && type != NULL);
+               mName = strdup(name);
+               mType = strdup(type);
+               mPort = port;
+       }
+       ~Service() {
+               free(mName);
+               free(mType);
+       }
     };
 
     typedef std::vector<Service> Services;
~~~

or even cleaner with std::string
~~~
diff --git a/src/agent/mdns_avahi.cpp b/src/agent/mdns_avahi.cpp
index 90a6419..c57f4d6 100644
--- a/src/agent/mdns_avahi.cpp
+++ b/src/agent/mdns_avahi.cpp
@@ -531,8 +531,7 @@ otbrError PublisherAvahi::PublishService(uint16_t aPort, const char *aName, cons
 
     for (Services::iterator it = mServices.begin(); it != mServices.end(); ++it)
     {
-        if (!strncmp(it->mName, aName, sizeof(it->mName)) && !strncmp(it->mType, aType, sizeof(it->mType)) &&
-            it->mPort == aPort)
+        if (it->mName == aName && it->mType == aType && it->mPort == aPort)
         {
             otbrLog(OTBR_LOG_INFO, "MDNS updating service %s...", aName);
             error = avahi_entry_group_update_service_txt_strlst(
@@ -549,10 +548,7 @@ otbrError PublisherAvahi::PublishService(uint16_t aPort, const char *aName, cons
     SuccessOrExit(error);
 
     {
-        Service service;
-        strncpy(service.mName, aName, sizeof(service.mName));
-        strncpy(service.mType, aType, sizeof(service.mType));
-        service.mPort = aPort;
+        Service service(aName, aType, aPort);
         mServices.push_back(service);
     }
 
diff --git a/src/agent/mdns_avahi.hpp b/src/agent/mdns_avahi.hpp
index 8e47416..bae518c 100644
--- a/src/agent/mdns_avahi.hpp
+++ b/src/agent/mdns_avahi.hpp
@@ -34,6 +34,8 @@
 #ifndef MDNS_AVAHI_HPP_
 #define MDNS_AVAHI_HPP_
 
+#include <cassert>
+#include <string>
 #include <vector>
 
 #include <avahi-client/client.h>
@@ -282,9 +284,12 @@ private:
 
     struct Service
     {
-        char     mName[kMaxSizeOfServiceName];
-        char     mType[kMaxSizeOfServiceType];
-        uint16_t mPort;
+        std::string mName;
+        std::string mType;
+        uint16_t   mPort;
+
+       Service(const char *name, const char *type, uint16_t port) : mName(name), mType(type), mPort(port) {}
+       ~Service() {}
     };
 
     typedef std::vector<Service> Services;

~~~

bukepo:
@doanac Thanks for the proposal. It seems most strings used in the agent(`src/agent`) are of expected size limit. Using pre-allocated buffer for these string seems to be simpler because we don't need to handle memory failures caused by `std::string`. The agent is supposed to be used in environments where **c++ exception** is disabled or not recommended(Android as an example). And I'm OK with using `std::string` in `src/web`.

bukepo:
Any update?

doanac:
@bukepo - sorry. i've been pulled into some other stuff lately and haven't had any cycles to look into this. I do hope to get back to fixing this soon. However, if you guys need a fix before that, you might want to re-assign the work.

jwhui:
Resolved by #299.

PR #248, opened 2019-02-15
[agent] add avahi-dbus mDNS backend

bukepo:
This PR adds a new backend for mDNS. It uses the d-bus API provided by avahi-daemon.


bukepo:
Not need any more

bukepo:
Not need any more

PR #252, opened 2019-02-20
[web] fix network prefix with wpantund

gjc13:


bukepo:
could you please also add the missing check for curl failure?

Issue #254, opened 2019-02-22
Steering Data is wrong


kwangeom:
**Describe the bug**
I tested with two devices to join a thread network.
First device was successful in commissioning and then tried commissioning for the second device.
But it does not start DTLS session because of wrong "Steering Data" from "Discovery Response".
From the below log captures, when second device tries commissioning, the steering data sent from NCP looks for the first device.

This is first device wireshark log when success
![image](https://user-images.githubusercontent.com/9152437/53267856-153a6700-36aa-11e9-9620-d0a5113c73b1.png)

This is second device wireshark log when success
![image](https://user-images.githubusercontent.com/9152437/53267939-56cb1200-36aa-11e9-80b7-55414295a752.png)

This is second device wireshark log when fail
![image](https://user-images.githubusercontent.com/9152437/53267796-ea501300-36a9-11e9-83b4-fc86351e7a91.png)

**To Reproduce**
Information to reproduce the behavior, including:
Network looks like this : BBB--NCP(leader)-----CLI MTD(1)
                                                                       |-- CLI MTD(2)
commission first device, and then commission the second device.
try this several times. sometimes second device does not start DTLS session after receiving Discovery Response. Then look into the streering data to see if it is ok.

**Expected behavior**

**Console/log output**
If applicable, add console/log output to help explain your problem.

**Additional context**

jwhui:
Moved this issue to `openthread/borderrouter` project, since this seems to be specific to border router.

bukepo:
@kwangeom It seems you are using a very old version, could you please try the latest version?

jwhui:
@bukepo, I was able to reproduce the behavior using the latest master.

bukepo:
@kwangeom Hi could you try apply openthread/openthread#3623 and #256 and see if they fix the issue you encountered?

kwangeom:
@Yakun Xu I will test this fix and let you know the result tomorrow.

kwangeom:
@bukepo  This fix is working for me. Thanks so much for the quick fix.

PR #256, opened 2019-02-23
[commissioner] ensure close notify on close

bukepo:
A quick fix to #254 when user use Ctrl^C to quit the first commissioner. For longer term, I think it makes sense to add an option that the commissioner quit on joiner session ends.


Gumidefektet:
836#

Issue #274, opened 2019-04-26
[docker] automatically release docker image


bukepo:
**Is your feature request related to a problem? Please describe.**
docker image is out of date, and nobody remembered to update it.

**Describe the solution you'd like**
Use Travis to periodically release new docker image for otbr.

**Describe alternatives you've considered**
Not sure how docker hub will authenticate. Maybe the job should be created in docker hub.

**Additional context**
None.

gjc13:
Reference: https://docs.docker.com/docker-hub/builds/

gjc13:
@jwhui We need to add a bot account to OpenThead team on github and docker hub so that docker hub can be linked to the code repository.

jwhui:
@gjc13, I enabled autobuild on docker hub for this repo. However, the builds are currently failing:
```
/bin/mkdir -p '/app/borderrouter/._stage/usr/include/CppUTest'
/usr/bin/install -c -m 644 include/CppUTest/CommandLineArguments.h include/CppUTest/CommandLineTestRunner.h include/CppUTest/CppUTestConfig.h include/CppUTest/JUnitTestOutput.h include/CppUTest/MemoryLeakDetector.h include/CppUTest/MemoryLeakDetectorMallocMacros.h include/CppUTest/MemoryLeakDetectorNewMacros.h include/CppUTest/MemoryLeakWarningPlugin.h include/CppUTest/PlatformSpecificFunctions.h include/CppUTest/PlatformSpecificFunctions_c.h include/CppUTest/SimpleString.h include/CppUTest/SimpleMutex.h include/CppUTest/StandardCLibrary.h include/CppUTest/TeamCityTestOutput.h include/CppUTest/TestFailure.h include/CppUTest/TestFilter.h include/CppUTest/TestHarness.h include/CppUTest/TestHarness_c.h include/CppUTest/TestMemoryAllocator.h include/CppUTest/TestOutput.h include/CppUTest/TestPlugin.h include/CppUTest/TestRegistry.h include/CppUTest/TestResult.h include/CppUTest/TestTestingFixture.h include/CppUTest/Utest.h include/CppUTest/UtestMacros.h generated/CppUTestGeneratedConfig.h '/app/borderrouter/._stage/usr/include/CppUTest'
/bin/mkdir -p '/app/borderrouter/._stage/usr/include/CppUTestExt'
[...]

PR #276, opened 2019-04-29
[docker] use git clone rather than copy

gjc13:
Fix https://github.com/openthread/ot-br-posix/issues/274


gjc13:
Duplicate with https://github.com/openthread/ot-br-posix/pull/277

PR #277, opened 2019-04-29
[script] fix script for docker

bukepo:
It seems docker hub builder does not recognize `[`.


bukepo:
@jwhui I moved the Dockerfile to `etc/docker/Dockerfile`.

PR #286, opened 2019-05-06
[docker] install git in Dockerfile

bukepo:
git should be installed before running `script/*`.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -105,7 +105,7 @@ install_packages_rpm()
     else
         PM=yum
     fi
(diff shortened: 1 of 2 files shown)


script/bootstrap
```diff
@@ -105,7 +105,7 @@ install_packages_rpm()
     else
         PM=yum
     fi
-    sudo $PM install -y gcc gcc-c++ libtool automake autoconf autoconf-archive git
```
jwhui: Should we also remove `git` from `install_packages_apt()`?

PR #288, opened 2019-05-08
[docker] remove src directory to reduce size

gjc13:


etc/docker/Dockerfile
```diff
@@ -39,8 +39,17 @@ WORKDIR /app
...
+RUN rm -rf /app/._build/
 RUN ./script/bootstrap && ./script/setup
 RUN chmod 644 /etc/bind/named.conf.options
+RUN mv /app/script /tmp
+RUN mv /app/etc /tmp
+RUN rm -rf /app/*
+RUN mv /tmp/script /app
+RUN mv /tmp/etc /app
+RUN mkdir -p /app/src/agent
+RUN touch /app/src/agent/main.cpp
```
bukepo: remove this.

etc/docker/Dockerfile
```diff
@@ -39,8 +39,17 @@ WORKDIR /app
 RUN apt-get update
 RUN apt-get install -y git lsb-core sudo
 
+RUN git reset --hard && git clean -xfd
+RUN rm -rf /app/._build/
```
bukepo: will this be cleaned by `git clean -dfx`?

etc/docker/Dockerfile
```diff
@@ -39,8 +39,17 @@ WORKDIR /app
...
 RUN ./script/bootstrap && ./script/setup
 RUN chmod 644 /etc/bind/named.conf.options
+RUN mv /app/script /tmp
```
bukepo: avoid use absolute path? we already set work dir.

etc/docker/Dockerfile
gjc13: No, it has to be manually removed.

PR #299, opened 2019-06-06
[misc] use safe wrapped version of strcpy instead

gjc13:
Fixes https://github.com/openthread/ot-br-posix/issues/243

### DIFF
--- a/src/utils/strcpy_utils.cpp
+++ b/src/utils/strcpy_utils.cpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2017, The OpenThread Authors.
+ *  Copyright (c) 2019, The OpenThread Authors.
  *  All rights reserved.
  *
  *  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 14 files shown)


src/utils/strcpy_utils.hpp
```diff
@@ -0,0 +1,47 @@
...
+#ifndef OTBR_STRCPY_UTILS_HPP
+#define OTBR_STRCPY_UTILS_HPP
+
+#include <string.h>
+
+/**
+ * This method is a safe version of strncpy always ensuring last byte to be 0
+ * If strlen(aSrc) >= n, the string will be truncated
+ *
+ * @param[in]  aDest      dest string
```
bukepo: This should be `[out]`?

src/utils/strcpy_utils.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * @param[in]  aSrc       src string
+ * @param[in]  aDestSize  size of dest string buffer
+ *
+ * @returns 0 on success, -1 on error
+ *
+ */
+extern "C" int strncpy_safe(char *aDest, const char *aSrc, size_t aDestSize);
```
bukepo: How about implementing a standard function, either `strcpy_s` or `strncpy_s`, so that once we switched to `C11`, we can simply rename the function.

src/utils/strcpy_utils.cpp
```diff
@@ -0,0 +1,43 @@
+/*
+ *  Copyright (c) 2017-2019, The OpenThread Authors.
```
bukepo: only `2019`?

src/utils/strcpy_utils.hpp
gjc13: Already implemented to be compatible with `strcpy_s`, use this name just to avoid potential linking issues.

src/utils/strcpy_utils.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "strcpy_utils.hpp"
+#include "common/code_utils.hpp"
+
+int strcpy_safe(char *aDest, size_t aDestSize, const char *aSrc)
+{
+    int ret = 0;
+    VerifyOrExit(aDest != NULL, ret = -1);
+    VerifyOrExit(aSrc != NULL, ret = -1, aDest[0] = 0);
+    VerifyOrExit(aDestSize > strnlen(aSrc, aDestSize), ret = -1, aDest[0] = 0);
```
bukepo: missing check aDestSize is 0.

src/utils/strcpy_utils.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * If strlen(aSrc) >= n, the string will be truncated
+ *
+ * @param[out] aDest      dest string
+ * @param[in]  aDestSize  size of dest string buffer
+ * @param[in]  aSrc       src string
+ *
+ * @returns 0 on success, -1 on error
+ *
+ */
+extern "C" int strcpy_safe(char *aDest, size_t aDestSize, const char *aSrc);
```
bukepo: use the common style?
```cpp
#ifdef __cplusplus__
extern "C" {
#endif
```

PR #304, opened 2019-06-19
[mdns] mojo test stubs for Android.mk

gjc13:
Add test stubs for Android.mk build test.
Set mojo as default mDNS implementation in Android.mk.


etc/libchrome_stub/Android.mk
```diff
@@ -0,0 +1,9 @@
+include $(CLEAR_VARS)
```
bukepo: missing license header

PR #312, opened 2019-07-12
[build] fix build without dbus

huamenggg:
This commit fix the case when it won't use dbus.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -420,7 +420,7 @@ AC_MSG_RESULT([$ac_cv_path_CTAGS])
 
 # Check if the DBus module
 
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -420,7 +420,9 @@ AC_MSG_RESULT([$ac_cv_path_CTAGS])
 
 # Check if the DBus module
 
-PKG_CHECK_MODULES(DBUS, [dbus-1 >= 1.4], , [AC_MSG_ERROR([could not find dbus(>= 1.4)])])
+if test "${with_mdns}" != none; then
```
bukepo: also detect if `${with_ncp} != wpantund`

PR #314, opened 2019-07-19
[build] add package otbr onto OpenWRT files.

huamenggg:
This commit adds three files for packaging otbr onto OpenWRT.
README.md is the instruction of operations.
feeds.conf is feed file for OpenWRT.
Makefile is to package the otbr project.

### DIFF
--- a/openthread/Makefile
+++ b/openthread/Makefile
@@ -1,0 +1,51 @@
+include $(TOPDIR)/rules.mk
+
+# Name, version and release number
+# The name and version of your package are used to define the variable to point to the build directory of your package: $(PKG_BUILD_DIR)
+PKG_NAME:=openthread
+PKG_VERSION:=1.0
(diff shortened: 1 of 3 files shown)


bukepo:
I tend to create a orphan branch for this, @jwhui Thoughts?

jwhui:
I think an orphan branch makes sense in this case.

bukepo:
@jwhui Just created the orphan branch. https://github.com/openthread/ot-br-posix/tree/openwrt/openthread

bukepo:
@jwhui any comments?

jwhui:
@bukepo, I reviewed this PR and it looks fine. However, it currently targets master branch rather than openwrt branch. Is that intentional?

huamenggg:
Wrong to push to master branch.

PR #325, opened 2019-09-05
[coap] fix write outside buffer bounds

kbogucki:
Lack of (unsigned char *) cast a pointer to structure pdu->hdr is
causing writes outside CoAP PDU buffer bounds.

### DIFF
--- a/src/common/coap_libcoap.cpp
+++ b/src/common/coap_libcoap.cpp
@@ -181,7 +181,7 @@ otbrError AgentLibcoap::Send(Message &       aMessage,
         VerifyOrExit(pdu->length + sizeof(meta) < pdu->max_size, errno = EMSGSIZE);
 
         tid = coap_send_confirmed(&mCoap, mCoap.endpoint, &remote, pdu);
-        memcpy(pdu->hdr + pdu->length, &meta, sizeof(meta));
+        memcpy((unsigned char *)pdu->hdr + pdu->length, &meta, sizeof(meta));
     }
     else
     {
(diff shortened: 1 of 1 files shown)


src/common/coap_libcoap.cpp
```diff
@@ -181,7 +181,7 @@ otbrError AgentLibcoap::Send(Message &       aMessage,
         VerifyOrExit(pdu->length + sizeof(meta) < pdu->max_size, errno = EMSGSIZE);
 
         tid = coap_send_confirmed(&mCoap, mCoap.endpoint, &remote, pdu);
-        memcpy(pdu->hdr + pdu->length, &meta, sizeof(meta));
+        memcpy((unsigned char *)pdu->hdr + pdu->length, &meta, sizeof(meta));
```
bukepo: nit: use `static_cast<uint8_t*>(pdu->hdr)`?

src/common/coap_libcoap.cpp
```diff
@@ -269,7 +269,7 @@ void AgentLibcoap::HandleResponse(coap_context_t *       aCoap,
 
     VerifyOrExit(aSent != NULL, otbrLog(OTBR_LOG_ERR, "request not found!"));
 
-    memcpy(&meta, aSent->hdr + aSent->length, sizeof(meta));
+    memcpy(&meta, (unsigned char *)aSent->hdr + aSent->length, sizeof(meta));
```
bukepo: nit: use `static_cast<uint8_t*>(pdu->hdr)`?

src/common/coap_libcoap.cpp
kbogucki: I used (unsigned char *) because this is how it's implemented in libcoap in exactly the same cases.

src/common/coap_libcoap.cpp
kbogucki: I used (unsigned char *) because this is how it's implemented in libcoap in exactly the same cases.

Issue #328, opened 2019-09-13
[agent] mojo publishes when Thread is not started


bukepo:
**Describe the bug**
mojo publishes when Thread is not started.

**To Reproduce**
1. Stop wpantund or `wpanctl leave`
2. start otbr-agent
3. use `avahi-browse -arp` to check services published by the DUT

**Expected behavior**
No service being published.

**Console/log output**
Thread is published with network empty and extended pan id all zero.

**Additional context**
None

jwhui:
Resolved by #329.

PR #329, opened 2019-09-16
[border-agent] fix mDNS with mojo initialization issue (#328)

gjc13:
Add retry timeout for connecting to mojo broker
Fix initialization issue when wpantund is not launched

### DIFF
--- a/src/agent/mdns_mojo.cpp
+++ b/src/agent/mdns_mojo.cpp
@@ -90,7 +90,6 @@ MdnsMojoPublisher::MdnsMojoPublisher(StateHandler aHandler, void *aContext)
     , mContext(aContext)
     , mStarted(false)
 {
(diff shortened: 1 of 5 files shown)


src/agent/mdns_mojo.cpp
```diff
@@ -125,14 +127,17 @@ void MdnsMojoPublisher::mMojoDisconnectedCb(void)
 
 otbrError MdnsMojoPublisher::Start(void)
 {
-    otbrError err = OTBR_ERROR_NONE;
-
-    VerifyOrExit(mConnector != nullptr, err = OTBR_ERROR_MDNS);
+    if (mConnector != nullptr)
+    {
+        mStateHandler(mContext, kStateReady);
```
bukepo: move this to MdnsMojoPublisher::mMojoConnectCb()?

gjc13: BorderAgent will call Start() itself when mojo is connected so we have to call handler here.

PR #330, opened 2019-09-18
[mojo] check broker access before connection

gjc13:


src/agent/mdns_mojo.cpp
```diff
@@ -39,6 +39,8 @@
 #include <mojo/core/embedder/embedder.h>
 #include <mojo/core/embedder/scoped_ipc_support.h>
 
+#include <unistd.h>
```
jwhui: style: include C standard library headers before third-party library headers.

PR #331, opened 2019-09-18
[dbus] Add dbus access for non-root users

gjc13:


Android.mk
```diff
@@ -141,6 +141,6 @@ LOCAL_MODULE_TAGS := eng
 LOCAL_MODULE_PATH := $(TARGET_OUT_ETC)/dbus-1/system.d
 LOCAL_SRC_FILES := src/agent/otbr-agent.conf
 $(LOCAL_PATH)/src/agent/otbr-agent.conf: $(LOCAL_PATH)/src/agent/otbr-agent.conf.in
-	sed 's/@OTBR_AGENT_USER@/root/g' $< > $@
+	sed 's/@OTBR_AGENT_USER@/wpan/g' $< > $@
```
bukepo: is it possible to get the user `wpan` from build environment?

PR #335, opened 2019-09-25
[build] fix on macOS

bukepo:
This PR fixed building otbr-agent on macOS and added a travis check for it.


bukepo:
@gjc13 I think you should install it, because it's required for compiling libcoap.

.travis/before_install.sh
```diff
@@ -150,6 +150,10 @@ linux)
     }
     ;;
 
+osx)
+    brew install dbus ctags
```
jwhui: Should this be included in `script/bootstrap`?

PR #338, opened 2019-10-09
[mdns] include thread version in query response

wgtdkp:
This PR includes the `Thread Version` is a DNS query response. According to the spec, `Thread Version` is **required**.

Currently supports only `ncp=openthread`.

Dependent PR: [4225](https://github.com/openthread/openthread/pull/4225).

This PR updated submodule openthread to include the dependent PR.

### DIFF
--- a/src/agent/border_agent.cpp
+++ b/src/agent/border_agent.cpp
@@ -136,6 +136,8 @@ otbrError BorderAgent::Start(void)
 #if OTBR_ENABLE_MDNS_AVAHI || OTBR_ENABLE_MDNS_MDNSSD || OTBR_ENABLE_MDNS_MOJO
     SuccessOrExit(error = mNcp->RequestEvent(Ncp::kEventNetworkName));
     SuccessOrExit(error = mNcp->RequestEvent(Ncp::kEventExtPanId));
+
+// Currently supports only NCP_OPENTHREAD
 #if OTBR_ENABLE_NCP_OPENTHREAD
     SuccessOrExit(error = mNcp->RequestEvent(Ncp::kEventThreadVersion));
 #endif // OTBR_ENABLE_NCP_OPENTHREAD
--- a/src/agent/border_agent.cpp
+++ b/src/agent/border_agent.cpp
@@ -429,7 +429,7 @@ void BorderAgent::HandleThreadVersion(void *aContext, int aEvent, va_list aArgum
     assert(aEvent == Ncp::kEventThreadVersion);
 
     // `uint16_t` has been promoted to `int`.
-    uint16_t threadVersion = va_arg(aArguments, int);
+    uint16_t threadVersion = static_cast<uint16_t>(va_arg(aArguments, int));
     static_cast<BorderAgent *>(aContext)->SetThreadVersion(threadVersion);
 }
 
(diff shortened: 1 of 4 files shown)


src/agent/border_agent.cpp
```diff
@@ -410,6 +424,28 @@ void BorderAgent::HandleExtPanId(void *aContext, int aEvent, va_list aArguments)
...
+}
+
+const char *BorderAgent::ThreadVersionToString(uint16_t aThreadVersion)
+{
+    switch (aThreadVersion)
+    {
+    case 2:
+        return "1.1.1";
+    case 3:
+        return "1.2.0";
```
librasungirl: what if there is '1.2.1' someday? how to differentiate it from the same version value 3?

wgtdkp: see the [JIRA](https://threadgroup.atlassian.net/browse/BHC-541?atlOrigin=eyJpIjoiZTJlZjczMTVkNjgxNDgwYzhmMzY4ZjQyZjgyYjFlN2MiLCJwIjoiaiJ9) about this.

If the mDNS query response truely needs the `MAJOR.MINOR.PATCH` format of the Thread version, we need new API from OpenThread to provide this version string.

@jwhui

wgtdkp: I think it is OK to always return "1.2.0" if we can't decide the `PATCH`.

src/agent/border_agent.hpp
```diff
@@ -128,24 +128,28 @@ class BorderAgent
...
 #if OTBR_ENABLE_NCP_WPANTUND
     int mSocket;
 #endif
-    uint8_t mExtPanId[kSizeExtPanId];
-    char    mNetworkName[kSizeNetworkName + 1];
-    bool    mThreadStarted;
-    bool    mPSKcInitialized;
+    uint8_t  mExtPanId[kSizeExtPanId];
+    char     mNetworkName[kSizeNetworkName + 1];
+    uint16_t mThreadVersion;
```
bukepo: move this after line 151 so that we can have better alignment.

src/agent/border_agent.cpp
```diff
@@ -410,6 +428,28 @@ void BorderAgent::HandleExtPanId(void *aContext, int aEvent, va_list aArguments)
...
+void BorderAgent::HandleThreadVersion(void *aContext, int aEvent, va_list aArguments)
+{
+    assert(aEvent == Ncp::kEventThreadVersion);
+
+    // `uint16_t` has been promoted to `int`.
+    uint16_t threadVersion = static_cast<uint16_t>(va_arg(aArguments, int));
+    static_cast<BorderAgent *>(aContext)->SetThreadVersion(threadVersion);
+}
+
+const char *BorderAgent::ThreadVersionToString(uint16_t aThreadVersion)
```
bukepo: I don't think this is necessary to be member method of BorderAgent. How about make this a local function?

wgtdkp: this is a stateless static member function.

src/agent/border_agent.hpp
wgtdkp: Does alignment matters here? `mThreadVersion` is placed near to `mNetworkName` because they are both properties requested from NCP/RCP device.

I think *place related fields together* is more meaningful than runtime size.
thoughts?

src/agent/border_agent.cpp
bukepo: use static local function, we reduce a symbol exposed to outside the class. And I think it can be compiled faster.

PR #339, opened 2019-10-11
[script] avoid bash extension

bukepo:
This PR removes dependency on `[[`, also fixed the broken travis test of **raspbian jessie**.


bukepo:
@jwhui current master travis test is broken, this PR can fix it.

PR #340, opened 2019-10-15
[travis] allow macOS check fail

bukepo:
travis-ci's macOS build system is broken when installing packages.
Revert after it recovers.


jwhui:
@bukepo, I just submitted #341 which seems to fix the macOS issue on Travis.

bukepo:
Thank you!

PR #343, opened 2019-11-04
[mdns] don't include thread version if it is not available

wgtdkp:


src/agent/border_agent.cpp
```diff
@@ -285,7 +285,7 @@ static const char *ThreadVersionToString(uint16_t aThreadVersion)
     case 3:
         return "1.2.0";
     default:
-        return "";
+        return NULL;
```
bukepo: assert here?

src/agent/border_agent.cpp
```diff
@@ -295,8 +295,19 @@ void BorderAgent::PublishService(void)
...
     assert(mNetworkName[0] != '\0');
     Utils::Bytes2Hex(mExtPanId, sizeof(mExtPanId), xpanid);
-    mPublisher->PublishService(kBorderAgentUdpPort, mNetworkName, kBorderAgentServiceType, "nn", mNetworkName, "xp",
-                               xpanid, "tv", ThreadVersionToString(mThreadVersion), NULL);
+
+#if OTBR_ENABLE_NCP_OPENTHREAD
+    if (ThreadVersionToString(mThreadVersion) != NULL)
+    {
+        mPublisher->PublishService(kBorderAgentUdpPort, mNetworkName, kBorderAgentServiceType, "nn", mNetworkName, "xp",
+                                   xpanid, "tv", ThreadVersionToString(mThreadVersion), NULL);
```
bukepo: The issue happened in this mode.

src/agent/border_agent.cpp
wgtdkp: `ThreadVersion` is typically requested from thread devices. Assert here is assuming a thread device will always return thread version of 2 or 3 and we shouldn't make this assumption.

wgtdkp:
@bukepo travis fixed.

PR #326, opened 2019-09-06
[openwrt] initial implementation

huamenggg:
This commit submits the initial implementation of ot-br-posix on OpenWRT.
Implemented functions are as follows:
* overview page
* setting page
* scan page
* view page
* add joiner page
* attach page

PR #345, opened 2019-11-06
[openwrt] Fix memory issues

gjc13:
Please review after https://github.com/openthread/ot-br-posix/pull/326 is merged.

### DIFF
--- a/src/agent/otubus.cpp
+++ b/src/agent/otubus.cpp
@@ -56,20 +56,19 @@ namespace BorderRouter {
 namespace ubus {
 static UbusServer *sUbusServerInstance = NULL;
 static int         sUbusEfd            = -1;
+// XXX: do not use static variable to pass parameters
 static void *      sJsonUri            = NULL;
 static int         sBufNum;
 static std::mutex *sNcpThreadMutex;
 
-const static int PANID_LENGTH     = 10;
-const static int XPANID_LENGTH    = 64;
-const static int MASTERKEY_LENGTH = 64;
+const static int PANID_LENGTH     = 10 + 1;
+const static int XPANID_LENGTH    = 64 + 1;
+const static int MASTERKEY_LENGTH = 64 + 1;
 
 UbusServer::UbusServer(Ncp::ControllerOpenThread *aController)
 {
     mController = aController;
     mSecond     = 0;
-    blob_buf_init(&mNetworkdataBuf, 0);
-    blob_buf_init(&mBuf, 0);
 }
 
 UbusServer &UbusServer::GetInstance(void)
@@ -236,7 +235,7 @@ void UbusServer::ProcessScan(void)
 
     sNcpThreadMutex->lock();
     SuccessOrExit(error = otLinkActiveScan(mController->GetInstance(), scanChannels, scanDuration,
-                                           &UbusServer::HandleActiveScanResult, this));
+                                           &UbusServer::HandleActiveScanResult, NULL));
 exit:
     sNcpThreadMutex->unlock();
     return;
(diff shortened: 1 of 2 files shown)


src/agent/otubus.cpp
```diff
@@ -1198,10 +1224,12 @@ int UbusServer::UbusGetInformation(struct ubus_context *     aContext,
             char                multicastAddr[10] = "ff03::2";
             long                value;
 
+            memset(&mNetworkdataBuf, 0, sizeof(mNetworkdataBuf));
             blob_buf_init(&mNetworkdataBuf, 0);
 
             SuccessOrExit(error = otIp6AddressFromString(multicastAddr, &address));
 
+            // XXX: cannot understand what this magic number refers to
```
huamenggg: This is the TLV type value, 5 is Route64 TLV (Thread1.1 spec 5-65), 16 is Child Table TLV (Thread 1.1 spec 10-26), otThreadSendDiagnosticGet() function needs parameters of tlv types.

src/agent/otubus.cpp
```diff
@@ -322,29 +326,26 @@ int UbusServer::UbusScanHandlerDetail(struct ubus_context *     aContext,
...
-    blob_buf_init(&mBuf, 0);
-    sJsonUri = blobmsg_open_array(&mBuf, "scan_list");
+    memset(&mScanReplyBuf, 0, sizeof(mScanReplyBuf));
+    blob_buf_init(&mScanReplyBuf, 0);
+    sJsonUri = blobmsg_open_array(&mScanReplyBuf, "scan_list");
 
     mIfFinishScan = 0;
     sUbusServerInstance->ProcessScan();
 
+    // XXX: Is there a neccessity to manually trigger openthread event loop?
```
huamenggg: If not trigger, the openthread event loop won't run.. (But this is the previous situation, maybe now is OK, needs some experiments).

src/agent/otubus.cpp
```diff
@@ -56,20 +56,19 @@ namespace BorderRouter {
...
-const static int MASTERKEY_LENGTH = 64;
+const static int PANID_LENGTH     = 10 + 1;
+const static int XPANID_LENGTH    = 64 + 1;
+const static int MASTERKEY_LENGTH = 64 + 1;
 
 UbusServer::UbusServer(Ncp::ControllerOpenThread *aController)
 {
     mController = aController;
     mSecond     = 0;
-    blob_buf_init(&mNetworkdataBuf, 0);
```
huamenggg: The init of mNetworkdataBuf is necessary, if not, the first time call "networkdata" function on OpenWRT, it will crash.
The logic of "networkdata" function is: 
    1. send reply through "ubus_send_reply(aContext, aReq, mNetworkdataBuf.head);" (if not init, this will crash)
    2. doing DiagnosticGet function to update the current networkdata. (will cost some time)
    3. Next time ask for networkdata, it will send the latest networkdata.

gjc13: I've moved the initialization to the request handling part, see https://github.com/openthread/ot-br-posix/pull/345/files#diff-44868e5af116c945b79bc922aa59c74fR1348

huamenggg: Sorry for ignore that, but in my opinion, the first time to call "networkdata" might still cause crash. The process of "networkdata" is UbusNetworkdataHandler() -> UbusGetInformation() -> "networkdata" branch [https://github.com/openthread/ot-br-posix/blob/master/src/agent/otubus.cpp#L1190](url) -> send mNetworkdataBuf.head (might crash)

gjc13: I've added a guard in the last patch.

PR #347, opened 2019-11-21
[Utils] Add concurrent queue class

gjc13:
This PR includes the concurrent queue template and its unit test. It
also fixes some build errors in the unit test module.

The concurrent queue class will be the task queue between the IPC
modules and the OpenThread core stack.

### DIFF
--- a/tests/unit/test_event_emitter.cpp
+++ b/tests/unit/test_event_emitter.cpp
@@ -26,11 +26,11 @@
  *    POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <CppUTest/TestHarness.h>
-
 #include <stdarg.h>
 
 #include "common/event_emitter.hpp"
+
+#include <CppUTest/TestHarness.h>
 
 static int   sCounter = 0;
 static int   sEvent   = 0;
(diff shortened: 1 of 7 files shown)


tests/unit/test_event_emitter.cpp
```diff
@@ -26,12 +26,12 @@
  *    POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <CppUTest/TestHarness.h>
-
 #include <stdarg.h>
 
 #include "common/event_emitter.hpp"
 
+#include <CppUTest/TestHarness.h>
```
gjc13: This line is moved intentionally to fix a build error, see https://stackoverflow.com/questions/54611227/compilation-error-after-including-jsoncpp

tests/unit/test_logging.cpp
```diff
@@ -85,7 +85,7 @@ TEST(Logging, TestLoggingLowerLevel)
 TEST(Logging, TestLoggingDump)
 {
     char ident[120];
-    char cmd[128];
+    char cmd[256];
```
gjc13: The size is modified to avoid a gcc warning about buffer overflow.

gjc13:
For the time being this patch is no longer needed.

PR #357, opened 2020-01-06
[build] recover code coverage

bukepo:
This PR recovers code coverage of this project.

Changes:
* Added the missing after_success.sh script.
* Added support for building out of source directory.
* Changed to use `byteswap.h` for swap64.
* Changed to use absolution path in `.gcno` so that sources can always be found.
* Enhanced to exit programs elegantly.
* Added `--enable-debug` when COVERAGE=1.
* Ignored third_party code in code coverage report.
* No failure for code coverage drop.


bukepo:
This depends on #358

.travis/after_success.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+#
+#  Copyright (c) 2019, The OpenThread Authors.
```
bukepo: This is borrowed from OpenThread, so kept 2019

bukepo:
@jwhui any comments? I know the current code coverage is still not good enough, but I hope to make this a new starting point. Otherwise, the PR will become more and more complex and hard to review.

PR #360, opened 2020-01-10
[dbus] introspect xml

gjc13:
This change contains a preview of Thread network management d-bus api.

PR #361, opened 2020-01-13
[dbus] property interface for dbus server

gjc13:
This change introduces the helper functions to implement standard d-bus
property apis.


src/dbus/constants.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "dbus/constants.hpp"
+
+#include <dbus/dbus.h>
+
+namespace otbr {
+namespace dbus {
+
+const std::string kDBusPropertyInterface = DBUS_INTERFACE_PROPERTIES;
+
+const std::string kDBusPropertyGetMethod          = "Get";
```
bukepo: I don't think these definitions will be used by other objects, how about moving these definitions into the DBusObject?

src/dbus/dbus_object.cpp
```diff
@@ -52,18 +54,41 @@ otbrError DBusObject::Init(void)
...
 {
     std::string fullPath = aInterfaceName + "." + aMethodName;
 
-    assert(mHandlers.find(fullPath) == mHandlers.end());
-    mHandlers.emplace(fullPath, aHandler);
+    assert(mMethodHandlers.find(fullPath) == mMethodHandlers.end());
+    mMethodHandlers.emplace(fullPath, aHandler);
+}
+
+void DBusObject::RegisterGetPropertyHandler(const std::string &        aInterfaceName,
```
bukepo:
```suggestion
void DBusObject::RegisterProperty(const std::string &aInterfaceName, const std::string &aPropertyName, const PropertyHandlerType &aGetHandler, PropertyHandlerType &aSetHandler);
```
And define a special NullHandler which means not supported? Thus, the object knows how to generate introspect.

src/dbus/dbus_object.cpp
```diff
@@ -81,15 +106,111 @@ DBusHandlerResult DBusObject::MessageHandler(DBusConnection *aConnection, DBusMe
...
+    VerifyOrExit(reply != nullptr);
+    VerifyOrExit(dbus_message_iter_init(aRequest.GetMessage().GetRaw(), &iter));
+    SuccessOrExit(DBusMessageExtract(&iter, interfaceName));
+    SuccessOrExit(DBusMessageExtract(&iter, propertyName));
+    propertyFullPath = interfaceName + "." + propertyName;
+
+    if (mSetPropertyHandlers.find(propertyFullPath) != mSetPropertyHandlers.end())
+    {
+        mSetPropertyHandlers.at(propertyFullPath)(iter);
+    }
```
bukepo: reply proper error messages, see: https://dbus.freedesktop.org/doc/api/html/group__DBusMessage.html#ga2ab896965aec97fb21293affeed36232

src/dbus/constants.cpp
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#include "dbus/constants.hpp"
+
+#include <dbus/dbus.h>
+
+namespace otbr {
+namespace dbus {
+
+const std::string kDBusPropertyInterface = DBUS_INTERFACE_PROPERTIES;
```
Moandor-y: I don't think defining `std::string` with static storage duration is a good idea. Maybe change them to `const char* const`?

src/dbus/error.cpp
```diff
@@ -0,0 +1,146 @@
...
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dbus/error.hpp"
+#include "dbus/constants.hpp"
+#include "dbus/dbus_message_helper.hpp"
+
+#include <unordered_map>
+
+static std::unordered_map<otError, std::string> sErrorNameMap = {
```
Moandor-y: The same as [this](https://github.com/openthread/ot-br-posix/pull/361#discussion_r366684362). Static `std::unordered_map`s are bad. It's better to use `new` to create a map and never `delete` it, in the form `static T& t = *new T;`

Moandor-y: https://github.com/openthread/ot-br-posix/blob/363feac43392b9655c86aaa2ee804cbfcacb4b6c/src/dbus/error.cpp#L72
This needs to be changed too.

src/dbus/dbus_message_helper.hpp
```diff
@@ -275,6 +275,35 @@ constexpr otbrError ConvertToTuple(DBusMessageIter *aIter, std::tuple<FieldTypes
...
+        err = OTBR_ERROR_DBUS);
+
+    SuccessOrExit(err = DBusMessageEncode(&subIter, aValue));
+
+    VerifyOrExit(dbus_message_iter_close_container(aIter, &subIter), err = OTBR_ERROR_DBUS);
+exit:
+    return err;
+}
+
+template <typename ValType> otbrError DBusMessageExtractFromVariant(DBusMessageIter *aIter, ValType &aValue)
```
bukepo:
```suggestion
template <typename ValueType> otbrError DBusMessageExtractFromVariant(DBusMessageIter *aIter, ValueType &aValue)
```

src/dbus/dbus_message_helper.cpp
```diff
@@ -27,17 +27,32 @@
...
+}
+
+otbrError DBusMessageExtract(DBusMessageIter *aIter, int8_t &aValue)
 {
     otbrError err = OTBR_ERROR_NONE;
     uint8_t   val;
 
     SuccessOrExit(err = DBusMessageExtract(aIter, val));
-    aValue = (val != 0);
+    aValue = static_cast<int8_t>(aValue);
```
Moandor-y: `aValue` is already `int8_t`. Casting is not needed.

Moandor-y:
```suggestion
    aValue = static_cast<int8_t>(val);
```
Maybe you mean this?

src/dbus/dbus_object.cpp
```diff
@@ -33,6 +33,8 @@
 
 #include "dbus/dbus_object.hpp"
 
+using namespace std::placeholders;
```
Moandor-y: [Do not use using-directives](https://google.github.io/styleguide/cppguide.html#Namespaces)

src/dbus/error.cpp
```diff
@@ -0,0 +1,121 @@
...
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dbus/error.hpp"
+
+#include "dbus/constants.hpp"
+#include "dbus/dbus_message_helper.hpp"
+
+#include <unordered_map>
```
bukepo:
```suggestion
```

src/dbus/error.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+static const std::pair<otError, const char *> sErrorNameMap[] = {
+    {OT_ERROR_GENERIC, "io.openthread.Error.GenericError"},
```
bukepo:
```suggestion
    {OT_ERROR_GENERIC, "io.openthread.Error.Generic"},
```

src/dbus/dbus_request.hpp
```diff
@@ -40,17 +41,26 @@ class DBusRequest
...
     /**
      * This method returns the message sent to call the d-bus method.
      *
      * @returns   The dbus message
      */
-    SharedDBusMessage GetMessage(void) { return mMessage; }
+    UniqueDBusMessage const &GetMessage(void) { return mMessage; }
+
+    /**
+     * This method returns underlying d-bus connection
```
bukepo: Missing `.`.

src/dbus/dbus_request.hpp
```diff
@@ -40,17 +41,26 @@ class DBusRequest
...
+     *
+     * @returns   The dbus connection
```
bukepo: Missing `.`

src/dbus/dbus_object.hpp
```diff
@@ -79,7 +82,31 @@ class DBusObject
...
      */
-    void RegisterMethod(const std::string &aInterfaceName, const std::string &aMethodName, const HandlerType &aHandler);
+    void RegisterMethod(const std::string &      aInterfaceName,
+                        const std::string &      aMethodName,
+                        const MethodHandlerType &aHandler);
+
+    /**
+     * This method the get handler for a property
+     *
+     * @param[in]   aInterfaceName    The interface name
```
bukepo: Missing `.`

src/dbus/dbus_object.cpp
```diff
@@ -81,15 +106,109 @@ DBusHandlerResult DBusObject::MessageHandler(DBusConnection *aConnection, DBusMe
     std::string       memberName = interface + "." + dbus_message_get_member(aMessage);
 
     if (dbus_message_get_type(aMessage) == DBUS_MESSAGE_TYPE_METHOD_CALL &&
-        mHandlers.find(memberName) != mHandlers.end())
+        mMethodHandlers.find(memberName) != mMethodHandlers.end())
     {
-        mHandlers.at(memberName)(request);
+        mMethodHandlers.at(memberName)(request);
```
Moandor-y: Changing to
```
auto iter = mMethodHandlers.find(memberName);
if (/*...*/ && iter != mMethodHandlers.end()) {
    (*iter)(request);
```
can save one hash table lookup.

src/dbus/error.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    for (auto &&p : sErrorNameMap)
+    {
+        if (p.second == aErrorName)
+        {
+            err = p.first;
+        }
+    }
+    return err;
+}
```
Moandor-y: Linear search is slow. Better to use 2 `unordered_map`s here.

src/dbus/dbus_object.cpp
```diff
@@ -79,17 +104,117 @@ DBusHandlerResult DBusObject::MessageHandler(DBusConnection *aConnection, DBusMe
...
+{
+    UniqueDBusMessage reply = UniqueDBusMessage(dbus_message_new_method_return(aRequest.GetMessage().get()));
+
+    DBusMessageIter iter;
+    std::string     interfaceName;
+    std::string     propertyName;
+    otError         err          = OT_ERROR_NONE;
+    auto            propertyIter = mGetPropertyHandlers.find(interfaceName);
+
+    VerifyOrExit(reply != nullptr, err = OT_ERROR_FAILED);
```
bukepo:
```suggestion
    VerifyOrExit(reply != nullptr, err = OT_ERROR_NO_BUFS);
```

src/dbus/dbus_object.cpp
```diff
@@ -79,17 +104,117 @@ DBusHandlerResult DBusObject::MessageHandler(DBusConnection *aConnection, DBusMe
...
+    VerifyOrExit(dbus_message_iter_init(aRequest.GetMessage().get(), &iter), err = OT_ERROR_FAILED);
+    VerifyOrExit(DBusMessageExtract(&iter, interfaceName) == OTBR_ERROR_NONE, err = OT_ERROR_FAILED);
+    VerifyOrExit(DBusMessageExtract(&iter, propertyName) == OTBR_ERROR_NONE, err = OT_ERROR_FAILED);
+
+    VerifyOrExit(propertyIter != mGetPropertyHandlers.end(), err = OT_ERROR_NOT_FOUND);
```
bukepo: move this before line 128?

src/dbus/error.cpp
```diff
@@ -0,0 +1,119 @@
...
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dbus/error.hpp"
+
+#include "dbus/constants.hpp"
+#include "dbus/dbus_message_helper.hpp"
+
+static const std::pair<otError, const char *> sErrorNameMap[] = {
```
wgtdkp: Why should we use `const char *` rather than `const std::string`?

src/dbus/error.cpp
```diff
@@ -0,0 +1,119 @@
...
+    {OT_ERROR_NOT_LOWPAN_DATA_FRAME, "io.openthread.Error.NonLowpanDatatFrame"},
+    {OT_ERROR_LINK_MARGIN_LOW, "io.openthread.Error.LinkMarginLow"},
+};
+
+namespace otbr {
+namespace dbus {
+
+const char *ConvertToDBusErrorName(otError aError)
+{
+    const char *name = sErrorNameMap[0].second;
```
wgtdkp: We can use `sErrorNameMap.find`.

src/dbus/error.cpp
```diff
@@ -0,0 +1,119 @@
...
+    {
+        if (p.first == aError)
+        {
+            name = p.second;
+        }
+    }
+    return name;
+}
+
+otError ConvertFromDBusErrorName(const std::string &aErrorName)
```
wgtdkp: Why don't we use `const char *` to keep the signature consistent with `ConvertToDBusErrorName`?

src/dbus/error.cpp
gjc13: We use std::string here to avoid static member destruction.

src/dbus/error.cpp
gjc13: It's an array so we don't use find. Has also renamed to avoid confusion.

src/dbus/error.cpp
gjc13: We use std::string here to support both const char* and std::string and to avoid the strcmp memory issue (you cannot really decide the length of a const char * that is safe to read).

src/dbus/dbus_object.hpp
```diff
@@ -67,44 +70,117 @@ class DBusObject
...
+     * @param[in]   aInterfaceName    The interface name.
+     * @param[in]   aMethodName       The method name.
+     * @param[in]   aHandler          The method handler.
+     */
+    void RegisterMethod(const std::string &      aInterfaceName,
+                        const std::string &      aMethodName,
+                        const MethodHandlerType &aHandler);
+
+    /**
+     * This method the get handler for a property.
```
librasungirl: nit: `the` ->`registers`

src/dbus/dbus_object.hpp
```diff
@@ -67,44 +70,117 @@ class DBusObject
...
+     * @param[in]   aInterfaceName    The interface name.
+     * @param[in]   aMethodName       The method name.
+     * @param[in]   aHandler          The method handler.
+     */
+    void RegisterGetPropertyHandler(const std::string &        aInterfaceName,
+                                    const std::string &        aMethodName,
+                                    const PropertyHandlerType &aHandler);
+
+    /**
+     * This method the set handler for a property.
```
librasungirl: nit: `the` ->`registers`

src/dbus/dbus_object.hpp
```diff
@@ -67,44 +70,117 @@ class DBusObject
...
+     *
+     * @param[in]   aInterfaceName    The interface name.
+     * @param[in]   aPropertyName     The property name.
+     * @param[in]   aValue            New value of the property.
+     *
+     * @retval OTBR_ERROR_NONE  Signal successfully sent.
+     * @retval OTBR_ERROR_DBUS  Failed to send the signal.
+     */
+    template <typename ValueType>
+    otbrError SignalPropertyChanged(const std::string &aInterfaceName,
```
librasungirl: nit:  `SignalPropertyChanged` -> `SignalPropertiesChanged` to match the macro `DBUS_PROPERTIES_CHANGED_SIGNAL`?

gjc13: This function only support sending the signal message of one property's change.

src/dbus/constants.hpp
```diff
@@ -0,0 +1,37 @@
...
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef OTBR_DBUS_CONSTANTS_HPP_
+#define OTBR_DBUS_CONSTANTS_HPP_
+
+#define DBUS_PROPERTY_GET_METHOD "Get"
+#define DBUS_PROPERTY_SET_METHOD "Set"
+#define DBUS_PROPERTY_GET_ALL_METHOD "GetAll"
```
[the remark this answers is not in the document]
gjc13: GetAll is the name defined by d-bus.

src/dbus/error.cpp
gjc13: Maps, sets, and other dynamic containers: if you require a static, fixed collection, such as a set to search against or a lookup table, you cannot use the dynamic containers from the standard library as a static variable, since they have non-trivial destructors. Instead, consider a simple array of trivial types, e.g. an array of arrays of ints (for a "map from int to int"), or an array of pairs (e.g. pairs of int and const char*). For small collections, linear search is entirely sufficient (and efficient, due to memory locality); consider using the facilities from absl/algorithm/container.h for the standard operations. If necessary, keep the collection in sorted order and use a binary search algorithm. If you do really prefer a dynamic container from the standard library, consider using a function-local static pointer, as described below.

src/agent/main.cpp
```diff
@@ -153,6 +153,12 @@ static void PrintVersion(void)
     printf("%s\n", PACKAGE_VERSION);
 }
 
+static void OnAllocateFailed(void)
+{
+    otbrLog(OTBR_LOG_INFO, "Allocate failure, exiting...");
```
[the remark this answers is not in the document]
wgtdkp: This seems not resolved? Are you changing at the wrong place (line 85)?

gjc13:
@wgtdkp @Moandor-y Any more comments?

src/dbus/dbus_message_helper.cpp
```diff
@@ -92,17 +110,29 @@ otbrError DBusMessageExtract(DBusMessageIter *aIter, std::vector<int64_t> &aValu
...
-    uint8_t val = aValue;
+    dbus_bool_t val = aValue ? 1 : 0;
+    otbrError   err = OTBR_ERROR_NONE;
+
+    VerifyOrExit(dbus_message_iter_append_basic(aIter, DBUS_TYPE_BOOLEAN, &val), err = OTBR_ERROR_DBUS);
+
+exit:
+    return err;
 
     return DBusMessageEncode(aIter, val);
```
wgtdkp: Should reverse the order of this `return` statement and `exit` label?

src/dbus/dbus_request.hpp
```diff
@@ -43,39 +44,72 @@ class DBusRequest
         : mConnection(aConnection)
         , mMessage(aMessage)
     {
+        dbus_message_ref(aMessage);
```
wgtdkp: This seems to be really weird that `unref` is called automatically by the `std::unique_ptr` deleter. To do resource management automatically, we can defines `DBusConnectHolder` and `DBusMessageHolder` as a wrapper of the two `C` structs. Like this:
```C++
class DBusConnectionHolder {
public:
        DBusConnectionHolder(DBusConnection *aConnection) 
            : mConnection(aConnection) {
                assert(mConnection != nullptr);
                dbus_connection_ref(mConnection);
        }
     
        ~DBusConnectionHolder() {
                dbus_connection_unref(mConnection);
        }

        // TODO: copy/move constructor, assignment

        DBusConnection *get() { return mConnection; }

private:
        DBusConnection *mConnection;
};
```

wgtdkp: Or we just use raw `DBusConnection` and `DBusMessage` pointer, and `unref` them in dtor of `DBusRequest`.

src/dbus/dbus_request.hpp
```diff
@@ -43,39 +44,72 @@ class DBusRequest
...
+    /**
+     * This method replies to the d-bus method call.
+     *
+     * @param[in] aReply  The tuple to be sent.
      *
      */
     template <typename... Args> void Reply(const std::tuple<Args...> &aReply)
     {
-        DBusMessage *reply = dbus_message_new_method_return(mMessage.GetRaw());
+        UniqueDBusMessage reply(dbus_message_new_method_return(mMessage.get()));
```
wgtdkp: I was wondering why a function call of `reply` has to declare the return type :(

Should we prefer C++11 `list initialization` to make it distinguishable?

```C++
 UniqueDBusMessage reply {dbus_message_new_method_return(mMessage.get())};
```

gjc13: The return type is void. The `template <typename... Args>` should be in a new line but the auto format seems to have its own idea.

src/dbus/dbus_request.hpp
gjc13: Updated

src/dbus/dbus_request.hpp
gjc13: Updated to prefer list initialization.

src/dbus/dbus_resources.hpp
```diff
@@ -29,61 +29,21 @@
...
-    }
-
-    T *mResource;
+    void operator()(kType *aPointer) { kUnref(aPointer); }
 };
 
-using SharedDBusConnection = SharedDBusResource<DBusConnection, dbus_connection_ref, dbus_connection_unref>;
-using SharedDBusMessage    = SharedDBusResource<DBusMessage, dbus_message_ref, dbus_message_unref>;
+using UniqueDBusMessage    = std::unique_ptr<DBusMessage, DBusDeleter<DBusMessage, &dbus_message_unref>>;
+using UniqueDBusConnection = std::unique_ptr<DBusConnection, DBusDeleter<DBusConnection, &dbus_connection_unref>>;
```
wgtdkp: Should eliminate `dbus_message_unref` anywhere else and always use `UniqueDBusMessage`? The same for `DBusConnection`?

src/dbus/dbus_server_test.cpp
```diff
@@ -29,19 +29,42 @@
...
 
 class TestObject : public DBusObject
 {
 public:
     TestObject(DBusConnection *aConnection)
         : DBusObject(aConnection, "/org/otbr/testobj")
+        , mCount(0)
     {
         RegisterMethod("org.otbr", "Ping", std::bind(&TestObject::PingHandler, this, _1));
+        RegisterGetPropertyHandler("org.otbr", "Count", std::bind(&TestObject::CountGetHandler, this, _1));
```
wgtdkp: Do we prefer lambda to `std::bind`?

src/dbus/error.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {OT_ERROR_DETACHED, "io.openthread.Error.Detached"},
+    {OT_ERROR_FCS, "io.openthread.Error.FcsErr"},
+    {OT_ERROR_NO_FRAME_RECEIVED, "io.openthread.Error.NoFrameReceived"},
+    {OT_ERROR_UNKNOWN_NEIGHBOR, "io.openthread.Error.UnknownNeighbor"},
+    {OT_ERROR_INVALID_SOURCE_ADDRESS, "io.openthread.Error.InvalidSourceAddress"},
+    {OT_ERROR_ADDRESS_FILTERED, "io.openthread.Error.AddressFiltered"},
+    {OT_ERROR_DESTINATION_ADDRESS_FILTERED, "io.openthread.Error.DestinationAddressFiltered"},
+    {OT_ERROR_NOT_FOUND, "io.openthread.Error.NotFound"},
+    {OT_ERROR_ALREADY, "io.openthread.Error.Already"},
+    {OT_ERROR_IP6_ADDRESS_CREATION_FAILURE, "io.openthread.Error.Ipv6AddressCreationFailure"},
```
wgtdkp: nit: make a macro for the prefix `io.openthread.Error`.

src/dbus/dbus_resources.hpp
gjc13: The only occurrence of `dbus_message_unref` is in dbus_request.hpp which we've agreed to use raw pointer.

src/dbus/dbus_server_test.cpp
gjc13: We don't really have a preference now so maybe we can keep it as it is.

src/dbus/dbus_request.hpp
```diff
@@ -53,14 +53,14 @@ class DBusRequest
      *
      * @returns   The dbus message.
      */
-    UniqueDBusMessage const &GetMessage(void) { return mMessage; }
+    DBusMessage *const &GetMessage(void) { return mMessage; }
```
wgtdkp: the reference is not necessary.

src/dbus/dbus_resources.hpp
wgtdkp: what about `ncp_wpantund.cpp` and `dbus_base.cpp` ?

src/dbus/dbus_request.hpp
gjc13: Done

src/dbus/dbus_resources.hpp
gjc13: We won't touch those code not under dbus/ folder in this PR.

wgtdkp:
LGTM 👍

gjc13:
@bukepo Rebased to master and added tests.

gjc13:
@bukepo Any more comments?

src/dbus/error.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+otError ConvertFromDBusErrorName(const std::string &aErrorName)
+{
+    otError err = OT_ERROR_GENERIC;
+
+    for (auto &&p : sErrorNames)
+    {
+        if (p.second == aErrorName)
+        {
+            err = p.first;
```
librasungirl: also `break`

tests/dbus/test-dbus-server.cpp
```diff
@@ -29,29 +29,52 @@
...
 class TestObject : public DBusObject
 {
 public:
     TestObject(DBusConnection *aConnection)
         : DBusObject(aConnection, "/org/otbr/testobj")
         , mEnded(false)
+        , mCount(0)
     {
         RegisterMethod("org.otbr", "Ping", std::bind(&TestObject::PingHandler, this, _1));
+        RegisterGetPropertyHandler("org.otbr", "Count", std::bind(&TestObject::CountGetHandler, this, _1));
```
librasungirl: how about keeping interface name consistent with https://github.com/openthread/ot-br-posix/pull/360/files# ?

PR #371, opened 2020-02-17
[utils] fix strcpy_safe behavior

wgtdkp:
`aDest` is the output provided by caller. `aDest` should not be touched If this function returns a failure (-1). It is unexpected that a failed call has any side effect on the output.

### DIFF
--- a/src/utils/strcpy_utils.cpp
+++ b/src/utils/strcpy_utils.cpp
@@ -33,8 +33,8 @@ int strcpy_safe(char *aDest, size_t aDestSize, const char *aSrc)
 {
     int ret = 0;
     VerifyOrExit(aDest != NULL, ret = -1);
-    VerifyOrExit(aSrc != NULL, ret = -1, aDest[0] = 0);
-    VerifyOrExit(aDestSize > strnlen(aSrc, aDestSize), ret = -1, aDest[0] = 0);
+    VerifyOrExit(aSrc != NULL, ret = -1);
+    VerifyOrExit(aDestSize > strnlen(aSrc, aDestSize), ret = -1);
 
     strncpy(aDest, aSrc, aDestSize);
 
--- a/src/utils/strcpy_utils.hpp
+++ b/src/utils/strcpy_utils.hpp
@@ -36,14 +36,14 @@ extern "C" {
 #endif
 
 /**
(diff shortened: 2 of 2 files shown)


bukepo:
This function is trying to mimic the behavior of `strcpy_s`, see: https://en.cppreference.com/w/c/string/byte/strcpy

I think what we want is for invalid arguments or data, we could detect the error. So maybe it's not the function needs to be changed, but where it is used.

wgtdkp:
@bukepo 
>I think what we want is for invalid arguments or data, we could detect the error.

This PR keeps this ability.

>This function is trying to mimic the behavior of strcpy_s

This is not necessary, because the behavior is not desired by our usage. because:
>strcpy_s is allowed to clobber the destination array from the last character written up to destsz in order to improve efficiency: it may copy in multibyte blocks and then check for null bytes.

set `aDest[0]` to `\0` is useful only in this case, and this is not how we are going to use `strcpy_safe`. Take one of the many usages as an example https://github.com/openthread/ot-br-posix/blob/master/src/agent/border_agent.cpp#L333:

```c++
void BorderAgent::SetNetworkName(const char *aNetworkName)
{
    strcpy_safe(mNetworkName, sizeof(mNetworkName), aNetworkName);

#if OTBR_ENABLE_MDNS_AVAHI || OTBR_ENABLE_MDNS_MDNSSD || OTBR_ENABLE_MDNS_MOJO
    if (mThreadStarted)
    {
        // Restart publisher to publish new service name.
        mPublisher->Stop();
        StartPublishService();
    }
#endif
}
```
There are two problems:
1. the return value of `strcpy_safe` is not handled;
2. The function should return a error; 
3. most importantly, the previous `mNetworkName` value will be erased even if an invalid argument is provided; it should be kept not changed in this case, right?

>So maybe it's not the function needs to be changed, but where it is used.
Everywhere needs to be changed and we don't benefit from the strange behavior of `strycpy_s`.

Thoughts?

bukepo:
Could you also update where the function is used?

wgtdkp:
> Could you also update where the function is used?

It leads to a significant change, will do later this week. Add `work-in-progress` for now.

jwhui:
Closing stale PR.

PR #373, opened 2020-02-20
[core] add openthread error code embedding into otbrError

gjc13:
This change adds embedded third-party error code to `otbrError` and provides an example with OpenThread error codes.


wgtdkp:
Why do we need this change?

PR #376, opened 2020-03-05
[dbus] more OpenThread properties

gjc13:
This CL contains more OpenThread properties.

The encoding of arrays in dbus messages are also changed so that we can send a vector as properties.


bukepo:
update src/dbus/server/instropect.xml?

src/dbus/common/types.hpp
```diff
@@ -401,6 +401,60 @@ struct otbrIpCounters
     uint32_t mRxFailure; ///< The number of IPv6 packets failed to receive.
 };
 
+struct otbrChannelQuality
+{
+    uint8_t  mChannel;
+    uint16_t mOccupancy;
+};
+
+struct otbrChildInfo
```
wgtdkp: Openthread public API has already defined those structures. Is there a good reason why we need this?

gjc13: I've discussed with @bukepo , we don't want to export openthread header to users so we need to define them here,

wgtdkp: Then, it seems we need to do this for all other openthread public structures in the future.

BTW, what are Cons of reusing openthread API here? @bukepo

src/dbus/common/dbus_message_helper_openthread.cpp
```diff
@@ -273,5 +273,127 @@ otbrError DBusMessageExtract(DBusMessageIter *aIter, otbrIpCounters &aCounters)
...
+}
+
+otbrError DBusMessageEncode(DBusMessageIter *aIter, const otbrLeaderData &aLeaderData)
+{
+    DBusMessageIter sub;
+    otbrError       err  = OTBR_ERROR_NONE;
+    auto            args = std::tie(aLeaderData.mPartitionId, aLeaderData.mWeighting, aLeaderData.mDataVersion,
+                         aLeaderData.mStableDataVersion, aLeaderData.mLeaderRouterId);
+
+    VerifyOrExit(dbus_message_iter_open_container(aIter, DBUS_TYPE_STRUCT, nullptr, &sub));
```
wgtdkp: Set error code.

src/dbus/common/dbus_message_helper_openthread.cpp
```diff
@@ -273,5 +273,127 @@ otbrError DBusMessageExtract(DBusMessageIter *aIter, otbrIpCounters &aCounters)
...
+    return err;
+}
+
+otbrError DBusMessageEncode(DBusMessageIter *aIter, const otbrChannelQuality &aQuality)
+{
+    DBusMessageIter sub;
+    otbrError       err  = OTBR_ERROR_NONE;
+    auto            args = std::tie(aQuality.mChannel, aQuality.mOccupancy);
+
+    VerifyOrExit(dbus_message_iter_open_container(aIter, DBUS_TYPE_STRUCT, nullptr, &sub));
```
wgtdkp: ditto.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -556,5 +589,248 @@ otError DBusThreadObject::GetSupportedChannelMaskHandler(DBusMessageIter &aIter)
...
+    VerifyOrExit(DBusMessageEncodeToVariant(&aIter, leaderData) == OTBR_ERROR_NONE, err = OT_ERROR_INVALID_ARGS);
+exit:
+    return err;
+}
+
+otError DBusThreadObject::GetNetworkDataHandler(DBusMessageIter &aIter)
+{
+    auto                 threadHelper = mNcp->GetThreadHelper();
+    otError              err          = OT_ERROR_NONE;
+    uint8_t              data[255]; // magic number from openthread/src/cli/cli.cpp:ProcessNetworkDataShow
```
wgtdkp: We can always define a constants if OT public API does offer it. for example:
```c++
static constexpr size_t kNetworkDataMaxSize = 254;
```

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -556,5 +589,248 @@ otError DBusThreadObject::GetSupportedChannelMaskHandler(DBusMessageIter &aIter)
...
+    uint8_t              len = sizeof(data);
+    std::vector<uint8_t> networkData;
+
+    otNetDataGet(threadHelper->GetInstance(), /*stable=*/false, data, &len);
```
wgtdkp: 1. We can write to `networkData` directly.
2. Catch the return error.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -556,5 +589,248 @@ otError DBusThreadObject::GetSupportedChannelMaskHandler(DBusMessageIter &aIter)
     return err;
 }
 
+otError DBusThreadObject::GetRloc16Handler(DBusMessageIter &aIter)
+{
+    auto     threadHelper = mNcp->GetThreadHelper();
+    otError  err          = OT_ERROR_NONE;
+    uint16_t rloc16       = otThreadGetRloc16(threadHelper->GetInstance());
+
+    VerifyOrExit(DBusMessageEncodeToVariant(&aIter, rloc16) == OTBR_ERROR_NONE, err = OT_ERROR_INVALID_ARGS);
```
wgtdkp: `OT_ERROR_INVALID_ARGS` doesn't seem to be the proper error code for this failure. I think we should return `otbrError` for this handlers.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -556,5 +589,248 @@ otError DBusThreadObject::GetSupportedChannelMaskHandler(DBusMessageIter &aIter)
...
+
+otError DBusThreadObject::GetRouterIdHandler(DBusMessageIter &aIter)
+{
+    auto         threadHelper = mNcp->GetThreadHelper();
+    otError      err          = OT_ERROR_NONE;
+    uint16_t     rloc16       = otThreadGetRloc16(threadHelper->GetInstance());
+    otRouterInfo info;
+
+    SuccessOrExit(err = otThreadGetRouterInfo(threadHelper->GetInstance(), rloc16, &info));
+    VerifyOrExit(DBusMessageEncodeToVariant(&aIter, info.mRouterId) == OTBR_ERROR_NONE, err = OT_ERROR_INVALID_ARGS);
```
wgtdkp: add an empty line before label `exit`.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -556,5 +589,248 @@ otError DBusThreadObject::GetSupportedChannelMaskHandler(DBusMessageIter &aIter)
...
+
+otError DBusThreadObject::GetStableNetworkDataHandler(DBusMessageIter &aIter)
+{
+    auto                 threadHelper = mNcp->GetThreadHelper();
+    otError              err          = OT_ERROR_NONE;
+    uint8_t              data[255]; // magic number from openthread/src/cli/cli.cpp:ProcessNetworkDataShow
+    uint8_t              len = sizeof(data);
+    std::vector<uint8_t> networkData;
+
+    otNetDataGet(threadHelper->GetInstance(), /*stable=*/true, data, &len);
```
wgtdkp: ditto.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -556,5 +589,248 @@ otError DBusThreadObject::GetSupportedChannelMaskHandler(DBusMessageIter &aIter)
...
+#if OPENTHREAD_CONFIG_CHANNEL_MONITOR_ENABLE
+    auto                            threadHelper = mNcp->GetThreadHelper();
+    otError                         err          = OT_ERROR_NONE;
+    uint32_t                        channelMask  = otLinkGetSupportedChannelMask(threadHelper->GetInstance());
+    constexpr uint8_t               kNumChannels = sizeof(channelMask) * 8; // 8 bit per byte
+    std::vector<otbrChannelQuality> quality;
+
+    for (uint8_t i = 0; i < kNumChannels; i++)
+    {
+        if (channelMask & (1 << i))
```
wgtdkp:
```suggestion
        if (channelMask & (1U << i))
```

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -556,5 +589,248 @@ otError DBusThreadObject::GetSupportedChannelMaskHandler(DBusMessageIter &aIter)
...
+
+otError DBusThreadObject::GetChildTableHandler(DBusMessageIter &aIter)
+{
+    auto                       threadHelper = mNcp->GetThreadHelper();
+    otError                    err          = OT_ERROR_NONE;
+    uint16_t                   childIndex   = 0;
+    otChildInfo                childInfo;
+    std::vector<otbrChildInfo> childTable;
+
+    while (otThreadGetChildInfoByIndex(threadHelper->GetInstance(), childIndex, &childInfo) == OT_ERROR_NONE)
```
wgtdkp: Is ignoring the error intentional?

tests/unit/test_dbus_message.cpp
```diff
@@ -147,7 +208,87 @@ TEST(DBusMessage, TestStructMessage)
...
+
+TEST(DBusMessage, TestOtbrLeaderData)
+{
+    DBusMessage *                     msg = dbus_message_new(DBUS_MESSAGE_TYPE_METHOD_RETURN);
+    tuple<otbr::DBus::otbrLeaderData> setVals({1, 2, 3, 4, 5});
+    tuple<otbr::DBus::otbrLeaderData> getVals;
+
+    assert(msg != NULL);
+
+    TupleToDBusMessage(*msg, setVals);
```
wgtdkp: `ASSERT_*` on those functions.

tests/unit/test_dbus_message.cpp
```diff
@@ -147,7 +208,87 @@ TEST(DBusMessage, TestStructMessage)
...
+    dbus_message_unref(msg);
+}
+
+TEST(DBusMessage, TestOtbrLeaderData)
+{
+    DBusMessage *                     msg = dbus_message_new(DBUS_MESSAGE_TYPE_METHOD_RETURN);
+    tuple<otbr::DBus::otbrLeaderData> setVals({1, 2, 3, 4, 5});
+    tuple<otbr::DBus::otbrLeaderData> getVals;
+
+    assert(msg != NULL);
```
wgtdkp: use gtest `ASSERT_*`?

tests/unit/test_dbus_message.cpp
```diff
@@ -147,7 +208,87 @@ TEST(DBusMessage, TestStructMessage)
...
+
+    TupleToDBusMessage(*msg, setVals);
+    DBusMessageToTuple(*msg, getVals);
+
+    assert(std::get<0>(setVals) == std::get<0>(getVals));
```
wgtdkp: use gtest `EXPECT_*`?

tests/unit/test_dbus_message.cpp
```diff
@@ -147,7 +208,87 @@ TEST(DBusMessage, TestStructMessage)
     TupleToDBusMessage(*msg, setVals);
     DBusMessageToTuple(*msg, getVals);
 
+    printf("%zu %s\n", std::get<2>(getVals).size(), std::get<2>(getVals)[0].c_str());
```
wgtdkp: remove this.

src/dbus/common/dbus_message_helper.hpp
```diff
@@ -182,20 +251,19 @@ template <typename T> otbrError DBusMessageExtract(DBusMessageIter *aIter, T &aV
...
+    VerifyOrExit(dbus_message_iter_get_arg_type(aIter) == DBUS_TYPE_ARRAY, err = OTBR_ERROR_DBUS);
+    dbus_message_iter_recurse(aIter, &subIter);
 
     aValue.clear();
-    for (uint32_t i = 0; i < size; i++)
+    while (dbus_message_iter_get_arg_type(&subIter) != DBUS_TYPE_INVALID)
     {
         aValue.emplace_back();
-        DBusMessageExtract(aIter, aValue.back());
+        SuccessOrExit(err = DBusMessageExtract(&subIter, aValue.back()));
```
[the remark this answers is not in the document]
gjc13: We never handle partial data, if you get an error code you should abort the parse.

src/dbus/server/dbus_thread_object.cpp
gjc13: We cannot since we cannot really decide the length of the network data.

src/dbus/server/dbus_thread_object.cpp
gjc13: Yes, the only error case is end of iteration.

tests/unit/test_dbus_message.cpp
gjc13: In fact it's cpputest but I'll replace the assert with the CHECK macro from the test framework.

src/dbus/common/dbus_message_helper_openthread.cpp
```diff
@@ -273,5 +273,127 @@ otbrError DBusMessageExtract(DBusMessageIter *aIter, otbrIpCounters &aCounters)
...
+{
+    DBusMessageIter sub;
+    otbrError       err  = OTBR_ERROR_NONE;
+    auto            args = std::tie(aChildInfo.mExtAddress, aChildInfo.mTimeout, aChildInfo.mAge, aChildInfo.mRloc16,
+                         aChildInfo.mChildId, aChildInfo.mNetworkDataVersion, aChildInfo.mLinkQualityIn,
+                         aChildInfo.mAverageRssi, aChildInfo.mLastRssi, aChildInfo.mFrameErrorRate,
+                         aChildInfo.mMessageErrorRate, aChildInfo.mRxOnWhenIdle, aChildInfo.mSecureDataRequest,
+                         aChildInfo.mFullThreadDevice, aChildInfo.mFullNetworkData, aChildInfo.mIsStateRestoring);
+
+    VerifyOrExit(dbus_message_iter_open_container(aIter, DBUS_TYPE_STRUCT, nullptr, &sub));
```
Moandor-y: Set error code.

src/dbus/common/dbus_message_helper_openthread.cpp
```diff
@@ -273,5 +273,127 @@ otbrError DBusMessageExtract(DBusMessageIter *aIter, otbrIpCounters &aCounters)
...
+{
+    DBusMessageIter sub;
+    otbrError       err  = OTBR_ERROR_NONE;
+    auto            args = std::tie(aNeighborInfo.mExtAddress, aNeighborInfo.mAge, aNeighborInfo.mRloc16,
+                         aNeighborInfo.mLinkFrameCounter, aNeighborInfo.mMleFrameCounter, aNeighborInfo.mLinkQualityIn,
+                         aNeighborInfo.mAverageRssi, aNeighborInfo.mLastRssi, aNeighborInfo.mFrameErrorRate,
+                         aNeighborInfo.mMessageErrorRate, aNeighborInfo.mRxOnWhenIdle, aNeighborInfo.mSecureDataRequest,
+                         aNeighborInfo.mFullThreadDevice, aNeighborInfo.mFullNetworkData, aNeighborInfo.mIsChild);
+
+    VerifyOrExit(dbus_message_iter_open_container(aIter, DBUS_TYPE_STRUCT, nullptr, &sub));
```
Moandor-y: ditto

tests/dbus/test_dbus_client.cpp
```diff
@@ -26,9 +26,11 @@
  *    POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <memory>
+#include <assert.h>
 #include <stdio.h>
```
Moandor-y:
```suggestion
#include <cassert>
#include <cstdio>
```

tests/unit/test_dbus_message.cpp
```diff
@@ -27,6 +27,8 @@
  */
 
 #include <assert.h>
+#include <string.h>
```
Moandor-y:
```suggestion
#include <cassert>
#include <cstring>
```

src/dbus/server/dbus_thread_object.cpp
gjc13: The handler function signature has been decided and it will requires quite some effort to change it. I'd prefer to change it after we've unified all the error codes.

tests/unit/test_dbus_message.cpp
gjc13: I'm not quite sure about this, does openthread prefer <xxx.h> to <cxxx> since I seldom see using the latter form in OpenThread code.
@bukepo @jwhui

src/dbus/client/thread_api_dbus.cpp
```diff
@@ -424,6 +424,76 @@ OtbrClientError ThreadApiDBus::GetSupportedChannelMask(uint32_t &aChannelMask)
     return GetProperty(OTBR_DBUS_SUPPORTED_CHANNEL_MASK_PROPERTY, aChannelMask);
 }
 
+OtbrClientError ThreadApiDBus::GetRloc16(uint16_t &aRloc16)
```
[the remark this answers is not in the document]
gjc13: Unfortunately OpenThread and otbr has been using passing output references foa a long time so I'm not sure about the style guide here. I choose to do as the Romans do. Modifying all of them to output pointers requires a large amount of effort.

src/dbus/server/dbus_thread_object.cpp
gjc13: Another consideration is that in fact all these handlers are calling some OpenThread apis that will return otError. So I decided to let the property handlers return directly an otError to reduce casts.

src/dbus/client/thread_api_dbus.cpp
```diff
@@ -350,78 +350,148 @@ OtbrClientError ThreadApiDBus::RemoveOnMeshPrefix(const otbrIp6Prefix &aPrefix)
...
+    return GetProperty(OTBR_DBUS_PROPERTY_EXTPANID, aExtPanId);
 }
 
 OtbrClientError ThreadApiDBus::GetChannel(uint16_t &aChannel)
 {
-    return GetProperty(OTBR_DBUS_CHANNEL_PROPERTY, aChannel);
+    return GetProperty(OTBR_DBUS_PROPERTY_CHANNEL, aChannel);
 }
 
 OtbrClientError ThreadApiDBus::GetMasterKey(std::vector<uint8_t> &aMasterKey)
```
bukepo: As c++ API, could we remove the prefix `OtbrClient`, just `Error`?

src/dbus/client/thread_api_dbus.cpp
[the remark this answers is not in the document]
Moandor-y: OK, I checked our style guide and it prefers reference to pointer. So we should keep passing by references.

src/dbus/common/types.hpp
bukepo: We want to decouple projects depends on this project, and preventing from them to use headers from this projects' dependencies. Thus we don't need to update all projects when OpenThread's APIs are changed. Take the mac counter as an example, if OpenThread adds a new fields, we don't want to change all projects up to service and database, because the new field may not be used in our products.

src/dbus/common/dbus_message_helper.hpp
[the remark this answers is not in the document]
wgtdkp: **We should keep the output `aValue` as it was before calling this function when this function failed**. Because it makes no sense to the caller that a failed function calling get its data modified, especially, this function will emplace an uninitialized data at the back. Suppose the user is calling this function on its member data, like this:

```C++
otbrError error = OTBR_ERROR_NONE;
SuccessOrExit(error = GetChannelMonitorAllChannelQualities(&aIter, mChannelQualities));

exit:

```

This calling and handling is common but leaves `mChannelQualities` in an invalid state. We can of cause use a temp variable for this case, but that will be tedious:

```C++
otbrError error = OTBR_ERROR_NONE;
ChannelQualityArray qualities;
SuccessOrExit(error = GetChannelMonitorAllChannelQualities(&aIter, qualities));
mChannelQualities = qualities;

exit:
```

What do you think?

src/dbus/server/dbus_thread_object.cpp
wgtdkp: I think we should make a function to cast from `otError` to `otbrError`. In this case, there are errors out of the scope of OT, but related to OTBR (OT error is subset of OTBR error I think). 

- It is okay to leave this to a future PR, but we should create an issue tracking it.
- `OT_ERROR_FAILED` fits better than `OT_ERROR_INVALID_ARGS` for compromised solution :)

src/dbus/common/dbus_message_helper.hpp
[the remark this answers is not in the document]
gjc13: There is a PR about these error encoding issues: https://github.com/openthread/ot-br-posix/pull/373

src/dbus/server/dbus_thread_object.cpp
gjc13: There is a PR about these error code: https://github.com/openthread/ot-br-posix/pull/373

tests/unit/test_dbus_message.cpp
bukepo: Let's use the OpenThread's style for now. @Moandor-y @gjc13

src/dbus/client/thread_api_dbus.cpp
gjc13: Calling it `Error` looks too generic, I'll call it `ClientError`.

src/dbus/common/dbus_message_helper.hpp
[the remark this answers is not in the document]
gjc13: Fixed

src/dbus/common/types.hpp
```diff
@@ -401,6 +401,60 @@ struct otbrIpCounters
...
+    uint16_t mFrameErrorRate;     ///< Frame error rate (0xffff->100%). Requires error tracking feature.
+    uint16_t mMessageErrorRate;   ///< (IPv6) msg error rate (0xffff->100%). Requires error tracking feature.
+    bool     mRxOnWhenIdle;       ///< rx-on-when-idle
+    bool     mSecureDataRequest;  ///< Secure Data Requests
+    bool     mFullThreadDevice;   ///< Full Thread Device
+    bool     mFullNetworkData;    ///< Full Network Data
+    bool     mIsStateRestoring;   ///< Is in restoring state
+};
+
+struct otbrNeighborInfo
```
bukepo: remove all `otbr` prefix?

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -31,8 +31,13 @@
 #include <string.h>
 
 #include "openthread/border_router.h"
+#include "openthread/channel_monitor.h"
 #include "openthread/instance.h"
 #include "openthread/joiner.h"
+#include "openthread/link_raw.h"
```
bukepo: use `<>` for public headers.

gjc13:
@Moandor-y any more suggestions?

gjc13:
@abtink Friendly reminder

abtink:
Will try to read/review by end of tomorrow (Tuesday)... if not @jwhui, @gjc13 please don't wait for me.

src/dbus/server/instropect.xml
```diff
@@ -182,7 +182,117 @@
...
+        uint16_t mFrameErrorRate;    ///< Frame error rate (0xffff->100%). Requires error tracking feature.
+        uint16_t mMessageErrorRate;  ///< (IPv6) msg error rate (0xffff->100%). Requires error tracking feature.
+        bool     mRxOnWhenIdle;      ///< rx-on-when-idle
+        bool     mSecureDataRequest; ///< Secure Data Requests
+        bool     mFullThreadDevice;  ///< Full Thread Device
+        bool     mFullNetworkData;   ///< Full Network Data
+        bool     mIsChild;           ///< Is the neighbor a child
+      }
+    -->
+    <property name="NeighborTable" type="a(tuquuyyyqqbbbb)" access="read">
```
librasungirl: 5 boolean fields in the end of the structure, however here only `b`x4?

librasungirl: seems also inconsistent with the NeighborInfo here
https://github.com/openthread/ot-br-posix/blob/810fdae2693d314c3fa0e8120db186d67d25d991/src/dbus/common/dbus_message_helper.hpp#L120-L125

src/dbus/server/instropect.xml
```diff
@@ -182,7 +182,117 @@
...
+      <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="false"/>
+    </property>
```
librasungirl: what's the criteria for adding one property?
I am thinking `RouterTable` might be useful as well?

gjc13: If I find a property used in products then I add it.

src/dbus/server/instropect.xml
gjc13: Thanks for pointing out. I've also updated the unit tests to check these type signatures.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -84,21 +84,21 @@ class ThreadApiDBus
      * @param[in]   aSeconds  The timeout to close the port, 0 for never close.
      *
      * @retval OT_ERROR_NONE successfully performed the dbus function call
-     * @returns Error value otherwise
+     * @retval Error value otherwise
```
jwhui: We should avoid using `@retval` like this. It doesn't work well with auto-generated documentation. See example: https://openthread.io/reference/group/api-thread-general#otthreadgetdevicerole

Prefer using `@returns` if we are not explicitly listing each possible return value.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -156,201 +156,356 @@ class ThreadApiDBus
...
+     * @retval OT_ERROR_NONE successfully performed the dbus function call
+     * @retval Error value otherwise
+     *
+     */
+    ClientError GetLocalLeaderWeight(uint8_t &aWeight);
+
+    /**
+     * This method gets the channel monitor sample count.
+     *
+     * @param[out]  aWeight     The local leader weight.
```
jwhui: Documentation does not match declaration.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -156,201 +156,356 @@ class ThreadApiDBus
...
+     * @retval OT_ERROR_NONE successfully performed the dbus function call
+     * @retval Error value otherwise
+     *
+     */
+    ClientError GetChannelMonitorSampleCount(uint32_t &aSampleCount);
+
+    /**
+     * This method gets the channel quality map.
+     *
+     * @param[out]  aQualityMap     The channel quality map.
```
jwhui: Documentation does not match declaration.

src/dbus/common/dbus_message_helper.hpp
```diff
@@ -91,12 +99,73 @@ template <size_t SIZE> struct DBusTypeTrait<std::array<uint8_t, SIZE>>
...
-template <> struct DBusTypeTrait<otbrIp6Prefix>
+template <> struct DBusTypeTrait<Ip6Prefix>
 {
     // struct of { arrray of bytes, byte}
     static constexpr const char *TYPE_AS_STRING = "(ayy)";
 };
 
+template <> struct DBusTypeTrait<LeaderData>
+{
+    // struct of { uint32, byte, byte, byte, byte, byte}
```
jwhui: Documentation does not match format string.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -350,7 +350,162 @@ class ThreadApiDBus
...
-    OtbrClientError GetSupportedChannelMask(uint32_t &aChannelMask);
+    ClientError GetSupportedChannelMask(uint32_t &aChannelMask);
+
+    /**
+     * This method gets the Thread routing locator
+     *
+     * @param[out]  aRloc16   The routing locator
+     *
+     * @retval OT_ERROR_NONE successfully performed the dbus function call
+     * @returns Error value otherwise
```
jwhui: Does this multi-line documentation provide value over something simple like:
`@returns A ClientError value.`
?

gjc13: Updated the doc string.
The error value behavior is returning ERROR_DBUS on dbus encoding/decoding errors. Otherwise we will return the return value of OpenThread api calls.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -83,22 +83,24 @@ class ThreadApiDBus
      * @param[in]   aPort     The port number.
      * @param[in]   aSeconds  The timeout to close the port, 0 for never close.
      *
-     * @retval OT_ERROR_NONE successfully performed the dbus function call
-     * @returns Error value otherwise
+     * @retval ERROR_NONE successfully performed the dbus function call
+     * @retval ERROR_DBUS dbus encode/decode error
+     * @returns OpenThread defined error value otherwise
```
jwhui: Sorry, I don't think I'm being clear. I want to avoid mixing `@retval` and `@returns` in the same block. If we want to call out a couple explicit values as you have done, I'd advocate:
```
     * @retval ERROR_NONE successfully performed the dbus function call
     * @retval ERROR_DBUS dbus encode/decode error
     * @retval other      OpenThread defined error value otherwise
```

PR #380, opened 2020-03-22
[dbus] add handlers for managing external route

gjc13:


src/dbus/common/constants.hpp
```diff
@@ -47,6 +47,8 @@
 #define OTBR_DBUS_ADD_UNSECURE_PORT_METHOD "AddUnsecurePort"
 #define OTBR_DBUS_JOINER_START_METHOD "JoinerStart"
 #define OTBR_DBUS_JOINER_STOP_METHOD "JoinerStop"
+#define OTBR_DBUS_ADD_EXTERNAL_ROUTE_METHOD "AddExternalRoute"
```
wgtdkp: Not necessary to be addressed in this PR, could you rename those macros to `OTBR_DBUS_METHOD_*`? Just as we what we do for those properties.

src/dbus/server/instropect.xml
```diff
@@ -54,6 +54,40 @@
...
+      <!--
+        struct {
+          uint8[] prefix_bytes
+          uint8 prefix_length
+        }
+      -->
+      <arg name="prefix" type="(ayy)"/>
+    </method>
+
+    <method name="RemoveExternalRoute">
```
bukepo: why there are two `RemoveExternalRoute`?

PR #381, opened 2020-03-23
[dbus] set steering data when permitting join

gjc13:
This change renames `AddUnsecurePort` to `PermitUnsecureJoin`.
This function will set steering data as well now.


gjc13:
@wgtdkp Friendly reminder

src/agent/thread_helper.cpp
```diff
@@ -352,11 +356,15 @@ otError ThreadHelper::AddUnsecurePort(uint16_t aPort, uint32_t aSeconds)
         }
 
         mNcp->PostTimerTask(triggerTime, [this, aPort]() {
-            auto now = std::chrono::steady_clock::now();
+            auto         now = std::chrono::steady_clock::now();
+            otExtAddress steeringClearData;
 
+            // 0 to clean steering data
+            memset(&steeringClearData.m8, 0, sizeof(steeringClearData.m8));
             if (now >= mPortTime[aPort])
```
wgtdkp: Will the timer be triggered only once or regularly? If once, we may have a chance here failed to remove the unsecure port. if `now >= mPortTime[aPort]` should always success, we should do assertion. Otherwise, we need to re-post the task.

wgtdkp:
```suggestion
            if (now >= mUnsecurePortTimeout[aPort])
```

PR #384, opened 2020-03-23
[agent] auto attach after reboot under daemon mode

gjc13:


src/agent/thread_helper.hpp
```diff
@@ -155,6 +155,14 @@ class ThreadHelper
                      const std::string &aVendorData,
                      ResultHandler      aHandler);
 
+    /**
+     * This method trys to restore the network after reboot
```
bukepo: Trys-> tries

src/agent/ncp_openthread.cpp
```diff
@@ -165,6 +166,12 @@ void ControllerOpenThread::Process(const otSysMainloopContext &aMainloop)
         mTimers.begin()->second();
         mTimers.erase(mTimers.begin());
     }
+
+    if (!mTriedAttach)
+    {
+        mThreadHelper->TryResumeNetwork();
```
bukepo: Can this be called when just entering the mainloop?

src/agent/thread_helper.cpp
```diff
@@ -336,6 +337,27 @@ void ThreadHelper::JoinerCallback(otError aError)
...
+{
+    otError err = OT_ERROR_NONE;
+
+    if (otLinkGetPanId(mInstance) != UINT16_MAX && otThreadGetDeviceRole(mInstance) == OT_DEVICE_ROLE_DISABLED)
+    {
+        if (!otIp6IsEnabled(mInstance))
+        {
+            SuccessOrExit(err = otIp6SetEnabled(mInstance, true));
+        }
+        SuccessOrExit(err = otThreadSetEnabled(mInstance, true));
```
bukepo:
```suggestion
```

PR #389, opened 2020-03-31
[beaglebone] disable nginx and apache

srickardti:
The BeagleBone Debian distribution has changed from Jessie to Strecth.
One of the changes in that migration was moving the server for the
Cloud9 IDE to nginx from apache.

One of the scripts called by the setup disables running webservers to
allow the OpenThread Border Router to run it's own server. This updates
the disable script to point to the current webserver on the BeagleBone.

### DIFF
--- a/script/_disable_services
+++ b/script/_disable_services
@@ -42,7 +42,7 @@ bbb_disable_services()
 
     for service in $SERVICE_LIST
     do
-        if [[ $(sudo systemctl is-active $service) != "inactive" ]];
+        if [ $(sudo systemctl is-active $service) != "inactive" ];
         then
             for action in stop disable
             do
--- a/script/_disable_services
+++ b/script/_disable_services
@@ -34,7 +34,7 @@
 bbb_disable_services()
 {
     SERVICE_LIST=""
-    SERVICE_LIST="${SERVICE_LIST} apache2 "
+    SERVICE_LIST="${SERVICE_LIST} nginx "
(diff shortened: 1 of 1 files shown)


srickardti:
This could be done based on the version of debian, 8 vs 9. But I am not sure what the correct way to do that in the setup script is.

bukepo:
> This could be done based on the version of debian, 8 vs 9. But I am not sure what the correct way to do that in the setup script is.

Will stop both work?

srickardti:
> Will stop both work?

Possibly, the setup process was failing because systemctl couldn't find the service to stop. `apache2` doesn't exist on the new versions of BeagleBone's Debian, so the commands below fail.

```
sudo systemctl apache2 stop
sudo systemctl apache2 disable
```

srickardti:
@bukepo , Looking through systemctl's man page I couldn't find an easy way to do it. Trying to stop and disable a non-existent service would always be an error.

Seems like `systemctl stop <service>` returns `inactive` if the service is stopped or it doesn't exist. We could probably check each one before stopping and disabling them.

srickardti:
@bukepo , This seems to work on debian stretch, I haven't tested it on the older distro.

I don't know which approach is better, this on or the one in _otbr. The function in `script/_otbr` accomplishes the same thing by ignoring the errors from `stop` and `disable`.

bukepo:
@srickardti  I think both approaches are acceptable.

Issue #374, opened 2020-02-21
Use lower cases for namespace names


gjc13:
**Is your feature request related to a problem? Please describe.**
Currently we are using camel case for inner namespaces. Such style is a little weird since you cannot determine at the first glance whether the `BorderRouter` in `otbr::BorderRouter` is a namespace or a class.

**Describe the solution you'd like**
Use lower case namespace names as suggested by the Google C++ style. The OpenThread code style suggests camel case but lower case is more natrual.

gjc13:
I raised this issue since we are improving the code style in https://github.com/openthread/ot-br-posix/pull/372
@bukepo @jwhui @abtink What's your idea on this renaming?

bukepo:
SGTM. @wgtdkp @librasungirl

wgtdkp:
SGTM

abtink:
I don't have any strong opinion here and fine either way but if asked to choose  would go with the existing (CamelCase) model for all namespaces (after the high level `ot` or `otbr`). The way I see it is that any `class` or `struct` definition is defining a namespace as well so it is fine if we cannot tell the difference and use same naming convention.

jwhui:
What is the benefit of being able to differentiate between namespace and class from the name itself? While the Google C++ style guide provides pros/cons/decision for many of the style choices, it does not provide any such rationale for the namespace naming choice. At the same time, C++ stdlib uses all lower case names for both namespace and class.

The existing [STYLE_GUIDE.md](https://github.com/openthread/ot-br-posix/blob/master/STYLE_GUIDE.md) states:

> All OpenThread class, namespace, structure, method, function, enumeration, and type names in the C/C++ language shall be in upper camel case. Exception: the top level OpenThread namespace 'ot'.

This is generally shared across all OpenThread projects. If there is good reason to change, we should consider making the change uniformly across all OpenThread projects.

PR #397, opened 2020-04-08
[style] Use lowercase namespace name

gjc13:
Resolves https://github.com/openthread/ot-br-posix/issues/374


gjc13:
I'm touching ot::Ip6 code and when I used `Ip6::`in th `Ip6`  and got an unexpected error. The complier thinks of it as the namespace name rather than the class name.

PR #401, opened 2020-04-14
[third_party] update openthread

gjc13:


gjc13:
Debugging broken travis builds.

gjc13:
OpenThread already updated.

PR #402, opened 2020-04-16
[dbus] close unsecure port when timeout is zero

gjc13:
This is the behavior of the current client code. Update implementation
to be compatible.

### DIFF
--- a/src/agent/thread_helper.cpp
+++ b/src/agent/thread_helper.cpp
@@ -382,17 +382,25 @@ otError ThreadHelper::PermitUnsecureJoin(uint16_t aPort, uint32_t aSeconds)
 
         mNcp->PostTimerTask(triggerTime, [this, aPort]() {
             auto         now = std::chrono::steady_clock::now();
-            otExtAddress steeringClearData;
+            otExtAddress noneAddress;
 
             // 0 to clean steering data
-            memset(&steeringClearData.m8, 0, sizeof(steeringClearData.m8));
+            memset(&noneAddress.m8, 0, sizeof(noneAddress.m8));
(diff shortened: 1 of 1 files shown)


src/agent/thread_helper.cpp
```diff
@@ -394,6 +394,14 @@ otError ThreadHelper::PermitUnsecureJoin(uint16_t aPort, uint32_t aSeconds)
             }
         });
     }
+    else
+    {
+        otExtAddress steeringClearData;
```
bukepo, line 399: Is this otExtAddress typo?

gjc13: No, the OpenThread function signature is
```
void otThreadSetSteeringData(otInstance *aInstance, const otExtAddress *aExtAddress);
```

bukepo: then the variable name should be extAddress?

src/agent/thread_helper.cpp
```diff
@@ -382,18 +382,26 @@ otError ThreadHelper::PermitUnsecureJoin(uint16_t aPort, uint32_t aSeconds)
...
             }
         });
     }
+    else
+    {
+        otExtAddress noneAddress;
+
+        memset(&noneAddress.m8, 0, sizeof(noneAddress.m8));
+        otIp6RemoveUnsecurePort(mInstance, aPort);
+        otThreadSetSteeringData(mInstance, &noneAddress);
```
bukepo, line 403: It seems this will remove pending joiner not added by this API. I think we need a new API in OpenThread to remove a single extended address.

gjc13: Can be addressed in another OpenThread PR.

PR #407, opened 2020-04-24
[doc] update wrong docstring

bukepo:


src/agent/otubus.hpp
```diff
@@ -129,7 +129,7 @@ class UbusServer
      *
      * @param[in]   aContext    A pointer to the ubus context.
      * @param[in]   aObj        A pointer to the ubus object.
-     * @param[in]   aReq        A pointer to the ubus request.
+     * @param[in]   aRequest    A pointer to the ubus request.
```
jwhui, line 132: Update parameter name below. Many other similar cases below.

PR #416, opened 2020-04-30
[dbus] add default initializer

gjc13:


bukepo:
@gjc13 tend to use [zero-initialization](https://docs.w3cub.com/cpp/language/zero_initialization/).

PR #422, opened 2020-05-09
[cmake] add ot-ctl

bukepo:
* add ot-ctl into package
* set log level and region of OpenThread
* use absolute path for web file path
* add more check to docker
* enable all warnings
* update docker entry point


bukepo:
This PR depends on https://github.com/openthread/openthread/pull/4938

src/agent/thread_helper.cpp
```diff
@@ -388,7 +398,7 @@ otError ThreadHelper::PermitUnsecureJoin(uint16_t aPort, uint32_t aSeconds)
             memset(&noneAddress.m8, 0, sizeof(noneAddress.m8));
             if (now >= mUnsecurePortCloseTime[aPort])
             {
-                otIp6RemoveUnsecurePort(mInstance, aPort);
+                (void)otIp6RemoveUnsecurePort(mInstance, aPort);
```
wgtdkp, line 401: Maybe we should `assert` on this?

src/agent/thread_helper.cpp
```diff
@@ -326,15 +326,25 @@ void ThreadHelper::sJoinerCallback(otError aError, void *aThreadHelper)
...
-        otThreadSetEnabled(mInstance, true);
+        otError error = otIp6SetEnabled(mInstance, false);
+
+        if (error != OT_ERROR_NONE)
+        {
+            otbrLog(OTBR_LOG_WARNING, "Failed to stop Thread network: %s", otThreadErrorToString(error));
+        }
+    }
+
+    if (aError != OT_ERROR_NONE)
```
wgtdkp, line 343: merge into `else {...}` above? and we should log "Failed to join Thread network..." before "Failed to stop Thread network...".

.travis/check-docker
```diff
@@ -65,7 +65,17 @@ main() {
...
     sudo lsof -i :8080
-    curl "http://127.0.0.1:8080"
+
+    readonly OTBR_WEB_URL="http://127.0.0.1:8080"
+    readonly OT_MASTER_KEY=00112233445566778899aabbccddeeff
+    readonly OT_XPANID=0011223344556677
+    readonly OT_PANID=0xface
+    readonly OT_AGENT_PASSPHRASE=123456
+    readonly OT_CHANNEL=12
+    readonly OT_NETWORK_NAME=OpenThreadDocker
```
wgtdkp, line 75: nit: move to the beginning or use `lcoal`?

src/agent/thread_helper.cpp
bukepo: This also logs error in line 331.

src/agent/thread_helper.cpp
bukepo: This can be already removed.

src/agent/thread_helper.cpp
wgtdkp: ok, better don't change pass-in `aError`.

Currently, the logging order is not correct if both log messages are printed. 

How about we log `aError` at the beginning of this function and log result of `otIp6SetEnabled` separately?

bukepo:
this PR depends on https://github.com/openthread/openthread/pull/4946

bukepo:
tests passed.

src/agent/CMakeLists.txt
```diff
@@ -51,10 +51,19 @@ target_link_libraries(otbr-agent PRIVATE
     otbr-common
     otbr-utils
 )
+add_dependencies(otbr-agent ot-ctl)
 install(
     TARGETS otbr-agent
     DESTINATION sbin
 )
+if(CMAKE_VERSION VERSION_LESS 3.13)
+install(PROGRAMS $<TARGET_FILE:ot-ctl> DESTINATION sbin)
```
jwhui, line 60: nit: indent body of if-statements?

src/agent/border_agent.cpp
```diff
@@ -280,7 +280,7 @@ void BorderAgent::SetThreadVersion(uint16_t aThreadVersion)
 #endif
 }
 
-void BorderAgent::HandlePSKc(void *aContext, int aEvent, va_list aArguments)
+void BorderAgent::HandlePSKc(void *aContext, OTBR_UNUSED int aEvent, va_list aArguments)
```
jwhui, line 283: nit: if we start using `OTBR_UNUSED`, how about putting after parameter declaration? I think it formats better when parameters are split across multiple lines.

Example from other projects: https://github.com/apple/HomeKitADK/search?q=UNUSED&unscoped_q=HAP_UNUSED

src/common/code_utils.hpp
```diff
@@ -104,4 +104,10 @@
         goto exit;   \
     } while (false)
 
+/**
+ * This flag indicates the parameter may be unused.
+ *
+ */
+#define OTBR_UNUSED __attribute__((unused))
```
jwhui, line 111: Is there a goal to support a wide range of toolchains (e.g. other than clang/gcc)?

bukepo: I tend to add the support when it is needed. Otherwise it may increase maintenance burden, thoughts?

src/agent/border_agent.cpp
bukepo: Ok, I will follow their style.

src/common/code_utils.hpp
jwhui: Are we confident that adding `OTBR_UNUSED` macro to each declaration is sufficient? OpenThread, LwIP, etc. went with a different model (i.e. after declaration) primarily due to compiler support issues.

bukepo: changed to reuse OT_UNUSED_VARIABLE

PR #435, opened 2020-05-25
[third_party] update openthread

gjc13:
Mainly to propagate the fix in https://github.com/openthread/openthread/pull/4994


gjc13:
Pending on https://github.com/openthread/openthread/pull/4996#pullrequestreview-417449979 top fix macos builds.

PR #436, opened 2020-05-25
[continuous-integration] move raspbian check to github actions

gjc13:


.github/workflows/build.yml
```diff
@@ -65,3 +65,19 @@ jobs:
...
+
+  raspbian-check:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Bootstrap
+      env:
+        IMAGE_NAME: 2018-04-18-raspbian-stretch-lite
```
bukepo, line 77: can we retrieve this from the script URL?

Issue #450, opened 2020-06-03
[web]  otbr-web binary command option require unwanted argument


tttttangTH:
After setting up ot-br-posix in Docker and start otbr-web, it will give an `option require an arrgument `  error when checking version number.

```
./build/src/web/otbr-web -v
## ./build/src/web/otbr-web: option requires an argument -- 'v'
## Usage: ./build/src/web/otbr-web [-d DEBUG_LEVEL] [-I interfaceName] [-p port] [-a 
listenAddress] [-v]
```

We know `-v` option is generally used to print version number which does not need an argument.
So the expected output should be like:

```
./build/src/web/otbr-web -v
#0.2.0-6413bac-dirty
```

bukepo:
I think there is a typo. In this project, we use `-v` for verbose, and `-V` for print version. Would you like to submit a PR to fix this issue?

tttttangTH:
Yeah, will fix it. Do I need to change the behavior in another issue/PR?

bukepo:
Either way works for me.

wgtdkp:
@tttttangTH Let's just fix the wrong output of `-v` option of in **`otbr-web`**. We can change `-v` / '-V' of **`otbr-agent`** in another PR.

PR #451, opened 2020-06-03
[web] fix option require error in otbr-web

tttttangTH:
As explained in #450, an option require error appears when we check version number of otbr-web.

The typo is fixed and now `-v` option could correctly print version information.


tttttangTH:
@googlebot I signed it!

googlebot <notifications@github.com> 于2020年6月3日周三 下午12:11写道：

tttttangTH:
Thanks everyone! will be familiar with OpenThread as soon as possible!

PR #460, opened 2020-06-05
[mdns] remove mojo support

gjc13:
Because of the code divergence of chromium mojo and open-source
libchrome. It's no longer possible to maintain an open-source version of
mojo-based mdns.


gjc13:
The android mojo check github action is removed so it cannot report.

PR #476, opened 2020-06-15
[setup] allow specify cmake options via OS env

simonlingoogle:
This PR allows developers to specify extra cmake options for `script/setup` using env `OTBR_OPTIONS`.

### DIFF
--- a/script/_otbr
+++ b/script/_otbr
@@ -53,10 +53,7 @@ otbr_install()
     local otbr_options=()
 
     if [[ ${OTBR_OPTIONS} ]]; then
-        oIFS=$IFS
-        IFS=" "
         read -r -a otbr_options <<< "${OTBR_OPTIONS}"
-        IFS=$oIFS
     fi
 
     otbr_options=(
--- a/script/_otbr
+++ b/script/_otbr
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 #  Copyright (c) 2017, The OpenThread Authors.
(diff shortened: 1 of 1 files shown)


script/_otbr
```diff
@@ -49,7 +49,8 @@ otbr_uninstall()
 
 otbr_install()
 {
-    local otbr_options=(
+    local otbr_options="${OTBR_OPTIONS[@]}"
```
bukepo, line 52: append instead of prepend?

script/_otbr
```diff
@@ -58,7 +59,7 @@ otbr_install()
 
     (mkdir -p "${OTBR_TOP_BUILDDIR}" \
         && cd "${OTBR_TOP_BUILDDIR}" \
-        && cmake -GNinja "${OTBR_TOP_SRCDIR}" "${otbr_options[@]}" \
+        && cmake -GNinja "${OTBR_TOP_SRCDIR}" "${otbr_options[@]}" ${OTBR_OPTIONS[@]} \
```
jwhui, line 62: Double quote array expansions to avoid re-splitting elements.
```suggestion
        && cmake -GNinja "${OTBR_TOP_SRCDIR}" "${otbr_options[@]}" "${OTBR_OPTIONS[@]}" \
```

simonlingoogle:
> Also, I think we want to update `#!/bin/sh` to `#!/bin/bash`.

I think it's a good idea. I will do it in a separate PR because there are many other scripts that need to be updated too.

PR #477, opened 2020-06-16
[style] remove NULL

bukepo:


Android.mk
```diff
@@ -40,20 +40,17 @@ LOCAL_C_INCLUDES := \
     $(LOCAL_PATH)/include \
     $(LOCAL_PATH)/src \
     external/openthread/include \
```
wgtdkp, line 42:
```suggestion
    external/openthread/include
```

Android.mk
```diff
@@ -40,20 +40,17 @@ LOCAL_C_INCLUDES := \
...
-    $(NULL)
```
wgtdkp, line 43: is this intentional?

bukepo: yes, intentional. Any concerns?

PR #480, opened 2020-06-16
[web] verbose invalid path exception

bukepo:
This commit outputs more exception information when the path is invalid.


bukepo:
background: https://groups.google.com/g/openthread-users/c/qnpKjh4_b2U/m/yiCccyn0BgAJ

PR #481, opened 2020-06-17
[script] set -x

bukepo:


script/_initrc
```diff
@@ -28,8 +28,12 @@
 #
 
 set -euo pipefail
-_HERE=$(dirname "$0")
-cd "${_HERE}/.."
+
+if [[ ${OTBR_VERBOSE:-} == 1 ]]; then
+    set -x
```
jwhui, line 33: Would it be useful to unconditionally set -x?

bukepo: I think this is only necessary when we want to debug the script. Otherwise, it just make it harder to find the real useful information.

jwhui: My primary concern is whether a new/casual user will know to set `OTBR_VERBOSE=1` when trying to debug something? There seems to be a lot of setup-related issues occurring with OTBR lately.

simonlingoogle: Indeed, it should be worthwhile for setup to always have `set -e`.

PR #499, opened 2020-06-30
[pretty] add support for shell scripts

jwhui:


jwhui:
Something in the last commit 1a873b5 is causing [check (mDNSResponder)](https://github.com/openthread/ot-br-posix/runs/820720751?check_suite_focus=true) to execute indefinitely. Anyone have any ideas?

/cc/ @bukepo

tests/mdns/test-stop
```diff
@@ -46,16 +47,16 @@ main()
     # stop service
     /bin/kill -USR1 $PID
     sleep 1
-    if [[ "${OTBR_MDNS}" = 'mDNSResponder' ]]; then
-        ! dns_sd_check SingleService _meshcop._udp 'nn=cool xp=ABCDEFGH'
+    if [[ ${OTBR_MDNS} == 'mDNSResponder' ]]; then
+        dns_sd_check SingleService _meshcop._udp 'nn=cool xp=ABCDEFGH' && exit 1
```
jwhui, line 51: Reverting this specific line causes "check (mDNSResponder)" to pass again. I changed this line to address the following shellcheck warning:
```
In tests/mdns/test-stop line 51:
        ! dns_sd_check SingleService _meshcop._udp 'nn=cool xp=ABCDEFGH'
        ^-- SC2251: This ! is not on a condition and skips errexit. Use `&& exit 1` instead, or make sure $? is checked.

Did you mean: 
         dns_sd_check SingleService _meshcop._udp 'nn=cool xp=ABCDEFGH' && exit 1
```

@bukepo , thoughts?

bukepo: Oh, I think the change is not identical to the original behavior. Maybe the following:
```suggestion
        (! dns_sd_check SingleService _meshcop._udp 'nn=cool xp=ABCDEFGH') || exit 1
```

jwhui: @bukepo , I pushed your suggested change and same result - "check (mDNSResponder)" executes indefinitely.

bukepo: I'll debug locally.

bukepo:
@gjc13 it seems the script get stuck at dbus-client test. Could you take a look?

tests/mdns/test-stop
bukepo: It seems deleting a mDNS service from mDNSResponder doesn't take effect immediately now. I tried adding sleep 200s before the check works.

jwhui: @bukepo , thanks for looking into this. Is waiting 200s before the service is actually removed the expected behavior? Is the suggestion to sleep 200 seconds before executing `dns_sd_check`?

bukepo: I don't think it is expected. But I also tested on macOS, closing the client doesn't remove the service from the server. You can reproduce this issue by:

1. Register the service
```bash
# Terminal 1
dns-sd -R SingleService _meshcop._udp local 12324
```
2. Query the service

```bash
# Terminal 2
dns-sd -L SingleService _meshcop._udp local
```
3. Stop the service by CTRL-C in terminal 1

4. Stop querying in terminal 2 and query the service again with the same command, the service is still listed.

My observation is the service disappears after about 3 minutes. That's why I suggest sleep 200s before the dns_sd_check.

jwhui:
> LGTM👍

@bukepo , thanks for your help on this!

PR #503, opened 2020-07-02
[log] fix openthread logging not availbe upon init

gjc13:
Fixes https://github.com/openthread/openthread/issues/5177

### DIFF
--- a/src/agent/ncp_openthread.cpp
+++ b/src/agent/ncp_openthread.cpp
@@ -34,6 +34,7 @@
 
 #include <openthread/cli.h>
 #include <openthread/dataset.h>
+#include <openthread/logging.h>
 #include <openthread/tasklet.h>
 #include <openthread/thread.h>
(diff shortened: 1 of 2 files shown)


src/agent/ncp_openthread.cpp
```diff
@@ -72,7 +73,34 @@ ControllerOpenThread::~ControllerOpenThread(void)
...
+        level = OT_LOG_LEVEL_NOTE;
+        break;
+    case OTBR_LOG_INFO:
+        level = OT_LOG_LEVEL_INFO;
+        break;
+    case OTBR_LOG_DEBUG:
+        level = OT_LOG_LEVEL_DEBG;
+        break;
+    default:
+        ExitNow(error = OTBR_ERROR_OPENTHREAD);
```
bukepo, line 100:
```suggestion
        assert(false);
```

PR #505, opened 2020-07-03
[dbus] fix handler registration under external reset

gjc13:
This change allows the dbus agent to send device role changed signals under unexpected external resets. (e.g. ot-ctl factoryreset and internal OpenThread resets)

### DIFF
--- a/src/agent/ncp_openthread.cpp
+++ b/src/agent/ncp_openthread.cpp
@@ -213,6 +213,10 @@ void ControllerOpenThread::Reset(void)
     otInstanceFinalize(mInstance);
     otSysDeinit();
     Init();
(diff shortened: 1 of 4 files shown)


src/agent/ncp_openthread.cpp
```diff
@@ -213,6 +213,10 @@ void ControllerOpenThread::Reset(void)
     otInstanceFinalize(mInstance);
     otSysDeinit();
     Init();
+    for (auto &handler : mResetHandlers)
```
bukepo, line 216: Do we need multiple notifications?

gjc13: Just for possibility of extension if we do support multiple IPC mechanisms.

PR #506, opened 2020-07-03
[mdns] fix avahi size check

gjc13:


src/mdns/mdns_avahi.cpp
```diff
@@ -517,7 +517,7 @@ otbrError PublisherAvahi::PublishService(uint16_t aPort, const char *aName, cons
         int         rval;
         const char *value       = va_arg(args, const char *);
         size_t      valueLength = va_arg(args, size_t);
-        size_t      needed      = sizeof(AvahiStringList) + strlen(name) + valueLength;
+        size_t      needed      = sizeof(AvahiStringList) + strlen(name) + valueLength + 1;
```
jwhui, line 520: nit: add comment to explain the `+ 1` and why the `< sizeof(buffer)` check below (vs. `<= sizeof(buffer)` is not sufficient.

gjc13: Checked the logic and found that the`<= sizeof(buffer)` suffices. Added the comments.

src/mdns/mdns_avahi.cpp
```diff
@@ -517,9 +517,10 @@ otbrError PublisherAvahi::PublishService(uint16_t aPort, const char *aName, cons
         int         rval;
         const char *value       = va_arg(args, const char *);
         size_t      valueLength = va_arg(args, size_t);
-        size_t      needed      = sizeof(AvahiStringList) + strlen(name) + valueLength;
+        // +1 for the size of "=", avahi doesn't need '\0' at the end of the entry
+        size_t needed = sizeof(AvahiStringList) + strlen(name) + valueLength + 1;
```
bukepo, line 521:
```suggestion
        size_t needed = sizeof(AvahiStringList) - sizeof(AvahiStringList::text) + strlen(name) + valueLength + 1;
```

PR #513, opened 2020-07-08
[web] fix reflected XSS

Moandor-y:
This PR fixes a reflected XSS vulnerability of `otbr-web`. The exception message may contain user inputs so needs to escape sensitive characters.

Reproducing:
1.  Start `otbr-web` service on `localhost`.
1.  Run:
    ```bash
    curl 'localhost/no_such_file.pl?"><script>alert(111);</script>'
    ```
    Then the response will contain executable scripts.

### DIFF
--- a/src/web/web-service/web_server.cpp
+++ b/src/web/web-service/web_server.cpp
@@ -59,8 +59,8 @@
 
 namespace otbr {
 namespace Web {
-namespace {
-void EscapeHtml(std::string &content)
+
+static void EscapeHtml(std::string &content)
 {
     std::string output;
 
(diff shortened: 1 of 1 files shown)


src/web/web-service/web_server.cpp
```diff
@@ -59,6 +59,40 @@
 
 namespace otbr {
 namespace Web {
+namespace {
```
bukepo, line 62: use static instead of anonymous namespace?

bukepo:
fixes #157

src/web/web-service/web_server.cpp
Moandor-y: Done.

PR #514, opened 2020-07-08
[cmake] move minimum version to 3.10.2

jwhui:
To support automated armv7 docker builds.

### DIFF
--- a/src/dbus/common/CMakeLists.txt
+++ b/src/dbus/common/CMakeLists.txt
@@ -38,6 +38,7 @@ target_link_libraries(otbr-dbus-common PUBLIC
     otbr-common
     openthread-ftd
     openthread-posix
-    ${DBUS_LINK_LIBRARIES}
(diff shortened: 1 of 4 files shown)


src/dbus/common/CMakeLists.txt
```diff
@@ -38,6 +38,6 @@ target_link_libraries(otbr-dbus-common PUBLIC
     otbr-common
     openthread-ftd
     openthread-posix
-    ${DBUS_LINK_LIBRARIES}
+    ${DBUS_LIBRARIES}
```
bukepo, line 41: Should we add DBUS_LINK_LIBRARIES, in case these libraries are not in the standard path. I think this is common on macOS. The homebrew prefix could be under home.

bukepo: sorry for the typo, it should be `DBUS_LIBRARY_DIRS`

https://cmake.org/cmake/help/v3.10/module/FindPkgConfig.html

PR #517, opened 2020-07-09
[joiner] fix joining behavior

gjc13:
* Should not close interface upon join failure
* Fix wrong API call to start Thread
* Add tests for joining

### DIFF
--- a/src/agent/thread_helper.cpp
+++ b/src/agent/thread_helper.cpp
@@ -317,13 +317,12 @@ void ThreadHelper::JoinerCallback(otError aError)
     if (aError != OT_ERROR_NONE)
     {
         otbrLog(OTBR_LOG_WARNING, "Failed to join Thread network: %s", otThreadErrorToString(aError));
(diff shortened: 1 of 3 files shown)


src/agent/thread_helper.cpp
```diff
@@ -317,13 +317,12 @@ void ThreadHelper::JoinerCallback(otError aError)
...
     {
         otbrLog(OTBR_LOG_WARNING, "Failed to join Thread network: %s", otThreadErrorToString(aError));
-        LogOpenThreadResult("Stop Thread network", otIp6SetEnabled(mInstance, false));
         mJoinerHandler(aError);
         mJoinerHandler = nullptr;
     }
     else
     {
-        LogOpenThreadResult("Start Thread network", otIp6SetEnabled(mInstance, true));
+        LogOpenThreadResult("Start Thread network", otThreadSetEnabled(mInstance, true));
```
bukepo, line 325: I think it's better to defer calling such APIs in case it causes some side effect inside this callback.

PR #522, opened 2020-07-14
[build] fix build with BUILD_SHARED_LIBS=ON

bukepo:


bukepo:
@pdgendt could you give it a try?

PR #519, opened 2020-07-10
[web] topology visualization

tttttangTH:
It is the first PR of OpenThread Network Manager.

In general, it implements a new module -- OTBR-REST, which is directly embedded in OTBR-AGENT and provides RESTful API for a new feature of WEB UI -- Topology Visualization.  

Topology Visualization aims to collect diagnostic information of all nodes in current network, visualize their relations and provide the display of all the details.

The frontend part is implemented based on the frontend of OTBR-WEB for the time being. We will then implement all the functionality OTBR-WEB provides and transplant the frontend static file of OTBR-WEB in this module. More details of the second PR is in #537 .

PR #528, opened 2020-07-28
[web] add REST APIs third parties (DON'T REVIEW)

tttttangTH:
a draft version for pre-review by @wgtdkp  and @simonlingoogle. Third party library is included.


src/rest/parser.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+class Parser
+{
+public:
+    Parser(Request *aRequest);
+    void Init();
+    void Process(const char *aBuf, int aLength);
+
+private:
+    std::shared_ptr<void> mParser;
```
tttttangTH, line 52: Is this a good practice? @wgtdkp

wgtdkp: obviously not. Why is this needed?

wgtdkp: This PR should equals https://github.com/openthread/ot-br-posix/pull/519 + third parties. Any comments or question should happen in https://github.com/openthread/ot-br-posix/pull/519, we simply don't review this PR, but for you to make sure GitHub action tests can success.

tttttangTH:
@wgtdkp @simonlingoogle 

Compared to #519, this PR have all three third party libraries included and it is created for checking whether OTBR-REST could pass Github Action test. Now these three libraries have been added within #519, so we no longer need it. It's time to close it.

Thanks for your review!

PR #534, opened 2020-08-06
[third_party] add third party library d3js 

tttttangTH:
This PR add third party library [d3js](https://github.com/d3/d3) into third_party directory. 

As is described by itself:

> D3.js is a JavaScript library for manipulating documents based on data. D3’s emphasis on web standards gives you the full capabilities of modern browsers without tying yourself to a proprietary framework, combining powerful visualization components and a data-driven approach to DOM manipulation.

We need this library for the frontend of OTBR-REST to build the topology graph ( a SVG graph drawn by d3js, nodes are represented by circles, triangles or polygons and their connections is represented by a line) for a OpenThread network


third_party/d3js/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+https://github.com/d3/d3/releases
+
+## Version
+
+5.16.0
+
+## Commit
+
+[7244e45e68af39c519f76667a03039d5b24dd453](https://github.com/d3/d3/releases/tag/v5.16.0)
```
wgtdkp, line 13: If the commit points to the 5.16.0 tag, I think we can just remove the commit info, and have the version above being a link.

tttttangTH: 👌

bukepo:
use git submodule?

tttttangTH:
> use git submodule?
@bukepo 
we think that directly including third_party library source code could make it easier when it is used in product environment without the need of making a mirroring of them. Also, we just use basic APIs of these libraries to format and form data and not expect any new features of them.

wgtdkp:
> > use git submodule?
> 
> @bukepo
> we think that directly including third_party library source code could make it easier when it is used in product environment without the need of making a mirroring of them. Also, we just use basic APIs of these libraries to format and form data and not expect any new features of them.

@tttttangTH Talked with @bukepo that we may don't need to put web stuff into products, so use gitsubmodules should be okay. Please also keep `Android.mk` unchanged.

tttttangTH:
> > > use git submodule?
> > 
> > 
> > @bukepo
> > we think that directly including third_party library source code could make it easier when it is used in product environment without the need of making a mirroring of them. Also, we just use basic APIs of these libraries to format and form data and not expect any new features of them.
> 
> @tttttangTH Talked with @bukepo that we may don't need to put web stuff into products, so use gitsubmodules should be okay. Please also keep `Android.mk` unchanged.

Ok, will do.

tttttangTH:
> > > use git submodule?
> > 
> > 
> > @bukepo
> > we think that directly including third_party library source code could make it easier when it is used in product environment without the need of making a mirroring of them. Also, we just use basic APIs of these libraries to format and form data and not expect any new features of them.
> 
> @tttttangTH Talked with @bukepo that we may don't need to put web stuff into products, so use gitsubmodules should be okay. Please also keep `Android.mk` unchanged.

So cJSON and http-parser should be included using git submodule.
 
However, d3js is a js library, it seems how ot-br-posix use angular, angular-material, mdl(they are all js library) is to directly  include their js source file, should we include d3js using the same method ? @wgtdkp @bukepo

wgtdkp:
> > > > use git submodule?
> > > 
> > > 
> > > @bukepo
> > > we think that directly including third_party library source code could make it easier when it is used in product environment without the need of making a mirroring of them. Also, we just use basic APIs of these libraries to format and form data and not expect any new features of them.
> > 
> > 
> > @tttttangTH Talked with @bukepo that we may don't need to put web stuff into products, so use gitsubmodules should be okay. Please also keep `Android.mk` unchanged.
> 
> So cJSON and http-parser should be included using git submodule.
> 
> However, d3js is a js library, it seems how ot-br-posix use angular, angular-material, mdl(they are all js library) is to directly include their js source file, should we include d3js using the same method ? @wgtdkp @bukepo

@tttttangTH I don't see why git submodule can't work in this situation.

tttttangTH:
@wgtdkp @simonlingoogle @bukepo

The same as #535 and #536, this library will be added using git submodule within #519. It's time to close it.

Thanks!

PR #535, opened 2020-08-06
[third_party] add third party library cJSON 

tttttangTH:
This PR adds third party library [cJSON](https://github.com/DaveGamble/cJSON) into third_party directory.

As is described by itself:

> a Ultralightweight JSON parser in ANSI C, cJSON aims to be the dumbest possible parser that you can get your job done with. It's a single file of C, and a single header file. It is a parser that tries to share the same philosophy as JSON itself. Simple, dumb, out of the way.

We need this library to parse and build Json object within a HTTP response and request in OTBR-REST. Compared to the current Json parser -- jsoncpp used in OTBR-WEB, the APIs provided by cJSON are more readable and easy to use. Besides, cJSON is very easy to include for it only has a header file and a source file and more faster. 

After OTBR-WEB is finally refactored, no other module will use jsoncpp. These two Json tools will only co-exist for a while and there will only be one left finally.


bukepo:
use git submodule?

tttttangTH:
@wgtdkp @simonlingoogle @bukepo

This PR is handled the same way as #536 and has been closed.

Thanks!

PR #536, opened 2020-08-06
[third_party] add third party library http-parser

tttttangTH:
This PR adds third party library [http-parser](https://github.com/nodejs/http-parser) into third_party directory. 

As is described by itself:

> This is a parser for HTTP messages written in C. It parses both requests and responses. It does not make any syscalls nor allocations, it does not buffer data, it can be interrupted at anytime. Depending on your architecture.

We need this library for the HTTP server of OTBR-REST to construct a request. To integrate the HTTP server of OTBR-REST into OTBR-AGENT, we implement the server start from socket IO, then a request parser is needed to parse the data from read buffer and then form a request. Therefore, we choose this library to do this job for us.

### DIFF
--- a/third_party/http-parser/README.md
+++ b/third_party/http-parser/README.md
@@ -6,11 +6,7 @@ https://github.com/nodejs/http-parser/releases
 
 ## Version 
 
-2.9.4
-
-## Commit
-
-[2343fd6b5214b2ded2cdcf76de2bf60903bb90cd](https://github.com/nodejs/http-parser/releases/tag/v2.9.4)
+[2.9.4](https://github.com/nodejs/http-parser/releases/tag/v2.9.4)
 
 ## License
 
--- a/third_party/http-parser/README.md
+++ b/third_party/http-parser/README.md
@@ -1,0 +1,27 @@
+# http-parser
+
+## URL
+
+https://github.com/nodejs/http-parser/releases
+
+## Version 
+
+2.9.4
+
+## Commit
+
+2343fd6b5214b2ded2cdcf76de2bf60903bb90cd
+
+## License
+
+MIT License
+
+## License File
+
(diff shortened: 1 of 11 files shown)


wgtdkp:
@sim

> LGTM 
> 
> BTW, should we run the test in our github action? @tttttangTH @wgtdkp

@simonlingoogle I think we this PR should only include the 3rd party itself. We can run GitHub Actions after merging these 3rd party PRs.

wgtdkp:
@tttttangTH Could you add a `README.md` file under exactly `third_party/http-parser` to describe the original repo/link and version of this library?

simonlingoogle:
> LGTM 
> 
> FYI, this is the http parser used by `node.js`, so I believe the quality is guaranteed.

Sounds great.

tttttangTH:
> @tttttangTH Could you add a `README.md` file under exactly `third_party/http-parser` to describe the original repo/link and version of this library?

Ok, will do this for all of them.

third_party/http-parser/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+https://github.com/nodejs/http-parser/releases
+
+## Version 
+
+2.9.4
+
+## Commit
+
+2343fd6b5214b2ded2cdcf76de2bf60903bb90cd
```
wgtdkp, line 13: It would be better if this commit hash can be a link to the release we are using.

tttttangTH: Sound good!

bukepo:
use git submodule?

tttttangTH:
Thanks for @wgtdkp @simonlingoogle @bukepo review, these three libraries have been added using git submodule and will be put together within one PR with topology visualization #519. 

This PR has been closed. Thanks again!

PR #546, opened 2020-08-20
[web] unify MasterKey and NetworkKey

tttttangTH:
As discussed in #545 , this PR changes all `NetworkKey` to `MasterKey` in Web module, so the naming of this terminology is consistent with the Thread Specification.


bukepo (approved):
LGTM:+1:

PR #554, opened 2020-09-11
[openthread] turn on 1.2 features by default

simonlingoogle:
I think OTBR should enable Thread 1.2 features by default since Thread 1.2 will be commonly used.


librasungirl:
Thread 1.2 OTBR with BBR does have more resource requirements than the one without BBR - specifically the ndproxy table and ma table. Commercial BR, which might have limited resources, may prefer to stay at Thread 1.1 version. 
I would suggest to leave the flexibility of thread version choices for user and enable the 1.2 specific feature flags when 1.2 version is specified.

Thoughts?

simonlingoogle:
> Thread 1.2 OTBR with BBR does have more resource requirements than the one without BBR - specifically the ndproxy table and ma table. Commercial BR, which might have limited resources, may prefer to stay at Thread 1.1 version.
> I would suggest to at least leave the flexibility of thread version choices for user and enable the 1.2 specific feature flags when 1.2 version is specified.
> 
> Thoughts?

Agree that it should be configurable to enable or disable 1.2 features. 
We can remove the `FORCE` so that these options can be turned off. 
But I would prefer to enable 1.2 features by default. Thoughts? @librasungirl @bukepo

librasungirl:
Since 1.2 feature on public repo are not tested thoroughly yet, I may prefer a bit to defer the default 1.2 features when 1.2 feature becomes more mature.

simonlingoogle:
> Since 1.2 feature on public repo are not tested thoroughly yet, I may prefer a bit to defer the default 1.2 features until 1.2 features become more mature.

Let's close this PR for now and reopen when the proper time has come.

PR #565, opened 2020-09-22
[dbus] add set link supported channel mask

gjc13:
This allows applications to set the channel mask based on its provisioned country.


gjc13:
Replaced by https://github.com/openthread/ot-br-posix/pull/567

PR #575, opened 2020-09-30
[github-actions] use docker/setup-buildx-action@v1

jwhui:


simonlingoogle:
docker/setup-buildx-action@v1 not work here?

jwhui:
> docker/setup-buildx-action@v1 not work here?

@simonlingoogle , it works, just need to do a little extra to enable qemu.

Issue #587, opened 2020-10-14
OTBR_REST=ON: compile error in hash table of status codes


markus-becker-tridonic-com:
**Describe the bug** 

When compiling ot-br-posix on an older OpenWrt toolchain and enabling `-DOTBR_REST=ON` I am getting the following compile error:

```
In file included from /usr/src/toolchain/staging_dir/toolchain-arm_cortex-a9_gcc-4.8-linaro_glibc-2.19_eabi/arm-openwrt-linux-gnueabi/include/c++/4.8.3/bits/hashtable.h:35:0,
                 from /usr/src/toolchain/staging_dir/toolchain-arm_cortex-a9_gcc-4.8-linaro_glibc-2.19_eabi/arm-openwrt-linux-gnueabi/include/c++/4.8.3/unordered_map:47,
                 from /usr/src/toolchain/feeds/otbr/src/rest/resource.hpp:37,
                 from /usr/src/toolchain/feeds/otbr/src/rest/connection.hpp:41,
                 from /usr/src/toolchain/feeds/otbr/src/rest/rest_web_server.hpp:37,
                 from /usr/src/toolchain/feeds/otbr/src/rest/rest_web_server.cpp:29:
/usr/src/toolchain/staging_dir/toolchain-arm_cortex-a9_gcc-4.8-linaro_glibc-2.19_eabi/arm-openwrt-linux-gnueabi/include/c++/4.8.3/bits/hashtable_policy.h: In instantiation of 'struct std::__detail::_Hash_code_base<otbr::rest::HttpStatusCode, std::pair<const otbr::rest::HttpStatusCode, std::basic_string<char> >, std::__detail::_Select1st, std::hash<otbr::rest::HttpStatusCode>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, true>':
[...]

PR #590, opened 2020-10-14
[web] fix OTBR-REST build fail in GCC 4.8.3

tttttangTH:
As discussed in #587 ,  we avoid using `std::hash` in OTBR-REST. 

Build check did not disclose this issue because OTBR-REST is not set as default enabled in test,  we also fix this in this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,7 +90,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        rest: ["rest", ""]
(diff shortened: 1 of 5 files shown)


wgtdkp (commented):
@tttttangTH  there are significant coverage decrease.

tttttangTH:
> @tttttangTH there are significant coverage decrease.

I will remove those unused HTTPStatusCode and see the result.

wgtdkp (approved):
Thanks! LGTM :+1:

jwhui (approved):

PR #532, opened 2020-08-04
[backbone-router] add Backbone multicast routing

simonlingoogle:
This PR uses `smcroute` to implement multicast forwarding for MLR Listeners.

Major changes include:
- [x] Implement multicast routing using [smcroute](https://github.com/troglobit/smcroute)
  - [x] Install/uninstall smcroute in `script/setup`
  - [x] Add OTBR_BACKBONE_ROUTER option and `otbr-backbone-router` library
- [x] Add new Dockerfile ARGs: `BACKBONE_ROUTER` and `NAT64`
- [x] **Tested by https://github.com/openthread/openthread/pull/5578 (merged)**

**Suggested merge steps:** @jwhui 
  1. Merge https://github.com/openthread/openthread/pull/5578 using latest commit of this PR, passing all checks
  2. Merge this PR using OpenThread master, passing all checks, after https://github.com/openthread/openthread/pull/5578 is merged

PR #537, opened 2020-08-12
[web] refactor OTBR-WEB

tttttangTH:
This PR refactors OTBR web with the new REST APIs:
- Remove standalone `otbr-web` binary.
- Add REST APIs that are required by the OTBR web page.
- Host the OTBR web page with Nginx.
- Remove unused third-party libraries.

For the new implementation, HTTP requests for static files are served by Nginx directly while requests for dynamic resources are served by OTBR agent REST APIs (requests are forwarded to OTBR agent by Nginx).

PR #593, opened 2020-10-19
[web] fix rest-test issues

tttttangTH:
This PR mainly fixes some issues on current tests of OTBR-REST, and makes some preparations for #532 .

- set timeout for rest-test in case it takes too much time 
- some modifications on rest-test principles & HTTP response, so this test will work properly when more error code of underlying APIs are defined
- bind `ReceiveDiagnosticGetCallback` to `resource` each time when `diagnostics API` is called(consistent with rest-test modifications and more logistic)


tttttangTH:
@simonlingoogle @wgtdkp  any comments?

tests/rest/test_rest.py
```diff
@@ -37,11 +37,16 @@
...
 
+    try:    
+        response = urllib.request.urlopen(urllib.request.Request(url))
+        body = response.read()
+        data = json.loads(body)
+        result[index] = data
+
+    
+    except urllib.error.HTTPError as e:
+        print(e.code)
```
simonlingoogle, line 49: why is raise exception not good?

src/rest/resource.cpp
```diff
@@ -518,31 +513,22 @@ void Resource::UpdateDiag(std::string aKey, std::vector<otNetworkDiagTlv> &aDiag
 
 void Resource::Diagnostic(const Request &aRequest, Response &aResponse) const
 {
-    otbrError error = OTBR_ERROR_NONE;
+    otThreadSetReceiveDiagnosticGetCallback(mInstance, &Resource::DiagnosticResponseHandler,
```
simonlingoogle, line 516: why setting callback for each `Diagnostic` is necessary?

tests/rest/test_rest.py
```diff
@@ -37,11 +37,16 @@
...
-    data = json.loads(body)
-    result[index] = data
 
+    try:    
+        response = urllib.request.urlopen(urllib.request.Request(url))
+        body = response.read()
+        data = json.loads(body)
+        result[index] = data
+
+
```
wgtdkp, line 47:
```suggestion
```

tests/rest/test_rest.py
wgtdkp: +1, why catching the exception?

src/rest/resource.cpp
```diff
@@ -518,31 +513,22 @@ void Resource::UpdateDiag(std::string aKey, std::vector<otNetworkDiagTlv> &aDiag
...
-    {
-        aResponse.SetStartTime(steady_clock::now());
-        aResponse.SetCallback();
-    }
-    else
-    {
-        ErrorHandler(aResponse, HttpStatusCode::kStatusInternalServerError);
-    }
+    aResponse.SetStartTime(steady_clock::now());
+    aResponse.SetCallback();
```
wgtdkp, line 531: For now, what happens if we fail in this function? Simply ignore the error?

wgtdkp (approved):
LGTM :+1:

tests/rest/test_rest.py
tttttangTH: I expect to just record the error here then raise one exception in main thread(on top of all subprocess), so for each subprocess, its job is either to get data(success) or to print error log(fail).

src/rest/resource.cpp
tttttangTH: For example, If `factoryreset` is called, the callback may be bind to another function. We should ensure that each time `diagnostic` API in `resource` is called, the response message can be collected by our `resource` handler.

src/rest/resource.cpp
tttttangTH: I found that It may fail when node is detached or no buffer for message. 

So I think one solution is to define more HTTP status code according to these newly added error code, I think I could do it after #532 is merged.

Another solution is like this, we ignore all errors(more simple, but still reasonable). 

For one thing, it could address the `no buffer` problem - although we fail in this API call, if we have diagnostic information(received in 4s but due to another call) left in `resource`,  this information is still valid and could be sent as response.
For another thing,  we could just view empty response as  "some issues happen, we can't get anything now".

src/rest/resource.cpp
simonlingoogle: After `factoryreset`, `Resource::Init` will run again and sets up the callback ? Is the original code causing real issues?

simonlingoogle:
> This PR mainly fixes some issues on current tests of OTBR-REST, and makes some preparations for #532 .
> 
> * set timeout for rest-test in case it takes too much time
> * some modifications on rest-test principles & HTTP response, so this test will work properly when more error code of underlying APIs are defined
> * bind `ReceiveDiagnosticGetCallback` to `resource` each time when `diagnostics API` is called(consistent with rest-test modifications and more logistic)

Thanks for the fix. I merged this PR to #532 and it did pass all the checks. 
But I am not clear what exactly does this PR fix. Any detailed explanation? @tttttangTH

src/rest/resource.cpp
wgtdkp: > Simply ignore the error?

Is this true for your changes? I didn't see why we cannot handle the possible failures in this PR. It looks to me that keeping 
```C+++
else
    {
        ErrorHandler(aResponse, HttpStatusCode::kStatusInternalServerError);
    }
```
should just work. Why did you remove this error handler?

wgtdkp, line 531: Don't need to address in this PR: the function name `SetCallback` is bad... I would expect this function accepts a callback function as an argument but it is actually a flag indicates whether we have/enable callback. Can we rename to `SetHasCallback`?

tests/rest/test_rest.py
wgtdkp: > then raise one exception in main thread

@tttttangTH  I didn't follow how this is implemented. 

Let me ask a question: Will the CI fail if a `urllib.error.HTTPError` exception is raised in this method?

simonlingoogle:
Labelled as P2 since #532 depends on this.

tests/rest/test_rest.py
tttttangTH: Yes, CI will fail.

src/rest/resource.cpp
tttttangTH: `Resource::Init` will only be run once when we start otbr-agent, so `factoryreset` will set the callback to the default(seems a handler for CLI) . will explain the detail later.

src/rest/resource.cpp
tttttangTH: because if we call `otThreadSendDiagnosticGet` too many times within a specific period, we will get a `no buffer` error for this call( for example, when we send 10 requests for `diagnostics` concurrently , the `no buffer` error usually occurs). 

But actually I prefer not viewing it as an error for HTTP response because the reason for calling `otThreadSendDiagnosticGet` each time when the server received a request for `diagnostics` is to update the information we maintained(we have a hash table to record all the diagnostic information we received within 4 seconds). If we have a 'no buffer' error, it means we have call the API many times recently so information in the hash table is almost the latest.

I know the right way to deal with this problem is to make an exception for `no buffer` rather than remove the error handler.  I expect to do this after #532 because I think I may need to rewrite error handler of all RESTful API according to the modifications in #532 . So here I just ignore it currently.

tttttangTH: > > This PR mainly fixes some issues on current tests of OTBR-REST, and makes some preparations for #532 .
> > 
> > * set timeout for rest-test in case it takes too much time
> > * some modifications on rest-test principles & HTTP response, so this test will work properly when more error code of underlying APIs are defined
> > * bind `ReceiveDiagnosticGetCallback` to `resource` each time when `diagnostics API` is called(consistent with rest-test modifications and more logistic)
> 
> Thanks for the fix. I merged this PR to #532 and it did pass all the checks.
> But I am not clear what exactly does this PR fix. Any detailed explanation? @tttttangTH

 After #532 is applied, several issues occurs.

- some Openthread API call fail for `detached`.  So I add basic configuration for the node before rest-test start.
- one  Openthread API still fails for `no buffer`. So I ignore this error currently (It does fail for an Openthread API call but I don't think it is a failure for our `diagnostics` API ), and will complete this part later after #532, 

- rest-test takes too much time, gives too much log.  It's strange, seems it is result from your PR, when I start the node, then stop `otbr-agent` and restart it again, many logs keep appearing and never stop. This is why we have too much logs in rest-test.  After `factoryreset`, it is back to normal. so I also add a `factoryreset` operation to the node before rest-test which leads to our discussion on callback bind.

tttttangTH:
> > This PR mainly fixes some issues on current tests of OTBR-REST, and makes some preparations for #532 .
> > 
> > * set timeout for rest-test in case it takes too much time
> > * some modifications on rest-test principles & HTTP response, so this test will work properly when more error code of underlying APIs are defined
> > * bind `ReceiveDiagnosticGetCallback` to `resource` each time when `diagnostics API` is called(consistent with rest-test modifications and more logistic)
> 
> Thanks for the fix. I merged this PR to #532 and it did pass all the checks.
> But I am not clear what exactly does this PR fix. Any detailed explanation? @tttttangTH


 After #532 is applied, several issues occur.

- some Openthread API call fail for `detached`.  So I add basic configuration for the node before rest-test starts.
- one  Openthread API still fails for `no buffer`. So I ignore this error currently (It does fail for an Openthread API call but I don't think it is a failure for our `diagnostics` API ), and will rewrite the error handler altogether after #532 according to the new return value principle.

- rest-test takes too much time, gives too much log.  It's strange, seems it is result from your PR, when I start the node, then stop `otbr-agent` and restart it again, many logs keep appearing and never stop. This is why we have too much logs in rest-test.  After `factoryreset`, it is back to normal. so I also add a `factoryreset` operation to the node before rest-test which leads to our discussion on callback.

src/rest/resource.cpp
simonlingoogle: > `Resource::Init` will only be run once when we start otbr-agent, then `factoryreset` will set the callback to the default(seems a handler for CLI), I am not sure whether there is another approach that set `otThreadSetReceiveDiagnosticGetCallback` to another function. do you think we do this each time before `diagnostics` API is called is acceptable? or is there an approach that we could detect whether `factoryreset` is called?

I think `factoryreset` re-exec the whole process, so the process restarts as if it was just launched normally. We should not blame `factoryreset` for any issue. If multiple components of ot-br-posix is using `otThreadSetReceiveDiagnosticGetCallback`, maybe they should coordinate with each other. 
Should we use `EventEmitter` so that ot-br-posix calls `otThreadSetReceiveDiagnosticGetCallback` just once and let other components subscribe to the corresponding event using `EventEmitter::On`? @gjc13 @tttttangTH

src/rest/resource.cpp
simonlingoogle: Why is `no buffer error` not good for such situation? I think `no buffer error` was exactly what happened and should be reflected by the HTTP response. 

Maybe we can return `507 Insufficient Storage` for `no buffer error`. 

But if the test is sending too much requests and is expecting `no buffer error` issue to happen sometimes, the test client can conclude with success even if there are some no buffer errors.

Thoughts? @tttttangTH @wgtdkp

tests/rest/test_rest.py
wgtdkp: > Yes, CI will fail.

@tttttangTH Could you explain how it works?
I don't see how it can make the CI fail, you are just printing a message...

src/rest/resource.cpp
wgtdkp: @tttttangTH You should not just remove the error handler. **You can at least add a log for this error**.

> I know the right way to deal with this problem is to make an exception for no buffer rather than remove the error handler. I expect to do this after #532 because I think I may need to rewrite error handler of all RESTful API according to the modifications in #532 . 

I don't see why we need to wait for #532. Are there any REST feature that depends on #532?

wgtdkp: Another point, I think `HasCallback` can be effectively represent by the nullness of a `mCallback` (a function pointer) member. If `mCallback` is `nullptr`, then we have a Callback, otherwise we don't have. Why do we need something like `HasCallback` ?

wgtdkp:
> * one  Openthread API still fails for `no buffer`. So I ignore this error currently (It does fail for an Openthread API call but I don't think it is a failure for our `diagnostics` API ), and will rewrite the error handler altogether after #532 according to the new return value principle.

@tttttangTH I think we are wrong about the principle of testing. To my understanding, you are just removing the error check in REST module to get #532 passed. But the purpose of tests is to find out bugs. We are not sure if #532 has bugs or not, so it could be #532 that causes existing testcases to fail. So I think you should not change the tests to make PRs happy. Otherwise, why bother have those tests?

It could be your tests issues that don't handle unexpected error if this error doesn't indicate a bug, it is even more likely that the new PR just has bugs. Before @simonlingoogle figuring out why #532 fails the tests, we should hold this PR.

tests/rest/test_rest.py
tttttangTH: There is an array `result`  which is used to record the response of each thread. if one thread gets HTTP error, the value in `result` will still be `None`, we will check this in main thread.

src/rest/resource.cpp
tttttangTH: Ok, will try to do like this.

src/rest/resource.cpp
tttttangTH: Have't found any other feature depend on #532 , but it seems we won't catch `no buffer`  at `otThreadSetReceiveDiagnosticGetCallback` before #532 is applied.

I noticed `no buffer` before when when I wrote this API, at that time, `no buffer` was just an `Info`, and I have chosen to ignore it at that time. But It‘s thoughtless for me to just remove the error handler here.

tttttangTH:
> > * one  Openthread API still fails for `no buffer`. So I ignore this error currently (It does fail for an Openthread API call but I don't think it is a failure for our `diagnostics` API ), and will rewrite the error handler altogether after #532 according to the new return value principle.
> 
> @tttttangTH I think we are wrong about the principle of testing. To my understanding, you are just removing the error check in REST tests to get #532 passed. But the purpose of tests is to find out bugs. We are not sure if #532 has bugs or not, so it could be #532 that causes existing testcases to fail. So I think you should not change the tests to make PRs happy. Otherwise, why bother have those tests?
> 
> Before figuring out why #532 fails the tests, we should hold this PR.

Ok, from my side, if we put aside the `no buffer` issue,  I think another issue is if there is no `factoryreset` after restarting otbr-agent,  rest-check always fails and output too many logs.  I am not sure whether no `factoryreset` is acceptable, if yes, this issue should be addressed by #532

wgtdkp:
> > > * one  Openthread API still fails for `no buffer`. So I ignore this error currently (It does fail for an Openthread API call but I don't think it is a failure for our `diagnostics` API ), and will rewrite the error handler altogether after #532 according to the new return value principle.
> > 
> > 
> > @tttttangTH I think we are wrong about the principle of testing. To my understanding, you are just removing the error check in REST tests to get #532 passed. But the purpose of tests is to find out bugs. We are not sure if #532 has bugs or not, so it could be #532 that causes existing testcases to fail. So I think you should not change the tests to make PRs happy. Otherwise, why bother have those tests?
> > Before figuring out why #532 fails the tests, we should hold this PR.
> 
> Ok, from my side, if we put aside the `no buffer` issue, I think another issue is if there is no `factoryreset` after restarting otbr-agent, rest-check always fails and output too many logs. I am not sure whether no `factoryreset` is acceptable, Seems it's result from #532, I will investigate on this and see why.

@tttttangTH OpenThread persists data across restarts. It expected that OpenThread will continue its function after resatrting. `factoryreset`, by its name, is to reset the device to the factory mode, which is an usual action in products. The too many logs issue is due to a bug of otbr with latest openthread, see the fail https://github.com/openthread/ot-br-posix/pull/594/checks?check_run_id=1290402004 in @simonlingoogle 's new PR.

simonlingoogle:
> The too many logs issue is due to a bug of otbr with latest openthread, see the fail https://github.com/openthread/ot-br-posix/pull/594/checks?check_run_id=1290402004 in @simonlingoogle 's new PR.

Yes. It's not an issue of otbr-rest. But we can still use this PR to enhance otbr-rest, e.g. configure test timeout.

tttttangTH:
> > The too many logs issue is due to a bug of otbr with latest openthread, see the fail https://github.com/openthread/ot-br-posix/pull/594/checks?check_run_id=1290402004 in @simonlingoogle 's new PR.
> 
> Yes. It's not an issue of otbr-rest. But we can still use this PR to enhance otbr-rest, e.g. configure test timeout.

ok, will update this PR according to our discussion.

wgtdkp:
> > > The too many logs issue is due to a bug of otbr with latest openthread, see the fail https://github.com/openthread/ot-br-posix/pull/594/checks?check_run_id=1290402004 in @simonlingoogle 's new PR.
> > 
> > 
> > Yes. It's not an issue of otbr-rest. But we can still use this PR to enhance otbr-rest, e.g. configure test timeout.
> 
> ok, will update this PR according to our discussion.

@tttttangTH I think it would be better to close this PR and create a new PR with clearer purpose.

tttttangTH:
> > > > The too many logs issue is due to a bug of otbr with latest openthread, see the fail https://github.com/openthread/ot-br-posix/pull/594/checks?check_run_id=1290402004 in @simonlingoogle 's new PR.
> > > 
> > > 
> > > Yes. It's not an issue of otbr-rest. But we can still use this PR to enhance otbr-rest, e.g. configure test timeout.
> > 
> > 
> > ok, will update this PR according to our discussion.
> 
> @tttttangTH I think it would be better to close this PR and create a new PR with clearer purpose.

Will do!

tttttangTH:
This PR will be closed and will be divided into several PRs with clear purpose in which they aim to:

- Enhance tests for OTBR_REST 
- Modify error handler for according to openthread API response
- (If possible)Deal with collaboration issues within different modules of ot-br-posix  on `otThreadSetReceiveDiagnosticGetCallback`
- Others( For example, does the change `SetCallback` need to be done in a individual PR or in #537 ? @wgtdkp )

wgtdkp:
> This PR will be closed and will be divided into several PRs with clear purpose in which they aim to:
> 
> * Enhance tests for OTBR_REST
> * Modify error handler for according to openthread API response
> * (If possible)Deal with collaboration issues within different modules of ot-br-posix  on `otThreadSetReceiveDiagnosticGetCallback`
> * Others( For example, does the change `SetCallback` need to be done in an individual PR or in #537 ? @wgtdkp )

@tttttangTH We need to make sure #537 always work and mergeable.

tttttangTH:
> > This PR will be closed and will be divided into several PRs with clear purpose in which they aim to:
> > 
> > * Enhance tests for OTBR_REST
> > * Modify error handler for according to openthread API response
> > * (If possible)Deal with collaboration issues within different modules of ot-br-posix  on `otThreadSetReceiveDiagnosticGetCallback`
> > * Others( For example, does the change `SetCallback` need to be done in an individual PR or in #537 ? @wgtdkp )
> 
> @tttttangTH We need to make sure #537 always work and mergeable.

👌！

jwhui:
Closing stale PR.

PR #601, opened 2020-10-29
[common] remove client dependency of OpenThread header

Moandor-y:
This removes the dependency of `openthread/ip6.h` from `common/types.hpp`.

### DIFF
--- a/src/common/types.hpp
+++ b/src/common/types.hpp
@@ -41,8 +41,6 @@
 #include <string>
 #include <vector>
 
-#include <openthread/ip6.h>
-
 #ifndef IN6ADDR_ANY
 /**
  * Any IPv6 address literal.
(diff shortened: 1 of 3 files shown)


bukepo (approved):
LGTM:+1:

src/common/types.hpp
```diff
@@ -41,8 +41,6 @@
 #include <string>
 #include <vector>
 
-#include <openthread/ip6.h>
```
simonlingoogle, line 44: Any background on the decision to not allow including openthread headers in otbr headers? @Moandor-y @bukepo

src/common/types.hpp
```diff
@@ -114,12 +112,12 @@ class Ip6Address
...
 
     /**
-     * Constructor with an Thread Ip6 address.
+     * Constructor with an Ip6 address.
      *
-     * @param[in]   aAddress    The Thread Ip6 address.
+     * @param[in]   aAddress    The Ip6 address.
      *
      */
-    Ip6Address(const otIp6Address &aAddress);
```
simonlingoogle, line 122: Is it possible to use forward declaration to implement `Ip6Address(const otIp6Address &aAddress);` without including `openthread/ip6.h` ?

src/common/types.hpp
bukepo: The goal is to keep the interface stable for internal projects. As you know, OpenThread itself iterates rapidly.

src/common/types.hpp
bukepo: I think it is possible. But since this is blocking our sync. I suggest we figure out a better solution later.

bukepo (approved):

simonlingoogle (approved):
lgtm 👍 \

Irving-cl (approved):
LGTM

jwhui (approved):

PR #607, opened 2020-11-10
[cli] add cli support for region code

gjc13:
This allows users to get/set the region code over OpenThread cli.

### DIFF
--- a/src/agent/ncp_openthread.cpp
+++ b/src/agent/ncp_openthread.cpp
@@ -60,6 +60,11 @@ using std::chrono::steady_clock;
 namespace otbr {
 namespace Ncp {
 
+const otCliCommand ControllerOpenThread::sRegionCommand = {
+    "region",
+    &ControllerOpenThread::HandleRegionCommand,
+};
+
 ControllerOpenThread::ControllerOpenThread(const char *aInterfaceName,
                                            const char *aRadioUrl,
                                            const char *aBackboneInterfaceName)
@@ -132,6 +137,7 @@ otbrError ControllerOpenThread::Init(void)
 #endif
 
     mThreadHelper = std::unique_ptr<otbr::agent::ThreadHelper>(new otbr::agent::ThreadHelper(mInstance, this));
+    otCliSetUserCommands(&sRegionCommand, 1, this);
 
 exit:
     return error;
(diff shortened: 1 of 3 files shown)


src/agent/ncp_openthread.cpp
```diff
@@ -60,6 +60,11 @@ using std::chrono::steady_clock;
 namespace otbr {
 namespace Ncp {
 
+otCliCommand ControllerOpenThread::sRegionCommand = {
```
bukepo, line 63: constexpr?

bukepo (approved):
LGTM:+1:

src/agent/ncp_openthread.cpp
gjc13: `const` will be more proper for these callback structs.

src/agent/ncp_openthread.cpp
```diff
@@ -322,6 +328,37 @@ void ControllerOpenThread::RegisterResetHandler(std::function<void(void)> aHandl
...
+    ControllerOpenThread *controller = static_cast<ControllerOpenThread *>(aContext);
+    controller->HandleRegionCommand(aArgLength, aArgs);
+}
+
+void ControllerOpenThread::HandleRegionCommand(uint8_t aArgLength, char **aArgs)
+{
+    if (aArgLength == 0)
+    {
+        otCliOutputFormat("%s\n", mRegionCode.c_str());
+        otCliOutputFormat("Done\n", mRegionCode.c_str());
```
bukepo, line 342:
```suggestion
        otCliOutputFormat("Done\n");
```

src/agent/ncp_openthread.cpp
```diff
@@ -322,6 +328,37 @@ void ControllerOpenThread::RegisterResetHandler(std::function<void(void)> aHandl
...
+    else if (aArgLength == 1)
+    {
+        if (strnlen(aArgs[0], 3) == 2)
+        {
+            mRegionCode = aArgs[0];
+            otCliOutputFormat("Done\n");
+        }
+        else
+        {
+            otCliOutputFormat("Error: InvalidArgs\n");
```
bukepo, line 353: not in this PR, but prefer change the prototype of otCliCommand to return otError and handle error message in OpenThread. Just like the diagnostic commands.

jwhui (approved):

PR #614, opened 2020-11-18
[docker] fix arm/v7 docker build

simonlingoogle:
Seems to be a known issue: https://gitlab.kitware.com/cmake/cmake/-/issues/20568, which is already fixed by https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5258. But this fix won't be available until cmake 3.19.0. 

I tried to downgrade the cmake version to 3.10.x or 3.13.x or 3.15.x but got a number of issues when building. 

Now, A reasonable fix I can see is using `ubuntu:bionic` for `armv7`, and keep using `ubuntu:focal` for arm64 and amd64.  This has been verified to work well. 

We can upgrade to `ubuntu:focal` or higher for `armv7` when cmake 3.19.0 is officially released. 

Sounds a good idea for you? @bukepo @wgtdkp @gjc13 @jwhui


bukepo (approved):
LGTM:+1:

wgtdkp (approved):
LGTM :+1:

jwhui (approved):
Thanks! 👍

rodrigobrochado:
@simonlingoogle Cmake 3.19 was released yesterday, right? :) would it be likely to update it on ot-br?

simonlingoogle:
> @simonlingoogle Cmake 3.19 was released yesterday, right? :) would it be likely to update it on ot-br?

Yes it is. But it's not installable via pip3 yet.

PR #624, opened 2020-11-24
[script] disable DNS of dnsmasq when DNS64 is on

bukepo:
This commit disables DNS of dnsmasq when DNS64 is enabled, because
bind9 provides DNS service already.

This commit also moves packages installation out from `script/bootstrap` into service setup script.

### DIFF
--- a/script/_network_manager
+++ b/script/_network_manager
@@ -333,7 +333,7 @@ network_manager_install()
     if with DNS64; then
         # bind9 provides DNS service
         sudo sed -i 's/^#port=5353/port=0/g' /etc/dnsmasq.conf
-        sudo systemctl restart dnsmasq
+        sudo systemctl stop dnsmasq
     fi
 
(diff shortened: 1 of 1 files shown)


script/_network_manager
```diff
@@ -330,6 +330,12 @@ network_manager_install()
         return 0
     fi
 
+    if with DNS64; then
+        # bind9 provides DNS service
+        sudo sed -i 's/^#port=5353/port=0/g' /etc/dnsmasq.conf
+        sudo systemctl restart dnsmasq
```
simonlingoogle, line 336: why do we install dnsmasq in the first place if `DNS64=1`?

bukepo: It provides other services in addition to DNS service.

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

Issue #631, opened 2020-11-30
reset command not work if DBUS server is not enabled


simonlingoogle:
**Describe the bug**
OT CLI `reset` command not effective if `OTBR_ENABLE_DBUS_SERVER` is not enabled.

```
> reset
Done
```

`reset` is not effective and output `Done`. 

**To Reproduce** Information to reproduce the behavior, including:
1. Build otbr-agent without OTBR_ENABLE_DBUS_SERVER
2. Input reset command in `ot-ctl`

**Expected behavior**

`otbr-agent` should reset the Thread stack (preferably restart the whole process).

**Additional context**

The reset code seems to be only effective if DBUS server is used:
```c++
#if OTBR_ENABLE_DBUS_SERVER
        if (ncpOpenThread->IsResetRequested())
        {
            ncpOpenThread->Reset();
            continue;
        }
#endif
```

The reset behavior itself (even when DBUS server) seems to be a `pseudo` reset, which I believe is very unsafe and should be avoided. 

Thoughts? @bukepo @gjc13 @jwhui

IHIDCHAOS:
Hi Simon,

I noticed that this problem also occurred in the OpenWrt platform with [UBUS](https://openwrt.org/docs/techref/ubus).

```
root@Hoo:/# 
root@Hoo:/# ot-ctl
> panid
0x1233
Done
> panid 0x1111
Done
> factoryreset
Done
> panid 
0x1111
Done
> reset
Done
> panid
0x1111
Done
>
```

Because the OpenWrt uses ubus instead of dbus, so we should also check "#if OTBR_ENABLE_OPENWRT".

https://github.com/openthread/ot-br-posix/blob/32e0914508f43e4991fee71efc46db69b5e2211c/src/agent/main.cpp#L62

Thanks.

Best,
Guo

simonlingoogle:
@IHIDCHAOS 
Thanks for the info. I think we need to support reset no matter `otbr-agent` is using `ubus`  or `dbus` or none. 

Per my understanding, otbr-agent tries to avoid **real** reset (via `execvp(argv[0], argv);`), which I believe is unsafe. 
Should we always reset the program by `execvp` ? @bukepo @jwhui

jwhui:
> Per my understanding, otbr-agent tries to avoid **real** reset (via `execvp(argv[0], argv);`) for DBUS server, which I believe is unsafe (we had bug where pseudo reset failed to initialize some static variables).
[...]

PR #639, opened 2020-12-02
[ncp] re-attach when soft reset is triggered

gjc13:


simonlingoogle:
@bukepo perfect example why we should not use soft reset. **We can never guarantee that there is no more bug.**

See also: https://github.com/openthread/ot-br-posix/issues/631#issuecomment-736247938

src/agent/ncp_openthread.cpp
```diff
@@ -257,7 +257,8 @@ void ControllerOpenThread::Reset(void)
     {
         handler();
     }
-    sReset = false;
+    mTriedAttach = false;
```
simonlingoogle, line 260: add a test to make sure that `reset` works?

simonlingoogle (approved):

librasungirl (approved):

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #640, opened 2020-12-02
[github-actions] switch to ubuntu 20.04

simonlingoogle:
to eliminate **Annotations**

- `package` is still using `ubuntu-18.04` because it installs `cmake==3.10.3`, which will fail on ubuntu 20.04.


Moandor-y (approved):
LGTM :+1:

jwhui (approved):

PR #646, opened 2020-12-08
[dhcp6] enable DHCP6 server and client for reference device

simonlingoogle:
**Enabled only for reference device.**
FYI: This is required by Thread 1.1 certification.


wgtdkp (approved):

bukepo:
FORCE means it cannot be disabled. I think it should be used only for required features in real products, but not for optional certification requirements.

librasungirl:
@simonlingoogle sorry, just remembered that it is not necessary to pass all the tests defined in Test Harness and it is allowed to skip the ones if the component claims to not support some optional feature.

Agree with @bukepo now.

simonlingoogle:
Close since DHCP6 client is optional

simonlingoogle:
> FORCE means it cannot be disabled. I think it should be used only for required features in real products, but not for optional certification requirements.

Agree.

The purpose of this PR has changed to enable dhcp6 client and server for reference device. Please take a look.

librasungirl (approved):

jwhui (approved):

PR #649, opened 2020-12-11
[agent] full reset using execvp

simonlingoogle:
Premature reset has causing a number of bugs.

This commit fix the bugs and avoid potential future bugs by resetting otbr-agent using `execvp`, which will bootstrap the whole process completely.

### DIFF
--- a/src/rest/rest_web_server.hpp
+++ b/src/rest/rest_web_server.hpp
@@ -62,9 +62,6 @@ public:
     /**
      * This method initializes the REST server.
      *
-     * @retval  OTBR_ERROR_NONE     REST server initialized successfully.
-     * @retval  OTBR_ERROR_REST     Failed due to rest error .
-     *
      */
     void Init(void);
 
--- a/src/rest/rest_web_server.hpp
+++ b/src/rest/rest_web_server.hpp
@@ -66,7 +66,7 @@ public:
      * @retval  OTBR_ERROR_REST     Failed due to rest error .
      *
      */
-    otbrError Init(void);
+    void Init(void);
 
     /**
      * This method updates the file descriptor sets and timeout for mainloop.
@@ -89,7 +89,7 @@ private:
     otbrError UpdateConnections(fd_set &aReadFdSet);
     void      CreateNewConnection(int32_t &aFd);
     otbrError Accept(int32_t aListenFd);
-    otbrError InitializeListenFd(void);
+    void      InitializeListenFd(void);
     bool      SetFdNonblocking(int32_t fd);
 
     // Resource handler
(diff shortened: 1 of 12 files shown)


bukepo (approved):
I think  `reset` is only needed for test. I'd suggest removing this command from OpenThread and implement this function in node.py or THCI script.

wgtdkp (approved):
LGTM 👍

simonlingoogle:
@jwhui 
https://github.com/openthread/openthread/pull/6204 depends on this PR. I think we can merge this PR first and then https://github.com/openthread/openthread/pull/6204.

src/rest/rest_web_server.hpp
```diff
@@ -66,7 +66,7 @@ class RestWebServer
      * @retval  OTBR_ERROR_REST     Failed due to rest error .
      *
      */
-    otbrError Init(void);
+    void Init(void);
```
jwhui, line 69: Update docs above.

src/utils/socket_utils.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <sys/socket.h>
+#include <unistd.h>
+
+enum SocketBlockOption
+{
+    kSocketBlock,
+    kSocketNonBlock,
+};
+
+int SocketWithCloseExec(int aDomain, int aType, int aProtocol, SocketBlockOption aBlockOption);
```
jwhui, line 47: Add docs?

src/utils/socket_utils.hpp
```diff
@@ -0,0 +1,49 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
```
jwhui, line 2:
```suggestion
 *  Copyright (c) 2021, The OpenThread Authors.
```

simonlingoogle: It was 2020 because this PR was actually created quite a time ago.

simonlingoogle:
> I think `reset` is only needed for test. I'd suggest removing this command from OpenThread and implement this function in node.py or THCI script.

Foremost, I think it deserves comprehensive discussion whether `reset` is a command that should exist in OpenThread, which is way beyond the scope of this PR (which just changes the behavior or reset command on OTBR).

However, there are some practical points:

1. `reset` command is required by TestHarness and many other test scripts, which can not be implemented totally in THCI or node.py. We can restart the otbr-agent process as a equivalent of `reset` but we can't do that for SOC devices. Anyway, we can't totally move the implementation of `reset` to scripts.
2. Software reset is a real thing on embedded devices. Many embedded platforms implemented `otPlatReset` for OpenThread using platform-specific **RESET** commands. So, if we can **reset** the SOC devices through OT CLI, we would also like to **reset** a otbr-agent process in the same way so that we provide consistent OT CLI user experience. 
3. If we remove `reset` from otbr-agent, it would make testing scripts (e.g. OTCI) more difficult to implement. 

Thoughts?

jwhui (approved):

PR #656, opened 2020-12-21
[radio] pass full power map in command line args

gjc13:
This allows the powermap to be dynamically reloaded when user sets the
region. Also fixes the reset behavior of the region code.

### DIFF
--- a/src/agent/ncp_openthread.cpp
+++ b/src/agent/ncp_openthread.cpp
@@ -41,6 +41,7 @@
 #include <openthread/thread_ftd.h>
 #include <openthread/platform/logging.h>
 #include <openthread/platform/misc.h>
+#include <openthread/platform/radio.h>
 #include <openthread/platform/settings.h>
 
 #include "common/code_utils.hpp"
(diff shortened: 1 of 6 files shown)


src/agent/ncp_openthread.cpp
```diff
@@ -345,8 +381,16 @@ void ControllerOpenThread::HandleRegionCommand(uint8_t aArgLength, char **aArgs)
     {
         if (strnlen(aArgs[0], 3) == 2)
         {
-            mRegionCode = aArgs[0];
-            otCliOutputFormat("Done\n");
+            otError error = SetRegionCode(aArgs[0]);
+
+            if (error != OT_ERROR_NONE)
+            {
+                otCliOutputFormat("Error: %s\n", otThreadErrorToString(error));
```
bukepo, line 388:
```suggestion
                otCliOutputFormat("Error %d: %s\n", error, otThreadErrorToString(error));
```

bukepo (approved):

PR #662, opened 2020-12-25
[dbus] add active dataset api

gjc13:
This change allows users to get and set the active dataset via the dbus
api. It also adds an `Attach` interface to allow users to start the
Thread network without specifying network parameters.

### DIFF
--- a/src/dbus/server/introspect.xml
+++ b/src/dbus/server/introspect.xml
@@ -31,6 +31,9 @@
       <arg name="channel_mask" type="u"/>
     </method>
 
+    <method name="Attach">
+    </method>
+
     <method name="PermitUnsecureJoin">
       <arg name="port" type="q"/>
       <arg name="timeout" type="u"/>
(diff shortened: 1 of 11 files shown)


src/dbus/server/introspect.xml
```diff
@@ -31,6 +31,9 @@
       <arg name="channel_mask" type="u"/>
     </method>
 
+    <method name="Attach">
```
bukepo, line 34: missing arguments?

gjc13: It doesn't have one. This is another `Attach` interface that takes no arguments. All the Thread network parameters are passed via the active dataset.

bukepo (approved):
LGTM:+1:

tests/dbus/test_dbus_client.cpp
```diff
@@ -178,10 +181,13 @@ int main()
                         {
                             exit(-1);
                         }
-                        api->Attach("Test", 0x5678, extpanid, {}, {}, UINT32_MAX, [&api](ClientError aErr) {
```
Moandor-y, line 181: Keep the old test so we can test both versions of `Attach`?

gjc13: The test ran test with the same parameter twice before this change. I turned the 2nd attach into the using active dataset.

Moandor-y (approved):
LGTM :+1:

jwhui (approved):

PR #663, opened 2020-12-26
[utils] add gitignores for CLion

wgtdkp:


bukepo (approved):

.gitignore
```diff
@@ -105,6 +105,10 @@ tests/unit/unittest
 # Vim auto complete helper
 .ycm_extra_conf.py
 
+# CLion
+.idea
```
simonlingoogle, line 109:
```suggestion
.idea/
```

.gitignore
```diff
@@ -105,6 +105,10 @@ tests/unit/unittest
...
+cmake-build*
```
simonlingoogle, line 110:
```suggestion
cmake-build-*/
```

simonlingoogle (approved):
LGTM :+1:

Irving-cl (approved):
LGTM 👍

.gitignore
jwhui, line 110: In openthread/openthread, we used:
```
.idea/**
cmake-build-*/**
```
https://github.com/openthread/openthread/blob/9ca79ddd9af3d4e3f78cb6e611a3117a71b2198c/.gitignore#L64

Should we use the same here?

wgtdkp: Did some search about ignoring with `.idea`, `.idea/`, `.idea/*`, `.idea/**` and `.idea/**/*`, There are [debates](https://stackoverflow.com/questions/8783093/gitignore-syntax-bin-vs-bin-vs-bin-vs-bin) and it is not clear to me why we need suffix `/*` or `/**` when `.idea` or `.idea/` works well.

I am trying to ignore anything in `.idea` and I think we should use the same pattern as we have already applied for `._build`. But if it is considered that `/*` or `/**` is a better practice, I think we should apply it to all files/directories listed in .gitignore file.

For `.idea` vs `.idea/`, `.idea/` will ignores only direcotries but I don't think we will have a `.idea` file and we don't want to ignore it. I am okay with both, if we consider `.idea/` better than `.idea`, then we should apply it to all items in .gitignore. Thoughts? @jwhui @simonlingoogle

simonlingoogle: I think `.idea/**` and `.idea/` are essentially same. 
Same with `cmake-build-*/**` and `cmake-build-*/`. 

I think both is fine and correct. 

BTW, I copied the pattern `.idea/**` from the CHIP project.

bukepo: From the official document of gitignore, `.idea` and `.idea/**` are different. The latter only matches `.idea` under the current directory, while the former matches `.idea` under any subdirectories as well. If the IDE generates such folders on multiple level directories, we should use `.idea`, otherwise `.idea/**` is preferred.

wgtdkp: Sounds like `.idea/**` is better here. Updated.

jwhui (approved):

PR #669, opened 2021-01-04
[mdns] accept a list of TXT entries

wgtdkp:
In cases the TXT record can not be determined at compile time (For services from Thread end devices, arbitrary TXT records may be provided at runntime), we need to support specifying a list of TXT name/value entries.

### DIFF
--- a/src/agent/border_agent.cpp
+++ b/src/agent/border_agent.cpp
@@ -161,11 +161,11 @@ BorderAgent::~BorderAgent(void)
     }
 }
 
-void BorderAgent::HandleMdnsState(Mdns::State aState)
+void BorderAgent::HandleMdnsState(Mdns::Publisher::State aState)
 {
     switch (aState)
     {
-    case Mdns::kStateReady:
+    case Mdns::Publisher::State::kReady:
         PublishService();
         break;
     default:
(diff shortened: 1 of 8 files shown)


bukepo (approved):

src/mdns/mdns.hpp
```diff
@@ -77,6 +60,50 @@ typedef void (*StateHandler)(void *aContext, State aState);
 class Publisher
 {
 public:
+    /**
+     * This structure represents a name/value pair of the TXT record.
+     *
+     */
+    struct TxtEntry
+    {
+        const char *   mName;        ///< The name of the TXT entry.
```
gjc13, line 69: Storing the pointer assumes the lifetime of the memory passed to the `TxtEntry`. Is it more favorable to use `std::unique_ptr<char[]>`?

src/agent/border_agent.cpp
```diff
@@ -217,19 +217,19 @@ static const char *ThreadVersionToString(uint16_t aThreadVersion)
 
 void BorderAgent::PublishService(void)
 {
+    Mdns::Publisher::TxtList txtList;
```
gjc13, line 220: Maybe we can use brace initialization when constructing the `TxtList` rather than `emplace_back`?
```
Mdns::Publisher::TxtList txtList { {"nn", mNetworkName}, {"xp", mExtPanId, sizeof(mExtPanId)}, {"tv", versionString}};
```

src/mdns/mdns.hpp
wgtdkp: I think it applies to other arguments: `const char *aName`, `const char *aType`. `std::unique_ptr` doesn't solve the problem unless we make a copy of the string and in this case we should use `std::string`. Thoughts?

src/agent/border_agent.cpp
wgtdkp: SGTM.

gjc13 (approved):

jwhui (approved):

PR #679, opened 2021-01-20
[dbus] add watch enable check

gjc13:
Some disabled watch will not be toggled causing writable file descriptor
 to be triggered infinitely. The toggle feature is optional in d-bus.
 This patch fixes this bug.


bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #686, opened 2021-01-26
[build] make dbus user and group configurable in Android.mk

gjc13:
This allow users to override the dbus user and group with non-root
configuration.


bukepo (approved):
LGTM:+1:

mspang (approved):

jwhui (approved):

PR #688, opened 2021-01-27
[dbus] fix wrong error code when error happens

gjc13:
The reply dbus message will not contain error code when error happens.
Currently all the error codes will be OTBR_ERROR_DBUS.
This CL checks separate DBusError to correctly handle the error codes
and adds corresponding tests.


bukepo (approved):
LGTM:+1:

Moandor-y (approved):
LGTM :+1:

jwhui (approved):

PR #690, opened 2021-02-02
[WIP] add HIDL server support

zhanglongxia:
This CL implements the HIDL server on the Android platform. The HIDL
server provides IThread and IThreadSettings HIDL interfaces for the
HIDL client.
The IThread interface is implemented by the HIDL server to provides
all control operations of the Thread network.
The IThreadSettings interface is mainly called by the HIDL server to
send sensitive settings to Thread System service. The Thread System
servie stores the sensitive settings to a secure area.

### DIFF
--- a/src/common/types.cpp
+++ b/src/common/types.cpp
@@ -30,11 +30,32 @@
 #include <sstream>
 #include <sys/socket.h>
 
+#ifdef __APPLE__
+#include <libkern/OSByteOrder.h>
+#define htobe16(x) OSSwapHostToBigInt16(x)
+#else
+#include <endian.h>
+#endif
+
 #include "common/code_utils.hpp"
 #include "common/logging.hpp"
 #include "common/types.hpp"
 
 namespace otbr {
+
+std::string HexToString(const uint8_t *aHex, uint16_t aLength)
+{
+    std::string str;
+    char        bytestr[3];
+
+    for (uint16_t i = 0; i < aLength; i++)
+    {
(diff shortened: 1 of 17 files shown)


src/common/types.cpp
```diff
@@ -30,12 +30,33 @@
 #include <sstream>
 #include <sys/socket.h>
 
+#ifdef __APPLE__
+#include <libkern/OSByteOrder.h>
+#define htobe16(x) OSSwapHostToBigInt16(x)
```
Moandor-y, line 35: Avoid defining macros. Functions should be preferred here.

Moandor-y (commented):
Most HIDL interfaces cannot be tested by CI. Should we have untested components in GitHub?

src/common/types.cpp
zhanglongxia: Are there any special reasons to avoid defining macros?

zhanglongxia:
> Most HIDL interfaces cannot be tested by CI. Should we have untested components in GitHub?
The HIDL interfaces are used in Android. When the code is synchronized to Android, they will be tested.

src/common/types.cpp
Moandor-y: Macros don't have namespace, so they are likely to conflict with other names. Macros also have many well known pitfalls. And this is also included in [the style guide](https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros).

PR #699, opened 2021-02-16
[common] add Task Runner

wgtdkp:
This PR adds a `TaskRunner` module that executes tasks on the mainloop.
The `TaskRunner` is useful to dispatch a callback back to the mainloop.
(e.g. handle mDNS callbacks which may be invoked from another thread).

We are not using [eventfd](https://man7.org/linux/man-pages/man2/eventfd.2.html) because it is linux-specific.

### DIFF
--- a/src/common/code_utils.hpp
+++ b/src/common/code_utils.hpp
@@ -33,6 +33,8 @@
 #ifndef OTBR_COMMON_CODE_UTILS_HPP_
 #define OTBR_COMMON_CODE_UTILS_HPP_
 
+#include "common/logging.hpp"
+
 /**
  *  This aligns the pointer to @p aAlignType.
  *
--- a/src/common/task_runner.cpp
+++ b/src/common/task_runner.cpp
@@ -1,0 +1,148 @@
+/*
+ *    Copyright (c) 2021, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ * This file implements the Task Runner that executes tasks on the mainloop.
+ */
+
+#include "common/task_runner.hpp"
+
+#include <algorithm>
+
+#include <fcntl.h>
+#include <unistd.h>
+
+#include "common/code_utils.hpp"
+
+namespace otbr {
+
+TaskRunner::TaskRunner(void)
+{
+    int flags;
+
+    // We do not handle failures when creating a pipe, simply die.
+    VerifyOrDie(pipe(mEventFd) != -1, strerror(errno));
+
+    flags = fcntl(mEventFd[kRead], F_GETFL, 0);
+    VerifyOrDie(fcntl(mEventFd[kRead], F_SETFL, flags | O_NONBLOCK) != -1, strerror(errno));
+    flags = fcntl(mEventFd[kWrite], F_GETFL, 0);
+    VerifyOrDie(fcntl(mEventFd[kWrite], F_SETFL, flags | O_NONBLOCK) != -1, strerror(errno));
+}
+
+TaskRunner::~TaskRunner(void)
+{
+    if (mEventFd[kRead] != -1)
+    {
+        close(mEventFd[kRead]);
+        mEventFd[kRead] = -1;
+    }
+    if (mEventFd[kWrite] != -1)
+    {
+        close(mEventFd[kWrite]);
+        mEventFd[kWrite] = -1;
+    }
+}
+
+void TaskRunner::Post(const Task<void> &aTask)
+{
+    PushTask(aTask);
+}
+
+void TaskRunner::UpdateFdSet(otSysMainloopContext &aMainloop)
+{
+    FD_SET(mEventFd[kRead], &aMainloop.mReadFdSet);
+    aMainloop.mMaxFd = std::max(mEventFd[kRead], aMainloop.mMaxFd);
+}
+
+void TaskRunner::Process(const otSysMainloopContext &aMainloop)
+{
+    if (FD_ISSET(mEventFd[kRead], &aMainloop.mReadFdSet))
+    {
+        uint8_t n;
+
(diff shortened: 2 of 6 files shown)


src/common/code_utils.hpp
```diff
@@ -33,6 +33,8 @@
 #ifndef OTBR_COMMON_CODE_UTILS_HPP_
 #define OTBR_COMMON_CODE_UTILS_HPP_
 
+#include "common/logging.hpp"
```
bukepo, line 36: Does this header depends on logging?

src/common/task_runner.cpp
```diff
@@ -0,0 +1,114 @@
...
+    // There are no chances that we can handle failure of creating
+    // the event fd, simply die.
+    VerifyOrDie(mEventFd != -1, strerror(errno));
+}
+
+TaskRunner::~TaskRunner(void)
+{
+    if (mEventFd != -1)
+    {
+        close(mEventFd);
```
bukepo, line 77: Just in case this destructor is called multiple times.
```suggestion
        close(mEventFd);
        mEventFd = -1;
```

src/common/code_utils.hpp
wgtdkp: The `SuccessOrDie` and `VerifyOrDie` refer to `otbrLog`.

src/common/task_runner.cpp
```diff
@@ -0,0 +1,132 @@
...
+    if (mEventFd[kRead] != -1)
+    {
+        close(mEventFd[kRead]);
+        mEventFd[kRead] = -1;
+    }
+    if (mEventFd[kWrite] != -1)
+    {
+        close(mEventFd[kWrite]);
+        mEventFd[kWrite] = -1;
+    }
```
simonlingoogle, line 70:
```suggestion
        close(mEventFd[kRead]);
        close(mEventFd[kWrite]);
```

I think the two FDs can never be `-1`

src/common/task_runner.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    if (rval != sizeof(kOne))
+    {
+        if (rval == -1)
+        {
+            otbrLog(OTBR_LOG_ERR, "failed to write event fd %d: %s", mEventFd, strerror(errno));
+        }
+        else
+        {
+            otbrLog(OTBR_LOG_ERR, "partially write event fd %d: %s", mEventFd, strerror(errno));
```
simonlingoogle, line 116:
```suggestion
            otbrLog(OTBR_LOG_ERR, "partially write event fd %d", mEventFd);
```
`errno` is not properly set if rval != -1 ?

src/common/task_runner.cpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+void TaskRunner::PopTasks(void)
+{
+    std::lock_guard<std::mutex> _(mTaskQueueMutex);
+
+    while (!mTaskQueue.empty())
+    {
+        mTaskQueue.front()();
```
simonlingoogle, line 127: If the Task calls `PushTask` to push another task when it's executing, it's going to be a deadlock ? 
I think we should allow posting new tasks in the tasks.

src/common/task_runner.cpp
```diff
@@ -0,0 +1,132 @@
...
+    std::lock_guard<std::mutex> _(mTaskQueueMutex);
+
+    mTaskQueue.emplace(aTask);
+    rval = write(mEventFd[kWrite], &kOne, sizeof(kOne));
+
+    if (rval != sizeof(kOne))
+    {
+        if (rval == -1)
+        {
+            otbrLog(OTBR_LOG_ERR, "failed to write event fd %d: %s", mEventFd, strerror(errno));
```
simonlingoogle, line 112: should we `die` if rval == -1 ? I think this is a critical error that we can't recover.

src/common/task_runner.cpp
simonlingoogle, line 116: Strictly speaking, if `rval == 0`, we should schedule a retry after some small delay or rescheduling (i.e. yielding to other threads). Otherwise, it might happen that the posted task never runs in  a long period of time (if there is no more task posted). 
I am not sure if this is necessary since it should rarely happen.

src/common/task_runner.cpp
wgtdkp: Just making the destructor self-contained.

src/common/task_runner.cpp
wgtdkp: Good thoughts, agree we should allow this case! :+1:

src/common/task_runner.cpp
wgtdkp: I don't feel it good to crash at runtime, rather it is acceptable to me to crash at initialization.

src/common/task_runner.cpp
wgtdkp: I don't feel this necessary.

src/common/task_runner.cpp
wgtdkp: Updated to allow posting new task in task, the unit tests are also updated to cover this case.

src/common/task_runner.cpp
simonlingoogle: > Just making the destructor self-contained.

But it's totally useless?

src/common/task_runner.cpp
simonlingoogle: SGTM

simonlingoogle (approved):
LGTM :+1: 

with nits

bukepo (approved):
LGTM:+1:

Moandor-y (approved):
LGTM overall.

A little performance improvement:

src/common/task_runner.cpp
```diff
@@ -0,0 +1,146 @@
...
+void TaskRunner::PushTask(const Task &aTask)
+{
+    int                         rval;
+    const uint8_t               kOne = 1;
+    std::lock_guard<std::mutex> _(mTaskQueueMutex);
+
+    mTaskQueue.emplace(aTask);
+    rval = write(mEventFd[kWrite], &kOne, sizeof(kOne));
+
+    if (rval != sizeof(kOne))
```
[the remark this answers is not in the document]
wgtdkp: I don't feel the caller can handle this error. Exposing this to the caller will introduce very verbose error handling code while the failure cases here are really rare.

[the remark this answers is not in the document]
simonlingoogle: Discussed with @wgtdkp that this is actually not an error when errno == `EAGAIN`, which is the only error that could happen here. In the case of `EAGAIN`, the read FD is still readable so that `PopTasks` will be called later.

zhanglongxia (approved):
LGTM

librasungirl (approved):

src/common/task_runner.hpp
```diff
@@ -0,0 +1,147 @@
...
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ * This file define the Task Runner that executes tasks on the mainloop.
```
jwhui, line 31:
```suggestion
 * This file defines the Task Runner that executes tasks on the mainloop.
```

src/common/task_runner.hpp
```diff
@@ -0,0 +1,147 @@
...
+     * Tasks are executed sequentially and follow the First-Come-First-Serve rule.
+     * It is safe to call this method in different threads concurrently.
+     *
+     * @param[in]  aTask  The task to be executed.
+     *
+     */
+    void Post(const Task<void> &aTask);
+
+    /**
+     * This method posts a task waits for the completion of the task.
```
jwhui, line 86:
```suggestion
     * This method posts a task and waits for the completion of the task.
```

src/common/task_runner.hpp
```diff
@@ -0,0 +1,147 @@
...
+    {
+        std::promise<T> pro;
+
+        Post([&]() { pro.set_value(aTask()); });
+
+        return pro.get_future().get();
+    }
+
+    /**
+     * This method updates the file descriptor sets and timeout for mainloop.
```
jwhui, line 105:
```suggestion
     * This method updates the file descriptor sets and timeout for the mainloop.
```

src/common/task_runner.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "common/code_utils.hpp"
+
+namespace otbr {
+
+TaskRunner::TaskRunner(void)
+{
+    int flags;
+
+    // There are no chances that we can handle failure of
+    // creating a pipe, simply die.
```
jwhui, line 50:
```suggestion
    // We do not handle failures when creating a pipe, simply die.
```

src/common/task_runner.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    if (rval != sizeof(kOne))
+    {
+        if (rval == -1)
+        {
+            otbrLog(OTBR_LOG_CRIT, "failed to write event fd %d: %s", mEventFd[kWrite], strerror(errno));
+        }
+        else
+        {
+            otbrLog(OTBR_LOG_ERR, "partially write event fd %d", mEventFd[kWrite]);
```
jwhui, line 116: What `rval` value represents a "partial" write?

wgtdkp: The [write](https://www.man7.org/linux/man-pages/man2/write.2.html) function returns the length of actual write or `-1` in case of error. we are here only when `rval != sizeof(kOne)` and `rval != -1`, and this is the case of partial write.

jwhui: But `sizeof(kOne)` is one byte. How can `write` partially write a single byte?

jwhui (approved):

Issue #520, opened 2020-07-14
[mdns] includes border agent discriminator in mDNS response


wgtdkp:
**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem is.
To uniquely identify a Thread Border Router, a discriminator is required to be advertised in the mDNS response.

**Describe the solution you'd like** A clear and concise description of what you want to happen.
Add Border Agent discriminator to the Border Agent and includes it in mDNS response.

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.

**Additional context** Add any other context or screenshots about the feature request here.

wgtdkp:
A default implementation is to use EUI64 as the discriminator.

PR #701, opened 2021-02-18
[mdns] advertise extended address as discriminator

wgtdkp:
addresses https://github.com/openthread/ot-br-posix/issues/520.

### DIFF
--- a/src/agent/border_agent.cpp
+++ b/src/agent/border_agent.cpp
@@ -102,13 +102,16 @@ void BorderAgent::Init(void)
 {
     memset(mNetworkName, 0, sizeof(mNetworkName));
     memset(mExtPanId, 0, sizeof(mExtPanId));
+    memset(mExtAddr, 0, sizeof(mExtAddr));
     mExtPanIdInitialized = false;
+    mExtAddrInitialized  = false;
     mThreadVersion       = 0;
 
 #if OTBR_ENABLE_MDNS_AVAHI || OTBR_ENABLE_MDNS_MDNSSD || OTBR_ENABLE_MDNS_MOJO
     mNcp->On(Ncp::kEventExtPanId, HandleExtPanId, this);
     mNcp->On(Ncp::kEventNetworkName, HandleNetworkName, this);
     mNcp->On(Ncp::kEventThreadVersion, HandleThreadVersion, this);
+    mNcp->On(Ncp::kEventExtAddr, HandleExtAddr, this);
 #endif
     mNcp->On(Ncp::kEventThreadState, HandleThreadState, this);
     mNcp->On(Ncp::kEventPSKc, HandlePSKc, this);
@@ -134,6 +137,7 @@ otbrError BorderAgent::Start(void)
     SuccessOrExit(error = mNcp->RequestEvent(Ncp::kEventNetworkName));
     SuccessOrExit(error = mNcp->RequestEvent(Ncp::kEventExtPanId));
     SuccessOrExit(error = mNcp->RequestEvent(Ncp::kEventThreadVersion));
+    SuccessOrExit(error = mNcp->RequestEvent(Ncp::kEventExtAddr));
 
 #if OTBR_ENABLE_SRP_ADVERTISING_PROXY
     mAdvertisingProxy.Start();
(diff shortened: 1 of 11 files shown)


src/agent/border_agent.cpp
```diff
@@ -229,10 +232,16 @@ void BorderAgent::PublishService(void)
...
     assert(mThreadVersion != 0);
 
     const char *             versionString = ThreadVersionToString(mThreadVersion);
-    Mdns::Publisher::TxtList txtList{{"nn", mNetworkName}, {"xp", mExtPanId, sizeof(mExtPanId)}, {"tv", versionString}};
+    Mdns::Publisher::TxtList txtList{{"nn", mNetworkName},
+                                     {"xp", mExtPanId, sizeof(mExtPanId)},
+                                     {"tv", versionString},
+                                     // "d" represents for discriminator which
+                                     // should always be the extended address.
+                                     {"d", mExtAddr, sizeof(mExtAddr)}};
```
bukepo, line 244:
```suggestion
                                     // "dd" represents for device discriminator which
                                     // should always be the extended address.
                                     {"dd", mExtAddr, sizeof(mExtAddr)}};
```

bukepo (approved):
LGTM:+1:

src/agent/border_agent.cpp
```diff
@@ -402,4 +425,14 @@ void BorderAgent::HandleThreadVersion(void *aContext, int aEvent, va_list aArgum
...
 }
 
+void BorderAgent::HandleExtAddr(void *aContext, int aEvent, va_list aArguments)
+{
+    OT_UNUSED_VARIABLE(aEvent);
+
+    assert(aEvent == Ncp::kEventExtPanId);
+
+    const uint8_t *extAddr = va_arg(aArguments, const uint8_t *);
+    static_cast<BorderAgent *>(aContext)->SetExtPanId(extAddr);
```
Moandor-y, line 435:
```suggestion
    static_cast<BorderAgent *>(aContext)->SetExtAddr(extAddr);
```

src/agent/border_agent.cpp
```diff
@@ -402,4 +425,14 @@ void BorderAgent::HandleThreadVersion(void *aContext, int aEvent, va_list aArgum
     static_cast<BorderAgent *>(aContext)->SetThreadVersion(threadVersion);
 }
 
+void BorderAgent::HandleExtAddr(void *aContext, int aEvent, va_list aArguments)
+{
+    OT_UNUSED_VARIABLE(aEvent);
+
+    assert(aEvent == Ncp::kEventExtPanId);
```
Moandor-y, line 432:
```suggestion
    assert(aEvent == Ncp::kEventExtAddr);
```

Moandor-y (approved):
LGTM :+1:

src/mdns/mdns_avahi.cpp
```diff
@@ -501,7 +501,7 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
     otbrError ret   = OTBR_ERROR_ERRNO;
     int       error = 0;
     // aligned with AvahiStringList
-    AvahiStringList  buffer[kMaxSizeOfTxtRecord / sizeof(AvahiStringList)];
+    AvahiStringList  buffer[kMaxSizeOfTxtRecord / sizeof(AvahiStringList) + 1];
```
simonlingoogle, line 504:
```suggestion
    AvahiStringList  buffer[(kMaxSizeOfTxtRecord + sizeof(AvahiStringList) - 1) / sizeof(AvahiStringList)];
```

src/agent/border_agent.cpp
```diff
@@ -229,10 +233,16 @@ void BorderAgent::PublishService(void)
...
     assert(mThreadVersion != 0);
 
     const char *             versionString = ThreadVersionToString(mThreadVersion);
-    Mdns::Publisher::TxtList txtList{{"nn", mNetworkName}, {"xp", mExtPanId, sizeof(mExtPanId)}, {"tv", versionString}};
+    Mdns::Publisher::TxtList txtList{{"nn", mNetworkName},
+                                     {"xp", mExtPanId, sizeof(mExtPanId)},
+                                     {"tv", versionString},
+                                     // "dd" represents for device discriminator which
+                                     // should always be the extended address.
+                                     {"dd", mExtAddr, sizeof(mExtAddr)}};
```
simonlingoogle, line 245: Should we re-publish the service when the extended address changes (e.g. after commission)

wgtdkp: I think the discriminator should never be changed even if the Border Agent has been commissioned into another Thread network.

simonlingoogle (approved):
lgtm

src/agent/border_agent.cpp
simonlingoogle: > I think the discriminator should never be changed even if the Border Agent has been commissioned into another Thread network.

Indeed. The extended address should not change after commissioning. However, currently we don't re-publish the service when the device reattaches (after which the extended address might change). I think we need to re-publish the mDNS service when the Thread state is changed from detached to attached (preferably in a separate PR).

wgtdkp: @simonlingoogle Actually, we are advertising the EUI64 of the border agent device.
```c++
case kEventEui64:
    {
        otExtAddress eui64;

        otLinkGetFactoryAssignedIeeeEui64(mInstance, &eui64);
        EventEmitter::Emit(kEventEui64, eui64.m8);
        break;
...
```

I changed the name from `mExtAddr` to `mEui64` to avoid confusion.

jwhui: I think we should use the Extended Address. A concern around static identifiers like IEEE EUI-64 is ability to track devices.

gabekassel: I gave this some thought yesterday as well, and agree to extended address (for whatever it's worth).

wgtdkp: > A concern around static identifiers like IEEE EUI-64 is ability to track devices.

Makes sense to me. Will change accordingly.

wgtdkp: Revised to use the Extended Address, PTAL.

wgtdkp:
@gabekassel @jwhui Changed to use `Extended Address`, PTAL.

jwhui (approved):

PR #704, opened 2021-02-19
[dns-sd] implement missing Avahi bindings

Damian-Nordic:
Currently, only mDNSResponder can be used to test the OpenThread SRP server. Provide the functionality necessary to test registration of new services using Avahi daemon.
    
Avahi forbids modifying entry groups which have already been committed. As a result, in order to support methods for publishing and removing a single host/service, each host and each service is processed as a separate Avahi entry group.

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -34,6 +34,7 @@
 #include "mdns/mdns_avahi.hpp"
 
 #include <algorithm>
+
 #include <avahi-common/alternative.h>
 #include <avahi-common/error.h>
 #include <avahi-common/malloc.h>
@@ -487,7 +488,7 @@ otbrError PublisherAvahi::CreateGroup(AvahiClient &aClient, AvahiEntryGroup *&aO
 {
     otbrError error = OTBR_ERROR_NONE;
 
-    VerifyOrExit(aOutGroup == nullptr, error = OTBR_ERROR_INVALID_ARGS);
+    assert(aOutGroup == nullptr);
 
     aOutGroup = avahi_entry_group_new(&aClient, HandleGroupState, this);
     VerifyOrExit(aOutGroup != nullptr, error = OTBR_ERROR_MDNS);
@@ -708,8 +709,7 @@ exit:
     }
     else if (error != OTBR_ERROR_NONE)
     {
-        otbrLog(OTBR_LOG_ERR, "Failed to publish service: %s!",
-                error == OTBR_ERROR_ERRNO ? strerror(errno) : otbrErrorString(error));
+        otbrLog(OTBR_LOG_ERR, "Failed to publish service: %s!", otbrErrorString(error));
     }
 
     if (error != OTBR_ERROR_NONE && serviceIt != mServices.end())
@@ -763,13 +763,11 @@ otbrError PublisherAvahi::PublishHost(const char *aName, const uint8_t *aAddress
     }
     else if (memcmp(hostIt->mAddress.data.ipv6.address, aAddress, aAddressLength))
     {
-        // Check if the update is necessary
-        VerifyOrExit(memcmp(hostIt->mAddress.data.ipv6.address, aAddress, aAddressLength));
         SuccessOrExit(error = ResetGroup(hostIt->mGroup));
     }
     else
     {
-        if (avahiError == 0 && mHostHandler != nullptr)
+        if (mHostHandler != nullptr)
         {
             // The handler should be called even if the request can be processed synchronously
             mHostHandler(aName, OTBR_ERROR_NONE, mHostHandlerContext);
@@ -800,8 +798,7 @@ exit:
     }
     else if (error != OTBR_ERROR_NONE)
     {
-        otbrLog(OTBR_LOG_ERR, "Failed to publish host: %s!",
-                error == OTBR_ERROR_ERRNO ? strerror(errno) : otbrErrorString(error));
+        otbrLog(OTBR_LOG_ERR, "Failed to publish host: %s!", otbrErrorString(error));
     }
 
     if (error != OTBR_ERROR_NONE && hostIt != mHosts.end())
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -534,7 +534,7 @@ otbrError PublisherAvahi::FreeGroup(AvahiEntryGroup *aGroup)
     return error;
 }
 
-void PublisherAvahi::FreeAllGroups()
+void PublisherAvahi::FreeAllGroups(void)
 {
     for (Service &service : mServices)
     {
@@ -622,7 +622,7 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
     Services::iterator serviceIt    = mServices.end();
     const char *       safeHostName = (aHostName != nullptr) ? aHostName : "";
     const char *       logHostName  = (aHostName != nullptr) ? aHostName : "localhost";
-    char               fullHostName[kMaxSizeOfHost];
+    std::string        fullHostName;
     // aligned with AvahiStringList
     AvahiStringList  buffer[kMaxSizeOfTxtRecord / sizeof(AvahiStringList)];
     AvahiStringList *last = nullptr;
@@ -636,8 +636,8 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
 
     if (aHostName != nullptr)
     {
-        SuccessOrExit(error = MakeFullName(fullHostName, sizeof(fullHostName), aHostName));
-        aHostName = fullHostName;
+        fullHostName = MakeFullName(aHostName);
+        aHostName    = fullHostName.c_str();
     }
 
     for (const auto &txtEntry : aTxtList)
@@ -678,6 +678,11 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
         otbrLog(OTBR_LOG_INFO, "[mdns] update service %s.%s for host %s", aName, aType, logHostName);
         avahiError = avahi_entry_group_update_service_txt_strlst(serviceIt->mGroup, AVAHI_IF_UNSPEC, mProtocol,
                                                                  AvahiPublishFlags{}, aName, aType, mDomain, last);
+        if (avahiError == 0 && mServiceHandler != nullptr)
+        {
+            // The handler should be called even if the request can be processed synchronously
+            mServiceHandler(aName, aType, OTBR_ERROR_NONE, mServiceHandlerContext);
+        }
         ExitNow();
     }
 
@@ -701,10 +706,10 @@ exit:
         error = OTBR_ERROR_MDNS;
         otbrLog(OTBR_LOG_ERR, "Failed to publish service for avahi error: %s!", avahi_strerror(avahiError));
     }
-
-    if (error == OTBR_ERROR_ERRNO)
+    else if (error != OTBR_ERROR_NONE)
     {
-        otbrLog(OTBR_LOG_ERR, "Failed to publish service: %s!", strerror(errno));
+        otbrLog(OTBR_LOG_ERR, "Failed to publish service: %s!",
+                error == OTBR_ERROR_ERRNO ? strerror(errno) : otbrErrorString(error));
     }
 
     if (error != OTBR_ERROR_NONE && serviceIt != mServices.end())
@@ -740,7 +745,7 @@ otbrError PublisherAvahi::PublishHost(const char *aName, const uint8_t *aAddress
     otbrError       error      = OTBR_ERROR_NONE;
     int             avahiError = 0;
     Hosts::iterator hostIt     = mHosts.end();
-    char            fullHostName[kMaxSizeOfHost];
+    std::string     fullHostName;
     AvahiAddress    address;
 
     VerifyOrExit(mState == State::kReady, errno = EAGAIN, error = OTBR_ERROR_ERRNO);
@@ -748,19 +753,28 @@ otbrError PublisherAvahi::PublishHost(const char *aName, const uint8_t *aAddress
     VerifyOrExit(aName != nullptr, error = OTBR_ERROR_INVALID_ARGS);
     VerifyOrExit(aAddress != nullptr, error = OTBR_ERROR_INVALID_ARGS);
     VerifyOrExit(aAddressLength == sizeof(address.data.ipv6.address), error = OTBR_ERROR_INVALID_ARGS);
-    SuccessOrExit(error = MakeFullName(fullHostName, sizeof(fullHostName), aName));
 
-    hostIt = FindHost(aName);
+    fullHostName = MakeFullName(aName);
+    hostIt       = FindHost(aName);
 
     if (hostIt == mHosts.end())
     {
         SuccessOrExit(error = CreateHost(*mClient, aName, hostIt));
     }
-    else
+    else if (memcmp(hostIt->mAddress.data.ipv6.address, aAddress, aAddressLength))
     {
         // Check if the update is necessary
         VerifyOrExit(memcmp(hostIt->mAddress.data.ipv6.address, aAddress, aAddressLength));
         SuccessOrExit(error = ResetGroup(hostIt->mGroup));
+    }
+    else
+    {
+        if (avahiError == 0 && mHostHandler != nullptr)
+        {
+            // The handler should be called even if the request can be processed synchronously
+            mHostHandler(aName, OTBR_ERROR_NONE, mHostHandlerContext);
+        }
+        ExitNow();
     }
 
     address.proto = AVAHI_PROTO_INET6;
@@ -768,7 +782,7 @@ otbrError PublisherAvahi::PublishHost(const char *aName, const uint8_t *aAddress
 
     otbrLog(OTBR_LOG_INFO, "[mdns] create host %s", aName);
     avahiError = avahi_entry_group_add_address(hostIt->mGroup, AVAHI_IF_UNSPEC, AVAHI_PROTO_INET6,
-                                               AVAHI_PUBLISH_NO_REVERSE, fullHostName, &address);
+                                               AVAHI_PUBLISH_NO_REVERSE, fullHostName.c_str(), &address);
     SuccessOrExit(avahiError);
 
     otbrLog(OTBR_LOG_INFO, "[mdns] commit host %s", aName);
@@ -784,10 +798,10 @@ exit:
         error = OTBR_ERROR_MDNS;
         otbrLog(OTBR_LOG_ERR, "Failed to publish host for avahi error: %s!", avahi_strerror(avahiError));
     }
-
-    if (error == OTBR_ERROR_ERRNO)
+    else if (error != OTBR_ERROR_NONE)
     {
-        otbrLog(OTBR_LOG_ERR, "Failed to publish host: %s!", strerror(errno));
+        otbrLog(OTBR_LOG_ERR, "Failed to publish host: %s!",
+                error == OTBR_ERROR_ERRNO ? strerror(errno) : otbrErrorString(error));
     }
 
     if (error != OTBR_ERROR_NONE && hostIt != mHosts.end())
@@ -817,20 +831,16 @@ exit:
     return error;
 }
 
-otbrError PublisherAvahi::MakeFullName(char *aFullName, size_t aFullNameLength, const char *aName)
+std::string PublisherAvahi::MakeFullName(const char *aName)
 {
-    assert(aFullName != nullptr);
     assert(aName != nullptr);
 
-    otbrError   error  = OTBR_ERROR_NONE;
-    const char *domain = mDomain == nullptr ? "local." : mDomain;
-    const int   count  = snprintf(aFullName, aFullNameLength, "%s.%s", aName, domain);
+    std::string fullHostName(aName);
 
-    VerifyOrExit(count >= 0, error = OTBR_ERROR_INVALID_ARGS);
-    VerifyOrExit(static_cast<size_t>(count) < aFullNameLength, error = OTBR_ERROR_INVALID_ARGS);
+    fullHostName += '.';
+    fullHostName += (mDomain == nullptr ? "local." : mDomain);
 
-exit:
-    return error;
+    return fullHostName;
 }
 
 Publisher *Publisher::Create(int aFamily, const char *aDomain, StateHandler aHandler, void *aContext)
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -504,11 +504,10 @@ exit:
 
 otbrError PublisherAvahi::ResetGroup(AvahiEntryGroup *aGroup)
 {
-    otbrError error = OTBR_ERROR_NONE;
-
     assert(aGroup != nullptr);
 
-    int avahiError = avahi_entry_group_reset(aGroup);
+    otbrError error      = OTBR_ERROR_NONE;
+    int       avahiError = avahi_entry_group_reset(aGroup);
 
     if (avahiError)
     {
@@ -519,16 +518,20 @@ otbrError PublisherAvahi::ResetGroup(AvahiEntryGroup *aGroup)
     return error;
 }
 
-void PublisherAvahi::FreeGroup(AvahiEntryGroup *aGroup)
+otbrError PublisherAvahi::FreeGroup(AvahiEntryGroup *aGroup)
 {
     assert(aGroup != nullptr);
 
-    int avahiError = avahi_entry_group_free(aGroup);
+    otbrError error      = OTBR_ERROR_NONE;
+    int       avahiError = avahi_entry_group_free(aGroup);
 
     if (avahiError)
     {
+        error = OTBR_ERROR_MDNS;
         otbrLog(OTBR_LOG_ERR, "Failed to free entry group for avahi error: %s", avahi_strerror(avahiError));
     }
+
+    return error;
 }
 
 void PublisherAvahi::FreeAllGroups()
@@ -715,8 +718,7 @@ exit:
 
 otbrError PublisherAvahi::UnpublishService(const char *aName, const char *aType)
 {
-    otbrError          error      = OTBR_ERROR_NONE;
-    int                avahiError = 0;
+    otbrError          error = OTBR_ERROR_NONE;
     Services::iterator serviceIt;
 
     VerifyOrExit(aName != nullptr, error = OTBR_ERROR_INVALID_ARGS);
@@ -726,16 +728,10 @@ otbrError PublisherAvahi::UnpublishService(const char *aName, const char *aType)
     VerifyOrExit(serviceIt != mServices.end());
 
     otbrLog(OTBR_LOG_INFO, "[mdns] unpublish service %s.%s", aName, aType);
-    avahiError = avahi_entry_group_free(serviceIt->mGroup);
+    error = FreeGroup(serviceIt->mGroup);
     mServices.erase(serviceIt);
 
 exit:
-    if (avahiError)
-    {
-        error = OTBR_ERROR_MDNS;
-        otbrLog(OTBR_LOG_ERR, "Failed to delete service for avahi error: %s!", avahi_strerror(avahiError));
-    }
-
     return error;
 }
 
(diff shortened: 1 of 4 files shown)


Damian-Nordic:
@googlebot I signed it!

src/mdns/mdns_avahi.cpp
```diff
@@ -590,13 +636,47 @@ otbrError PublisherAvahi::UnpublishService(const char *aName, const char *aType)
...
+    SuccessOrExit(ret = MakeFullName(fullHostName, sizeof(fullHostName), aName));
+    SuccessOrExit(ret = GetOrCreateHost(mClient, aName, &host));
+    // Check if the update is needed
+    VerifyOrExit(!host->mIsCommitted || memcmp(host->mAddress.data.ipv6.address, aAddress, aAddressLength));
+
+    address.proto = AVAHI_PROTO_INET6;
+    memcpy(&address.data.ipv6.address[0], aAddress, aAddressLength);
+
+    otbrLog(OTBR_LOG_INFO, "MDNS set address for host %s", fullHostName);
+    error = avahi_entry_group_add_address(host->mGroup, AVAHI_IF_UNSPEC, AVAHI_PROTO_INET6, AVAHI_PUBLISH_NO_REVERSE,
```
wgtdkp, line 659: We need to remove existing addresses. Current `Publisher::PublishHost` API has the semantic of `SET` but not `ADD`. Although multiple addresses can be associated with a single hostname, current API doesn't support it yet.

src/mdns/mdns_avahi.hpp
```diff
@@ -266,6 +266,21 @@ class PublisherAvahi : public Publisher
      */
     otbrError UnpublishHost(const char *aName) override;
 
+    /**
+     * This method commits recent updates for a host.
+     *
+     * Some implementations require an additional step to apply recent updates for a host or its services.
+     * This method should be called after all changes with
+     * PublishHost/UnpublishHost/PublishService/UnpublishService methods are done for a given host name.
```
wgtdkp, line 274: Why do we need to wait for `PublishService/UnpublishService`? Does avahi forbids changes to a service entry group when a different host entry group is committed?

src/mdns/mdns_avahi.hpp
```diff
@@ -335,22 +350,38 @@ class PublisherAvahi : public Publisher
...
-    int              mProtocol;
-    const char *     mDomain;
-    State            mState;
-    StateHandler     mStateHandler;
-    void *           mContext;
+    otbrError MakeFullName(char *aFullName, size_t aFullNameLength, const char *aName);
+
+    AvahiClient *mClient;
+    Host         mDefaultHost;
+    Hosts        mHosts;
```
wgtdkp, line 378: I think we just need a `mServicesGroup`.

wgtdkp (commented):
@Damian-Nordic Thanks very much for your contribution!

Sorry that I am not fully following your intention of adding `CommitHost`, see my comments below.

src/mdns/mdns_avahi.cpp
```diff
@@ -590,13 +636,47 @@ otbrError PublisherAvahi::UnpublishService(const char *aName, const char *aType)
...
+    VerifyOrExit(mState == State::kReady, errno = EAGAIN, ret = OTBR_ERROR_ERRNO);
+    VerifyOrExit(aName != nullptr, ret = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aAddress != nullptr, ret = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aAddressLength == sizeof(address.data.ipv6.address), ret = OTBR_ERROR_INVALID_ARGS);
 
-    return OTBR_ERROR_NONE;
+    SuccessOrExit(ret = MakeFullName(fullHostName, sizeof(fullHostName), aName));
+    SuccessOrExit(ret = GetOrCreateHost(mClient, aName, &host));
+    // Check if the update is needed
+    VerifyOrExit(!host->mIsCommitted || memcmp(host->mAddress.data.ipv6.address, aAddress, aAddressLength));
```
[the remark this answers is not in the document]
Damian-Nordic: Right, an attempt to call `avahi_entry_group_add_address` on an entry group which has already been committed results in a "Bad state" Avahi error. In theory, we could reset the entry group and commit it again every time `PublishHost` or `PublishService` is called, but it causes an unnecessary overhead on the Avahi side, so I thought it would be more elegant to introduce the "CommitHost" method which allows to do the commit explicitly. Let me know if you have a better idea on to handle this.

src/mdns/mdns_avahi.cpp
Damian-Nordic: Given that Avahi forbids modifying a committed entry group, in order to override an existing address we would need to store all service information in the Publisher class itself and rebuild the Avahi entry group from scratch every time any modification is needed. Do you think it would be a good approach?

src/mdns/mdns_avahi.hpp
Damian-Nordic: I already commented on that above, but I'll try to put it a little more clear:
1) an Avahi entry group can store address and service information for a single host
2) once an entry group has been committed, its services can be modified, but they can't be removed (unpublished), new services can't be added and the IP address of the host can't be changed. In case any of the forbidden operations is needed, the group must be rebuilt from scratch (at least I couldn't find any other way to do it).
3) `CommitHost` has been added to reduce number of `reset/commit` operations in case there's a need to do several changes to the given host.

src/mdns/mdns_avahi.hpp
Damian-Nordic: Could elaborate more? Do you just propose to change the name (why?) or a different structure of the data?

src/mdns/mdns_avahi.cpp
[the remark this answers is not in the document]
wgtdkp: What happens if I want to change my address for hostname `ot-my-host`? For example, `ot-my-host` was published with address `0xfd00::1` and I committed it. But later I want to change the address to `0xfd00::2` with another call to `Publisher::PublishHost()`, can we success?

src/mdns/mdns_avahi.hpp
wgtdkp: Thanks, I get you now.

It seems you are using the same entry group for a host and all its associated services, right? What if we use separate entry groups for each services and the host? Although we manage services as a member list of host, there are no strict requirements for the service and host be in the same avahi entry group. To be more precise, service entries (TXT RR, SRV RR) and host entries (address RRs) are in different entry groups, because they have different DNS names.

I would suggest we group DNS RRs (represented by the `Service` and `Host` objects) which different DNS names in different avahi entry groups. This makes it easier to modify each service and host. Thoughts?

Damian-Nordic: I'll check if the service resolution will work if we do so.

src/mdns/mdns_avahi.hpp
wgtdkp: I mean we can keep the `AvahiEntryGroup *mGroup;`, but just rename it to `AvahiEntryGroup *mServicesGroup;`

src/mdns/mdns_avahi.hpp
wgtdkp: Thanks, I believe this should work :smiley:

src/mdns/mdns_avahi.hpp
Damian-Nordic: I'll check if it will remain valid after the rework we talked above, but I changed it to `Host` as we not only needed the avahi group, but also other host-related fields like services connected with the host. But probably I'll organize it differently if we go with "group-per-service" approach :)

wgtdkp: @Damian-Nordic Thanks, Looking forward to the change.

src/mdns/mdns_avahi.hpp
Damian-Nordic: Seems to be working :) I've reworked the PR according to your suggestions.

src/mdns/mdns_avahi.cpp
```diff
@@ -533,34 +629,35 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
-        }
+        SuccessOrExit(ret = CreateService(*mClient, aName, aType, serviceIt));
+    }
+    else if (serviceIt->mHostName != safeHostName || serviceIt->mPort != aPort)
+    {
+        SuccessOrExit(ret = ResetGroup(serviceIt->mGroup));
+    }
+    else
+    {
+        otbrLog(OTBR_LOG_INFO, "MDNS update service %s for host '%s'", aName, safeHostName);
```
wgtdkp, line 644: nit:
```suggestion
        otbrLog(OTBR_LOG_INFO, "[mdns] update service %s for host '%s'", aName, safeHostName);
```

We use `[mdns]` in the mDNSResponder implementation, let's follow it :).

src/mdns/mdns_avahi.cpp
```diff
@@ -533,34 +629,35 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+        error = avahi_entry_group_update_service_txt_strlst(serviceIt->mGroup, AVAHI_IF_UNSPEC, mProtocol,
+                                                            AvahiPublishFlags{}, aName, aType, mDomain, last);
+        ExitNow();
     }
 
-    otbrLog(OTBR_LOG_INFO, "MDNS create service %s", aName);
-    error = avahi_entry_group_add_service_strlst(mGroup, AVAHI_IF_UNSPEC, mProtocol, static_cast<AvahiPublishFlags>(0),
+    otbrLog(OTBR_LOG_INFO, "MDNS create service %s for host '%s'", aName, safeHostName);
```
wgtdkp, line 650: nit: it will print `... for host ''` when we are not advertising a service for a custom host.

src/mdns/mdns_avahi.cpp
```diff
@@ -580,32 +677,127 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
 
-    VerifyOrDie(false, "UnpublishService is not implemented with avahi");
+    VerifyOrExit(aName != nullptr, ret = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aType != nullptr, ret = OTBR_ERROR_INVALID_ARGS);
 
-    return OTBR_ERROR_NONE;
+    serviceIt = FindService(aName, aType);
+    VerifyOrExit(serviceIt != mServices.end());
+
+    otbrLog(OTBR_LOG_INFO, "MDNS delete service '%s'", aName);
```
wgtdkp, line 690: nit:
```suggestion
    otbrLog(OTBR_LOG_INFO, "MDNS unpublish service '%s.%s'", aName, aType);
```

src/mdns/mdns_avahi.cpp
```diff
@@ -533,34 +629,35 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+        otbrLog(OTBR_LOG_INFO, "MDNS update service %s for host '%s'", aName, safeHostName);
+        error = avahi_entry_group_update_service_txt_strlst(serviceIt->mGroup, AVAHI_IF_UNSPEC, mProtocol,
+                                                            AvahiPublishFlags{}, aName, aType, mDomain, last);
+        ExitNow();
     }
 
-    otbrLog(OTBR_LOG_INFO, "MDNS create service %s", aName);
-    error = avahi_entry_group_add_service_strlst(mGroup, AVAHI_IF_UNSPEC, mProtocol, static_cast<AvahiPublishFlags>(0),
+    otbrLog(OTBR_LOG_INFO, "MDNS create service %s for host '%s'", aName, safeHostName);
+    error = avahi_entry_group_add_service_strlst(serviceIt->mGroup, AVAHI_IF_UNSPEC, mProtocol, AvahiPublishFlags{},
```
wgtdkp, line 651: Do we want to remove the service from the service list if failed to publish the service?

src/mdns/mdns_avahi.cpp
```diff
@@ -580,32 +677,127 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+
+    hostIt->mAddress = address;
+
+exit:
+
+    if (error)
+    {
+        ret = OTBR_ERROR_MDNS;
+        otbrLog(OTBR_LOG_ERR, "Failed to publish host for avahi error: %s!", avahi_strerror(error));
+    }
```
wgtdkp, line 752: Do we want to remove the host from the host list if failed to publish the host?

src/mdns/mdns_avahi.cpp
```diff
@@ -580,32 +677,127 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+otbrError PublisherAvahi::MakeFullName(char *aFullName, size_t aFullNameLength, const char *aName)
+{
+    otbrError ret = OTBR_ERROR_NONE;
+
+    assert(aFullName != nullptr);
+    assert(aName != nullptr);
+
+    const char *domain = mDomain == nullptr ? "local." : mDomain;
+    const int   count  = snprintf(aFullName, aFullNameLength, "%s.%s", aName, domain);
+    VerifyOrExit(count >= 0, ret = OTBR_ERROR_INVALID_ARGS);
```
wgtdkp, line 796:
```suggestion
    assert(aFullName != nullptr);
    assert(aName != nullptr);

    otbrError ret = OTBR_ERROR_NONE;
    const char *domain = mDomain == nullptr ? "local." : mDomain;
    const int   count  = snprintf(aFullName, aFullNameLength, "%s.%s", aName, domain);

    VerifyOrExit(count >= 0, ret = OTBR_ERROR_INVALID_ARGS);
```

wgtdkp (commented):
Looks good overall, some nits:

We need to add tests for the new avahi mDNS implementation, missing tests can be found here:
https://github.com/openthread/ot-br-posix/blob/f6404ab573bd410f779304b4fa5187826cf4b87b/tests/mdns/test-single-custom-host#L44-L47

https://github.com/openthread/ot-br-posix/blob/f6404ab573bd410f779304b4fa5187826cf4b87b/tests/mdns/test-multiple-custom-hosts#L49-L52

src/mdns/mdns_avahi.cpp
wgtdkp: I think a better practice is that we add the service to the list only after successfully publishing it. Thoughts?

src/mdns/mdns_avahi.cpp
```diff
@@ -412,19 +398,126 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
...
+{
+    assert(aHostName != nullptr);
+
+    return std::find_if(mHosts.begin(), mHosts.end(), [aHostName](const Host &h) { return h.mHostName == aHostName; });
+}
+
+otbrError PublisherAvahi::CreateHost(AvahiClient &aClient, const char *aHostName, Hosts::iterator &aOutHostIt)
 {
-    VerifyOrExit(mGroup == nullptr);
+    otbrError ret = OTBR_ERROR_NONE;
```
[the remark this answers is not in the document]
Damian-Nordic: I wanted to keep it consistent with the existing code in the file which used `ret` for `otbrError` and `error` for avahi error codes. Should I change it throughout the file?

src/mdns/mdns_avahi.cpp
Damian-Nordic: Yes, that was my intention. Do you think we should add "for host" part only for custom hosts?

src/mdns/mdns_avahi.cpp
Damian-Nordic: Ok, the advertising proxy seems to call the `Publish*` methods periodically, so I thought it would be better to allocate the structures once and just repeat the Avahi calls, but I can change it.

src/mdns/mdns_avahi.cpp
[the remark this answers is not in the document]
wgtdkp: It's will be great if you can change it throughout the file.

src/mdns/mdns_avahi.cpp
wgtdkp: What about printing `... for localhost` if `aHostName` is null?

src/mdns/mdns_avahi.cpp
```diff
@@ -378,10 +367,7 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
 
 void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupState aState)
```
wgtdkp, line 368: I think the service is not actually multicasted on the link immediately after calling `avahi_entry_group_commit`
when publishing a service (mDNS needs to do probing before announcing new RRs).

The Advertising Proxy waits for the publisher to call the `mServiceHandler` to notify the completion of the mDNS registration:
https://github.com/openthread/ot-br-posix/blob/a1e0c7344e137bda1143bf72a252e36f60d392e6/src/mdns/mdns.hpp#L335-L339

I think the proper place to call the `mServiceHandler` is at here. Thoughts?

Damian-Nordic: Good catch, I forgot about the handlers. Thanks!

src/mdns/mdns_avahi.cpp
```diff
@@ -533,79 +660,188 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+    Services::iterator serviceIt;
+
+    VerifyOrExit(aName != nullptr, error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aType != nullptr, error = OTBR_ERROR_INVALID_ARGS);
+
+    serviceIt = FindService(aName, aType);
+    VerifyOrExit(serviceIt != mServices.end());
+
+    otbrLog(OTBR_LOG_INFO, "[mdns] unpublish service %s.%s", aName, aType);
+    avahiError = avahi_entry_group_free(serviceIt->mGroup);
```
wgtdkp, line 729:
```suggestion
    avahiError = FreeGroup(serviceIt->mGroup);
```

src/mdns/mdns_avahi.cpp
```diff
@@ -533,79 +660,188 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+
+    VerifyOrExit(aName != nullptr, error = OTBR_ERROR_INVALID_ARGS);
 
-    VerifyOrDie(false, "UnpublishHost is not implemented with avahi");
+    hostIt = FindHost(aName);
+    VerifyOrExit(hostIt != mHosts.end());
 
-    return OTBR_ERROR_NONE;
+    otbrLog(OTBR_LOG_INFO, "[mdns] delete host %s", aName);
+    avahiError = avahi_entry_group_free(hostIt->mGroup);
```
wgtdkp, line 818:
```suggestion
    avahiError = FreeGroup(hostIt->mGroup);
```

wgtdkp (approved):
LGTM :+1: 
Thanks @Damian-Nordic nice work!

simonlingoogle (approved):
LGTM :+1: 

nit

src/mdns/mdns_avahi.hpp
```diff
@@ -328,29 +328,55 @@ class PublisherAvahi : public Publisher
...
-    AvahiEntryGroup *mGroup;
-    Poller           mPoller;
-    int              mProtocol;
-    const char *     mDomain;
-    State            mState;
-    StateHandler     mStateHandler;
-    void *           mContext;
+    void        CallHostOrServiceCallback(AvahiEntryGroup *aGroup, otbrError aError) const;
+
+    otbrError MakeFullName(char *aFullName, size_t aFullNameLength, const char *aName);
```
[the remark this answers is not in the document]
Damian-Nordic: I would still like to handle "too small buffer" case, but I will change it to `MakeFullName(const char* aName, std::string& aFullName)`. Would that be ok?

src/mdns/mdns_avahi.cpp
```diff
@@ -533,79 +663,174 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+    {
+        error = OTBR_ERROR_MDNS;
+        otbrLog(OTBR_LOG_ERR, "Failed to publish host for avahi error: %s!", avahi_strerror(avahiError));
+    }
 
-    return OTBR_ERROR_NONE;
+    if (error == OTBR_ERROR_ERRNO)
+    {
+        otbrLog(OTBR_LOG_ERR, "Failed to publish host: %s!", strerror(errno));
+    }
```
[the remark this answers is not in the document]
Damian-Nordic: Avahi errors are handled a few lines above where the message explicitly states that it's an error from avahi. Without that it would require from a user to go through both otbr and avahi error messages to figure out where it comes from. Can we leave it as it is? :)

src/mdns/mdns_avahi.hpp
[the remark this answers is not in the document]
wgtdkp: I think there will be no "too small buffer" errors if you use std::string for the full name. Thoughts?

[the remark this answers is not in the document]
Damian-Nordic: Right!

src/mdns/mdns_avahi.cpp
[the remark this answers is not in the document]
simonlingoogle: If so, we won't show error messages for errors like `OTBR_ERROR_INVALID_ARGS` ? Thoughts?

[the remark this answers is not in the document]
Damian-Nordic: Ah, sorry, I must haven't woken up already and misunderstood your comment. I tried to stay consistent with the existing code which also didn't print errors for OTBR_ERROR_INVALID_ARGS etc. If you want me to change it, I will, but there are more places which require such a change.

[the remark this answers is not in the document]
Damian-Nordic: Ok, done.

src/mdns/mdns_avahi.cpp
```diff
@@ -533,79 +663,184 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+    {
+        error = OTBR_ERROR_MDNS;
+        otbrLog(OTBR_LOG_ERR, "Failed to publish service for avahi error: %s!", avahi_strerror(avahiError));
+    }
+    else if (error != OTBR_ERROR_NONE)
     {
-        ret = OTBR_ERROR_MDNS;
-        otbrLog(OTBR_LOG_ERR, "Failed to publish service for avahi error: %s!", avahi_strerror(error));
+        otbrLog(OTBR_LOG_ERR, "Failed to publish service: %s!",
+                error == OTBR_ERROR_ERRNO ? strerror(errno) : otbrErrorString(error));
```
simonlingoogle, line 712:
```suggestion
        otbrLog(OTBR_LOG_ERR, "Failed to publish service: %s!",
                otbrErrorString(error));
```

src/mdns/mdns_avahi.cpp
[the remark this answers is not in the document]
simonlingoogle: Thanks @Damian-Nordic ,

I am not sure if there is a convention of not printing `OTBR_ERROR_INVALID_ARGS`. But I think we should print all errors so that we don't miss any errors that could happen.

BTW, I found that `otbrErrorString` function already handled `OTBR_ERROR_ERRNO` differently, so we can make the code even simpler.

bukepo (approved):

src/mdns/mdns_avahi.cpp
```diff
@@ -533,79 +663,184 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+    hostIt       = FindHost(aName);
+
+    if (hostIt == mHosts.end())
+    {
+        SuccessOrExit(error = CreateHost(*mClient, aName, hostIt));
+    }
+    else if (memcmp(hostIt->mAddress.data.ipv6.address, aAddress, aAddressLength))
+    {
+        // Check if the update is necessary
+        VerifyOrExit(memcmp(hostIt->mAddress.data.ipv6.address, aAddress, aAddressLength));
```
jwhui, line 767: Redundant with above?

src/mdns/mdns_avahi.cpp
```diff
@@ -533,79 +663,184 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+        error = OTBR_ERROR_MDNS;
+        otbrLog(OTBR_LOG_ERR, "Failed to publish host for avahi error: %s!", avahi_strerror(avahiError));
+    }
+    else if (error != OTBR_ERROR_NONE)
+    {
+        otbrLog(OTBR_LOG_ERR, "Failed to publish host: %s!",
+                error == OTBR_ERROR_ERRNO ? strerror(errno) : otbrErrorString(error));
+    }
+
+    if (error != OTBR_ERROR_NONE && hostIt != mHosts.end())
```
jwhui, line 807: Is there a case where `avahiError != 0` and `error != OTBR_ERROR_NONE`? If not, we can combine in above else-if statement?

src/mdns/mdns_avahi.cpp
```diff
@@ -33,6 +33,7 @@
 
 #include "mdns/mdns_avahi.hpp"
 
+#include <algorithm>
```
jwhui, line 36: nit - separate into own block:
```suggestion
#include <algorithm>

```

src/mdns/mdns_avahi.cpp
```diff
@@ -412,17 +401,154 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
...
 exit:
-    if (mGroup == nullptr)
+    return error;
+}
+
+otbrError PublisherAvahi::CreateGroup(AvahiClient &aClient, AvahiEntryGroup *&aOutGroup)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(aOutGroup == nullptr, error = OTBR_ERROR_INVALID_ARGS);
```
jwhui, line 490: Can we use `assert()` instead?

src/mdns/mdns_avahi.cpp
```diff
@@ -533,79 +663,184 @@ otbrError PublisherAvahi::PublishService(const char *   aHostName,
...
+    else if (error != OTBR_ERROR_NONE)
     {
-        ret = OTBR_ERROR_MDNS;
-        otbrLog(OTBR_LOG_ERR, "Failed to publish service for avahi error: %s!", avahi_strerror(error));
+        otbrLog(OTBR_LOG_ERR, "Failed to publish service: %s!",
+                error == OTBR_ERROR_ERRNO ? strerror(errno) : otbrErrorString(error));
     }
 
-    if (ret == OTBR_ERROR_ERRNO)
+    if (error != OTBR_ERROR_NONE && serviceIt != mServices.end())
```
jwhui, line 715: Is there a case where `avahiError != 0` and `error != OTBR_ERROR_NONE`? If not, we can combine in above else-if statement?

src/mdns/mdns_avahi.hpp
```diff
@@ -328,29 +328,55 @@ class PublisherAvahi : public Publisher
...
+
+    Services::iterator FindService(const char *aName, const char *aType);
+    otbrError          CreateService(AvahiClient &       aClient,
+                                     const char *        aName,
+                                     const char *        aType,
+                                     Services::iterator &aOutServiceIt);
+
+    otbrError   CreateGroup(AvahiClient &aClient, AvahiEntryGroup *&aOutGroup);
+    otbrError   ResetGroup(AvahiEntryGroup *aGroup);
+    otbrError   FreeGroup(AvahiEntryGroup *aGroup);
```
jwhui, line 363: Declare these methods as `static`?

jwhui (commented):
Overall, looks great. Thanks! Some minor comments inline.

src/mdns/mdns_avahi.cpp
Damian-Nordic: Thanks! I forgot to remove the line :)

src/mdns/mdns_avahi.cpp
Damian-Nordic: When `avahiError != 0`, `error` is assigned `OTBR_ERROR_MDNS` so after the `if-else-if- block it's possible.

src/mdns/mdns_avahi.cpp
Damian-Nordic: See the comment above.

jwhui (approved):
Thanks! 👍🏼

PR #706, opened 2021-02-22
[dnssd] implement DNS-SD Discovery Proxy

simonlingoogle:
This commit implements DNS-SD Discovery Proxy ([RFC 8766](https://tools.ietf.org/html/rfc8766)) over mDNSResponder.

### DIFF
--- a/src/agent/discovery_proxy.cpp
+++ b/src/agent/discovery_proxy.cpp
@@ -1,0 +1,390 @@
+/*
+ *    Copyright (c) 2021, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   The file implements the DNS-SD Discovery Proxy.
+ */
+
+#if OTBR_ENABLE_DNSSD_DISCOVERY_PROXY
+
+#include "agent/discovery_proxy.hpp"
+
+#include <algorithm>
+#include <string>
+
+#include <assert.h>
+
+#include <openthread/dnssd_server.h>
+
+#include "common/code_utils.hpp"
+#include "common/dns_utils.hpp"
+#include "common/logging.hpp"
+
+namespace otbr {
+namespace Dnssd {
+
+DiscoveryProxy::DiscoveryProxy(Ncp::ControllerOpenThread &aNcp, Mdns::Publisher &aPublisher)
+    : mNcp(aNcp)
+    , mMdnsPublisher(aPublisher)
+{
+}
+
+void DiscoveryProxy::Start(void)
+{
+    otDnssdQuerySetCallbacks(mNcp.GetInstance(), &DiscoveryProxy::OnDiscoveryProxySubscribe,
+                             &DiscoveryProxy::OnDiscoveryProxyUnsubscribe, this);
+
+    mMdnsPublisher.SetSubscriptionCallbacks(
+        [this](const std::string &aType, const Mdns::Publisher::DiscoveredInstanceInfo &aInstanceInfo) {
+            OnServiceDiscovered(aType, aInstanceInfo);
+        },
+
+        [this](const std::string &aHostName, const Mdns::Publisher::DiscoveredHostInfo &aHostInfo) {
+            OnHostDiscovered(aHostName, aHostInfo);
+        });
+
+    otbrLog(OTBR_LOG_INFO, "[discproxy] started");
+}
+
+void DiscoveryProxy::Stop(void)
+{
+    otDnssdQuerySetCallbacks(mNcp.GetInstance(), nullptr, nullptr, nullptr);
+    mMdnsPublisher.SetSubscriptionCallbacks(nullptr, nullptr);
+
+    otbrLog(OTBR_LOG_INFO, "[discproxy] stopped");
+}
+
+void DiscoveryProxy::OnDiscoveryProxySubscribe(void *aContext, const char *aFullName)
+{
+    reinterpret_cast<DiscoveryProxy *>(aContext)->OnDiscoveryProxySubscribe(aFullName);
+}
+
+void DiscoveryProxy::OnDiscoveryProxySubscribe(const char *aFullName)
+{
+    std::string                    fullName(aFullName);
+    std::string                    instanceName, serviceName, hostName, domain;
+    otbrError                      error = OTBR_ERROR_NONE;
+    MdnsSubscriptionList::iterator it;
+    DnsNameType                    nameType = GetDnsNameType(fullName);
+
+    otbrLog(OTBR_LOG_INFO, "[discproxy] subscribe: %s", fullName.c_str());
+
+    switch (nameType)
+    {
+    case kDnsNameTypeService:
+        SuccessOrExit(error = SplitFullServiceName(fullName, serviceName, domain));
+        break;
+    case kDnsNameTypeInstance:
+        SuccessOrExit(error = SplitFullServiceInstanceName(fullName, instanceName, serviceName, domain));
+        break;
+    case kDnsNameTypeHost:
+        SuccessOrExit(error = SplitFullHostName(fullName, hostName, domain));
+        break;
+    default:
+        ExitNow(error = OTBR_ERROR_NOT_IMPLEMENTED);
+    }
+
+    it = std::find_if(mSubscriptions.begin(), mSubscriptions.end(), [&](const MdnsSubscription &aSubscription) {
+        return aSubscription.Matches(instanceName, serviceName, hostName, domain);
+    });
+
+    VerifyOrExit(it == mSubscriptions.end(), it->mSubscriptionCount++);
+
+    mSubscriptions.emplace_back(instanceName, serviceName, hostName, domain);
+
+    {
+        MdnsSubscription &subscription = mSubscriptions.back();
+
+        otbrLog(OTBR_LOG_DEBUG, "[discproxy] subscriptions: %sx%d", subscription.ToString().c_str(),
+                subscription.mSubscriptionCount);
+
+        if (GetServiceSubscriptionCount(instanceName, serviceName) == 1)
+        {
+            if (subscription.mHostName.empty())
+            {
+                mMdnsPublisher.SubscribeService(serviceName, instanceName);
+            }
+            else
+            {
+                mMdnsPublisher.SubscribeHost(hostName);
+            }
+        }
+    }
+
+exit:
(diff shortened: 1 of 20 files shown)


.gitmodules
```diff
@@ -1,7 +1,3 @@
-[submodule "third_party/openthread/repo"]
-	path = third_party/openthread/repo
-	url = https://github.com/openthread/openthread.git
-	branch = master
```
wgtdkp, line 4: This must be an accident :)

src/agent/discovery_proxy.cpp
```diff
@@ -0,0 +1,369 @@
...
+            {
+                mMdnsPublisher.UnsubscribeHost(hostName);
+            }
+        }
+    }
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        otbrLog(OTBR_LOG_WARNING, "[dnssd discovery proxy] failed to unsubscribe %s.%s: %s",
+                instanceName.empty() ? "ANY" : instanceName.c_str(), aFullName, otbrErrorString(error));
```
wgtdkp, line 214:
```suggestion
        otbrLog(OTBR_LOG_WARNING, "[dnssd discovery proxy] failed to unsubscribe %s: %s", aFullName, otbrErrorString(error));
```

`aFullName` should already include the instance name, right?

src/agent/discovery_proxy.cpp
```diff
@@ -0,0 +1,369 @@
...
+    }
+}
+
+void DiscoveryProxy::OnServiceDiscovered(const std::string &                            aType,
+                                         const Mdns::Publisher::DiscoveredInstanceInfo &aInstanceInfo)
+{
+    otDnssdServiceInstanceInfo instanceInfo;
+
+    otbrLog(OTBR_LOG_INFO,
+            "[dnssd discovery proxy] service discovered: %s, instance %s hostname %s address %s:%d priority %d "
```
wgtdkp, line 224:
```suggestion
            "[dnssd discovery proxy] service discovered: %s, instance %s hostname %s address [%s]:%d priority %d "
```

The port will look like a portion of the IPv6 address without `[]`.

src/agent/discovery_proxy.cpp
```diff
@@ -0,0 +1,369 @@
...
+        if (aSubscription.MatchesServiceInstance(aType, aInstanceInfo.mName))
+        {
+            std::string serviceFullName  = aType + "." + aSubscription.mDomain;
+            std::string hostName         = TranslateDomain(aInstanceInfo.mHostName, aSubscription.mDomain);
+            std::string instanceFullName = aInstanceInfo.mName + "." + serviceFullName;
+
+            instanceInfo.mFullName = instanceFullName.c_str();
+            instanceInfo.mHostName = hostName.c_str();
+
+            otDnssdQueryNotifyServiceInstance(mNcp.GetInstance(), serviceFullName.c_str(), &instanceInfo);
```
wgtdkp, line 251: The user could subscribe to a specific service instance, right? Shouldn't we notify OT with the full service instance name  in this case?

src/agent/discovery_proxy.cpp
```diff
@@ -0,0 +1,369 @@
...
+{
+    otDnssdHostInfo hostInfo;
+
+    otbrLog(OTBR_LOG_INFO, "[dnssd discovery proxy] host discovered: %s hostname %s address %s", aHostName.c_str(),
+            aHostInfo.mHostName.c_str(), aHostInfo.mAddress.ToString().c_str());
+
+    CheckHostnameSanity(aHostInfo.mHostName);
+
+    hostInfo.mAddress = *reinterpret_cast<const otIp6Address *>(&aHostInfo.mAddress);
+    hostInfo.mTtl     = std::min(aHostInfo.mTtl, static_cast<uint32_t>(kServiceTtlCapLimit));
```
wgtdkp, line 267: nit: define a short function which returns the correct TTL.

src/agent/discovery_proxy.cpp
```diff
@@ -0,0 +1,369 @@
...
+    instanceInfo.mPriority  = aInstanceInfo.mPriority;
+    instanceInfo.mWeight    = aInstanceInfo.mWeight;
+    instanceInfo.mTxtLength = static_cast<uint16_t>(aInstanceInfo.mTxtData.size());
+    instanceInfo.mTxtData   = aInstanceInfo.mTxtData.data();
+    instanceInfo.mTtl       = std::min(aInstanceInfo.mTtl, static_cast<uint32_t>(kServiceTtlCapLimit));
+
+    std::for_each(mSubscriptions.begin(), mSubscriptions.end(), [&](const MdnsSubscription &aSubscription) {
+        if (aSubscription.MatchesServiceInstance(aType, aInstanceInfo.mName))
+        {
+            std::string serviceFullName  = aType + "." + aSubscription.mDomain;
```
wgtdkp, line 244: Do we want to translate domain for service names as well?

src/agent/discovery_proxy.cpp
```diff
@@ -0,0 +1,369 @@
...
+    return targetName;
+}
+
+int DiscoveryProxy::GetServiceSubscriptionCount(const std::string &aInstanceName, const std::string &aServiceName)
+{
+    return std::accumulate(
+        mSubscriptions.begin(), mSubscriptions.end(), 0, [&](int aAccum, const MdnsSubscription &aSubscription) {
+            return aAccum + (aSubscription.mInstanceName == aInstanceName && aSubscription.mServiceName == aServiceName
+                                 ? aSubscription.mSubscriptionCount
+                                 : 0);
```
wgtdkp, line 300: I think there should only one subscription that has the same instance and service name, right? Seems we don't to accumulate.

src/agent/discovery_proxy.hpp
```diff
@@ -0,0 +1,143 @@
...
+#define OTBR_DISCOVERY_PROXY_HPP_
+
+#if OTBR_ENABLE_DNSSD_DISCOVERY_PROXY
+
+#include <set>
+#include <utility>
+#include <openthread/dns.h>
+#include <openthread/instance.h>
+
+#include <stdint.h>
```
wgtdkp, line 44:
```suggestion
#include <set>
#include <utility>

#include <stdint.h>

#include <openthread/dns.h>
#include <openthread/instance.h>
```

src/agent/discovery_proxy.hpp
```diff
@@ -0,0 +1,143 @@
...
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definition for DNS-SD Discovery Proxy.
+ */
+
+#ifndef OTBR_DISCOVERY_PROXY_HPP_
+#define OTBR_DISCOVERY_PROXY_HPP_
```
wgtdkp, line 35:
```suggestion
#ifndef OTBR_AGENT_DISCOVERY_PROXY_HPP_
#define OTBR_AGENT_DISCOVERY_PROXY_HPP_
```

The header guard should match the source directory.

src/mdns/mdns.hpp
```diff
@@ -93,6 +94,68 @@ class Publisher : public MainloopProcessor
...
+        Ip6Address  mAddress;  ///< IP6 address.
+        uint32_t    mTtl;      ///< Host TTL.
+    };
+
+    /**
+     * This function is called to notify a discovered service instance.
+     *
+     */
+    typedef std::function<void(const std::string &aType, const DiscoveredInstanceInfo &aInstanceInfo)>
+        DiscoveredServiceInstanceCallback;
```
wgtdkp, line 150: nit:
```suggestion
    using DiscoveredServiceInstanceCallback =  std::function<void(const std::string &aType, const DiscoveredInstanceInfo &aInstanceInfo)>;
```

More C++11 style :)

src/mdns/mdns_mdnssd.cpp
```diff
@@ -703,6 +877,290 @@ void Publisher::Destroy(Publisher *aPublisher)
...
+    otbrLog(OTBR_LOG_INFO, "DNSServiceGetAddrInfo reply: %d, flags=%u, host=%s, sa_family=%d", aErrorCode, aFlags,
+            aHostName, aAddress->sa_family);
+
+    VerifyOrExit(aErrorCode == kDNSServiceErr_NoError);
+    VerifyOrExit((aFlags & kDNSServiceFlagsAdd) && aAddress->sa_family == AF_INET6);
+
+    address.CopyFrom(*reinterpret_cast<const struct sockaddr_in6 *>(aAddress));
+    VerifyOrExit(!address.IsLinkLocal(),
+                 otbrLog(OTBR_LOG_DEBUG, "[mdns] DNSServiceGetAddrInfo ignore link-local address %s",
+                         address.ToString().c_str()));
```
wgtdkp, line 1064: What will happen in this case? Will it result in no address or will we get more IPv6 addresses when the mDNSResponder finds more?

src/agent/discovery_proxy.cpp
simonlingoogle: The full service instance name is in `instanceInfo` already

simonlingoogle: Technically `serviceFullName` is redundant info because `instanceInfo` already contains the full instance name. However, we notify OT core with both service full name and instance full name that OT core don't need to parse or split.

src/agent/discovery_proxy.cpp
simonlingoogle: I think this statement is already doing it.

src/agent/discovery_proxy.cpp
simonlingoogle: There can be duplicate subscription from OT core to Discovery Proxy.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -693,6 +747,126 @@ PublisherMDnsSd::HostIterator PublisherMDnsSd::FindPublishedHost(const char *aHo
...
+    }
+}
+
+void PublisherMDnsSd::UnsubscribeService(const std::string &aType, const std::string &aInstanceName)
+{
+    otbrError error = OTBR_ERROR_NOT_FOUND;
+
+    OTBR_UNUSED_VARIABLE(error);
+
+    while (true)
```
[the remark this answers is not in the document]
simonlingoogle: There are.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -693,6 +747,126 @@ PublisherMDnsSd::HostIterator PublisherMDnsSd::FindPublishedHost(const char *aHo
...
+    mSubscribedHosts.back().Resolve();
+}
+
+void PublisherMDnsSd::UnsubscribeHost(const std::string &aHostName)
+{
+    otbrError error = OTBR_ERROR_NOT_FOUND;
+
+    OTBR_UNUSED_VARIABLE(error);
+
+    while (true)
```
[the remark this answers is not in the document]
simonlingoogle: There are.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -703,6 +877,290 @@ void Publisher::Destroy(Publisher *aPublisher)
...
+    OTBR_UNUSED_VARIABLE(aServiceRef);
+
+    otbrLog(OTBR_LOG_INFO, "[mdns] DNSServiceBrowse reply: %s.%s%s inf %u, flags=%u, error=%d", aInstanceName, aType,
+            aDomain, aInterfaceIndex, aFlags, aErrorCode);
+
+    VerifyOrExit(aErrorCode == kDNSServiceErr_NoError);
+    VerifyOrExit(aFlags & kDNSServiceFlagsAdd);
+
+    DeallocateServiceRef();
+    Resolve(aInterfaceIndex, aInstanceName, aType, aDomain);
```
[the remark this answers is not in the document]
simonlingoogle: I didn't put `DeallocateServiceRef()` inside `Resolve` because it's this `ServiceRef` is created by `DNSServiceBrowse`, so it's `DNSServiceBrowseReply`'s duty to clean the `ServiceRef` when it's not needed. 
So, we need to `DeallocateServiceRef()` in `HandleBrowseResult` (which is the DNSServiceBrowseReply), no matter `HandleBrowseResult` calls `Resolve` or not.

src/mdns/mdns_mdnssd.cpp
simonlingoogle: it will wait for more IPv6 addresses. The process stop when `kDNSServiceFlagsMoreComing` flag is 1.

src/mdns/mdns_mdnssd.cpp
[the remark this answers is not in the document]
wgtdkp: Also makes sense to me. Thanks!

src/agent/discovery_proxy.cpp
wgtdkp: Why not have a `mServiceName` pointer which points to middle of `mFullName` of `instanceInfo`. I believe the user will expect that the same name should be used in the notify callback.

wgtdkp (approved):
LGTM :+1:

abtink (approved):
Overall looks good to me. 👍 
But would suggest other who are more familiar with the orbr core to also have a look.

src/agent/discovery_proxy.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * This method starts the Discovery Proxy.
+     *
+     */
+    void Start(void);
+
+    /**
+     * This method stops the Discovery Proxy.
+     */
```
jwhui, line 77: nit: be consistent with the empty line spacing?

src/common/types.hpp
```diff
@@ -189,6 +200,23 @@ class Ip6Address
      */
     bool IsMulticast(void) const { return m8[0] == 0xff; }
 
+    /**
+     * This method returns if the Ip6 address is a link-local address.
```
jwhui, line 204: nit: be consistent with Ip6 vs. IPv6.

src/common/dns_utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+otbrError SplitFullServiceInstanceName(const std::string &aFullName,
+                                       std::string &      aInstanceName,
+                                       std::string &      aType,
+                                       std::string &      aDomain)
+{
+    otbrError error = OTBR_ERROR_INVALID_ARGS;
+    size_t    dotPos[3];
+
+    dotPos[0] = aFullName.find_first_of('.');
```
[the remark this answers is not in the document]
simonlingoogle: This issue is not introduced by this PR.
I prefer to solve it in another following PR.
Thoughts?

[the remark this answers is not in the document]
jwhui: Resolving in a separate PR is OK for me.

jwhui (approved):

PR #712, opened 2021-02-25
[setup] enable IPv6 routing and DNS-based service discovery

wgtdkp:
This PR builds OTBR with IPv6 routing and DNS-based service discovery by default:
1. Enables the `Border Routing`, `SRP` features in default setup.
2. change `DHCPV6_PD` to default `0` for raspberry Pi platform because it requires IPv6 supports in `dhcpcd` while the Border Routing feature needs to disable it.
3. change `NETWORK_MANAGER` to default `0` for raspberry Pi platform because the `otbr-agent` will not be able to receive ICMPv6 Router Solicitations if it is enabled ( Not clear of the root causes yet).
4. Auto-config `accept_ra` options.

Depends-On: https://github.com/openthread/ot-br-posix/pull/715.


script/bootstrap
```diff
@@ -54,6 +54,12 @@ install_packages_apt()
 
     # mDNS
     sudo apt-get install --no-install-recommends -y libavahi-client3 libavahi-common-dev libavahi-client-dev avahi-daemon
+    (MDNS_RESPONDER_SOURCE_NAME=mDNSResponder-878.30.4 \
+        && cd /tmp \
+        && wget -4 --no-check-certificate https://opensource.apple.com/tarballs/mDNSResponder/$MDNS_RESPONDER_SOURCE_NAME.tar.gz \
+        && tar xvf $MDNS_RESPONDER_SOURCE_NAME.tar.gz \
+        && cd $MDNS_RESPONDER_SOURCE_NAME/mDNSPosix \
+        && make os=linux && sudo make install os=linux)
```
simonlingoogle, line 62: If avahi is not used by our reference release, we should remove these avahi packages to save image space.

Also, should we install one of mDSNResponder and Avahi instead of installing both?

simonlingoogle (commented):
LGTM overall

script/bootstrap
wgtdkp: The `bootstrap` script is not only for creating images. All OTBR developer's will need to run the script to set up their environment, such as who working on `avahi` mDNS implementations.

simonlingoogle: It's apparently not ideal to install two mDNS apps, especially for the reference BR release, which will always use mDNSResponder and should better uninstall Avahi packages. 

I wonder if `installing mDNS` should be part of `setup` script rather than `bootstrap`.

Thoughts? @wgtdkp @bukepo

etc/docker/Dockerfile
```diff
@@ -36,6 +38,8 @@ ARG NAT64
 ARG REFERENCE_DEVICE
 ARG RELEASE
 
+ENV INFRA_IF_NAME=${INFRA_IF_NAME:-eth0}
+ENV BORDER_ROUTING=${BORDER_ROUTING}:-0}
```
wgtdkp, line 42:
```suggestion
ENV BORDER_ROUTING=${BORDER_ROUTING:-0}
```

simonlingoogle (approved):
LGTM :+1:

Moandor-y (approved):
LGTM :+1:

third_party/openthread/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@
 
 set(OT_BORDER_AGENT ON CACHE BOOL "enable border agent" FORCE)
 set(OT_BORDER_ROUTER ON CACHE BOOL "enable border router feature" FORCE)
+set(OT_BORDER_ROUTING ${OTBR_BORDER_ROUTING} CACHE BOOL "enable border routing feature" FORCE)
```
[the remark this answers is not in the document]
wgtdkp: I think it's better to follow `OTBR_ADVERTISING_PROXY` as `SRP_SERVER` is required by the Advertising Proxy but not the Routing Manager.

script/bootstrap
wgtdkp: There will be a dozen dependencies that are not necessary if some options are not enabled, and it will be too much to define an variable for each such case...  I don't feel this really necessary.

wgtdkp: Anyway, we will switch to `avahi` when tests for avahi version is added and we can have the mDNSResponder library removed from here.

script/_initrc
```diff
@@ -104,6 +104,13 @@ if [[ ! ${PLATFORM+x} ]]; then
 fi
 echo "Current platform is $PLATFORM"
 
+# The DHCPV6_PD feature requires IPv6 features in dhcpcd but RIO
+# is not supported within dhcpcd.
+with BORDER_ROUTING && with DHCPV6_PD && die "BORDER_ROUTING and DHCPV6_PD cannot coexist!"
+
+# OTBR cannot receive RS messages when NETWORK_MANAGER is enabled.
+with BORDER_ROUTING && with NETWORK_MANAGER && die "BORDER_ROUTING and DHCPV6_PD cannot coexist!"
```
bukepo, line 112:
```suggestion
with BORDER_ROUTING && with NETWORK_MANAGER && die "BORDER_ROUTING and NETWORK_MANAGER cannot coexist!"
```

third_party/openthread/CMakeLists.txt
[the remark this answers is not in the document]
librasungirl: Sorry, missed below `if (OTBR_SRP_ADVERTISING_PROXY)` block previously.

librasungirl (approved):

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #713, opened 2021-03-01
[dbus] remove inclusion of code_utils.hpp in dbus api header

wgtdkp:
`common/code_utils.hpp` includes macros like `VerifiOrDie` which are not protected by C++ namespace, so it will easily get conflicted with other definitions.

We should not include `common/code_utils.hpp` in public headers.

Blocking: https://github.com/project-chip/connectedhomeip/pull/5062

### DIFF
--- a/src/dbus/client/thread_api_dbus.hpp
+++ b/src/dbus/client/thread_api_dbus.hpp
@@ -38,7 +38,6 @@
 
 #include <dbus/dbus.h>
 
-#include "common/code_utils.hpp"
 #include "common/types.hpp"
 #include "dbus/common/constants.hpp"
 #include "dbus/common/error.hpp"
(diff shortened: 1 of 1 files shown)


src/dbus/client/thread_api_dbus.hpp
```diff
@@ -38,7 +38,6 @@
 
 #include <dbus/dbus.h>
 
-#include "common/code_utils.hpp"
```
Moandor-y, line 41: We need `OTBR_UNUSED_VARIABLE` from this header file. If we remove this inclusion then `EmptyFree` needs to be moved into the cpp file.

wgtdkp: I removed `aData` from `EmptyFree`. How do you think?

Moandor-y (approved):
LGTM :+1:

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #716, opened 2021-03-04
[mdns] rename `Publisher` to `MDns`

simonlingoogle:
This commit renames `otbr::Mdns::Publisher` to `otbr::Mdns::MDns` because the "Publisher" will also allow subscribing to services, which is beyond the scope of "Publisher".

Related to #706 

Alternative names to consider:
- `otbr::Mdns::Client` - Our mDNS implementations are effectively mDNS clients (to Avahi/mDNSResponder), so I think it's also a good choice.
- `otbr::Mdns::Responder`

### DIFF
--- a/src/agent/advertising_proxy.cpp
+++ b/src/agent/advertising_proxy.cpp
@@ -130,9 +130,9 @@ static otError OtbrErrorToOtError(otbrError aError)
     return error;
 }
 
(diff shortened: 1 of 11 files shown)


simonlingoogle:
Close since no one seems to care about the name.

Issue #702, opened 2021-02-18
mDNS announce is missing necessary fields


mikhail-pridushchenko:
**Description:**
otbr-agent doesn't announce on mDNS the information required by the Thread 1.2 specification.
Particular fields missing:
* rv - record format.
* sb - state bitmap.

Fields missing in case of BBR function enabled (not required in non-CCM case, though looking sane):
* dn -Thread Domain Name.
* sq - BBR sequence number.
* bb - BBR port.

**To Reproduce:**

1. Git commit id: 35e9327b8e6c2dc2b9d0efbc3b64ea23d5db9894
2. IEEE 802.15.4 hardware platform: nRF52840.
3. Build steps
    a. Build OTBR with script parameters: `NETWORK_MANAGER=0 BACKBONE_ROUTER=1`
    b. Build RCP as: `make -f examples/Makefile-nrf52840 USB=1 BOOTLOADER=1 THREAD_VERSION=1.2`
4. Network topology
    OTBR running with BBR function enabled, right after network formation.

**Steps to check:**
In the OTBR shell:
```
# Install avahi-browse
$ sudo apt-get install avahi-utils
# The test itself
$ avahi-browse --cache --resolve _meshcop._udp
```

**Expected behavior:**
As mentioned in the 'To Reproduce' section, the TXT Answer Record from mDNS should contain the following fields (in addition to the present 'tv', 'xp', and 'nn' ones): 'rv', 'sb', 'dn', 'sq', 'bb'. As result the `avahi-browse` output should look like (main attention to the last line):
```
$ avahi-browse --cache --resolve _meshcop._udp
+   eth0 IPv6 OpenThreadDemo                                _meshcop._udp        local
+   eth0 IPv4 OpenThreadDemo                                _meshcop._udp        local
=   eth0 IPv6 OpenThreadDemo                                _meshcop._udp        local
   hostname = [otc-otbr-05.local]
   address = [fdaa:bb::ba27:ebff:fe58:9fa2]
   port = [49191]
   txt = ["tv=1.2.0" "xp=""""" "nn=OpenThreadDemo"]
=   eth0 IPv4 OpenThreadDemo                                _meshcop._udp        local
   hostname = [otc-otbr-05.local]
   address = [172.16.1.5]
   port = [49191]
   **txt = ["rv=1" "tv=1.2.0" "sb=""""" "xp=""""" "nn=OpenThreadDemo" "dn=Thread" "sq=""""" "bb="""""]**
```

[...]

PR #721, opened 2021-03-06
[border-agent] advertise more meshcop service TXT entries

wgtdkp:
This commit refactors the `Border Agent` module to advertise the full list of
TXT entries of the `_meshcop._udp` mDNS service. The event emitter is removed
and we rely on the `otThreadStateChanged` event to notify changes of extended
PAN ID, network name and backbone router state...

resolves https://github.com/openthread/ot-br-posix/issues/702.


simonlingoogle:
Removing `event emitter` is not directly related to publishing mDNS service and should be done in a separate PR?

Furthermore, I am not sure if it's a good idea to remove `event emitter`. 
The `event emitter` separate OTBR code from direct OT API calls. For example, if we are to support NCP mode in otbr-agent, we can't call OT API calls directly in OTBR code. I think it's better to keep OTBR code not directly dependent on the current RCP model to allow potential future enhancements. 

Thoughts? @wgtdkp @bukepo

src/agent/border_agent.cpp
```diff
@@ -227,29 +228,91 @@ static const char *ThreadVersionToString(uint16_t aThreadVersion)
...
+        else
+        {
+            txtList.emplace_back("sq", &bbrConfig.mSequenceNumber, sizeof(bbrConfig.mSequenceNumber));
+        }
+
+        txtList.emplace_back("dn", otThreadGetDomainName(instance));
+    }
+#endif
+
+    // TODO: append BBR sequence number and BBR port.
```
librasungirl, line 307: nit: sequence number is appended already
```suggestion
    // TODO: append BBR port.
```

Fyi, BBR port is 61631 by default so far, and not configurable so far.
https://github.com/openthread/openthread/blob/3852ac3d3e10602470a52d98c5c243595c8d9eb2/src/core/backbone_router/backbone_tmf.hpp#L44

src/agent/border_agent.cpp
```diff
@@ -277,158 +339,57 @@ void BorderAgent::StopPublishService(void)
     otbrLog(OTBR_LOG_INFO, "Stop publishing service");
 }
 
-void BorderAgent::SetNetworkName(const char *aNetworkName)
+void BorderAgent::HandleThreadStateChanged(otChangedFlags aFlags)
 {
-    strcpy_safe(mNetworkName, sizeof(mNetworkName), aNetworkName);
+    VerifyOrExit(IsPskcInitialized() && mPublisher != nullptr);
```
bukepo, line 344: Will this catch PSKc change?

wgtdkp: We wanted to stop publishing the `meshcop` service if `PSKc` is set to all-zeros, right?

bukepo: yes, and recover when PSKc is set.

wgtdkp: okay. but we cannot simply stop the mDNS publisher now, because the advertising proxy should continue working when `PSKc` is zeroed. I will revise to unpublish the `_meshcop._udp` service instead.

librasungirl (approved):

src/agent/border_agent.cpp
```diff
@@ -225,210 +228,179 @@ static const char *ThreadVersionToString(uint16_t aThreadVersion)
...
+void BorderAgent::HandleThreadStateChanged(otChangedFlags aFlags)
 {
-    VerifyOrExit(mThreadStarted != aStarted);
-
-    mThreadStarted = aStarted;
+    VerifyOrExit(mPublisher != nullptr);
+    VerifyOrExit((aFlags & OT_CHANGED_THREAD_ROLE) || (aFlags & OT_CHANGED_THREAD_EXT_PANID) ||
+                 (aFlags & OT_CHANGED_THREAD_NETWORK_NAME) || (aFlags & OT_CHANGED_ACTIVE_DATASET) ||
+                 (aFlags & OT_CHANGED_THREAD_PARTITION_ID) || (aFlags & OT_CHANGED_THREAD_BACKBONE_ROUTER_STATE) ||
+                 (aFlags & OT_CHANGED_PSKC));
```
simonlingoogle, line 358:
```suggestion
    VerifyOrExit(aFlags & (OT_CHANGED_THREAD_ROLE | OT_CHANGED_THREAD_EXT_PANID | OT_CHANGED_THREAD_NETWORK_NAME | OT_CHANGED_ACTIVE_DATASET | OT_CHANGED_THREAD_PARTITION_ID | OT_CHANGED_THREAD_BACKBONE_ROUTER_STATE | OT_CHANGED_PSKC ));
```

simonlingoogle (approved):
LGTM  :+1: 

Thanks

bukepo (approved):
LGTM:+1:

src/agent/agent_instance.cpp
```diff
@@ -40,7 +40,7 @@
 
 namespace otbr {
 
-AgentInstance::AgentInstance(Ncp::Controller *aNcp)
+AgentInstance::AgentInstance(otbr::Ncp::ControllerOpenThread &aNcp)
```
jwhui, line 43: Is it necessary to be fully-qualified? We are already in `otbr` namespace here.
```suggestion
AgentInstance::AgentInstance(Ncp::ControllerOpenThread &aNcp)
```

src/agent/border_agent.cpp
```diff
@@ -81,89 +84,91 @@ enum
...
-BorderAgent::BorderAgent(Ncp::Controller *aNcp)
+uint32_t BorderAgent::StateBitmap::ToUint32(void) const
+{
+    uint32_t bitmap = 0;
+
+    bitmap |= mConnectionMode & (0x07 << 0);
+    bitmap |= mThreadIfStatus & (0x03 << 3);
+    bitmap |= mAvailability & (0x03 << 5);
+    bitmap |= mBbrIsActive & (0x01 << 7);
+    bitmap |= mBbrIsPrimary & (0x01 << 8);
```
jwhui, line 95: Is the order of operations correct here? It seems we are shifting the mask first and then applying bit-wise & operator?

src/mdns/mdns_avahi.hpp
```diff
@@ -305,7 +305,7 @@ class PublisherAvahi : public Publisher
 private:
     enum
     {
-        kMaxSizeOfTxtRecord   = 128,
+        kMaxSizeOfTxtRecord   = 256,
```
jwhui, line 308: Does this include the length byte?

wgtdkp: This is actually the `rdata` length of the TXT RR. The maximum should be `65535` but the implementation seems want to avoid too big buffer.

There will be some issues when caller provides longer TXT entries. I think we can refactor to use `stl` later.

src/agent/border_agent.cpp
wgtdkp: @jwhui Thanks! You are right and I think we actually don't need to bit-wise & since those member fields are already bitfields. I updated, PTAL.

jwhui (approved):

PR #745, opened 2021-03-20
[dbus] add dbus logs

zhanglongxia:
This CL adds logs for each Dbus commands. The log helps developer get
more OpenThread stack information when the otbr-agent is crashed.


src/common/types.hpp
```diff
@@ -304,6 +345,45 @@ class Ip6Prefix
...
+     */
+    Ip6NetworkPrefix(void) { m64[0] = 0; }
+
+    /**
+     * Constructor with an 8-bytes prefix.
+     *
+     * @param[in] aPrefix  A pointer to the buffer containing the 8-bytes prefix.
+     *
+     */
+    Ip6NetworkPrefix(const uint8_t *aPrefix) { memcpy(m8, aPrefix, sizeof(*this)); }
```
jinran-google, line 368: Add a `length` parameter to avoid buffer overflow?

jinran-google (approved):
LGTM :+1:

src/common/types.cpp
```diff
@@ -30,12 +30,42 @@
...
+#include <openthread/dataset.h>
+
+#include "common/byteswap.hpp"
 #include "common/code_utils.hpp"
 #include "common/logging.hpp"
 #include "common/types.hpp"
 
 namespace otbr {
 
+std::string BytesToHexString(const uint8_t *aHex, uint16_t aLength)
```
simonlingoogle, line 42:
```suggestion
std::string BytesToHexString(const uint8_t *aBytes, uint16_t aLength)
```

src/common/types.cpp
```diff
@@ -30,12 +30,42 @@
...
+{
+    std::string str;
+    char        bytestr[3];
+
+    for (uint16_t i = 0; i < aLength; i++)
```
simonlingoogle, line 47:
```suggestion
    str.reserve(aLength * 2);

    for (uint16_t i = 0; i < aLength; i++)
```
use [string::reserve](https://www.cplusplus.com/reference/string/string/reserve/) for efficiency

src/common/types.cpp
```diff
@@ -114,13 +150,13 @@ std::string Ip6Prefix::ToString() const
...
 }
 
-std::string MacAddress::ToString(void) const
+std::string Ip6NetworkPrefix::ToString() const
 {
-    char strbuf[sizeof(m8) * 3];
+    char strbuf[INET6_ADDRSTRLEN];
 
-    snprintf(strbuf, sizeof(strbuf), "%02x:%02x:%02x:%02x:%02x:%02x", m8[0], m8[1], m8[2], m8[3], m8[4], m8[5]);
+    snprintf(strbuf, sizeof(strbuf), "%x:%x:%x:%x::0/64", htobe16(m16[0]), htobe16(m16[1]), htobe16(m16[2]),
```
simonlingoogle, line 157:
```suggestion
    snprintf(strbuf, sizeof(strbuf), "%x:%x:%x:%x::/64", htobe16(m16[0]), htobe16(m16[1]), htobe16(m16[2]),
```

simonlingoogle: Btw, this method only supports prefixes of length 64. Do we need to support other lengths?

src/common/types.hpp
```diff
@@ -96,6 +108,27 @@ enum
...
+/**
+ * This method converts the hex array to the string in hex format.
+ *
+ * @param[in]  aHex     A pointer to the hex array.
+ * @param[in]  aLength  The length of the hex array.
+ *
+ * @returns The string in hex format.
+ *
+ */
+std::string BytesToHexString(const uint8_t *aHex, uint16_t aLength);
```
simonlingoogle, line 120:
```suggestion
std::string BytesToHexString(const uint8_t *aBytes, uint16_t aLength);
```

src/common/types.cpp
```diff
@@ -30,12 +30,42 @@
 #include <sstream>
 #include <sys/socket.h>
 
+#include <openthread/dataset.h>
```
bukepo, line 33: We tried to avoid propagate dependency on OpenThread to the consuming projects.

src/common/types.cpp
zhanglongxia: The length of the `otIp6NetworkPrefix ` is a fixed length 64. So this class only supports 64 bits prefix.

src/common/types.cpp
```diff
@@ -126,13 +150,13 @@ std::string Ip6Prefix::ToString() const
...
 }
 
-std::string MacAddress::ToString(void) const
+std::string Ip6NetworkPrefix::ToString() const
 {
-    char strbuf[sizeof(m8) * 3];
+    char strbuf[INET6_ADDRSTRLEN];
 
-    snprintf(strbuf, sizeof(strbuf), "%02x:%02x:%02x:%02x:%02x:%02x", m8[0], m8[1], m8[2], m8[3], m8[4], m8[5]);
+    snprintf(strbuf, sizeof(strbuf), "%x:%x:%x:%x::/64", htobe16(m16[0]), htobe16(m16[1]), htobe16(m16[2]),
```
simonlingoogle, line 157: Can we use `inet_pton`?

The problem with current code is that it does not use compression properly.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -88,6 +91,19 @@ static uint64_t ConvertOpenThreadUint64(const uint8_t *aValue)
...
 
+static otExtendedPanId Uint64ToExtendedPanId(uint64_t aExtPanId)
+{
+    otExtendedPanId extPanId;
+    uint64_t        mask = UINT8_MAX;
+
+    for (size_t i = 0; i < sizeof(uint64_t); i++)
+    {
+        extPanId.m8[i] = static_cast<uint8_t>((aExtPanId >> ((sizeof(uint64_t) - i - 1) * 8)) & mask);
+    }
```
simonlingoogle, line 102: Can we utilize htobe64?

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -296,12 +328,14 @@ void DBusThreadObject::DetachHandler(DBusRequest &aRequest)
 
 void DBusThreadObject::FactoryResetHandler(DBusRequest &aRequest)
 {
+    otbrLogInfo("FactoryReset");
```
simonlingoogle, line 331: Instead of add log for each command, can we log for all commands in a unified way? 
And seems we are already printing dbus log like `Handling method FactoryReset` ?

zhanglongxia: It is hard for us to use a unified way. Because the parameters of each command are different, we need to print out all the parameters to get detailed info when the program crashed.
This log only record the Dbus API operations, other modules print out the `FactoryReset` log may not be caused by the Dbus. And I didn't find other modules have print the log `Handling method FactoryReset`.

simonlingoogle (approved):
LGTM :+1:

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -88,6 +91,16 @@ static uint64_t ConvertOpenThreadUint64(const uint8_t *aValue)
     return val;
 }
 
+static otExtendedPanId Uint64ToExtendedPanId(uint64_t aExtPanId)
+{
+    otExtendedPanId extPanId;
+    uint64_t *      pExtPanId = reinterpret_cast<uint64_t *>(&extPanId.m8[0]);
```
jinran-google, line 97: It's possible that `extPanId` is not aligned properly.

Use `alignas` when defining `extPanId`, or use `memcpy` instead.

src/common/types.hpp
```diff
@@ -381,6 +452,69 @@ class MacAddress
...
+     */
+    ExtAddress(void) { memset(m8, 0, sizeof(m8)); }
+
+    /**
+     * Constructor with an extended MAC address.
+     *
+     * @param[in]  aExtAddress  The extended MAC address.
+     *
+     */
+    ExtAddress(const otExtAddress &aExtAddress) { memcpy(reinterpret_cast<void *>(this), &aExtAddress, sizeof(*this)); }
```
jinran-google, line 474:
```suggestion
    ExtAddress(const otExtAddress &aExtAddress) { memcpy(this, &aExtAddress, sizeof(*this)); }
```
This is redundant.

src/common/types.hpp
```diff
@@ -381,6 +452,69 @@ class MacAddress
...
+     */
+    ExtPanId(void) { memset(m8, 0, sizeof(m8)); }
+
+    /**
+     * Constructor with an extended Pan Id.
+     *
+     * @param[in] aExtPanId  The extended Pan Id.
+     *
+     */
+    ExtPanId(const otExtendedPanId &aExtPanId) { memcpy(reinterpret_cast<void *>(this), &aExtPanId, sizeof(*this)); }
```
jinran-google, line 506:
```suggestion
    ExtPanId(const otExtendedPanId &aExtPanId) { memcpy(this, &aExtPanId, sizeof(*this)); }
```
ditto

bukepo (approved):

zhanglongxia:
Do not need this PR anymore.

PR #751, opened 2021-03-22
[web] improve add/delete prefix functionality

LuDuda:
This PR introduces two changes:
- Set default prefix length to 64 if not given (similar to form and join operation)
- Call `netdata register` to remove/add prefix immediatlelly


jwhui (approved):
Thanks! 👍🏼

PR #760, opened 2021-03-29
[mdns] fix deallocation of null pointers

jinran-google:


wgtdkp (approved):
LGTM 👍 
It is okay to deallocate the ref but mDNSResponder will give a warning, thanks @jinran-google for catching this!

jwhui (approved):

PR #765, opened 2021-03-31
[agent] add vendor bus server support

zhanglongxia:
This commit adds Vendor Bus APIs so that users can implement their
own Bus Server for otbr-agent.

### DIFF
--- a/src/agent/main.cpp
+++ b/src/agent/main.cpp
@@ -59,6 +59,9 @@ using otbr::rest::RestWebServer;
 #include "dbus/server/dbus_agent.hpp"
 using otbr::DBus::DBusAgent;
 #endif
+#ifdef OTBR_ENABLE_VENDOR_BUS_SERVER
+#include "agent/vendor_bus.hpp"
+#endif
 using otbr::Ncp::ControllerOpenThread;
 
 #if OTBR_ENABLE_OPENWRT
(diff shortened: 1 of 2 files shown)


jinran-google (approved):
LGTM overall, with a few questions:
- I believe OTBR can't be compiled on its own if `OTBR_ENABLE_VENDOR_BUS_SERVER` is turned on. Consider adding a weak implementation that does nothing?
- Just curious, why is this called "vendor bus"?

src/agent/main.cpp
```diff
@@ -124,6 +127,10 @@ static int Mainloop(otbr::AgentInstance &aInstance, const char *aInterfaceName)
     RestWebServer *restServer = RestWebServer::GetRestWebServer(&ncpOpenThread);
     restServer->Init();
 #endif
+#ifdef OTBR_ENABLE_VENDOR_BUS_SERVER
+    otbrVendorBusInit(ncpOpenThread);
```
wgtdkp, line 131: Should we follow the style of `dbusAgent` and `restServer` which use C++ methods of an object but not `C` style functions? e.g.:
```c++
#if OTBR_ENABLE_VENDOR_BUS_SERVER
VendorBusServer *vendorBus = VendorBusServer::GetVendorBusServer(&ncpOpenThread);
vendorBus->Init();
#endif
```

And this `VendorBusServer` can inherit `otbr::MainloopProcessor`. Thoughts?

zhanglongxia:
Do not need this PR anymore.

Issue #755, opened 2021-03-24
Add Detach command in DBus API


erjiaqing:
**Is your feature request related to a problem? Please describe.** Currently, we have a Reset() command, however, it will restart the whole process and cannot wait for its status. We should have a API for disable thread without reset the whole program.

**Describe the solution you'd like** Add Detach() command.

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.

**Additional context** Add any other context or screenshots about the feature request here.

PR #767, opened 2021-04-01
[dbus] add "Detach" API

jinran-google:
Fixes <https://github.com/openthread/ot-br-posix/issues/755>.

### DIFF
--- a/src/agent/thread_helper.cpp
+++ b/src/agent/thread_helper.cpp
@@ -294,7 +294,13 @@ exit:
 
 otError ThreadHelper::Detach(void)
 {
-    return otThreadSetEnabled(mInstance, false);
+    otError error = OT_ERROR_NONE;
+
+    SuccessOrExit(error = otThreadSetEnabled(mInstance, false));
+    SuccessOrExit(error = otIp6SetEnabled(mInstance, false));
+
+exit:
+    return error;
 }
 
 otError ThreadHelper::Reset(void)
--- a/src/agent/thread_helper.cpp
+++ b/src/agent/thread_helper.cpp
@@ -292,6 +292,11 @@ exit:
     }
 }
 
+otError ThreadHelper::Detach(void)
+{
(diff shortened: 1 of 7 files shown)


src/agent/thread_helper.cpp
```diff
@@ -292,6 +292,11 @@ void ThreadHelper::Attach(ResultHandler aHandler)
     }
 }
 
+otError ThreadHelper::Detach(void)
+{
+    return otThreadSetEnabled(mInstance, false);
```
zhanglongxia, line 297: The "Attach" function calls the function `otIp6SetEnabled` to start the Ipv6 stack, should we  also call the `otIp6SetEnabled` to disable the ipv6 stack in the `Detach` function?

src/dbus/common/constants.hpp
```diff
@@ -46,6 +46,7 @@
 
 #define OTBR_DBUS_SCAN_METHOD "Scan"
 #define OTBR_DBUS_ATTACH_METHOD "Attach"
+#define OTBR_DBUS_DETACH_METHOD "Detach"
```
bukepo, line 49: update https://github.com/openthread/ot-br-posix/blob/main/src/dbus/server/introspect.xml?

jinran-google: It's already updated <https://github.com/openthread/ot-br-posix/pull/767/files#diff-64258c71b54a165182be9a632cb0e131147fa82efdce34730e5a5c1b40c58c3bR49-R52>

src/agent/thread_helper.cpp
jinran-google: Updated. I think both are OK, but disabling the stack may save some resources.

zhanglongxia (approved):
LGTM

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #797, opened 2021-04-15
[tests] wait longer for services to be ready

### LINKED ISSUE #786: [github-actions] script check fails frequently
[logs_6926.zip](https://github.com/openthread/ot-br-posix/files/6297333/logs_6926.zip)

jwhui:
Addresses some of the failures in `check-scripts`.

Fixes #786.

### DIFF
--- a/tests/scripts/check-scripts
+++ b/tests/scripts/check-scripts
@@ -72,7 +72,7 @@ main()
     SERVICES_PID=$!
     sudo service tayga start
     echo 'Waiting for services to be ready...'
-    sleep 10
+    sleep 30
     check_otbr_agent
     netstat -an | grep 80
     pidof tayga


tests/scripts/check-scripts
```diff
@@ -72,7 +72,7 @@ main()
     SERVICES_PID=$!
     sudo service tayga start
     echo 'Waiting for services to be ready...'
-    sleep 10
+    sleep 30
```
wgtdkp, line 75: Thanks @jwhui ! Ididn't really get a chance to reproduce the failure in my docker environment perhaps because I was testing on the workstation which should be much more powerful than the GitHub Action virtual machine. But it is still surprising that it takes 30 seconds to start the tayga service :)

I didn't rerun the checks for another 2 times and it passes consistently so I think this should be the fix, Thanks!

wgtdkp (approved):
LGTM 👍

tests/scripts/check-scripts
jwhui: The problem is actually waiting for `./script/console` to launch all the services. `check_otbr_agent` below fails when not all the OTBR services have been started yet.

PR #799, opened 2021-04-15
[mdns] fix pointer to object on stack in TxtList

Damian-Nordic:
`TxtList` is a list of DNS TXT records to publish, represented as memory ranges for keys and values of the records.
While memory ranges for the values point directly to a `otSrpServerService` object's member whose lifetime is longer
than lifetime of the `TxtList`, memory ranges for the keys pointed to a member of `otDnsTxtEntryIterator` structure,
allocated on stack during the translation.

It resulted in publishing garbage TXT records.

Fix the issue by using `std::string` to store the keys instead of raw pointers/lengths.

Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -63,10 +63,10 @@ otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, uint8_t *aTxtData, u
 
     for (const auto &txtEntry : aTxtList)
     {
-        const char *   name        = txtEntry.mName;
-        const size_t   nameLength  = txtEntry.mNameLength;
-        const uint8_t *value       = txtEntry.mValue;
-        const size_t   valueLength = txtEntry.mValueLength;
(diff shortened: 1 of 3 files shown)


wgtdkp (commented):
Thanks @Damian-Nordic for catching this issue!

Do you mind to change the `mValue` to use std container (e.g. `std::vector<uint8_t>`) as well? Even if the `mValue` is pointing to a buffer in `otSrpServerService` for now, it may change in the future. To avoid such issues and to align with the `mName` field, I think worth the change :)

Damian-Nordic:
@wgtdkp Right, it's better to be on the safe side. :) Fixed as you suggeted.

wgtdkp (approved):
LGTM :+1:  Thanks @Damian-Nordic !

jwhui (approved):
Thanks! 👍🏼

PR #803, opened 2021-04-20
[trel] enable TREL by default

simonlingoogle:
This commit enables TREL by default.

**TREL is enabled by default using the Backbone interface**

- [x] Make sure OTBR setup works with TREL enabled

### DIFF
--- a/.github/workflows/border_router.yml
+++ b/.github/workflows/border_router.yml
@@ -53,11 +53,11 @@ jobs:
       matrix:
         include:
           - name: "Border Router"
-            otbr_options: "-DOT_MLR=ON -DOTBR_COVERAGE=ON -DOT_SRP_SERVER=ON -DOT_ECDSA=ON -DOT_SERVICE=ON -DOTBR_DUA_ROUTING=OFF"
+            otbr_options: "-DOT_TREL=OFF -DOT_MLR=ON -DOTBR_COVERAGE=ON -DOT_SRP_SERVER=ON -DOT_ECDSA=ON -DOT_SERVICE=ON -DOTBR_DUA_ROUTING=OFF"
             border_routing: 1
             cert_scripts: ./tests/scripts/thread-cert/border_router/*.py
           - name: "Backbone Router"
-            otbr_options: "-DOT_DUA=ON -DOT_MLR=ON -DOTBR_COVERAGE=ON -DOT_SRP_SERVER=ON -DOT_ECDSA=ON -DOT_SERVICE=ON -DOTBR_DUA_ROUTING=ON"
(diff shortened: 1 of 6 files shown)


bukepo (approved):
I prefer a more extensible syntax:

```bash
ot-cli trel://eth0
```

simonlingoogle:
> I prefer a more extensible syntax:
> 
> ```shell
> ot-cli trel://eth0
> ```

URL is more flexible if we have arguments for TREL. 
But there isn't any for now. 

**I'd prefer to add features only when they are useful.**

abtink (approved):
LGTM. Thanks. 👍

simonlingoogle:
As suggested by @wgtdkp , I removed `-T <trel-ifname>` and use existing argument `-B <backbone-ifname>` for TREL interface because we don't have a use case to use different infrastructure links for `Border/Backbone Router` and `TREL`.

Thoughts?

wgtdkp (approved):
LGTM :+1:

abtink:
> As suggested by @wgtdkp , I removed -T <trel-ifname> and use existing argument -B <backbone-ifname> for TREL interface because we don't have a use case to use different infrastructure links for Border/Backbone Router and TREL.
> 
> Thoughts?

Thanks @simonlingoogle, I feel keeping the two (backbone and trel) separate as you have originally in this PR may be a good idea.

- While I agree we may not use them together, fundamentally (from what each of these do) they are two unrelated functionality/features in OT and Thread and can potentially be used together.
- Also having different argument names makes the intention of what is being used more explicit and clear (e.g., `-T <trel-ifname>` makes it clear it is setting trel ifname).
- The one advantage I see with combining the args is to reuse code but not sure if code size is a big factor for `ot-br-posix`.

simonlingoogle:
> * While I agree we may not use them together, fundamentally (from what each of these do) they are two unrelated functionality/features in OT and Thread and can potentially be used together.
> * Also having different argument names makes the intention of what is being used more explicit and clear (e.g., `-T <trel-ifname>` makes it clear it is setting trel ifname).
> * The one advantage I see with combining the args is to reuse code but not sure if code size is a big factor for `ot-br-posix`.

Agree with you on the benefits. And code size is not an issue for `otbr`. 

Will restore it back to use `-T <trel-ifname>`. Thanks.

gabekassel:
Is this case sensitive? Should this be `-t` to match [Openthread](https://github.com/openthread/openthread/pull/4440/files#diff-0e68f9e173660fad0412aa18c799123ae39a8b618bae1a05ec54315912f6d3a8R145)?

simonlingoogle:
> Is this case sensitive? Should this be `-t` to match [Openthread](https://github.com/openthread/openthread/pull/4440/files#diff-0e68f9e173660fad0412aa18c799123ae39a8b618bae1a05ec54315912f6d3a8R145)? 

Good question. Actually we are going to change to use url like trek://eth0. Neither -t nor -T will be used.

jwhui (approved):

PR #821, opened 2021-04-29
[dnssd] remove reference counting logic

gjc13:
The reference counting logic has been moved to the core library.

Depends-On: https://github.com/openthread/openthread/pull/6533


wgtdkp (approved):
LGTM :+1:

gjc13:
@simonlingoogle Rebased to latest master, please take a look.

simonlingoogle (commented):
Reference Couting is still necessary because domain may be different?
For example, if OT core browses these two services:
- _srv._tcp.default.service.arpa.
- _srv._tcp.example.com.

Then, Discovery Proxy still needs to use `Reference Couting` to make sure that `SubscribeService("_srv._tcp")` is called only once.

PR #827, opened 2021-05-02
[mbedtls] enable MBEDTLS_SSL_KEEP_PEER_CERTIFICATE define

LuDuda:
The mbedTLS 2.17 [introduced](https://github.com/ARMmbed/mbedtls/blob/12f93f4fc25055163b42cd91c215ff2c594f036c/ChangeLog#L1042) a new option called `MBEDTLS_SSL_KEEP_PEER_CERTIFICATE` which allows saving RAM by not keeping the peer's certificate after completion of the handshake. This define is enabled by default in reference configuration but not in the `mbedtls-config.h`. This functionality is needed when building the OpenThread Posix application (more specifically `Dtls::GetPeerCertificateBase64` method).

This PR is necessary to introduce newer versions of mbedTLS in the OpenThread tree.


jwhui:
> The mbedTLS 2.17 [introduced](https://github.com/ARMmbed/mbedtls/blob/12f93f4fc25055163b42cd91c215ff2c594f036c/ChangeLog#L1042) a new option called `MBEDTLS_SSL_KEEP_PEER_CERTIFICATE` which allows saving RAM by not keeping the peer's certificate after completion of the handshake. This define is enabled by default in reference configuration but not in the `mbedtls-config.h`. This functionality is needed when building the OpenThread Posix application (more specifically `Dtls::GetPeerCertificateBase64` method).

@LuDuda , thanks for raising this issue. I'm not aware of any users of `Dtls::GetPeerCertificateBase64()` currently. Not for this PR, but should we consider removing this API to support saving RAM?

LuDuda:
I think this function is used by the [otCoapSecureGetPeerCertificateBase64](https://github.com/openthread/openthread/blob/b7751460e04a233589393fd4faccef5bc6e8e88b/include/openthread/coap_secure.h#L132) API. The documentation already states that `MBEDTLS_BASE64_C` has to be enabled to take advantage of this feature. So i do see two options, either, we will require both `MBEDTLS_BASE64_c` and  `MBEDTLS_SSL_KEEP_PEER_CERTIFICATE` defines to be enabled (disabled by default), or completely remove it - I'm fine either way, do you have any preference?

jwhui:
> I think this function is used by the [otCoapSecureGetPeerCertificateBase64](https://github.com/openthread/openthread/blob/b7751460e04a233589393fd4faccef5bc6e8e88b/include/openthread/coap_secure.h#L132) API. The documentation already states that `MBEDTLS_BASE64_C` has to be enabled to take advantage of this feature. So i do see two options, either, we will require both `MBEDTLS_BASE64_c` and `MBEDTLS_SSL_KEEP_PEER_CERTIFICATE` defines to be enabled (disabled by default), or completely remove it - I'm fine either way, do you have any preference?

I will merge this PR first. But if `otCoapSecureGetPeerCertificateBase64()` is not useful and we can save RAM doing so, then I'd prefer to remove the API. We can discuss in a separate GitHub Issue / PR.

jwhui (approved):

PR #829, opened 2021-05-04
[tests] build with clang/clang++ in CI tests

wgtdkp:
clang/clang++ applies stricter error/warning check. Use clang for CI builds so that there is less chance that we will miss issues like https://github.com/openthread/openthread/pull/6560.


wgtdkp:
Replaced by https://github.com/openthread/openthread/pull/6570.

PR #834, opened 2021-05-08
[openthread] enable DNS client and SRP client by default

simonlingoogle:
Reason to enable SRP and DNS clients:
- SRP and DNS servers are already enabled on OTBR by default. The clients can be used to test with servers or to verify if the servers are working as expected.
- The added code size of clients are trivial for OTBR.
- One can still disable the clients using OT options.


wgtdkp (approved):

librasungirl:
deprecate now ?@simonlingoogle

PR #840, opened 2021-05-11
[scripts] add send_udp.py for certification

simonlingoogle:


wgtdkp (commented):
Define a function and call the function at end of the script?

script/reference-device/send_udp.py
```diff
@@ -0,0 +1,79 @@
...
+log = 'payload %s' % payload
+syslog.syslog(syslog.LOG_ERR, log)
+print(log)
+
+byte_payload = binascii.unhexlify(payload)
+
+try:
+    sock = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
+
+    if sys.version_info[0] < 3:
```
librasungirl, line 65: Do we still want to support python 2.x?
Btw, `print()` above is the python3 format.

script/reference-device/send_udp.py
```diff
@@ -0,0 +1,79 @@
...
+        sock.setsockopt(socket.SOL_SOCKET, 25, interface)
+    else:
+        sock.setsockopt(socket.SOL_SOCKET, 25, str(interface).encode('utf-8'))
```
librasungirl, line 68: Magic `25` should be avoided, use `socket.SO_BINDTODEVICE` instead?

librasungirl (commented):
`+x` permission

script/reference-device/send_udp.py
simonlingoogle: Sure. We should only support Python3?

wgtdkp (approved):
LGTM :+1:

jwhui (approved):

PR #841, opened 2021-05-11
[dnssd] use queries in dns-sd server rather than local map

gjc13:
The newly introduces `otDnssdGetQuery` apis allows platform
implementations to use the queries stored in the dns-sd server. Remove
the local map to reduce complexity.

Depends-On: https://github.com/openthread/openthread/pull/6611

### DIFF
--- a/src/agent/discovery_proxy.cpp
+++ b/src/agent/discovery_proxy.cpp
@@ -89,41 +89,24 @@ void DiscoveryProxy::OnDiscoveryProxySubscribe(void *aContext, const char *aFull
 
 void DiscoveryProxy::OnDiscoveryProxySubscribe(const char *aFullName)
 {
-    std::string                    fullName(aFullName);
-    otbrError                      error = OTBR_ERROR_NONE;
-    MdnsSubscriptionList::iterator it;
-    DnsNameInfo                    nameInfo = SplitFullDnsName(fullName);
+    std::string fullName(aFullName);
+    otbrError   error    = OTBR_ERROR_NONE;
+    DnsNameInfo nameInfo = SplitFullDnsName(fullName);
 
     otbrLogInfo("subscribe: %s", fullName.c_str());
 
-    it = std::find_if(mSubscriptions.begin(), mSubscriptions.end(), [&](const MdnsSubscription &aSubscription) {
-        return aSubscription.Matches(nameInfo.mInstanceName, nameInfo.mServiceName, nameInfo.mHostName,
-                                     nameInfo.mDomain);
-    });
-
-    VerifyOrExit(it == mSubscriptions.end(), it->mSubscriptionCount++);
-
-    mSubscriptions.emplace_back(nameInfo.mInstanceName, nameInfo.mServiceName, nameInfo.mHostName, nameInfo.mDomain);
-
+    if (GetServiceSubscriptionCount(nameInfo) == 1)
     {
-        MdnsSubscription &subscription = mSubscriptions.back();
-
-        otbrLogDebug("subscriptions: %sx%d", subscription.ToString().c_str(), subscription.mSubscriptionCount);
-
-        if (GetServiceSubscriptionCount(nameInfo.mInstanceName, nameInfo.mServiceName, nameInfo.mHostName) == 1)
+        if (nameInfo.mHostName.empty())
         {
-            if (subscription.mHostName.empty())
-            {
-                mMdnsPublisher.SubscribeService(nameInfo.mServiceName, nameInfo.mInstanceName);
-            }
-            else
-            {
-                mMdnsPublisher.SubscribeHost(nameInfo.mHostName);
-            }
+            mMdnsPublisher.SubscribeService(nameInfo.mServiceName, nameInfo.mInstanceName);
+        }
+        else
+        {
+            mMdnsPublisher.SubscribeHost(nameInfo.mHostName);
         }
     }
 
-exit:
     if (error != OTBR_ERROR_NONE)
     {
         otbrLogWarning("failed to subscribe %s: %s", fullName.c_str(), otbrErrorString(error));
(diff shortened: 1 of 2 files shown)


src/agent/discovery_proxy.cpp
```diff
@@ -214,25 +175,52 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string &
     instanceInfo.mTxtData   = aInstanceInfo.mTxtData.data();
     instanceInfo.mTtl       = CapTtl(aInstanceInfo.mTtl);
 
-    std::for_each(mSubscriptions.begin(), mSubscriptions.end(), [&](const MdnsSubscription &aSubscription) {
-        if (aSubscription.MatchesServiceInstance(aType, aInstanceInfo.mName))
+    while ((query = otDnssdGetNextQuery(mNcp.GetInstance(), query)) != nullptr)
+    {
+        std::string      instanceName, serviceName, hostName, domain;
```
wgtdkp, line 180:
```suggestion
        std::string      instanceName;
        std::string      serviceName;
        std::string      hostName;
        std::string      domain;
```

src/agent/discovery_proxy.cpp
```diff
@@ -251,14 +239,29 @@ void DiscoveryProxy::OnHostDiscovered(const std::string &
 
     hostInfo.mTtl = CapTtl(aHostInfo.mTtl);
 
-    std::for_each(mSubscriptions.begin(), mSubscriptions.end(), [&](const MdnsSubscription &aSubscription) {
-        if (aSubscription.MatchesHost(aHostName))
+    while ((query = otDnssdGetNextQuery(mNcp.GetInstance(), query)) != nullptr)
+    {
+        std::string      hostName, domain;
```
wgtdkp, line 244:
```suggestion
        std::string      hostName;
        std::string      domain;
```

wgtdkp (approved):
LGTM :+1:

src/agent/discovery_proxy.cpp
```diff
@@ -214,25 +175,52 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string &
...
+            break;
+        case OT_DNSSD_QUERY_TYPE_RESOLVE:
+            splitError = SplitFullServiceInstanceName(queryName, instanceName, serviceName, domain);
+            break;
+        case OT_DNSSD_QUERY_TYPE_RESOLVE_HOST:
+            splitError = OTBR_ERROR_PARSE;
+            break;
+        default:
+            splitError = OTBR_ERROR_NOT_IMPLEMENTED;
+            break;
```
simonlingoogle, line 198:
```suggestion
        default:
            splitError = OTBR_ERROR_NOT_FOUND;
            break;
```
seems we don't need to differentiate `OT_DNSSD_QUERY_TYPE_RESOLVE_HOST` and default?

src/agent/discovery_proxy.cpp
```diff
@@ -214,25 +175,52 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string &
...
+    {
+        std::string      instanceName, serviceName, hostName, domain;
+        char             queryName[OT_DNS_MAX_NAME_SIZE];
+        otDnssdQueryType type = otDnssdGetQueryTypeAndName(query, &queryName);
+        otbrError        splitError;
+
+        switch (type)
+        {
+        case OT_DNSSD_QUERY_TYPE_BROWSE:
+            splitError = SplitFullServiceName(queryName, serviceName, domain);
```
simonlingoogle, line 188:
```suggestion
            splitError = SplitFullServiceName(queryName, serviceName, domain);
            OT_ASSERT(splitError == OTBR_ERROR_NONE);
```
There should never be an error?

src/agent/discovery_proxy.cpp
```diff
@@ -214,25 +175,52 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string &
...
+            break;
+        case OT_DNSSD_QUERY_TYPE_RESOLVE:
+            splitError = SplitFullServiceInstanceName(queryName, instanceName, serviceName, domain);
```
simonlingoogle, line 191:
```suggestion
            splitError = SplitFullServiceInstanceName(queryName, instanceName, serviceName, domain);
            OT_ASSERT(splitError == OTBR_ERROR_NONE);
```

src/agent/discovery_proxy.cpp
```diff
@@ -251,14 +239,29 @@ void DiscoveryProxy::OnHostDiscovered(const std::string &
...
+        otbrError        splitError = OTBR_ERROR_PARSE;
+
+        if (type == OT_DNSSD_QUERY_TYPE_RESOLVE_HOST)
+        {
+            splitError = SplitFullHostName(queryName, hostName, domain);
+        }
+        if (splitError != OTBR_ERROR_NONE)
+        {
+            continue;
+        }
```
simonlingoogle, line 256:
```suggestion
        otbrError splitError;

        if (type != OT_DNSSD_QUERY_TYPE_RESOLVE_HOST)
        {
            continue;
        }
        
        splitError = SplitFullHostName(queryName, hostName, domain);
        OT_ASSERT(splitError == OT_ERROR_NONE);
```
`SplitFullHostName` should never fail.

simonlingoogle:
One minor concern I have for this PR is that it may be better if OTBR features can just work with OT callbacks, without extra calling to OT APIs (e.g. `otDnssdGetNextQuery` in this PR). 

The reason is that in long term we might want `otbr-agent` to work with both RCP and NCP mode. In RCP mode, the whole OpenThread runs in the same process space of otbr-agent, so it's fine to call any OT APIs. But in NCP mode, every communication between otbr-agent and OpenThread would have to be using UART/SPI communications.
So, avoid using OT APIs in otbr-agent can help reduce the future effort on support NCP. 

Thoughts? @bukepo

bukepo:
For supporting NCP mode, I'd like to leverage some RPC mechanism to provide the same OpenThread API for otbr.

simonlingoogle (commented):
> For supporting NCP mode, I'd like to leverage some RPC mechanism to provide the same OpenThread API for otbr.

Using RCP may allow certain APIs to continue work, but it does not address the problem that every API call will involve communication between NCP and Host (now in the form of RPC), which leads to inefficiency if called frequently. 

And for APIs like `otDnssdGetQuery` which returns a pointer to memory object hold by OpenThread, I think it's infeasible for RPC to provides such APIs.

Thoughts?

wgtdkp:
Do we really need to support NCP-based Border Router?

bukepo:
> > For supporting NCP mode, I'd like to leverage some RPC mechanism to provide the same OpenThread API for otbr.
> 
> Using RPC may allow certain APIs to continue work, but it does not address the problem that every API call will involve communication between NCP and Host (now in the form of RPC), which leads to inefficiency if called frequently.
> 
> And for APIs like `otDnssdGetQuery` which returns a pointer to memory object hold by OpenThread, I think it's infeasible for RPC to provides such APIs.
> 
> Thoughts?

Agree. Maybe we can optimize frequently called APIs then if they exist and cause issues. 

> Do we really need to support NCP-based Border Router?

No plans in the near future yet.

gjc13:
@bukepo The build fails because I'm depending on the api introduced in a OpenThread PR. Can the `Depend-On` feature handle such cases?

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #867, opened 2021-05-25
[dns] fix potential Discovery Proxy crash

simonlingoogle:
This commit fixes potential Discovery Proxy crash when the mDNS implementation replied with an empty host name.

### DIFF
--- a/src/agent/discovery_proxy.cpp
+++ b/src/agent/discovery_proxy.cpp
@@ -222,12 +222,18 @@ void DiscoveryProxy::OnHostDiscovered(const std::string &                       
                                       const Mdns::Publisher::DiscoveredHostInfo &aHostInfo)
 {
     otDnssdHostInfo     hostInfo;
-    const otDnssdQuery *query = nullptr;
+    const otDnssdQuery *query            = nullptr;
+    std::string         resolvedHostName = aHostInfo.mHostName;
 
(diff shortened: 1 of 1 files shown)


src/agent/discovery_proxy.cpp
```diff
@@ -222,12 +222,18 @@ void DiscoveryProxy::OnHostDiscovered(const std::string &
...
     otDnssdHostInfo     hostInfo;
-    const otDnssdQuery *query = nullptr;
+    const otDnssdQuery *query            = nullptr;
+    std::string         resolvedHostName = aHostInfo.mHostName;
 
     otbrLogInfo("host discovered: %s hostname %s addresses %zu", aHostName.c_str(), aHostInfo.mHostName.c_str(),
                 aHostInfo.mAddresses.size());
 
-    CheckHostnameSanity(aHostInfo.mHostName);
+    if (resolvedHostName.empty())
```
wgtdkp, line 231: In what cases will the `resolvedHostName` be empty?

gjc13 (approved):

src/agent/discovery_proxy.cpp
simonlingoogle: This is possible when mDNS found no IPv6 address for this host name.
In such case, I think it's meaningful to notify the Discovery Proxy with empty hostname and zero addresses.

Thoughts?

PR #875, opened 2021-06-03
[web] add pskd support for OTBR-WEB when joining an existing network

xwang146:


src/web/web-service/wpan_service.cpp
```diff
@@ -66,20 +70,37 @@ std::string WpanService::HandleJoinNetworkRequest(const std::string &aJoinReques
...
 
     VerifyOrExit(reader.parse(aJoinRequest.c_str(), root) == true, ret = kWpanStatus_ParseRequestFailed);
-    index        = root["index"].asUInt();
-    masterKey    = root["masterKey"].asString();
-    prefix       = root["prefix"].asString();
-    defaultRoute = root["defaultRoute"].asBool();
+    index          = root["index"].asUInt();
+    credentialType = root["credentialType"].asString();
+    masterKey      = root["masterKey"].asString();
+    pskd           = root["pskd"].asString();
```
Irving-cl, line 76:
```
if (credentialType == CREDENTIAL_TYPE_MASTER_KEY)
{
    masterKey = root["masterKey"].asString();
}
else if (credentialType == CREDENTIAL_TYPE_PSKD)
{
    pskd = root["pskd"].asString();
}
    
```
Not a MUST now. Probably we can enhance the UI later to update the table according user's choice of `credentialType`.

Irving-cl (approved):
LGTM :+1:

src/web/web-service/wpan_service.cpp
xwang146: Thanks. Will consider this way in later versions when I add more credential types.

bukepo (approved):

src/web/web-service/wpan_service.cpp
```diff
@@ -66,20 +70,37 @@ std::string WpanService::HandleJoinNetworkRequest(const std::string &aJoinReques
...
+    {
+        VerifyOrExit(
+            (ret = commitActiveDataset(client, masterKey, mNetworks[index].mNetworkName, mNetworks[index].mChannel,
+                                       mNetworks[index].mExtPanId, mNetworks[index].mPanId)) == kWpanStatus_Ok);
+        VerifyOrExit(client.Execute("ifconfig up") != nullptr, ret = kWpanStatus_JoinFailed);
+    }
+    else if (credentialType == CREDENTIAL_TYPE_PSKD)
+    {
+        VerifyOrExit(client.Execute("ifconfig up") != nullptr, ret = kWpanStatus_JoinFailed);
+        VerifyOrExit(client.Execute("joiner start %s", pskd.c_str()) != nullptr, ret = kWpanStatus_JoinFailed);
```
jwhui, line 97: Do we need to wait until the joiner operation completes? I don't know if calling `thread start` before the joiner operation completes will produce the intended result.

xwang146: That works during my experiment. 

Currently the "Execute()" method checks whether OpenThread returns "Done" for a command then returns. Are you suggesting checking the "Join success" message?

jwhui: I think it would be safer to wait for the "Join success" message.

@bukepo , thoughts?

bukepo: Agree to check the joiner success message. And better to tell user what error is and how to fix. For example, the not found error should be caused by the joiner is not added by commissioner, so the user should use a commissioner to add it. For security error, it probably indicates the pskd does not match the one given to the commissioner, and we should ask the user double check the pskd. Ideally we generate the standard qrcode and show it in the dialogue when user types the pskd so that users can scan it with an external commissioner app.

bukepo: This is a qrcode generator supports embedded ``: http://goqr.me/

xwang146: Sounds good. Thanks for the suggestions. I've added the error message for the 2 different cases and embedded the QR code generator.

brianljt (approved):
LGTM, Thanks!

bukepo:
@xwang146 there are merge conflicts.

bukepo (approved):

src/web/web-service/ot_client.hpp
```diff
@@ -101,6 +101,17 @@ class OpenThreadClient
...
+    /**
+     * This method reads from OpenThread CLI.
+     *
+     * @param[in]   response    The expected read response.
+     * @param[in]   timeout     Timeout for the read, in ms.
+     *
+     * @returns A pointer to the output if the expected response is found, otherwise nullptr.
+     *
+     */
+    char *Read(const char *response, int timeout);
```
jwhui, line 113:
```suggestion
    char *Read(const char *aResponse, int aTimeout);
```

src/web/web-service/ot_client.hpp
```diff
@@ -101,6 +101,17 @@ class OpenThreadClient
      */
     char *Execute(const char *aFormat, ...);
 
+    /**
+     * This method reads from OpenThread CLI.
+     *
+     * @param[in]   response    The expected read response.
+     * @param[in]   timeout     Timeout for the read, in ms.
```
jwhui, line 108:
```suggestion
     * @param[in]   aResponse    The expected read response.
     * @param[in]   aTimeout     Timeout for the read, in ms.
```

xwang146: Thanks. Updated.

jwhui (approved):

PR #881, opened 2021-06-08
[web-service] fix form/join operations

jwhui:
- fix join operation to use Active Timestamp = 0
- include PSKc as part of dataset (not separate cmd)


bukepo:
Should we include the current unix time as the active timestamp?

jwhui:
> Should we include the current unix time as the active timestamp when forming network?

@bukepo, the OT API currently only allows setting the seconds component and not the ticks or U bit. Are you suggesting we just set the Active Timestamp field in the dataset structure using the return value of `time()`? I think anything more should probably done in a separate PR.

bukepo (approved):
LGTM:+1:

PR #887, opened 2021-06-10
[advertising-proxy] refactor AdvertisingHandler to reuse its functionality

superwhd:
- Add a handler when the mDNS instance connects to the provider.
- Refactor `AdvertisingHandler` to make it reusable by other parts.
- This is useful when the mDNS provider goes down and reboots. In such situation we need to trigger that handler and re-advertise all the registered hosts and services.


superwhd:
closing this PR since it's messy

PR #896, opened 2021-06-16
[cmake] include CoAPS PSK support for OTBR

suveshpratapa:
Turn on mbedtls PSK support for CoAP testing on reference OTBR (similar to ECDSA)


jwhui (approved):
Thanks! 👍🏼

PR #906, opened 2021-06-22
[dnssd] set TTL as 10 for entries discovered by avahi

superwhd:
Currently I don't see an easy way to get TTL information via avahi API. This PR is a workaround to set TTL as 10 all the time to avoid zero TTLs.

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -1074,7 +1074,7 @@ void PublisherAvahi::ServiceSubscription::HandleResolveResult(AvahiServiceResolv
     // TODO priority
     // TODO weight
     // TODO use a more proper TTL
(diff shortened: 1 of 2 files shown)


src/mdns/mdns_avahi.cpp
```diff
@@ -1073,7 +1073,8 @@ void PublisherAvahi::ServiceSubscription::HandleResolveResult(AvahiServiceResolv
 
     // TODO priority
     // TODO weight
-    // TODO ttl
+    // TODO use a more proper TTL
+    mInstanceInfo.mTtl = 10;
```
simonlingoogle, line 1077: Define a constant?

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

Issue #735, opened 2021-03-16
[border-agent] use vendor and product name as MeshCop service instance name


wgtdkp:
Current Border Agent implementation is using the Thread network name as the `_meshcop._udp` service instance name.
The Thread network name is not a proper service instance name because we need to publish a new service instance when
the Thread network name changes.

Actually, the Thread network name is advertised as a mDNS TXT entry (`"nn"`) of the `_meshcop._udp` service and only the TXT entry needs to be updated when the network name changes.

The Thread specification recommends a combination of vendor and product name for the service instance name. We may consider follow it.

Issue #826, opened 2021-04-30
MeshCoP Service does not have conflict resolution with Avahi daemon


gabekassel:
**Describe the bug**
When multiple BRs exist on the same infrastructure network with the same network name, OTBR does not conduct conflict resolution to allow all services to coexist. 

**To Reproduce**
Place two BRs on same wifi/ethernet network with same network name

**Expected behavior**
Append integer or some unique identifier to resolve conflict. Believe Avahi has something built-in for this.

PR #946, opened 2021-07-28
[border-agent] name meshcop service after "VendorName_ProductName" 

superwhd:
This PR aims to fix https://github.com/openthread/ot-br-posix/issues/735. 

The recommended format of service instance is `VendorName ProductName._meshcop._udp_.local.`. However, it's quite tricky to handle the space in service name properly. Also, the space in the name will be converted to `\032` after the registration, which is not a very elegant form. Therefore, I chose `VendorName_ProductName._meshcop._udp_.local.` as the default service name.

This PR should be merged together with https://github.com/openthread/openthread/pull/6894.


superwhd:
The `Depends-On` in the description doesn't seem to let CI patch the depended PR before running workflows. Is there any way to let CI run tests with both PRs patched?

jwhui:
> The `Depends-On` in the description doesn't seem to let CI patch the depended PR before running workflows. Is there any way to let CI run tests with both PRs patched?

@superwhd , the `Depends-On` requires the text to be of the form `openthread/openthread#6865`, rather than the `https://github.com/openthread/openthread/pull/6865` link. I have updated the text above and kicked GitHub Actions tests.

superwhd:
@jwhui Thanks! 

However, it seems the test case stayed unchanged after specifying `Depends-On`. I was expecting [this block](https://github.com/openthread/openthread/pull/6865/files#diff-e67178cb8388ffced2417ed0ad9519b2b4099e9dd20f26a1c8fc814466feaa03R229) would be patched so the test case could be passed.

jwhui:
> However, it seems the test case stayed unchanged after specifying `Depends-On`. I was expecting [this block](https://github.com/openthread/openthread/pull/6865/files#diff-e67178cb8388ffced2417ed0ad9519b2b4099e9dd20f26a1c8fc814466feaa03R229) would be patched so the test case could be passed.

@bukepo , can you help take a look?

bukepo:
we use submodule in this project, and depends-on doesn't work on this project yet. We can get OpenThread merged first and get back to this pr later. Thoughts?

superwhd:
> we use submodule in this project, and depends-on doesn't work on this project yet. We can get OpenThread merged first and get back to this pr later. Thoughts?

SGTM. Will do.

src/agent/border_agent.hpp
```diff
@@ -49,6 +49,18 @@
...
+#ifndef OTBR_VENDOR_NAME
+#define OTBR_VENDOR_NAME "OpenThread"
+#endif
+
+#ifndef OTBR_PRODUCT_NAME
+#define OTBR_PRODUCT_NAME "BorderRouter"
+#endif
+
+#ifndef OTBR_MESHCOP_SERVICE_NAME
+#define OTBR_MESHCOP_SERVICE_NAME OTBR_VENDOR_NAME "_" OTBR_PRODUCT_NAME
```
wgtdkp, line 61: There will be name conflicts if two OTBRs are placed on the same network link. mDNS Responder can auto rename the conflicted names, but avahi seems not. See https://github.com/openthread/ot-br-posix/issues/826.

superwhd: I think this issue can be regarded as independent of the current PR.

For the issue #826, I can think of two ways to fix it:
- Generating unique service names (e.g. using random digits for the suffix) before registering the service.
- Leveraging avahi's utils to handle name conflicts. There's an example in their doc: https://www.avahi.org/doxygen/html/client-publish-service_8c-example.html.

wgtdkp (approved):
LGTM :+1:

src/agent/border_agent.cpp
```diff
@@ -63,7 +63,8 @@
 
 namespace otbr {
 
-static const char kBorderAgentServiceType[] = "_meshcop._udp"; ///< Border agent service type of mDNS
+static const char kBorderAgentServiceType[] = "_meshcop._udp";           ///< Border agent service type of mDNS
+static const char kBorderAgentServiceName[] = OTBR_MESHCOP_SERVICE_NAME; ///< Border agent service name of mDNS
```
simonlingoogle, line 67:
```suggestion
static const char kBorderAgentServiceInstanceName[] = OTBR_MESHCOP_SERVICE_INSTANCE_NAME; ///< Border agent service name of mDNS
```
Prefer to add `Instance` for clarity. 
I think `service name` is quite misleading. RFC 6763 refers to service types (e.g. `_meshcop._udp`) as the service names, and always use `instance name` for service instance names.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #949, opened 2021-07-30
[mdns] resolve service name conflicts when using avahi

superwhd:
Previously we didn't implement the strategy for resolving service name conflicts while using avahi mdns. Therefore, when there is a service name conflict, the registration will fail silently. Especially, this issue affects the situation when their are two border routers on the same infra link, where there will only have one meshcop service successfully registered.

Have tested locally, will add the test case once https://github.com/openthread/openthread/pull/6865 is checked in.

This PR should fix issue: https://github.com/openthread/ot-br-posix/issues/826

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -680,6 +680,7 @@ otbrError PublisherAvahi::PublishService(const char *       aHostName,
     // aligned with AvahiStringList
     AvahiStringList  buffer[(kMaxSizeOfTxtRecord - 1) / sizeof(AvahiStringList) + 1];
     AvahiStringList *head = nullptr;
+
     SuccessOrExit(error = TxtListToAvahiStringList(aTxtList, buffer, sizeof(buffer), head));
 
     VerifyOrExit(mState == State::kReady, errno = EAGAIN, error = OTBR_ERROR_ERRNO);
@@ -879,6 +880,7 @@ otbrError PublisherAvahi::TxtListToAvahiStringList(const TxtList &   aTxtList,
     size_t           used  = 0;
     AvahiStringList *last  = nullptr;
     AvahiStringList *curr  = aBuffer;
+
     aHead                  = nullptr;
     for (const auto &txtEntry : aTxtList)
     {
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -367,9 +367,20 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
         break;
 
     case AVAHI_ENTRY_GROUP_COLLISION:
+    {
         otbrLogErr("Name collision!");
-        CallHostOrServiceCallback(aGroup, OTBR_ERROR_DUPLICATED);
+        Services::iterator serviceIt = FindService(aGroup);
+        if (serviceIt != mServices.end())
+        {
+            char *alternativeName = avahi_alternative_service_name(serviceIt->mName.c_str());
+            otbrLogInfo("Re-publishing service %s using an alternative name: %s", serviceIt->mName.c_str(),
+                        alternativeName);
+            serviceIt->mName = alternativeName;
+            avahi_free(alternativeName);
+            RetryPublishService(serviceIt);
+        }
         break;
+    }
 
     case AVAHI_ENTRY_GROUP_FAILURE:
         /* Some kind of failure happened while we were registering our services */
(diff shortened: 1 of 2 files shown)


simonlingoogle (commented):
What's the preferred strategy in Thread SRP server in the case of name conflict?

1. Automatically rename in mDNS and notify SRP clients that the registration was successful. 
2. Fail with name conflict and notify SRP clients

@wgtdkp @jwhui @superwhd @abtink

wgtdkp:
> What's the preferred strategy in Thread SRP server in the case of name conflict?
> 
> 1. Automatically rename in mDNS and notify SRP clients that the registration was successful.
> 2. Fail with name conflict and notify SRP clients
> 
> @wgtdkp @jwhui @superwhd @abtink

1. We want auto-rename for meshcop services as there are no dependency on unique service/host names;
2. We may want to report name conflicts to SRP clients so that the client get a chance to properly handle it. For example, the client may switch the SRP server;

So we should have a flag in the API which indicates whether we want auto-rename or not.

jwhui:
> What's the preferred strategy in Thread SRP server in the case of name conflict?
> 
> 1. Automatically rename in mDNS and notify SRP clients that the registration was successful.
> 2. Fail with name conflict and notify SRP clients

[draft-sctl-advertising-proxy-02 Section 2.1](https://datatracker.ietf.org/doc/html/draft-sctl-advertising-proxy-02) specifies the name conflict behavior:

```
   Specifically, a name conflict can occur:

   1.  During the SRP validation process, because another SRP client has
       already registered the same name.

   2.  Immediately while the Advertising Proxy is registering the name,
       if the Multicast DNS uniqueness probes detect a conflicting
       record.

   3.  After the name has been successfully registered, but before the
       response has been sent to the client.

   4.  After the initial response has been sent to the client.

   In the first three cases, the client can be notified of the conflict
   at the time of registration, and is expected to choose a new name.
   In the last case, SRP clients must be coded defensively to handle the
   case where an apparently successful record registration is later
   determined to be in conflict, just as existing Multicast DNS clients
   have to be coded defensively to handle late conflicts gracefully.
   With a sleepy SRP client there may be no way to notify it of the
   conflict until it next re-registers.  In the case of late conflicts,
   the service registry with Advertising Proxy capability is responsible
   for selecting a temporary new name to be used until the client
   renews.  When the client next renews, the service registry informs
   the client of the new name the service registry selected on its
   behalf.  The client can choose to accept that new name, or select a
   new name of its own choosing.
```

wgtdkp (approved):
LGTM :+1:

superwhd:
I'll send a future PR to implement different renaming strategies for the meshcop service and SRP services.

src/mdns/mdns_avahi.cpp
```diff
@@ -367,9 +367,20 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
         break;
 
     case AVAHI_ENTRY_GROUP_COLLISION:
+    {
         otbrLogErr("Name collision!");
-        CallHostOrServiceCallback(aGroup, OTBR_ERROR_DUPLICATED);
+        Services::iterator serviceIt = FindService(aGroup);
```
simonlingoogle, line 372: Can it be a Host name conflict?

src/mdns/mdns_avahi.cpp
```diff
@@ -367,9 +367,20 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
...
+        if (serviceIt != mServices.end())
+        {
+            char *alternativeName = avahi_alternative_service_name(serviceIt->mName.c_str());
```
simonlingoogle, line 375: nit: handle `alternativeName == nullptr`?

src/mdns/mdns_avahi.cpp
```diff
@@ -413,6 +424,51 @@ void PublisherAvahi::CallHostOrServiceCallback(AvahiEntryGroup *aGroup, otbrErro
     }
 }
 
+otbrError PublisherAvahi::RetryPublishService(const Services::iterator &aServiceIt)
```
simonlingoogle, line 427: This method contains duplicated logic of PublishService. 

Can we reuse the same logic?

src/mdns/mdns_avahi.cpp
```diff
@@ -829,6 +869,44 @@ otbrError PublisherAvahi::UnpublishHost(const char *aName)
     return error;
 }
 
+otbrError PublisherAvahi::TxtListToAvahiStringList(const TxtList &   aTxtList,
```
simonlingoogle, line 872: Is it related change?

src/mdns/mdns_avahi.cpp
superwhd: I tried doing so initially but I found it hard to achieve it, because there are some logic about updating the service in `PublishService`will result in a wrong behavior when retrying publishing the service.

src/mdns/mdns_avahi.cpp
superwhd: Yes. I extracted this function to let it be reused.

src/mdns/mdns_avahi.cpp
superwhd: I think host name conflict should trigger a `AVAHI_SERVER_COLLISION`.

src/mdns/mdns_avahi.cpp
simonlingoogle: SGTM if it's hard to do.

simonlingoogle (approved):
LGTM 👍

src/mdns/mdns_avahi.cpp
```diff
@@ -829,6 +870,44 @@ otbrError PublisherAvahi::UnpublishHost(const char *aName)
...
+                                                   AvahiStringList * aBuffer,
+                                                   size_t            aBufferSize,
+                                                   AvahiStringList *&aHead)
+{
+    otbrError        error = OTBR_ERROR_NONE;
+    size_t           used  = 0;
+    AvahiStringList *last  = nullptr;
+    AvahiStringList *curr  = aBuffer;
+    aHead                  = nullptr;
```
jwhui, line 882: nit: empty line after declaration block
```suggestion
    AvahiStringList *curr  = aBuffer;

    aHead                  = nullptr;
```

src/mdns/mdns_avahi.cpp
```diff
@@ -616,9 +679,8 @@ otbrError PublisherAvahi::PublishService(const char *       aHostName,
     std::string        fullHostName;
     // aligned with AvahiStringList
     AvahiStringList  buffer[(kMaxSizeOfTxtRecord - 1) / sizeof(AvahiStringList) + 1];
-    AvahiStringList *last = nullptr;
-    AvahiStringList *curr = buffer;
-    size_t           used = 0;
+    AvahiStringList *head = nullptr;
+    SuccessOrExit(error = TxtListToAvahiStringList(aTxtList, buffer, sizeof(buffer), head));
```
jwhui, line 683: nit: empty line after declaration block
```suggestion
    AvahiStringList *head = nullptr;

    SuccessOrExit(error = TxtListToAvahiStringList(aTxtList, buffer, sizeof(buffer), head));
```

jwhui (approved):

PR #951, opened 2021-07-30
fix the problem that the web page may get wrong results prefixed with…

UnionType:
fix the problem that the web page may get wrong results prefixed with "> "，the problem occurs when multiple commands are sent to otbr-agent continuously and quickly .


jwhui:
Superseded by #1671

PR #976, opened 2021-08-18
[mdns] fix unused variable of aServiceRef

superwhd:
Without this change clang may complain `aServiceRef` is unused in some situations.

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -435,12 +435,13 @@ exit:
 
 void PublisherMDnsSd::DiscardService(const char *aName, const char *aType, DNSServiceRef aServiceRef)
 {
+    OT_UNUSED_VARIABLE(aServiceRef);
+
     ServiceIterator service = FindPublishedService(aName, aType);
 
     if (service != mServices.end())
     {
         assert(aServiceRef == nullptr || aServiceRef == service->mService);
-        OTBR_UNUSED_VARIABLE(aServiceRef);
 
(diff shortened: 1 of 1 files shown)


src/mdns/mdns_mdnssd.cpp
```diff
@@ -435,12 +435,13 @@ void PublisherMDnsSd::HandleServiceRegisterResult(DNSServiceRef         aService
...
 void PublisherMDnsSd::DiscardService(const char *aName, const char *aType, DNSServiceRef aServiceRef)
 {
+    OT_UNUSED_VARIABLE(aServiceRef);
+
     ServiceIterator service = FindPublishedService(aName, aType);
 
     if (service != mServices.end())
     {
         assert(aServiceRef == nullptr || aServiceRef == service->mService);
-        OTBR_UNUSED_VARIABLE(aServiceRef);
```
jinran-google, line 443: Is it better to update `OTBR_UNUSED_VARIABLE` to make it work with `clang`?

superwhd: I don't see a way to achieve that. Do you have any idea?

librasungirl (approved):

src/mdns/mdns_mdnssd.cpp
jinran-google: OK, I see. The reason clang complains about it is that it's in an if statement? That's weird.

jinran-google (approved):

bukepo (approved):

jwhui (approved):

PR #986, opened 2021-08-25
[mdns-avahi] browse or resolve using unspecified protocol

simonlingoogle:
This commit fixes the bug that mDNS using Avahi only browses or resolves services using IPv6 sockets. 

Note that `avahi-daemon` may be configured to not use IPv6 socket:

```
       use-ipv6=  Takes a boolean value ("yes" or "no"). If set to "no" avahi-daemon will not use
       IPv6 sockets. Default is "yes".
```

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -935,7 +935,7 @@ void PublisherAvahi::SubscribeService(const std::string &aType, const std::strin
     }
     else
     {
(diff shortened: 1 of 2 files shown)


superwhd (approved):
LGTM:+1:

superwhd (changes requested):
~After a second thought, I suspect we are missing some necessary conversions between system protocol enums and Avahi protocol enums.~

superwhd (approved):
LGTM:+1:

bukepo (approved):

jwhui (approved):

Issue #736, opened 2021-03-16
[border-agent] advertise the meshcop service even when the Thread If is disabled.


wgtdkp:
Current implementation stops advertising the `_meshcop._udp` service when the BR is detached or commissioning with `SPKc` is disabled.

While the Thread specification requires the `_meshcop._udp` service being published when the BR is detached but update the State Bitmap (`sb`) TXT entry to reflect the Thread interface status.

wgtdkp:
@gabekassel  aha, that could be a workaround but there are service parameters that varies depending that the Border Router state and configuration (e.g. the `discriminator` value and `ConnectionMode`).

The expected solution is to advertise the meshcop service once the OTBR process is started and update the service parameters when the Border Router formed a Thread network by itself or is attached to existing network. I will create a PR later.

PR #991, opened 2021-08-26
[border-agent] always publish meshcop service

superwhd:
This PR aims to fix https://github.com/openthread/ot-br-posix/issues/736.

In this PR, otbr-agent will always publish the meshcop service regardless of the state of thread. When thread is disabled, the meshcop service port is a dummy port.

Tested by: https://github.com/openthread/openthread/pull/6962


src/agent/border_agent.cpp
```diff
@@ -66,6 +66,7 @@ namespace otbr {
 static const char kBorderAgentServiceType[] = "_meshcop._udp"; ///< Border agent service type of mDNS
 static const char kBorderAgentServiceInstanceName[] =
     OTBR_MESHCOP_SERVICE_INSTANCE_NAME; ///< Border agent service name of mDNS
+static constexpr int kBorderAgentServiceDummyPort = 65535;
```
wgtdkp, line 69: `65535` is a valid port, right? Consider using zero.

src/border_agent/border_agent.cpp
```diff
@@ -287,6 +288,14 @@ void BorderAgent::PublishMeshCopService(void)
 
     txtList.emplace_back("dn", otThreadGetDomainName(instance));
 #endif
+    if (otBorderAgentGetState(instance) != OT_BORDER_AGENT_STATE_STOPPED)
+    {
+        port = otBorderAgentGetUdpPort(instance);
+    }
+    else
+    {
+        port = kBorderAgentServiceDummyPort;
```
wgtdkp, line 297: Add comments for why using a "dummy" port, why it's acceptable, are there potential issues.

superwhd: Updated

wgtdkp (approved):
LGTM :+1:

src/border_agent/border_agent.cpp
```diff
@@ -311,35 +324,21 @@ void BorderAgent::PublishMeshCopService(void)
...
 }
 
 void BorderAgent::UnpublishMeshCopService(void)
 {
-    assert(IsThreadStarted());
-    VerifyOrExit(!mNetworkName.empty());
-
     otbrLogInfo("Unpublish meshcop service %s.%s.local.", kBorderAgentServiceInstanceName, kBorderAgentServiceType);
 
     mPublisher->UnpublishService(kBorderAgentServiceInstanceName, kBorderAgentServiceType);
```
wgtdkp, line 335: Not related to this PR, just a question: What will happen if the published MeshCoP service has been renamed?

superwhd: First of all. We are not unpublishing MeshCoP services any more. 
For mDNSResponder, the published MeshCoP service will be unpublished, since the implementation stores the original instance name in `mServices`. 
For avahi, we're storing the updated names in `mServices` so that for MeshCoP services, we cannot unpublish the renamed ones. This may be a problem.

wgtdkp: > First of all. We are not unpublishing MeshCoP services any more.

The same concern applies to case where we want to update the renamed MeshCoP service.

wgtdkp: Resolved by https://github.com/openthread/ot-br-posix/pull/1104

src/border_agent/border_agent.cpp
```diff
@@ -68,6 +68,7 @@ static const char kProductName[]            = OTBR_PRODUCT_NAME;
 static const char kBorderAgentServiceType[] = "_meshcop._udp"; ///< Border agent service type of mDNS
 static const char kBorderAgentServiceInstanceName[] =
     OTBR_MESHCOP_SERVICE_INSTANCE_NAME; ///< Border agent service name of mDNS
+static constexpr int kBorderAgentServiceDummyPort = 49152;
```
jwhui, line 71: 49152 is a valid port. Can we use something like 0 instead?

superwhd: Some mDNS publishers treat port 0 in special ways. Taking mDNSResponder as an example:
```
* port:            The port, in network byte order, on which the service accepts connections.
 *                  Pass 0 for a "placeholder" service (i.e. a service that will not be discovered
 *                  by browsing, but will cause a name conflict if another client tries to
 *                  register that same name). Most clients will not use placeholder services.
```

jwhui (approved):

PR #1016, opened 2021-09-16
[logging] fix OpenThread dynamic logging

wgtdkp:
OTBR is providing custom `otPlatLog` function which overwrites the default posix logging function. The problem is that the OTBR `otPlatLog` function will filter out logs based on current OTBR log level, this will result in the situation that we cannot print more OT logs with command `ot-ctl log level <level-value>` (when dynamic logging is enabled).


bukepo (approved):

simonlingoogle:
Should OTBR just use the OpenThread log level instead of having a separate log level ? 
I think it's error prone to have to manage two log levels.

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #1020, opened 2021-09-20
[advertising-proxy] harden duplicate SRP update handling

### LINKED ISSUE #1011: "Operation refused for security reasons" when updating the same SRP host twice in a row
**Describe the bug**
https://github.com/openthread/ot-br-posix/pull/1008 was supposed to solve an issue with duplicate SRP updates but it doesn't help in my case. In fact, it can make things even worse in some scenarios.

Suppose that the following SRP updates are sent within a short period of time:
```
UPDATE1:
    EXISTING_HOST
        SERVICE1
UPDATE2:
    EXISTING_HOST
        SERVICE1
```

Since `PublishServiceHandler` and `PublishHostHandler` simply decrement the number of outstanding operations for a matching SRP update and the handlers can be called synchronously when calling `PublishHost` or `PublishService`, `UPDATE1` will be considered completed after processing `UPDATE2.EXISTING_HOST` even if `UPDATE1.EXISTING_HOST.SERVICE1` is still pending. Moreover, too early removal of `UPDATE1` will invalidate a pointer to `UPDATE2` and `UPDATE2.EXISTING_HOST.SERVICE1` will be put in a wrong place in the memory. As a result, the SRP client will never receive a result of `UPDATE2`. 

**To Reproduce**

1. Build and run OTBR Docker container using commit `685a1e0f483febc26127c8c2f2bd8f922b682a06`.
2. Form a Thread network using default settings.
3. Build and flash [Matter Lock](https://github.com/project-chip/connectedhomeip/tree/master/examples/lock-app/nrfconnect) onto a nRF52840 DK.
4. Press Button 3 on the DK to join the formed Thread network.
5. Reboot the DK. After each reboot the firmware adds a new SRP service and due to the issue described above, an SRP update may fail.

**Expected behavior**
SRP updates succeed no matter how frequently they are sent.

**Console/log output**
To be added.

**Additional context**
I prepared a fix which is available here: https://github.com/Damian-Nordic/ot-br-posix/commit/08cb66612477eb5ceb0574b01e7069e855261807, so I can create a PR, but since it reverts https://github.com/openthread/ot-br-posix/pull/1008 I would like to first discuss if that's a good direction.

Damian-Nordic:
Currently, handlers called after completing an asynchronous
update of a DNS host or service simply decrement the number
of outstanding operations. It may lead to incorrect
handling of several SRP updates related to a single host,
sent within a short period of time, such as the ones below:

```
UPDATE1:
  EXISTING_HOST
    SERVICE1
UPDATE2:
  EXISTING_HOST
    SERVICE1
```

It may happen that both UPDATE1.EXISTING_HOST and
UPDATE2.EXISTING_HOST decrement the number of outstanding
operations for UPDATE1 and UPDATE1 is considered completed
even before UPDATE1.EXISTING_HOST.SERVICE1 is published.

Moreover, since the handlers can be called synchronously
in certain scenarios, it may occur that an update object
is released before PublishHostAndItsServices() exits which
may lead to accessing an invalid memory location. Make sure
that the handlers are called in the main loop.

Due the the mentioned problems, an SRP client may not
receive a response for UPDATE2 and keep re-sending the same
update.

Fixes #1011

Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>

### DIFF
--- a/src/agent/advertising_proxy.cpp
+++ b/src/agent/advertising_proxy.cpp
@@ -156,7 +156,12 @@ void AdvertisingProxy::AdvertisingHandler(otSrpServerServiceUpdateId aId,
 
 void AdvertisingProxy::PublishServiceHandler(const char *aName, const char *aType, otbrError aError, void *aContext)
 {
-    static_cast<AdvertisingProxy *>(aContext)->PublishServiceHandler(aName, aType, aError);
+    std::string       name(aName);
+    std::string       type(aType);
+    AdvertisingProxy *thisPtr = static_cast<AdvertisingProxy *>(aContext);
+
+    thisPtr->mTaskRunner.Post(
+        [name, type, aError, thisPtr]() { thisPtr->PublishServiceHandler(name.c_str(), type.c_str(), aError); });
 }
 
 void AdvertisingProxy::PublishServiceHandler(const char *aName, const char *aType, otbrError aError)
@@ -168,9 +173,10 @@ void AdvertisingProxy::PublishServiceHandler(const char *aName, const char *aTyp
     // TODO: there may be same names between two SRP updates.
     for (auto update = mOutstandingUpdates.begin(); update != mOutstandingUpdates.end(); ++update)
     {
-        for (const auto &nameAndType : update->mServiceNames)
+        for (auto nameAndType = update->mServiceNames.begin(); nameAndType != update->mServiceNames.end();
+             ++nameAndType)
         {
-            if (aName != nameAndType.first || !Mdns::Publisher::IsServiceTypeEqual(aType, nameAndType.second.c_str()))
+            if (aName != nameAndType->first || !Mdns::Publisher::IsServiceTypeEqual(aType, nameAndType->second.c_str()))
             {
                 continue;
             }
@@ -187,6 +193,7 @@ void AdvertisingProxy::PublishServiceHandler(const char *aName, const char *aTyp
             }
             else
             {
+                update->mServiceNames.erase(nameAndType);
                 --update->mCallbackCount;
             }
             ExitNow();
@@ -202,7 +209,10 @@ exit:
 
 void AdvertisingProxy::PublishHostHandler(const char *aName, otbrError aError, void *aContext)
 {
-    static_cast<AdvertisingProxy *>(aContext)->PublishHostHandler(aName, aError);
+    std::string       name(aName);
+    AdvertisingProxy *thisPtr = static_cast<AdvertisingProxy *>(aContext);
+
+    thisPtr->mTaskRunner.Post([name, aError, thisPtr]() { thisPtr->PublishHostHandler(name.c_str(), aError); });
 }
 
 void AdvertisingProxy::PublishHostHandler(const char *aName, otbrError aError)
@@ -213,7 +223,7 @@ void AdvertisingProxy::PublishHostHandler(const char *aName, otbrError aError)
 
     for (auto update = mOutstandingUpdates.begin(); update != mOutstandingUpdates.end(); ++update)
     {
-        if (aName != update->mHostName)
+        if (update->mHostNamePublished || aName != update->mHostName)
         {
             continue;
         }
@@ -230,6 +240,7 @@ void AdvertisingProxy::PublishHostHandler(const char *aName, otbrError aError)
         }
         else
         {
+            update->mHostNamePublished = true;
             --update->mCallbackCount;
         }
         ExitNow();
@@ -256,6 +267,16 @@ void AdvertisingProxy::PublishAllHostsAndServices(void)
 
 exit:
     return;
+}
+
+void AdvertisingProxy::Update(MainloopContext &aMainloop)
+{
+    mTaskRunner.Update(aMainloop);
+}
+
+void AdvertisingProxy::Process(const MainloopContext &aMainloop)
+{
+    mTaskRunner.Process(aMainloop);
 }
 
 otbrError AdvertisingProxy::PublishHostAndItsServices(const otSrpServerHost *aHost, OutstandingUpdate *aUpdate)
--- a/src/agent/advertising_proxy.hpp
+++ b/src/agent/advertising_proxy.hpp
@@ -42,6 +42,8 @@
 #include <openthread/srp_server.h>
 
 #include "agent/ncp_openthread.hpp"
+#include "common/mainloop.hpp"
+#include "common/task_runner.hpp"
 #include "mdns/mdns.hpp"
 
 namespace otbr {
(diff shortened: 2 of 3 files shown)


wgtdkp:
@Damian-Nordic What about adding a test case for #1011 ?

Damian-Nordic:
> @Damian-Nordic What about adding a test case for #1011 ?

Do we have any testing framework for SRP updates in this repository? I have only found the publisher integration tests. Perhaps, writing unit tests for advertising proxy would be easier. What would you recommend?

wgtdkp:
> > @Damian-Nordic What about adding a test case for #1011 ?
> 
> Do we have any testing framework for SRP updates in this repository? I have only found the publisher integration tests. Perhaps, writing unit tests for advertising proxy would be easier. What would you recommend?

I'd prefer functional tests that make sure this case will not fail with your fix :)
The Border Router functions are actually tested in OpenThread. For example, the Advertising Proxy is tested in https://github.com/openthread/openthread/blob/main/tests/scripts/thread-cert/border_router/test_advertising_proxy.py. I think you can make a copy of this test script and change it to test the #1011 case. Then, you can create a new OpenThread pull request and have that OT PR depends on this OTBR PR. Take a look at https://github.com/openthread/openthread/pull/6894 for how to depends on another PR by using `Depends-On` in PR description.

Let me know if you need more help.

src/agent/advertising_proxy.cpp
```diff
@@ -156,7 +156,13 @@ void AdvertisingProxy::AdvertisingHandler(otSrpServerServiceUpdateId aId,
 
 void AdvertisingProxy::PublishServiceHandler(const char *aName, const char *aType, otbrError aError, void *aContext)
 {
-    static_cast<AdvertisingProxy *>(aContext)->PublishServiceHandler(aName, aType, aError);
+    std::string       name(aName);
+    std::string       type(aType);
+    AdvertisingProxy *thisPtr = static_cast<AdvertisingProxy *>(aContext);
+
+    thisPtr->mNcp.PostTimerTask(Milliseconds(0), [name, type, aError, thisPtr]() {
```
simonlingoogle, line 163: Should we add a `TaskRunner` to `AdvertisingProxy` rather than using `mNcp` ?

@wgtdkp comments?

src/agent/advertising_proxy.cpp
```diff
@@ -187,6 +194,7 @@ void AdvertisingProxy::PublishServiceHandler(const char *aName, const char *aTyp
             }
             else
             {
+                update->mServiceNames.erase(nameAndType);
```
simonlingoogle, line 197:
```suggestion
                update->mPendingServiceNames.erase(nameAndType);
```
I suggest to rename to `mPendingServiceNames` to better reflect its usage.

src/agent/advertising_proxy.cpp
```diff
@@ -230,6 +242,7 @@ void AdvertisingProxy::PublishHostHandler(const char *aName, otbrError aError)
         }
         else
         {
+            update->mHostName.clear();
```
simonlingoogle, line 245:
```suggestion
            update->mHostNamePublished = true;
```

Maybe add a flag rather than clear the host name?

src/agent/advertising_proxy.cpp
```diff
@@ -282,6 +295,7 @@ otbrError AdvertisingProxy::PublishHostAndItsServices(const otSrpServerHost *aHo
         aUpdate->mCallbackCount += !hostDeleted;
         aUpdate->mHostName = hostName;
         service            = nullptr;
+
```
simonlingoogle, line 298:
```suggestion
```

nit: avoid unrelated change

simonlingoogle (commented):
Thanks for the fix!

src/agent/advertising_proxy.cpp
wgtdkp: Agree.

src/agent/advertising_proxy.cpp
wgtdkp: `mServiceNames` is already a field of `OutstandingUpdate` which should already indicate that the services and hosts are pending.

simonlingoogle: SGTM

Damian-Nordic:
> Let me know if you need more help.

@wgtdkp Looks like I may need more help as I've spent quite some time trying to run the tests locally and they always fail for me. Is there any instruction how to run tests from thread-cert/border_router reliably? I don't have any issues with tests outside the border_router directory. When I try to run `test_advertising_proxy.py` with command:
`sudo VIRTUAL_TIME=0 ./script/test cert tests/scripts/thread-cert/border_router/test_advertising_proxy.py`

I get the following errors:
```
ERROR: test (__main__.SingleHostAndService)
  File "/home/damian/src/openthread/tests/scripts/thread-cert/border_router/test_advertising_proxy.py", line 89, in test
    self.assertEqual('leader', server.get_state())
  File "/home/damian/src/openthread/tests/scripts/thread-cert/node.py", line 1455, in get_state
    self.send_command('state')
  File "/home/damian/src/openthread/tests/scripts/thread-cert/node.py", line 695, in send_command
    self.pexpect.send(cmd + '\n')
  File "/usr/lib/python3/dist-packages/pexpect/popen_spawn.py", line 142, in send
    return self.proc.stdin.write(b)
BrokenPipeError: [Errno 32] Broken pipe
```
When I try to use `VIRTUAL_TIME=1` or other settings I get other errors.

Also, when the test fails it leaves a lot of open processes, docker containers and temporary files, so one has to always clean everything up. Do you run these tests locally or use some VM/docker/whatever?

simonlingoogle:
> Is there any instruction how to run tests from thread-cert/border_router reliably?

Running tests in `border_router` directory is more complex.

You can follow steps in `.github/workflows/border_router.yml` `border_router` job. It can be either `Border Router (mDNSResponder)` or `Border Router (Avahi)`.

wgtdkp:
Hi @Damian-Nordic, please see https://github.com/openthread/openthread/pull/7044 for instructions of running BR tests locally.

src/agent/advertising_proxy.cpp
Damian-Nordic: Added a new `TaskRunner`.

Damian-Nordic:
@simonlingoogle @wgtdkp Thanks for comments. I added a test: https://github.com/openthread/openthread/pull/7049 and applied your suggestions.

wgtdkp (approved):
Thanks! :+1:

LuDuda (approved):

simonlingoogle (approved):
LGTM 👍

jwhui (approved):
Thanks! 👍🏼

PR #1025, opened 2021-09-22
Refactor otbr agent

wangjm92:
Added library `otbr-border-agent` and `otbr-sdp-proxy`
Moved ncp_openthread.cpp/hpp and thread_helper.cpp/hpp from `otbr-agent` to `otbr-utils`


Android.mk
```diff
@@ -131,7 +127,11 @@ LOCAL_SRC_FILES := \
     src/dbus/server/dbus_thread_object.cpp \
     src/dbus/server/error_helper.cpp \
     src/mdns/mdns.cpp \
+    src/sdp_proxy/advertising_proxy.cpp \
```
simonlingoogle, line 130: what does `sdp` mean?

Android.mk
```diff
@@ -131,7 +127,11 @@ LOCAL_SRC_FILES := \
...
+    src/sdp_proxy/discovery_proxy.cpp \
     src/utils/hex.cpp \
+    src/utils/ncp_openthread.cpp \
```
simonlingoogle, line 133:
```suggestion
    src/ncp/ncp_openthread.cpp \
```

move to ncp ?

simonlingoogle (commented):
Totally agree that we should move these files out of the `agent` directory.

Android.mk
wangjm92: service discovery protocol

Android.mk
wangjm92: From fuchsia's perspective, we will need to include //src/utils anyway, so I did not add another directory for ncp. I am ok with adding a //src/ncp if that looks better to you.

src/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 #
-#  Copyright (c) 2020, The OpenThread Authors.
+#  Copyright (c) 2021, The OpenThread Authors.
```
simonlingoogle, line 2:
```suggestion
#  Copyright (c) 2020, The OpenThread Authors.
```

Do we change copyright version when modifying files? @jwhui

src/border_agent/border_agent.hpp
```diff
@@ -38,12 +38,12 @@
 
 #include <stdint.h>
 
-#include "agent/advertising_proxy.hpp"
-#include "agent/discovery_proxy.hpp"
 #include "agent/instance_params.hpp"
-#include "agent/ncp_openthread.hpp"
 #include "common/mainloop.hpp"
 #include "mdns/mdns.hpp"
+#include "sdp_proxy/advertising_proxy.hpp"
```
simonlingoogle, line 44: I think eventually we should move these components out of border_agent.

simonlingoogle (approved):
LGTM :+1:

wgtdkp (approved):
LGTM :+1:

bukepo (approved):

PR #1038, opened 2021-10-01
[tests] use `--no-check-certificates` when invoking `wget`

jwhui:
To address the following issue:
```
get -nv https://dbus.freedesktop.org/releases/dbus/dbus-1.4.26.tar.gz
ERROR: cannot verify dbus.freedesktop.org's certificate, issued by '/C=US/O=Let\'s Encrypt/CN=R3':
  Issued certificate has expired.
To connect to dbus.freedesktop.org insecurely, use `--no-check-certificate'.
Error: Process completed with exit code 5.
```

### DIFF
--- a/tests/scripts/check-android-build
+++ b/tests/scripts/check-android-build
@@ -42,7 +42,7 @@ set -x
 prepare_dbus()
 {
     DBUS_SOURCE=dbus-1.4.26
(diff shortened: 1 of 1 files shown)


abtink (approved):
LGTM. 👍

PR #1055, opened 2021-10-19
[dua] increase number of supported DUA registrations

edmont:
Increase ancillary buffer size to allow for a larger number of multicast groups
Required for NdProxyManager::JoinSolicitedNodeMulticastGroup

### DIFF
--- a/script/_rt_tables
+++ b/script/_rt_tables
@@ -51,5 +51,6 @@ rt_tables_install()
         sudo sh -c 'echo "" >>/etc/sysctl.conf'
         sudo sh -c 'echo "# OpenThread configuration" >>/etc/sysctl.conf'
         sudo sh -c 'echo "net.core.optmem_max=65536" >>/etc/sysctl.conf'
+        sudo sh -c 'sysctl -p /etc/sysctl.conf'
     fi
 }


simonlingoogle:
> Any idea on how to overcome the CI issues?

Maybe we can skip it if it's building in Docker. We can always configure this in the Host when we are running OTBR in Docker.

edmont:
> Maybe we can skip it if it's building in Docker. We can always configure this in the Host when we are running OTBR in Docker.

I don't fully understand your proposal here. Should I change something in my code?

simonlingoogle:
> > Maybe we can skip it if it's building in Docker. We can always configure this in the Host when we are running OTBR in Docker.
> 
> I don't fully understand your proposal here. Should I change something in my code?

Apparently `sysctl -w net.core.optmem_max=65536` failed to run in Docker. I am proposing that the setup script can check if it's in Docker, and skip setting `sysctl -w net.core.optmem_max=65536` if it is.

edmont:
@simonlingoogle thanks for the tip, now CI is green. I understand that the setting must be applied on the host side in order to be available in the Docker container.

@jwhui could you label this PR as `bug`? Without this fix the default `OPENTHREAD_CONFIG_NDPROXY_TABLE_ENTRY_NUM` minimum limit for DUA registrations can't be reached.

PR #1063, opened 2021-10-26
[firewall] ingress filtering

superwhd:
Re-sending the PR that implements firewall. It's mostly the same as https://github.com/openthread/ot-br-posix/pull/1030, which was previously reverted due to flaky CI.

Besides the previous changes, I also added one more rule to drop unicast packets from wpan0 to wpan0.

The following is the description copied from https://github.com/openthread/ot-br-posix/pull/1030.

------------

This PR implements part of the OTBR firewall. It should be merged together with https://github.com/openthread/openthread/pull/7107. This implementation focuses on the ingress filtering part. We may also introduce egress filtering part when necessary.

For security purpose, there are some packet forwarding rules to follow, which were originally introduced in the spec.

Inbound packets initiated with On-Link addresses source (OMR and mesh local prefix based addresses) should be blocked.
Inbound unicast packets whose destination address is not OMR address or DUA should be blocked.
Inbound unicast packets whose source address or destination address is link-local should be blocked. Note that we don’t need to explicitly add rules for link-local addresses since this should already be handled by the kernel.
These rules can be easily implemented by iptables and ipset.

Before otbr-agent starts, there is a script creating the iptables rules. The rules themselves are constant so we don't need to change them dynamically.
During the runtime of otbr-agent, otbr-agent updates ipsets accordingly whenever there's a change of on-link prefixes.

### DIFF
--- a/examples/platforms/beagleboneblack/default
+++ b/examples/platforms/beagleboneblack/default
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-#  Copyright (c) 2017, The OpenThread Authors.
+#  Copyright (c) 2017-2021, The OpenThread Authors.
 #  All rights reserved.
 #
 #  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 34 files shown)


simonlingoogle (approved):
LGTM :+1:

superwhd:
@jwhui  Hi Jonathan, this PR and https://github.com/openthread/openthread/pull/7107 are waiting for your review. Could you take a look when you get a chance?

jwhui (approved):

PR #1065, opened 2021-10-26
[meshcop] add vn and mn fields in meshcop service's TXT record

superwhd:
vn: vendor name
mn: model name

These fields are defined as optional in the spec.

PR #1068, opened 2021-10-28
[WIP] [meshcop] use space instead of underscore in meshcop service instance name

superwhd:


superwhd:
> @superwhd what's the motivation for this change?

I previously thought it would be more readable to use space instead of underscore in the service instance names. Also it would follow the recommendation (not mandatory) by Thread spec. Since we've introduced the CMake option to completely customize the service instance name (https://github.com/openthread/ot-br-posix/pull/1065), I think we don't need to move forward with the current PR. I'm closing it.

PR #1072, opened 2021-10-30
[docs] remove redundant documentation

wgtdkp:
There are few benefits of documenting the same interface again in derived
classes but only introduce the burden of keeping them consistent when the
base class documentations are updated.

PR #1073, opened 2021-10-30
[mdns] unify mDNS interface to use std::string

wgtdkp:
We are inconsistenly using both const char * and std::string in the
code which is sure bad practice. We have been seen bugs caused by
raw char * strings and I think it should be safer to use std::string
for easier life-time management.
    
This commit unifies the mDNS and advertising proxy modules to always
use std::string against const char * and std::vector<uint8_t> against
const uint8_t * whenever appliable. This makes the code more maintainable
and avoids potential memory segments.

This commit also removes the unused `aProtocol` and `aDomain` parameters
of `Publisher::Create()`.

------
This PR depends on https://github.com/openthread/ot-br-posix/pull/1072 so please review only the second commit, thanks.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -39,53 +39,38 @@ namespace otbr {
 
 namespace Mdns {
 
-bool Publisher::IsServiceTypeEqual(const char *aFirstType, const char *aSecondType)
+bool Publisher::IsServiceTypeEqual(std::string aFirstType, std::string aSecondType)
 {
-    size_t firstLength  = strlen(aFirstType);
-    size_t secondLength = strlen(aSecondType);
-
-    if (firstLength > 0 && aFirstType[firstLength - 1] == '.')
+    if (!aFirstType.empty() && aFirstType.back() == '.')
     {
-        --firstLength;
-    }
-    if (secondLength > 0 && aSecondType[secondLength - 1] == '.')
-    {
-        --secondLength;
+        aFirstType.pop_back();
     }
 
-    return firstLength == secondLength && memcmp(aFirstType, aSecondType, firstLength) == 0;
+    if (!aSecondType.empty() && aSecondType.back() == '.')
+    {
+        aSecondType.pop_back();
+    }
+
+    return aFirstType == aSecondType;
 }
 
-otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, uint8_t *aTxtData, uint16_t &aTxtLength)
+otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, std::vector<uint8_t> &aTxtData)
 {
     otbrError error = OTBR_ERROR_NONE;
-    uint8_t * cur   = aTxtData;
 
     for (const auto &txtEntry : aTxtList)
     {
-        const char *   name        = txtEntry.mName.c_str();
-        const size_t   nameLength  = txtEntry.mName.length();
-        const uint8_t *value       = txtEntry.mValue.data();
-        const size_t   valueLength = txtEntry.mValue.size();
-        const size_t   entryLength = nameLength + 1 + valueLength;
+        const auto & name        = txtEntry.mName;
+        const auto & value       = txtEntry.mValue;
+        const size_t entryLength = name.length() + 1 + value.size();
 
-        VerifyOrExit(nameLength > 0 && nameLength <= kMaxTextEntrySize, error = OTBR_ERROR_INVALID_ARGS);
-        VerifyOrExit(cur + entryLength + 1 <= aTxtData + aTxtLength, error = OTBR_ERROR_INVALID_ARGS);
+        VerifyOrExit(entryLength <= kMaxTextEntrySize, error = OTBR_ERROR_INVALID_ARGS);
 
-        cur[0] = static_cast<uint8_t>(entryLength);
-        ++cur;
-
-        memcpy(cur, name, nameLength);
-        cur += nameLength;
-
-        cur[0] = '=';
-        ++cur;
-
-        memcpy(cur, value, valueLength);
-        cur += valueLength;
+        aTxtData.push_back(static_cast<uint8_t>(entryLength));
+        aTxtData.insert(aTxtData.end(), name.begin(), name.end());
+        aTxtData.push_back('=');
+        aTxtData.insert(aTxtData.end(), value.begin(), value.end());
     }
-
-    aTxtLength = cur - aTxtData;
 
 exit:
     return error;
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -75,6 +75,8 @@ namespace otbr {
 
 namespace Mdns {
 
+static const char kDomain[] = "local.";
+
 Poller::Poller(void)
 {
     mAvahiPoller.userdata         = this;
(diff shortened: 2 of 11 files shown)


src/mdns/mdns.cpp
```diff
@@ -39,54 +39,39 @@ namespace otbr {
...
 
-    return firstLength == secondLength && memcmp(aFirstType, aSecondType, firstLength) == 0;
+    return aFirstType == aSecondType;
 }
 
-otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, uint8_t *aTxtData, uint16_t &aTxtLength)
+otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, std::vector<uint8_t> &aTxtData)
 {
     otbrError error = OTBR_ERROR_NONE;
-    uint8_t * cur   = aTxtData;
```
superwhd, line 60: nit: Use `vector::reserve` to avoid too many memory allocations.

src/mdns/mdns.cpp
```diff
@@ -39,54 +39,39 @@ namespace otbr {
...
-        --secondLength;
+        aSecondType.pop_back();
     }
 
-    return firstLength == secondLength && memcmp(aFirstType, aSecondType, firstLength) == 0;
+    return aFirstType == aSecondType;
 }
 
-otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, uint8_t *aTxtData, uint16_t &aTxtLength)
+otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, std::vector<uint8_t> &aTxtData)
```
superwhd, line 57: nit: Use `vector::reserve` to reduce the times of memory allocations.

src/mdns/mdns_avahi.cpp
```diff
@@ -864,18 +837,6 @@ otbrError PublisherAvahi::TxtListToAvahiStringList(const TxtList &   aTxtList,
     return error;
 }
 
-std::string PublisherAvahi::MakeFullName(const char *aName)
```
superwhd, line 867: I'm inclined to keep this function.

src/mdns/mdns_avahi.cpp
```diff
@@ -373,21 +372,16 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
         if (serviceIt != mServices.end())
         {
             Service &   service         = *serviceIt;
-            const char *hostName        = service.mHostName.c_str();
+            std::string hostName        = service.mHostName;
```
superwhd, line 375: We may not need this variable since hostName is not changed in this block.

src/mdns/mdns_avahi.cpp
```diff
@@ -459,14 +449,10 @@ otbrError PublisherAvahi::CreateHost(AvahiClient &aClient, const char *aHostName
...
+PublisherAvahi::Services::iterator PublisherAvahi::FindService(const std::string &aName, const std::string &aType)
 {
-    assert(aName != nullptr);
-    assert(aType != nullptr);
-
-    return std::find_if(mServices.begin(), mServices.end(), [aName, aType](const Service &aService) {
-        return aService.mName == aName && aService.mType == aType;
-    });
+    return std::find_if(mServices.begin(), mServices.end(),
+                        [&](const Service &aService) { return aService.mName == aName && aService.mType == aType; });
```
superwhd, line 455: nit: Explicitly specify what are captured.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -710,55 +697,36 @@ void PublisherMDnsSd::HandleRegisterHostResult(DNSServiceRef       aHostsConnect
...
+    return regType;
 }
 
-PublisherMDnsSd::ServiceIterator PublisherMDnsSd::FindPublishedService(const char *aName, const char *aType)
+PublisherMDnsSd::ServiceIterator PublisherMDnsSd::FindPublishedService(const std::string &aName,
+                                                                       const std::string &aType)
 {
-    return std::find_if(mServices.begin(), mServices.end(), [&aName, aType](const Service &service) {
-        return strcmp(aName, service.mName) == 0 && IsServiceTypeEqual(aType, service.mType);
+    return std::find_if(mServices.begin(), mServices.end(), [&](const Service &service) {
```
superwhd, line 728: I'm inclined to list the captures explicitly.

src/sdp_proxy/advertising_proxy.cpp
```diff
@@ -154,29 +154,29 @@ void AdvertisingProxy::AdvertisingHandler(otSrpServerServiceUpdateId aId,
...
+                                             otbrError          aError,
+                                             void *             aContext)
 {
-    std::string       name(aName);
-    std::string       type(aType);
     AdvertisingProxy *thisPtr = static_cast<AdvertisingProxy *>(aContext);
 
-    thisPtr->mTaskRunner.Post(
-        [name, type, aError, thisPtr]() { thisPtr->PublishServiceHandler(name.c_str(), type.c_str(), aError); });
+    thisPtr->mTaskRunner.Post([=]() { thisPtr->PublishServiceHandler(aName, aType, aError); });
```
superwhd, line 164: I'd suggest to list the captures explicitly.

src/sdp_proxy/advertising_proxy.cpp
```diff
@@ -203,23 +203,22 @@ void AdvertisingProxy::PublishServiceHandler(const char *aName, const char *aTyp
...
 }
 
-void AdvertisingProxy::PublishHostHandler(const char *aName, otbrError aError, void *aContext)
+void AdvertisingProxy::PublishHostHandler(const std::string &aName, otbrError aError, void *aContext)
 {
-    std::string       name(aName);
     AdvertisingProxy *thisPtr = static_cast<AdvertisingProxy *>(aContext);
 
-    thisPtr->mTaskRunner.Post([name, aError, thisPtr]() { thisPtr->PublishHostHandler(name.c_str(), aError); });
+    thisPtr->mTaskRunner.Post([=]() { thisPtr->PublishHostHandler(aName, aError); });
```
superwhd, line 214: I'd suggest to list the captures explicitly.

src/sdp_proxy/advertising_proxy.cpp
```diff
@@ -302,15 +301,17 @@ otbrError AdvertisingProxy::PublishHostAndItsServices(const otSrpServerHost *aHo
 
     if (!hostDeleted)
     {
+        std::vector<uint8_t> firstHostAddress{hostAddress[0].mFields.m8,
+                                              hostAddress[0].mFields.m8 + sizeof(hostAddress[0])};
```
superwhd, line 305: For simplicity, you can use `std::end(hostAddress[0].mFields.m8)` for the end iterator. I'm not sure if `std::end` is considered allowed in the codebase though.

tests/mdns/main.cpp
```diff
@@ -98,7 +98,7 @@ void PublishSingleServiceWithCustomHost(void *aContext, Mdns::Publisher::State a
         Mdns::Publisher::TxtList txtList{
             {"nn", "cool"}, {"xp", xpanid, sizeof(xpanid)}, {"tv", "1.1.1"}, {"dd", extAddr, sizeof(extAddr)}};
 
-        error = sContext.mPublisher->PublishHost(hostName, hostAddr, sizeof(hostAddr));
+        error = sContext.mPublisher->PublishHost(hostName, {hostAddr, hostAddr + sizeof(hostAddr)});
```
superwhd, line 101: nit: make `hostAddr` a vector so you don't need to repeat this pattern in the test code.

simonlingoogle (approved):
LGTM :+1: 

Thanks

src/mdns/mdns.cpp
wgtdkp: May not easy to estimate the size, I think `vector::reserve` has its advantages only if we already know the target size this vector is going to grow to.

src/mdns/mdns_avahi.cpp
wgtdkp: reverted.

src/sdp_proxy/advertising_proxy.cpp
wgtdkp: updated!

bukepo (approved):

zhanglongxia (approved):
LGTM

superwhd (approved):
LGTM :+1:

jwhui (approved):

PR #1074, opened 2021-10-31
[docs] unify spaces in documentations

wgtdkp:
There are different numbers of spaces among parameter tag (i.e. `@param`), the parameter
and parameter comment. This PR unifies the spaces:
1. Use 1 space after tags (e.g. @param @returns).
2. Use 2 spaces between the parameter/retval and comment.
3. Capitalizing the first char of the param and return value comment.

An example of the new documentation is:
```c++
    /**
     * This method publishes or updates a service.
     *
     * @param[in] aHostName     The name of the host which this service resides on. If an empty string is
     *                          provided, this service resides on local host and it is the implementation
     *                          to provide specific host name. Otherwise, the caller MUST publish the host
     *                          with method PublishHost.
     * @param[in] aPort         The port number of this service.
     * @param[in] aName         The name of this service.
     * @param[in] aType         The type of this service.
     * @param[in] aSubTypeList  A list of service subtypes.
     * @param[in] aTxtList      A list of TXT name/value pairs.
     *
     * @retval OTBR_ERROR_NONE   Successfully published or updated the service.
     * @retval OTBR_ERROR_ERRNO  Failed to publish or update the service.
     *
     */
```

This PR also fixes trivial typos.

------
This PR depends on https://github.com/openthread/ot-br-posix/pull/1073 so please reviews only the third commit.

### DIFF
--- a/src/border_agent/border_agent.cpp
+++ b/src/border_agent/border_agent.cpp
@@ -199,7 +199,7 @@ void BorderAgent::HandleMdnsState(Mdns::Publisher::State aState)
 #endif
         break;
     default:
-        otbrLogWarning("mDNS service not available!");
+        otbrLogWarning("mDNS publisher not available!");
         break;
     }
 }
--- a/src/common/mainloop_manager.hpp
+++ b/src/common/mainloop_manager.hpp
@@ -61,8 +61,6 @@ public:
     /**
      * This method returns the singleton instance of the mainloop manager.
      *
-     * @retval  A reference to the mainloop manager.
-     *
      */
     static MainloopManager &GetInstance(void)
     {
@@ -73,7 +71,7 @@ public:
     /**
      * This method adds a mainloop processors to the mainloop managger.
      *
-     * @param[in] aMainloopProcessor   A pointer to the mainloop processor.
+     * @param[in] aMainloopProcessor  A pointer to the mainloop processor.
      *
      */
     void AddMainloopProcessor(MainloopProcessor *aMainloopProcessor);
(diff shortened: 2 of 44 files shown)


src/common/mainloop_manager.hpp
```diff
@@ -61,7 +61,7 @@ class MainloopManager
     /**
      * This method returns the singleton instance of the mainloop manager.
      *
-     * @retval  A reference to the mainloop manager.
+     * @retval A reference to the mainloop manager.
```
superwhd, line 64:
```suggestion
     * @returns A reference to the mainloop manager.
```

src/border_agent/border_agent.cpp
```diff
@@ -278,9 +278,8 @@ void BorderAgent::PublishMeshCopService(void)
     txtList.emplace_back("dn", otThreadGetDomainName(instance));
 #endif
 
-    mPublisher->PublishService(/* aHostName */ nullptr, otBorderAgentGetUdpPort(instance),
-                               kBorderAgentServiceInstanceName, kBorderAgentServiceType, Mdns::Publisher::SubTypeList{},
-                               txtList);
+    mPublisher->PublishService(/* aHostName */ "", otBorderAgentGetUdpPort(instance), kBorderAgentServiceInstanceName,
```
superwhd, line 281: It seems this is not a doc change. Could you clarify the purpose of doing so? Or you may want to make this change in another PR.

simonlingoogle (approved):
LGTM :+1:

src/border_agent/border_agent.cpp
wgtdkp: This isn't change in this PR, see https://github.com/openthread/ot-br-posix/pull/1073 instead.

abtink (approved):
👍

superwhd (approved):
Thanks!

jwhui (approved):

PR #1078, opened 2021-11-03
[web-service] flush any lingering output before issuing command

jwhui:
This commit ensures to flush any received CLI output before issuing
the command to make parsing the output of the command more reliable.

### DIFF
--- a/src/web/web-service/ot_client.cpp
+++ b/src/web/web-service/ot_client.cpp
@@ -110,6 +110,32 @@ exit:
     return ret == 0;
 }
 
+void OpenThreadClient::DiscardRead(void)
+{
+    fd_set  readFdSet;
+    timeval timeout = {0, 0};
+    ssize_t count;
+    int     ret;
+
+    for (;;)
+    {
+        FD_ZERO(&readFdSet);
+        FD_SET(mSocket, &readFdSet);
+
+        ret = select(mSocket + 1, &readFdSet, nullptr, nullptr, &timeout);
+        if (ret <= 0)
+        {
+            break;
+        }
+
+        count = read(mSocket, mBuffer, sizeof(mBuffer));
+        if (count <= 0)
+        {
+            break;
+        }
+    }
+}
+
 char *OpenThreadClient::Execute(const char *aFormat, ...)
 {
     va_list args;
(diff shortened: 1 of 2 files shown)


src/web/web-service/ot_client.cpp
```diff
@@ -118,6 +118,25 @@ char *OpenThreadClient::Execute(const char *aFormat, ...)
...
+
+        ret = select(mSocket + 1, &readFdSet, nullptr, nullptr, &timeout);
+        if (ret <= 0)
+        {
+            break;
+        }
+
+        count = read(mSocket, mBuffer, sizeof(mBuffer));
+        VerifyOrExit(count > 0);
+    }
```
simonlingoogle, line 138: nit: encapsulate with a function? `OpenThreadClient::Execute` seems to be getting large.

simonlingoogle (approved):
LGTM :+1:

simonlingoogle:
If `otbr-web` connects to the OT CLI socket, will it prohibit users from opening another `ot-ctl` command in interactive mode? @bukepo

jwhui:
> If `otbr-web` connects to the OT CLI socket, will it prohibit users from opening another `ot-ctl` command in interactive mode? @bukepo

Yes. The same is true for the reverse - if `ot-ctl` command is in interactive mode, it prevents `otbr-web` from functioning.

simonlingoogle:
> > If `otbr-web` connects to the OT CLI socket, will it prohibit users from opening another `ot-ctl` command in interactive mode? @bukepo
> 
> Yes. The same is true for the reverse - if `ot-ctl` command is in interactive mode, it prevents `otbr-web` from functioning.

I tend to think that's a bug. Thoughts?

bukepo (approved):

bukepo:
> > > If `otbr-web` connects to the OT CLI socket, will it prohibit users from opening another `ot-ctl` command in interactive mode? @bukepo
> > 
> > 
> > Yes. The same is true for the reverse - if `ot-ctl` command is in interactive mode, it prevents `otbr-web` from functioning.
> 
> I tend to think that's a bug. Thoughts?

`ot-ctl` keeps connection to server, while otbr-web will not. We can enhance this by close connection when `Done` is received in response and re-establish when a new command is issued. Thoughts?

simonlingoogle:
> > > > If `otbr-web` connects to the OT CLI socket, will it prohibit users from opening another `ot-ctl` command in interactive mode? @bukepo
> > > 
> > > 
> > > Yes. The same is true for the reverse - if `ot-ctl` command is in interactive mode, it prevents `otbr-web` from functioning.
> > 
> > 
> > I tend to think that's a bug. Thoughts?
> 
> `ot-ctl` keeps connection to server, while otbr-web will not. We can enhance this by close connection when `Done` is received in response and re-establish when a new command is issued. Thoughts?

I believe it's a good idea.

PR #1083, opened 2021-11-04
[build] provide a sample Yocto recipe

gabekassel:
Provides a sample Yocto recipe to build otbr-agent and set specific build flags.

### DIFF
--- a/etc/yocto/otbr_git.bb
+++ b/etc/yocto/otbr_git.bb
@@ -1,0 +1,37 @@
+# Sample Yocto recipe to build OTBR
+
+DESCRIPTION = "OpenThread Border Router Agent"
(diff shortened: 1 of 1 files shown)


etc/yocto/otbr_git.bb
```diff
@@ -0,0 +1,37 @@
+# Sample Yocto recipe to build OTBR
```
wgtdkp, line 1: Add the OpenThread license header?

wgtdkp (approved):
:+1:

etc/yocto/otbr_git.bb
gabekassel: added

jwhui (approved):

PR #1085, opened 2021-11-09
[discovery-proxy] unescape service instance name with spaces

simonlingoogle:


src/utils/dns_utils.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * This function unescapes a DNS Service Instance name according to "DNS Name Escaping".
+ *
+ * @sa  "Notes on DNS Name Escaping" at
+ *      https://opensource.apple.com/source/mDNSResponder/mDNSResponder-1310.140.1/mDNSShared/dns_sd.h.auto.html.
+ *
+ * @param[in] aName  The DNS Service Instance name to unescape.
+ *
+ * @returns  The unescaped DNS Service Instance name.
```
superwhd, line 53:
```suggestion
 * @returns The unescaped DNS Service Instance name.
```

src/utils/dns_utils.cpp
```diff
@@ -0,0 +1,73 @@
...
+            else if (i + 1 < nameLen)
+            {
+                newName.push_back(aName[i + 1]);
+                i += 1;
+                continue;
+            }
+        }
+
+        // append all not escaped characters
+        newName.push_back(c);
```
superwhd, line 65: What about using `if` and `else` instead of `continue`? Or use `continue` instead of `else`?

src/sdp_proxy/discovery_proxy.cpp
```diff
@@ -154,6 +155,8 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string &
                 aType.c_str(), aInstanceInfo.mName.c_str(), aInstanceInfo.mHostName.c_str(),
                 aInstanceInfo.mAddresses.size(), aInstanceInfo.mPort, aInstanceInfo.mPriority, aInstanceInfo.mWeight);
 
+    std::string unescapedInstanceName = DnsUtils::UnescapeInstanceName(aInstanceInfo.mName);
```
superwhd, line 158: Move this line above the logging?

superwhd (approved):

superwhd:
LGTM with some nits.

src/sdp_proxy/discovery_proxy.cpp
simonlingoogle: done

src/utils/dns_utils.hpp
simonlingoogle: We don't seem to have a strict rule that the spacing after `@returns` should be exactly one.

src/utils/dns_utils.cpp
simonlingoogle: Changed to use `continue` consistently. 

We are using `continue` rather than else because for special cases like:
- `\2` should unescaped to `2`
- `\` at the end of line should **not** be unescaped

wgtdkp (approved):
LGTM :+1:

jwhui (approved):

PR #1090, opened 2021-11-12
[dns-sd] let Avahi resolver handle IPv4 services

superwhd:
This PR lets Avahi resolve a service instance in both IPv4 and IPv6. 

`otbr-agent` will not process the resolved IPv4 addresses but other information like port, TXT will be handled in the discovery proxy.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -86,7 +86,8 @@ install_packages_apt()
     sudo apt-get install --no-install-recommends -y libjsoncpp-dev
 
     # reference device
-    without REFERENCE_DEVICE || sudo apt-get install --no-install-recommends -y radvd dnsutils
+    without REFERENCE_DEVICE || sudo apt-get install --no-install-recommends -y radvd dnsutils python3-pip
+    without REFERENCE_DEVICE || pip3 install zeroconf
 
     # backbone-router
     without BACKBONE_ROUTER || sudo apt-get install --no-install-recommends -y iptables libnetfilter-queue1 libnetfilter-queue-dev
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -53,7 +53,7 @@ install_packages_apt()
     sudo apt-get install --no-install-recommends -y libdbus-1-dev
 
     # mDNS
-    sudo apt-get install --no-install-recommends -y libavahi-client3 libavahi-common-dev libavahi-client-dev avahi-daemon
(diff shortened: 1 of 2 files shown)


simonlingoogle:
@superwhd Is the PR still draft?

script/bootstrap
```diff
@@ -86,7 +86,8 @@ install_packages_apt()
     sudo apt-get install --no-install-recommends -y libjsoncpp-dev
 
     # reference device
-    without REFERENCE_DEVICE || sudo apt-get install --no-install-recommends -y radvd dnsutils
+    without REFERENCE_DEVICE || sudo apt-get install --no-install-recommends -y radvd dnsutils python3-pip
+    without REFERENCE_DEVICE || pip3 install zeroconf
```
simonlingoogle, line 90: why do we need zeroconf in reference devce?

superwhd:
> @superwhd Is the PR still draft?

Yes. I'll request a review once it passes the test case https://github.com/openthread/openthread/pull/7177.

src/mdns/mdns_avahi.cpp
```diff
@@ -1024,7 +1024,7 @@ void PublisherAvahi::ServiceSubscription::Resolve(uint32_t           aInterfaceI
     otbrLogInfo("resolve service %s %s inf %d", aInstanceName.c_str(), aType.c_str(), aInterfaceIndex);
     mServiceResolver = avahi_service_resolver_new(
         mPublisherAvahi->mClient, aInterfaceIndex, aProtocol, aInstanceName.c_str(), aType.c_str(),
-        /* domain */ nullptr, AVAHI_PROTO_INET6, static_cast<AvahiLookupFlags>(0), HandleResolveResult, this);
+        /* domain */ nullptr, AVAHI_IF_UNSPEC, static_cast<AvahiLookupFlags>(0), HandleResolveResult, this);
```
simonlingoogle, line 1027: How does the change work? Do we have any test results?

One small concern about using `AVAHI_IF_UNSPEC` is that if the service has both IPv6 and IPv4 addresses, it may return any of them, which is not ideal because we wanto to prefer IPv6 address.

But it's fine to enhance it in a future PR.

simonlingoogle (approved):
LGTM :+1:

superwhd:
Closing this PR since it has conflict a bit with the recent pending changes. I'll send another PR later.

PR #1103, opened 2021-11-20
[common] add OnceCallback which can be invoked for at most once

wgtdkp:
This commit adds a new facility which wraps up a std::function
and makes sure that a function can be invoked for at most once.

This is useful for async operations to guarantee that the callback
for receiving the result will be only invoked once.

### DIFF
--- a/tests/unit/test_once_callback.cpp
+++ b/tests/unit/test_once_callback.cpp
@@ -1,0 +1,67 @@
+/*
+ *    Copyright (c) 2021, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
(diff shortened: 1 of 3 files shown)


simonlingoogle (commented):
why not use Abseil?

tests/unit/test_once_callback.cpp
```diff
@@ -0,0 +1,56 @@
...
+    otbr::OnceCallback<void(void)> noop = [](void) {};
+
+    CHECK_FALSE(noop.IsNull());
+}
+
+TEST(IsNull, IsNullAfterInvoking)
+{
+    otbr::OnceCallback<int(int)> square = [](int x) { return x * x; };
+
+    std::move(square)(5);
```
superwhd, line 53: Check the result of function call.

superwhd (approved):
LGTM :+1:

wgtdkp:
> why not use Abseil?

There can be platforms which Abseil doesn't support.

simonlingoogle (approved):
LGTM :+1:

tests/unit/test_once_callback.cpp
wgtdkp: Added one more test case for verifying the result of the callback. Current test case is for checking the `IsNull` function, feel not good to add it directly here.

wgtdkp:
@jwhui There is another significant enhancement PR (https://github.com/openthread/ot-br-posix/pull/1113) which is depending on this commit, PTAL :)

bukepo (approved):
Glad to see we don't need to remove the Android check.

jwhui (approved):

PR #1104, opened 2021-11-22
[mdns] fix bug that avahi may publish redundant meshcop services

superwhd:
Avahi may rename the meshcop service and publish it when there's a name conflict. However, our current logic has a bug that it may publish redundant meshcop services after some renaming.

For example there are two BRs on the infra link:

1. BR1 starts. It publishes `instance._meshcop._udp`.
2. BR2 starts. It publishes `instance #1._meshcop._udp`.
3. BR1 stops. It unpublishes `instance._meshcop._udp`.
3. BR2 wants to update the existing meshcop service. However, it actually publishes a new service `instance._meshcop._udp`. There are two meshcop services existing at the same time both published by BR2.

To fix this bug, we will keep track of the original name in the avahi publisher. When we call `PublishService`, we should always pass in the original name and let the publisher handle the renaming. In this way we will be able to find the wanted service when updating it or unpublishing it. 

The test case is covered in https://github.com/openthread/openthread/pull/7192.


simonlingoogle (approved):
LGTM :+1:

wgtdkp (approved):

Damian-Nordic (approved):

jwhui (approved):
Thanks! 👍🏼

PR #1111, opened 2021-11-25
[meshcop] use xa instead of dd according to latest spec

superwhd:
`xa` stands for the extended mac address of the thread interface of the border agent.

We also need https://github.com/openthread/openthread/pull/7204 to update the test case in openthread repo.


wgtdkp (approved):
LGTM 👍

simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #1113, opened 2021-11-25
Mdns callback for each service

wgtdkp:


src/mdns/mdns_avahi.cpp
```diff
@@ -843,9 +767,38 @@ otbrError PublisherAvahi::TxtListToAvahiStringList(const TxtList &   aTxtList,
...
 
-std::string PublisherAvahi::MakeFullName(const std::string &aName)
+Publisher::ServiceRegistrationPtr PublisherAvahi::FindServiceRegistration(const AvahiEntryGroup *aEntryGroup) const
 {
-    return aName + "." + kDomain;
+    ServiceRegistrationPtr result = nullptr;
+
+    for (auto &kv : mServiceRegistrations)
+    {
+        auto &serviceReg = static_cast<AvahiServiceRegistration &>(*kv.second);
```
superwhd, line 776:
```suggestion
    for (const auto &kv : mServiceRegistrations)
    {
        const auto &serviceReg = static_cast<const AvahiServiceRegistration &>(*kv.second);
```

nit. Feel free to leave it as is.

src/mdns/mdns_avahi.cpp
```diff
@@ -843,9 +767,38 @@ otbrError PublisherAvahi::TxtListToAvahiStringList(const TxtList &   aTxtList,
...
+    return result;
+}
+
+Publisher::HostRegistrationPtr PublisherAvahi::FindHostRegistration(const AvahiEntryGroup *aEntryGroup) const
+{
+    HostRegistrationPtr result = nullptr;
+
+    for (auto &kv : mHostRegistrations)
+    {
+        auto &hostReg = static_cast<AvahiHostRegistration &>(*kv.second);
```
superwhd, line 793:
```suggestion
    for (const auto &kv : mHostRegistrations)
    {
        const auto &hostReg = static_cast<const AvahiHostRegistration &>(*kv.second);
```
nit. Feel free to leave it as is.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -340,6 +334,79 @@ void PublisherMDnsSd::Process(const MainloopContext &aMainloop)
...
+
+    DNSServiceRemoveRecord(mServiceRef, mRecordRef, /* flags */ 0);
+    // TODO: ?
+    // DNSRecordRefDeallocate(mRecordRef);
+
+exit:
+    return;
+}
+
+Publisher::ServiceRegistrationPtr PublisherMDnsSd::FindServiceRegistration(const DNSServiceRef &aServiceRef)
```
superwhd, line 371: Just a sudden thought: we may be able to put this method in the base class. For each derived `ServiceRegistration`, we should provide an interface for comparison. 

It may or may not be worth doing. No need to change it in this PR.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -361,16 +428,20 @@ void PublisherMDnsSd::HandleServiceRegisterResult(DNSServiceRef         aService
...
+    std::string            originalInstanceName;
+    ServiceRegistrationPtr serviceReg = FindServiceRegistration(aServiceRef);
+
+    VerifyOrExit(serviceReg != nullptr);
 
-    VerifyOrExit(service != mServices.end());
+    if (type.back() == '.')
+    {
+        type.pop_back();
+    }
```
superwhd, line 440: I wonder why we didn't need this before?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -361,16 +428,20 @@ void PublisherMDnsSd::HandleServiceRegisterResult(DNSServiceRef         aService
...
 
-    // mDNSResponder could auto-rename the service instance name when name conflict
-    // is detected. In this case, `aName` may not match `service->mName` and we
-    // should use the original `service->mName` to find associated SRP service.
-    originalInstanceName = service->mName;
+    // mDNSResponder could auto-rename the service instance
+    // name when name conflict is detected.
```
superwhd, line 443: Is it still true? I saw you've turned on `kDNSServiceFlagsNoAutoRename`.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -382,310 +453,194 @@ void PublisherMDnsSd::HandleServiceRegisterResult(DNSServiceRef         aService
...
 }
 
-std::string PublisherMDnsSd::MakeFullName(const std::string &aName)
-{
-    return aName + "." + kDomain;
-}
-
 // See `regtype` parameter of the DNSServiceRegister() function for more information.
-std::string PublisherMDnsSd::MakeRegType(const std::string &aType, const SubTypeList &aSubTypeList)
+std::string PublisherMDnsSd::MakeRegType(const std::string &aType, SubTypeList aSubTypeList)
```
superwhd, line 638: nit: Maybe this can be a method in the base class.

src/border_agent/border_agent.hpp
```diff
@@ -140,25 +140,23 @@ class BorderAgent : private NonCopyable
...
 #if OTBR_ENABLE_DBUS_SERVER
     void HandleUpdateVendorMeshCoPTxtEntries(std::map<std::string, std::vector<uint8_t>> aUpdate);
 #endif
 
     void HandleThreadStateChanged(otChangedFlags aFlags);
 
-    bool IsThreadStarted(void) const;
-    bool IsPskcInitialized(void) const;
+    bool        IsThreadStarted(void) const;
+    std::string getAlternativeServiceInstanceName() const;
```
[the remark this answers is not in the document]
wgtdkp: If we want to make it a common utility, we need to specify current name as parameters:
```c++
std::string GetAlternativeServiceInstanceName(const std::string &aCurrentName) const;
```
And it will not be straightforward for us to create a new name based on current name, for example, we need to consider if current name is ready renamed.

That said, it will be much easier to accept a `aOriginalName` parameter name but that doesn't make sense and the user need to save the original name for this purpose.

This method can be simple if we keep it private to `BorderAgent` and we can think about making it public accessible when there is another usage. Thoughts?

src/mdns/mdns.cpp
```diff
@@ -159,6 +162,160 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+    }
+    else
+    {
+        // If the same service is being registered with the same parameters,
+        // let's join the waiting queue for the result.
+        hostReg->mCallback = std::bind(
+            [](std::shared_ptr<ResultCallback> aExistingCallback, std::shared_ptr<ResultCallback> aNewCallback,
+               otbrError aError) {
+                std::move (*aExistingCallback)(aError);
+                std::move (*aNewCallback)(aError);
```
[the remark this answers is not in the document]
wgtdkp: Somehow, the formatting tool wants a space there :)

src/mdns/mdns.cpp
```diff
@@ -159,6 +162,160 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+    }
+    else
+    {
+        // If the same service is being registered with the same parameters,
+        // let's join the waiting queue for the result.
+        serviceReg->mCallback = std::bind(
+            [](std::shared_ptr<ResultCallback> aExistingCallback, std::shared_ptr<ResultCallback> aNewCallback,
+               otbrError aError) {
+                std::move (*aExistingCallback)(aError);
+                std::move (*aNewCallback)(aError);
```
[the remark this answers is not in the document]
wgtdkp: ditto.

src/mdns/mdns.cpp
```diff
@@ -57,7 +60,7 @@ bool Publisher::IsServiceTypeEqual(std::string aFirstType, std::string aSecondTy
     return aFirstType == aSecondType;
 }
 
-otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, std::vector<uint8_t> &aTxtData)
+otbrError Publisher::EncodeTxtData(TxtList aTxtList, std::vector<uint8_t> &aTxtData)
```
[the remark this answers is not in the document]
wgtdkp: Reverted, I think it's left after refactoring.

src/mdns/mdns_avahi.cpp
```diff
@@ -601,207 +584,148 @@ void PublisherAvahi::HandleClientState(AvahiClient *aClient, AvahiClientState aS
...
-    {
-        SuccessOrExit(error = ResetGroup(serviceIt->mGroup));
+        // Make sure that the host has been published.
+        VerifyOrExit(hostReg != nullptr, error = OTBR_ERROR_INVALID_ARGS);
+        fullHostName = MakeFullHostName(aHostName);
     }
-    else
+
+    aCallback = HandleDuplicateServiceRegistration(aHostName, aName, aType, sortedSubTypeList, aPort, sortedTxtList,
+                                                   std::move(aCallback));
```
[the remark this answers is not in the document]
wgtdkp: It will introduce more changes to current inheritances structure and I propose we consider it in future smaller changes. Thoughts?

src/mdns/mdns_mdnssd.cpp
wgtdkp: Because we were using `IsServiceTypeEqual` to handle this case but now we are removing this method in this PR (I didn't remove it before your comment) as it's that useful if we can handle the mdnssd case explicitly.

src/mdns/mdns_mdnssd.cpp
wgtdkp: Good catch! I cleaned up this logic.

src/mdns/mdns_mdnssd.cpp
wgtdkp: The `regType` format (separate by ',') is specific to mDNSResponder.

src/border_agent/border_agent.cpp
```diff
@@ -349,23 +362,16 @@ bool BorderAgent::IsThreadStarted(void) const
...
-            break;
-        }
-    }
-
-    return initialized;
+    std::random_device                      r;
+    std::default_random_engine              engine(r());
+    std::uniform_int_distribution<uint16_t> uniform_dist(1, 0xFFFF);
+    uint16_t                                rand = uniform_dist(engine);
+    char                                    nameBuffer[sizeof(kBorderAgentServiceInstanceName) + 9];
```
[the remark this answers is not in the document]
wgtdkp: Done!

src/mdns/mdns.hpp
```diff
@@ -417,23 +362,125 @@ class Publisher : private NonCopyable
...
+                            uint16_t         aPort,
+                            TxtList          aTxtList,
+                            ResultCallback &&aCallback)
+            : Registration(std::move(aCallback))
+            , mHostName(std::move(aHostName))
+            , mName(std::move(aName))
+            , mType(std::move(aType))
+            , mSubTypeList(SortSubTypeList(aSubTypeList))
+            , mPort(aPort)
+            , mTxtList(SortTxtList(aTxtList))
```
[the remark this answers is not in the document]
wgtdkp: Done!

src/mdns/mdns.hpp
```diff
@@ -417,23 +362,125 @@ class Publisher : private NonCopyable
...
+                            std::string      aType,
+                            SubTypeList      aSubTypeList,
+                            uint16_t         aPort,
+                            TxtList          aTxtList,
+                            ResultCallback &&aCallback)
+            : Registration(std::move(aCallback))
+            , mHostName(std::move(aHostName))
+            , mName(std::move(aName))
+            , mType(std::move(aType))
+            , mSubTypeList(SortSubTypeList(aSubTypeList))
```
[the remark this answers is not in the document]
wgtdkp: Done!

src/mdns/mdns_avahi.hpp
```diff
@@ -211,25 +98,46 @@ class PublisherAvahi : public Publisher
...
-        SubTypeList      mSubTypeList;
-        std::string      mHostName;
-        uint16_t         mPort  = 0;
-        AvahiEntryGroup *mGroup = nullptr;
-        TxtList          mTxtList;
-    };
+    public:
+        AvahiServiceRegistration(const std::string &aHostName,
+                                 const std::string &aName,
+                                 const std::string  aType,
```
superwhd, line 106:
```suggestion
                                 const std::string &aType,
```

src/mdns/mdns_mdnssd.hpp
```diff
@@ -96,20 +93,50 @@ class PublisherMDnsSd : public MainloopProcessor, public Publisher
...
     {
-        std::string   mName;
-        std::string   mType;
-        std::string   mRegType; ///< Service type with optional subtypes separated by commas
-        DNSServiceRef mService;
-        uint16_t      mPort;
+    public:
+        DnssdServiceRegistration(const std::string &aHostName,
+                                 const std::string &aName,
+                                 const std::string  aType,
```
superwhd, line 101:
```suggestion
                                 const std::string &aType,
```

src/mdns/mdns.cpp
```diff
@@ -159,6 +147,160 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+Publisher::HostRegistrationPtr Publisher::FindHostRegistration(const std::string &aName)
+{
+    return mHostRegistrations[MakeFullHostName(aName)];
+}
+
+Publisher::Registration::~Registration(void)
+{
+    if (!mCallback.IsNull())
+    {
+        std::move(mCallback)(OTBR_ERROR_ABORTED);
```
superwhd, line 284:
```suggestion
        Complete(OTBR_ERROR_ABORTED);
```

src/mdns/mdns.cpp
```diff
@@ -159,6 +147,160 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+void Publisher::RemoveServiceRegistration(const std::string &aName, const std::string &aType)
+{
+    otbrLogInfo("Removing service %s.%s", aName.c_str(), aType.c_str());
+    mServiceRegistrations.erase(MakeFullServiceName(aName, aType));
+}
+
+Publisher::ServiceRegistrationPtr Publisher::FindServiceRegistration(const std::string &aName, const std::string &aType)
+{
+    return mServiceRegistrations[MakeFullServiceName(aName, aType)];
+}
```
superwhd, line 187: In all such `Find*` methods, Maybe we should add a `const` qualifier for this method. Also, we may use `find` to avoid creating dummy entries in the map.

src/mdns/mdns_avahi.cpp
```diff
@@ -295,24 +380,37 @@ void Poller::Process(const MainloopContext &aMainloop)
     }
 }
 
-PublisherAvahi::PublisherAvahi(StateHandler aHandler, void *aContext)
+PublisherAvahi::PublisherAvahi(StateCallback aStateCallback)
     : mClient(nullptr)
+    , mPoller(std::unique_ptr<AvahiPoller>(new AvahiPoller()))
```
superwhd, line 385: nit: We have `MakeUnique` in `common/code_utils.hpp`.

src/mdns/mdns.cpp
```diff
@@ -159,6 +147,160 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+               otbrError aError) {
+                std::move (*aExistingCallback)(aError);
+                std::move (*aNewCallback)(aError);
+            },
+            std::make_shared<ResultCallback>(std::move(serviceReg->mCallback)),
+            std::make_shared<ResultCallback>(std::move(aCallback)), std::placeholders::_1);
+    }
+
+exit:
+    return std::move(aCallback);
```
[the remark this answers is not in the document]
wgtdkp: `aCallback` is passed in parameter list and it's move-only. It gives compiling error if I change it to `return aCallback;`:
```
../src/mdns/mdns.cpp:226:12: fatal error: local variable 'aCallback' will be copied despite being returned by name [-Wreturn-std-move]
    return aCallback;
           ^~~~~~~~~
../src/mdns/mdns.cpp:226:12: note: call 'std::move' explicitly to avoid copying
    return aCallback;
           ^~~~~~~~~
           std::move(aCallback)
1 error generated.
```

src/mdns/mdns.cpp
```diff
@@ -159,6 +147,160 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+               otbrError aError) {
+                std::move (*aExistingCallback)(aError);
+                std::move (*aNewCallback)(aError);
+            },
+            std::make_shared<ResultCallback>(std::move(hostReg->mCallback)),
+            std::make_shared<ResultCallback>(std::move(aCallback)), std::placeholders::_1);
+    }
+
+exit:
+    return std::move(aCallback);
```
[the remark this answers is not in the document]
wgtdkp: ditto

src/mdns/mdns.cpp
wgtdkp: We are returning a non-const pointer to an element in `mServiceRegistrations`. Even if we can add a `const` qual, will it be misleading?

src/mdns/mdns_mdnssd.cpp
wgtdkp: We don't want to expose `DNSServiceRef` to the base class `ServiceRegistration`.

superwhd (approved):
LGTM :+1:

src/mdns/mdns_mdnssd.cpp
```diff
@@ -361,331 +428,203 @@ void PublisherMDnsSd::HandleServiceRegisterResult(DNSServiceRef         aService
...
 {
     std::string regType = aType;
 
+    std::sort(aSubTypeList.begin(), aSubTypeList.end());
```
simonlingoogle, line 626: Isn't aSubTypeList already sorted?

src/mdns/mdns.cpp
```diff
@@ -159,6 +147,164 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+}
+
+Publisher::ServiceRegistrationPtr Publisher::FindServiceRegistration(const std::string &aName, const std::string &aType)
+{
+    auto it = mServiceRegistrations.find(MakeFullServiceName(aName, aType));
+
+    return it != mServiceRegistrations.end() ? it->second : nullptr;
+}
+
+Publisher::ResultCallback Publisher::HandleDuplicateServiceRegistration(const std::string &aHostName,
```
simonlingoogle, line 191:
```suggestion
Publisher::ResultCallback Publisher::CheckDuplicateServiceRegistration(const std::string &aHostName,
```
nit

src/mdns/mdns.cpp
```diff
@@ -159,6 +147,164 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+               otbrError aError) {
+                std::move (*aExistingCallback)(aError);
+                std::move (*aNewCallback)(aError);
+            },
+            std::make_shared<ResultCallback>(std::move(serviceReg->mCallback)),
+            std::make_shared<ResultCallback>(std::move(aCallback)), std::placeholders::_1);
+    }
+
+exit:
+    return std::move(aCallback);
```
simonlingoogle, line 228: can we `std::move(aCallback)` twice if it `serviceReg->IsCompleted()` ?

src/mdns/mdns.cpp
```diff
@@ -159,6 +147,164 @@ void Publisher::OnHostResolved(const std::string &aHostName, const Publisher::Di
...
+    {
+        // Returns success if the same service has already been
+        // registered with exactly the same parameters.
+        std::move(aCallback)(OTBR_ERROR_NONE);
+    }
+    else
+    {
+        // If the same service is being registered with the same parameters,
+        // let's join the waiting queue for the result.
+        serviceReg->mCallback = std::bind(
```
simonlingoogle, line 217: Instead of using std::bind, can we simply use a container like vector<ResultCallback> too allow multiple callbacks?

src/mdns/mdns.cpp
simonlingoogle, line 228: just return `true` if the callback is completed and no further action is required?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -361,331 +428,203 @@ void PublisherMDnsSd::HandleServiceRegisterResult(DNSServiceRef         aService
...
-
-        RecordHost(aName, aAddress, host->mRecord);
-        if (mHostHandler != nullptr)
-        {
-            mHostHandler(aName, DNSErrorToOtbrError(error), mHostHandlerContext);
-        }
-    }
-    else
+    aCallback = HandleDuplicateHostRegistration(aName, aAddress, std::move(aCallback));
+    if (aCallback.IsNull())
```
simonlingoogle, line 539:
```suggestion
    VerifyOrExit(!aCallback.IsNull());
```

src/border_agent/border_agent.cpp
```diff
@@ -294,15 +293,30 @@ void BorderAgent::PublishMeshCopService(void)
     }
 #endif
 
-    mPublisher->PublishService(/* aHostName */ "", port, kBorderAgentServiceInstanceName, kBorderAgentServiceType,
-                               Mdns::Publisher::SubTypeList{}, txtList);
+    mPublisher->PublishService(/* aHostName */ "", mServiceInstanceName, kBorderAgentServiceType,
+                               Mdns::Publisher::SubTypeList{}, port, txtList, [this](otbrError aError) {
+                                   otbrLogResult(aError, "Result of publish meshcop service %s.%s.local",
+                                                 mServiceInstanceName.c_str(), kBorderAgentServiceType);
+                                   if (aError == OTBR_ERROR_DUPLICATED)
```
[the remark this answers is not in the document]
wgtdkp: All error values other than `OTBR_ERROR_NONE` indicate a failure. I documented this on the `PublishService` and `PublishHost` API declarations.

src/border_agent/border_agent.cpp
```diff
@@ -349,23 +363,16 @@ bool BorderAgent::IsThreadStarted(void) const
...
-    }
-
-    return initialized;
+    std::random_device                      r;
+    std::default_random_engine              engine(r());
+    std::uniform_int_distribution<uint16_t> uniform_dist(1, 0xFFFF);
+    uint16_t                                rand = uniform_dist(engine);
+    std::stringstream                       ss;
+
+    ss << kBorderAgentServiceInstanceName << " (" << rand << ")";
```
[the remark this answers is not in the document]
wgtdkp: Tend to use decimals because it's more human readable.

src/mdns/mdns_mdnssd.cpp
wgtdkp: We don't have the hard requirement on the parameter `aSubTypeList` that it must be sorted.

src/mdns/mdns.cpp
wgtdkp: `Check` usually means a method doesn't change things, but this is not the fact for `HandleDuplicateServiceRegistration`.

src/mdns/mdns.cpp
wgtdkp: Yes we can, only that we are moving a null Callback for the second `std::move` here.

wgtdkp: I think it won't work, because `aCallback` is moved into method.

src/mdns/mdns_mdnssd.cpp
wgtdkp: updated.

src/mdns/mdns.cpp
wgtdkp: It should work, but chaining callbacks with lambdas is just simpler and straightforward (it's also a little more code to handle a vector of callbacks :)

simonlingoogle (approved):
LGTM :+1: 

Thanks

abtink (approved):
Looks great. Thanks. 👍

tests/mdns/main.cpp
```diff
@@ -176,87 +174,77 @@ void PublishMultipleServices(void *aContext, Mdns::Publisher::State aState)
...
-            subTypeList.back() = "_SUBTYPE3";
-        }
+    Mdns::Publisher::SubTypeList subTypeList{"_subtype1", "_SUBTYPE2"};
 
-        error = sContext.mPublisher->PublishService("", 12345, "ServiceWithSubTypes", "_meshcop._udp.", subTypeList,
-                                                    Mdns::Publisher::TxtList{});
-        assert(error == OTBR_ERROR_NONE);
-    }
+    // if (sContext.mUpdate)
+    // {
```
jwhui, line 232: nit: remove commented code?

wgtdkp: Thanks, resolved!

jwhui (approved):

PR #1117, opened 2021-11-28
[mdns] support multiple subscribers

simonlingoogle:
Mdns::Publisher used to allow only one subscriber to add subscription callbacks via `SetSubscriptionCallbacks`. 

This PR enhances Mdns::Publisher to allow multiple subscribers to add and remove subscription callbacks. 

Hint: `Discovery Proxy` and `TREL DNSSD` will both need to add subscription callbacks.

### DIFF
--- a/src/mdns/mdns.hpp
+++ b/src/mdns/mdns.hpp
@@ -35,6 +35,7 @@
 #define OTBR_AGENT_MDNS_HPP_
 
 #include <functional>
+#include <map>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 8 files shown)


src/mdns/mdns.hpp
```diff
@@ -415,14 +422,18 @@ class Publisher
...
     void *                mServiceHandlerContext = nullptr;
 
     PublishHostHandler mHostHandler        = nullptr;
     void *             mHostHandlerContext = nullptr;
 
-    DiscoveredServiceInstanceCallback mDiscoveredServiceInstanceCallback = nullptr;
-    DiscoveredHostCallback            mDiscoveredHostCallback            = nullptr;
+    uint64_t mNextSubscriberId = 1;
+
+    std::map<uint64_t, std::pair<DiscoveredServiceInstanceCallback, DiscoveredHostCallback>> mDiscoveredCallbacks;
```
superwhd, line 436: nit: Make a struct to hold `DiscoveredServiceInstanceCallback` and `DiscoveredHostCallback` so it can be more expressive.

superwhd (commented):
`std::list` should also work in this use case. Although `std::map` may be a more popular choice.

superwhd (commented):
`std::list` should also work in this use case. Although `std::map` may be a more popular choice.

superwhd (approved):
LGTM overall.

src/mdns/mdns.hpp
simonlingoogle: We may have a major refactoring for mdns modules. 
We can do it when that happens.

jwhui (approved):

PR #1120, opened 2021-11-29
[mdns-avahi] support continuous browsing

simonlingoogle:
The `SubscribeService` interface of `Mdns::Publisher` was designed to be continuously discovering services. 

However, it's not supported by Publisher over Avahi yet because `Discovery Proxy` only needs one-shot browsing. 

This commit enhances Publisher over Avahi  to support continuous browsing.

Based on #1117 and #1119

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -883,30 +883,9 @@ void PublisherAvahi::UnsubscribeService(const std::string &aType, const std::str
                 mSubscribedServices.size());
 }
 
-void PublisherAvahi::OnServiceResolved(ServiceSubscription &aService)
-{
-    otbrLogInfo("Service %s is resolved successfully: %s host %s addresses %zu", aService.mType.c_str(),
-                aService.mInstanceInfo.mName.c_str(), aService.mInstanceInfo.mHostName.c_str(),
-                aService.mInstanceInfo.mAddresses.size());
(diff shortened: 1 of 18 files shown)


src/mdns/mdns_avahi.cpp
```diff
@@ -1077,33 +1091,43 @@ void PublisherAvahi::ServiceSubscription::HandleResolveResult(AvahiServiceResolv
...
+{
+    assert(aServiceResolver != nullptr);
+    mServiceResolvers.insert(aServiceResolver);
+}
+
+void PublisherAvahi::ServiceSubscription::RemoveServiceResolver(AvahiServiceResolver *aServiceResolver)
+{
+    assert(aServiceResolver != nullptr);
+    avahi_service_resolver_free(aServiceResolver);
+    mServiceResolvers.erase(aServiceResolver);
```
superwhd, line 1130: I'd suggest to move this line before `avahi_service_resolver_free`. There is no bug with current code, but if `avahi_service_resolver_free` uses `T *&` as the parameter type this is dangerous to consider `aServiceResolver` as  valid after freeing it.

superwhd (approved):
LGTM :+1:

src/mdns/mdns_avahi.cpp
simonlingoogle: SGTM.

jwhui (approved):

PR #1121, opened 2021-11-29
[mdns-avahi] fix potential crash on browsing

simonlingoogle:
This commit fixes a bug that `aName` is `nullptr` when `aEvent` is not `AVAHI_BROWSER_NEW`. So, we should only use 
`aName` on `AVAHI_BROWSER_NEW`.


wgtdkp (approved):
LGTM :+1:

superwhd (approved):
LGTM :+1:

jwhui (approved):

PR #1122, opened 2021-11-29
[mdns-mdnssd] support continuous browsing

simonlingoogle:


src/mdns/mdns_mdnssd.cpp
```diff
@@ -883,57 +895,73 @@ void PublisherMDnsSd::ServiceSubscription::HandleBrowseResult(DNSServiceRef
...
+{
+    mResolvingInstances.emplace_back(*this, aInstanceName, aType, aDomain, aInterfaceIndex);
+
+    mResolvingInstances.back().Resolve();
+}
+
+void PublisherMDnsSd::ServiceSubscription::RemoveInstanceResolution(
+    PublisherMDnsSd::ServiceInstanceResolution &aInstanceResolution)
+{
+    std::remove_if(
```
superwhd, line 921: `std::remove_if` doesn't really erase elements from the container. Instead, it moves removed elements to the end of the container.  You may want to `a.erase(std::remove_if(a.begin(), a.end(), ...), a.end())`.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -830,12 +842,12 @@ void Publisher::Destroy(Publisher *aPublisher)
     delete static_cast<PublisherMDnsSd *>(aPublisher);
 }
 
-void PublisherMDnsSd::Subscription::Release(void)
+void PublisherMDnsSd::ServiceRef::Release(void)
```
superwhd, line 845: Looks like `Release` and `DeallocateServiceRef` are doing the same thing. Is it necessary to keep both?

src/mdns/mdns_mdnssd.cpp
superwhd: Looks like the predicate of `std::remove_if` is depending on the index of the target element. In that case I think it's safer to use `std::vector::erase` and `std::vector::find_if`. `std::remove_if` **might** cause issues if it changes elements' indexes during the look up.

src/mdns/mdns_mdnssd.cpp
simonlingoogle: May not necessary. But it's beyond the scope of this PR.

We can refactor it later.

src/mdns/mdns_mdnssd.cpp
simonlingoogle: Fixed.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1019,22 +1074,37 @@ void PublisherMDnsSd::ServiceSubscription::HandleGetAddrInfoResult(DNSServiceRef
 
     address.CopyFrom(*reinterpret_cast<const struct sockaddr_in6 *>(aAddress));
     VerifyOrExit(!address.IsUnspecified() && !address.IsLinkLocal() && !address.IsMulticast() && !address.IsLoopback(),
-                 otbrLogDebug("DNSServiceGetAddrInfo ignores address %s", address.ToString().c_str()));
+                 otbrLogDebug("DNSServiceGetAddrInfoq ignores address %s", address.ToString().c_str()));
```
superwhd, line 1077:
```suggestion
                 otbrLogDebug("DNSServiceGetAddrInfo ignores address %s", address.ToString().c_str()));
```

src/mdns/mdns_mdnssd.cpp
```diff
@@ -333,42 +319,24 @@ void PublisherMDnsSd::Process(const MainloopContext &aMainloop)
         }
     }
 
-    for (auto &subscription : mSubscribedServices)
+    for (DNSServiceRef serviceRef : readyServices)
```
superwhd, line 322:
```suggestion
    for (const DNSServiceRef &serviceRef : readyServices)
```

simonlingoogle: Actually this is not change of this PR.

Since `DNSServiceRef` is already a pointer, no need to use `const &`.

superwhd (approved):
LGTM overall. 

Just felt a little bit hard to understand the newly introduced classes `ServiceRef` and `ServiceInstanceResolution`. Maybe we can refactor them in the future so they will be more intuitive.

simonlingoogle:
> LGTM overall.
> 
> Just felt a little bit hard to understand the newly introduced classes `ServiceRef` and `ServiceInstanceResolution`. Maybe we can refactor them in the future so they will be more intuitive.

`ServiceRef` is a renaming of the original class `Subscription` with no significant logic change. 

ServiceInstanceResolution is introduced to manage one service instance resolving. We used to do service resolution in ServiceSubscription object directly because `ServiceSubscription` only handles one resolved service. However, now that we are resolving multiple services continuously, each ServiceSubscription would have a list of ongoing `ServiceInstanceResolution` objects to continuously resolve multiple service instances.

jwhui (approved):

PR #1129, opened 2021-12-02
[mdns-avahi] use assert to make sure Service Resolver is valid

simonlingoogle:
Background:
We used to use `assert` in `HandleResolveResult` to make sure the Avahi resolver is a valid resolver, which make sense because only valid resolvers can have callbacks. But there was a crash issue with the original code so we changed to use `VerifyOrExit`. Turn out the actual issue is that we should call `RemoveServiceResolver` as early as possible so that the resolver won't be freed twice.


superwhd (approved):
LGTM :+1:

jwhui (approved):

PR #1131, opened 2021-12-02
[mdns-avahi] fix potential invalid memory access

simonlingoogle:
We are passing a pointer to the `ServiceSubscription` to Avahi methods like `avahi_service_browser_new` and `avahi_service_resolver_new` as the userdata. 
However, since `ServiceSubscription` objects were directly stored in a `vector<ServiceSubscription>`, any update to the vector causes the `ServiceSubscription` objects to move, leading to potential invalid pointer access. 

This commit fixes this issue by storing `ServiceSubscription` pointers in the vector. 

This commit also fixes the bug that the `ServiceSubscription` and `HostSubscription` objects may be freed in resolved callbacks.

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -851,18 +851,20 @@ std::string PublisherAvahi::MakeFullName(const std::string &aName)
 
 void PublisherAvahi::SubscribeService(const std::string &aType, const std::string &aInstanceName)
 {
-    mSubscribedServices.emplace_back(*this, aType, aInstanceName);
+    auto service = MakeUnique<ServiceSubscription>(*this, aType, aInstanceName);
+
+    mSubscribedServices.push_back(std::move(service));
 
     otbrLogInfo("subscribe service %s.%s (total %zu)", aInstanceName.c_str(), aType.c_str(),
                 mSubscribedServices.size());
 
     if (aInstanceName.empty())
     {
-        mSubscribedServices.back().Browse();
+        mSubscribedServices.back()->Browse();
     }
     else
     {
-        mSubscribedServices.back().Resolve(AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, aInstanceName, aType);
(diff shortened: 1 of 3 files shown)


superwhd (commented):
Is there an example how the old code would cause invalid memory access?

simonlingoogle:
> Is there an example how the old code would cause invalid memory access?

It's easy to comprehend in mind, isn't?

simonlingoogle:
> Is there an example how the old code would cause invalid memory access?

Every time we remove an element from the vector<ServiceSubscription>, which happens in `UnsubscribeService`, will cause other ServiceSubscription pointers to be invalid. 

Using unique_ptr<ServiceSubscription> makes sure the pointer to a ServiceSubscription never change during its lifetime.

src/mdns/mdns_avahi.cpp
```diff
@@ -851,33 +851,39 @@ std::string PublisherAvahi::MakeFullName(const std::string &aName)
...
 
     assert(it != mSubscribedServices.end());
 
-    it->Release();
-    mSubscribedServices.erase(it);
+    {
+        std::unique_ptr<ServiceSubscription> service = std::move(*it);
+
+        mSubscribedServices.erase(it);
+        service->Release();
```
superwhd, line 885: nit: I think you can directly do following:

```
(*it)->Release();
mSubscribedServices.erase(it);
```

src/mdns/mdns_avahi.cpp
```diff
@@ -895,23 +901,29 @@ void PublisherAvahi::OnHostResolveFailed(const HostSubscription &aHost, int aErr
...
 
     assert(it != mSubscribedHosts.end());
 
-    it->Release();
-    mSubscribedHosts.erase(it);
+    {
+        std::unique_ptr<HostSubscription> host = std::move(*it);
+
+        mSubscribedHosts.erase(it);
+        host->Release();
```
superwhd, line 925: ditto

superwhd (approved):
LGTM :+1:

src/mdns/mdns_avahi.cpp
simonlingoogle: The current code should be fine. Prefer to leave it as is.

jwhui (approved):

PR #1134, opened 2021-12-03
[mdns-mdnssd] fix potential invalid memory access

simonlingoogle:
Please only review the last commit @superwhd @jwhui 

Based on #1133  and #1131

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -1092,12 +1092,6 @@ exit:
 
         mMDnsSd->OnHostResolveFailed(*this, aErrorCode);
     }
-    else if (mHostInfo.mAddresses.empty() && (aFlags & kDNSServiceFlagsMoreComing) == 0)
-    {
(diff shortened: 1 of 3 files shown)


src/mdns/mdns_mdnssd.cpp
```diff
@@ -284,23 +284,23 @@ void PublisherMDnsSd::Update(MainloopContext &aMainloop)
         aMainloop.mMaxFd = std::max(aMainloop.mMaxFd, fd);
     }
 
-    for (Subscription &subscription : mSubscribedServices)
+    for (auto &subscription : mSubscribedServices)
```
superwhd, line 287: nit: here and everywhere below, you may use `const auto &`.

superwhd (approved):
LGTM :+1:

src/mdns/mdns_mdnssd.cpp
simonlingoogle: I am doing that in #1122

I prefer to keep this PR focus on the memaccess fixing

jwhui (approved):

PR #1135, opened 2021-12-03
[meshcop] enforce vo entry when vendor specific entry is set

superwhd:
According to the spec, we need to include the vo entry when the there is any other vendor specific entry present.


superwhd:
According to offline discussion with @wgtdkp, this check potentially brings more difficulty to those who don't currently have an OUI. Therefore we decided not to include this check.

PR #1140, opened 2021-12-06
[mdns] notify service resolved with interface index

simonlingoogle:
Required by TREL DNS-SD #1109


superwhd (approved):
LGTM :+1:

superwhd (commented):
I found that `OnServiceRemoved` depends on `OnServiceResolved`. In the future, it may be better to refactor the `OnServiceResolved` and `OnServiceRemoved`:

- When the resolved service is a new service, call `OnNewServiceResolved`.
- When the resolved service is a removed service, call `OnRemovedServiceResolved`.
- Both `OnNewServiceResolved` and `OnRemovedServiceResolved` calls `OnServiceResolved`.

In this way the dependency is more clear.

jwhui (approved):

PR #1142, opened 2021-12-07
[dbus] fix and enhance energy scan api

Irving-cl:
Fixes a bug in energy scan api and enhance according to @simonlingoogle 's comment.

### DIFF
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -136,7 +136,7 @@ void ThreadHelper::EnergyScan(uint32_t aScanDuration, EnergyScanHandler aHandler
     otError  error             = OT_ERROR_NONE;
     uint32_t preferredChannels = otPlatRadioGetPreferredChannelMask(mInstance);
 
-    VerifyOrExit(aHandler != nullptr, error = OT_ERROR_BUSY);
+    VerifyOrExit(aHandler != nullptr, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aScanDuration < UINT16_MAX, error = OT_ERROR_INVALID_ARGS);
     mEnergyScanHandler = aHandler;
     mEnergyScanResults.clear();
@@ -147,7 +147,7 @@ void ThreadHelper::EnergyScan(uint32_t aScanDuration, EnergyScanHandler aHandler
 exit:
     if (error != OT_ERROR_NONE)
     {
-        if (aHandler)
+        if (mEnergyScanHandler)
         {
             mEnergyScanHandler(error, {});
         }
@@ -199,10 +199,10 @@ void ThreadHelper::EnergyScanCallback(otEnergyScanResult *aResult)
 {
     if (aResult == nullptr)
     {
-        if (mEnergyScanHandler != nullptr)
-        {
-            mEnergyScanHandler(OT_ERROR_NONE, mEnergyScanResults);
-        }
+        assert(mEnergyScanHandler != nullptr);
+
+        mEnergyScanHandler(OT_ERROR_NONE, mEnergyScanResults);
+        mEnergyScanHandler = nullptr;
     }
     else
     {


src/utils/thread_helper.cpp
```diff
@@ -136,7 +136,7 @@ void ThreadHelper::EnergyScan(uint32_t aScanDuration, EnergyScanHandler aHandler
     otError  error             = OT_ERROR_NONE;
     uint32_t preferredChannels = otPlatRadioGetPreferredChannelMask(mInstance);
 
-    VerifyOrExit(aHandler != nullptr, error = OT_ERROR_BUSY);
+    VerifyOrExit(aHandler != nullptr, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(aScanDuration < UINT16_MAX, error = OT_ERROR_INVALID_ARGS);
     mEnergyScanHandler = aHandler;
```
simonlingoogle, line 141: If `mEnergyScanHandler` was not `nullptr`, would it be a OT_ERROR_BUSY?

src/utils/thread_helper.cpp
```diff
@@ -147,7 +147,7 @@ void ThreadHelper::EnergyScan(uint32_t aScanDuration, EnergyScanHandler aHandler
 exit:
     if (error != OT_ERROR_NONE)
     {
-        if (aHandler)
+        if (mEnergyScanHandler)
```
simonlingoogle, line 150: When `error != OT_ERROR_NONE`, `mEnergyScanHandler` is likely the callback set by previous `EnergyScan`. Why are we calling the callback of previous `EnergyScan` ?

src/utils/thread_helper.cpp
```diff
@@ -199,10 +199,10 @@ void ThreadHelper::EnergyScanCallback(otEnergyScanResult *aResult)
...
     if (aResult == nullptr)
     {
-        if (mEnergyScanHandler != nullptr)
-        {
-            mEnergyScanHandler(OT_ERROR_NONE, mEnergyScanResults);
-        }
+        assert(mEnergyScanHandler != nullptr);
+
+        mEnergyScanHandler(OT_ERROR_NONE, mEnergyScanResults);
+        mEnergyScanHandler = nullptr;
```
simonlingoogle, line 205: We should be extremely careful when calling callbacks. The callback may be reset during the handler. 

```suggestion
        auto callback = mEnergyScanHandler;
        mEnergyScanHandler = nullptr;
        callback(OT_ERROR_NONE, mEnergyScanResults);
```

src/utils/thread_helper.cpp
```diff
@@ -199,10 +199,10 @@ void ThreadHelper::EnergyScanCallback(otEnergyScanResult *aResult)
 {
     if (aResult == nullptr)
     {
-        if (mEnergyScanHandler != nullptr)
-        {
-            mEnergyScanHandler(OT_ERROR_NONE, mEnergyScanResults);
-        }
+        assert(mEnergyScanHandler != nullptr);
+
+        mEnergyScanHandler(OT_ERROR_NONE, mEnergyScanResults);
```
simonlingoogle, line 204: should also clear mEnergyScanResults after callback?

src/utils/thread_helper.cpp
```diff
@@ -136,7 +136,7 @@ void ThreadHelper::EnergyScan(uint32_t aScanDuration, EnergyScanHandler aHandler
     otError  error             = OT_ERROR_NONE;
     uint32_t preferredChannels = otPlatRadioGetPreferredChannelMask(mInstance);
 
-    VerifyOrExit(aHandler != nullptr, error = OT_ERROR_BUSY);
+    VerifyOrExit(aHandler != nullptr, error = OT_ERROR_INVALID_ARGS);
```
jinran-google, line 139: We'd better add test cases covering all these error cases, like null handler, scanning again before the previous one ends, starting scan in the callback, etc.

src/utils/thread_helper.cpp
```diff
@@ -147,7 +147,7 @@ void ThreadHelper::EnergyScan(uint32_t aScanDuration, EnergyScanHandler aHandler
 exit:
     if (error != OT_ERROR_NONE)
```
simonlingoogle, line 148: Usually, we install callbacks only when the error is `OT_ERROR_NONE` , and only then the callbacks will be invoked later. 
Why are we using a different strategy here, where callbacks seem to be called no matter it's successful or not?

Issue #1149, opened 2021-12-11
"publish meshcop service when thread is not started" segfaults on start for openwrt


leojrfs:
The following commit 288fecd56081b0d5bdb7ca5e20075c9ed1e798e4 makes `otbr-agent` segfault on start.

```
[  815.232530] do_page_fault(): sending SIGSEGV to otbr-agent for invalid read access from 00000000
[  815.240027] epc = 004409dd in otbr-agent[400000+77000]
[  815.245048] ra  = 00440a35 in otbr-agent[400000+77000]
Segmentation fault
```

Reproduced on OpenWRT 21.02.1, on the following devices using multiple RCP configurations:
- raspberrypi 4 (aarch64)
- gl.inet ar150 (ath79)

It was built using the OpenWRT build instructions in this repo.

Using the `main` branch head of this repo at the time of writing (a11d76b9bee828167757a764b4b123d90baacd38) and reverting 288fecd56081b0d5bdb7ca5e20075c9ed1e798e4 works.

wgtdkp:
@leojrfs Thanks for raising this issue, I submitted https://github.com/openthread/ot-br-posix/pull/1150 as the fix. Can you help confirm if the PR works for you?

leojrfs:
I can confirm that #1150 fixes it!

Thank you

PR #1150, opened 2021-12-12
[border-agent] make border agent feature configurable

wgtdkp:
The Thread Border Agent feature is for advertising the MeshCoP
mDNS service (i.e. `_meshcop._udp`). There may be cases this
function is not supported, so making it optional.

Fixes https://github.com/openthread/ot-br-posix/issues/1149.


leojrfs:
I can confirm this PR fixes #1149.

src/agent/application.hpp
```diff
@@ -105,7 +109,13 @@ class Application
 
     static void HandleSignal(int aSignal);
 
+    Ncp::ControllerOpenThread &mPlaceHolder;
```
jwhui, line 112: What is the purpose of `mPlaceHolder`?

wgtdkp: All members/features in class `Application` is compile-time optional, we need to put something behind the member initializer list symbol (i.e. `:`) to get it compile:
```c++
Application::Application(Ncp::ControllerOpenThread &aOpenThread)
    : mPlaceHolder(aOpenThread)
#if OTBR_ENABLE_BORDER_AGENT
    , mBorderAgent(aOpenThread)
#endif
#if OTBR_ENABLE_BACKBONE_ROUTER
    , mBackboneAgent(aOpenThread)
#endif
#if OTBR_ENABLE_OPENWRT
    , mUbusAgent(aOpenThread)
#endif
#if OTBR_ENABLE_REST_SERVER
    , mRestWebServer(aOpenThread)
#endif
#if OTBR_ENABLE_DBUS_SERVER
    , mDBusAgent(aOpenThread)
#endif
#if OTBR_ENABLE_VENDOR_SERVER
    , mVendorServer(aOpenThread)
#endif
{
    OTBR_UNUSED_VARIABLE(mPlaceHolder);
}
```

Any suggestions to get it work without an additional member?


// @simonlingoogle 
// @bukepo

bukepo: Make Application into a singleton and change `sShouldTerminate` to non-static?

wgtdkp: > Make Application into a singleton and change `sShouldTerminate` to non-static?


It may not be preferred because:
1. It requires more changes;
2. Should avoid creating singleton if it doesn't improve readability or performance - singleton lifecycle management needs more attentions.

bukepo: `sShouldTerminate` needs to be access in signal handler. A singleton seems to be safer.

wgtdkp: > `sShouldTerminate` needs to be access in signal handler. A singleton seems to be safer.


`sShouldTerminate` is a static field already.

simonlingoogle (commented):
Should we make OTBR_ENABLE_BORDER_AGENT by default?

wgtdkp:
> Should we make OTBR_ENABLE_BORDER_AGENT ON by default?

Updated to enable it by default in `Android.mk`.

bukepo (approved):
fine with switch to singleton solution after this PR is merged.

jwhui (approved):

PR #1153, opened 2021-12-14
[continuous-integration] use the openthread repo's border router & backbone router workflows

superwhd:
This is a **hotfix** for the border router test cases in CI. The CI failed in `test_firewall.py` because it hasn't turned on the DUA routing.

I thought the border router CI behave the same in ot-br-posix and openthread repos but that wasn't the case.

### DIFF
--- a/.github/workflows/border_router.yml
+++ b/.github/workflows/border_router.yml
@@ -46,81 +46,177 @@ jobs:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
       if: "github.ref != 'refs/heads/main'"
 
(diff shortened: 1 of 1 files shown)


simonlingoogle (approved):
LGTM :+1:

simonlingoogle:
I think it's fine to have `DUA routing` feature in Border Router tests.
`DUA Routing` does not affect `Border Routing` and has no real effect when there is no Domain Prefix. 
So, should be fine to have both `DUA Routing` and `Border Routing`.

superwhd:
obsolete

PR #1155, opened 2021-12-14
[main] add `--no-auto-resume-thread` argument

simonlingoogle:
This commit adds `--no-auto-resume-thread` argument and use it for resetting.


simonlingoogle:
close until it's really useful

PR #1160, opened 2021-12-16
[nat64] configure nat64 prefix in Tayga

xuyirio:
This commits configures nat64 prefix /etc/tayga.conf when starting border agent. Tayga is restarted after configuration.

The _nat64 script is also updated to be compatible with prefix configuration by otbr-agent.

OTBR_TAYGA_NAT64 & OTBR_ENABLE_TAYGA_NAT64 are defined to guard the enablement of configuration.

Tests will be updated in a following PR.

### DIFF
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -42,6 +42,11 @@ endif()
 
 option(OTBR_BORDER_ROUTING "Enable Border Routing Manager" OFF)
 
+option(OTBR_TAYGA_NAT64 "Enable NAT64 support via TAYGA" OFF)
+if(OTBR_TAYGA_NAT64)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_TAYGA_NAT64=1)
+endif()
+
 option(OTBR_DBUS "Enable DBus support" OFF)
 if(OTBR_DBUS)
     pkg_check_modules(DBUS REQUIRED dbus-1)
@@ -69,7 +74,7 @@ if (OTBR_SRP_ADVERTISING_PROXY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_SRP_ADVERTISING_PROXY=1)
 endif()
 
-option(OTBR_DNSSD_DISCOVERY_PROXY   "Enable DNS-SD Discovery Proxy support" OFF)
+option(OTBR_DNSSD_DISCOVERY_PROXY  "Enable DNS-SD Discovery Proxy support" OFF)
 if (OTBR_DNSSD_DISCOVERY_PROXY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_DNSSD_DISCOVERY_PROXY=1)
 endif()
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -42,10 +42,7 @@ endif()
 
 option(OTBR_BORDER_ROUTING "Enable Border Routing Manager" OFF)
 
-option(OTBR_TAYGA_NAT64 "Enable NAT64 support via TAYGA" OFF)
-if(OTBR_TAYGA_NAT64)
-    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_TAYGA_NAT64=1)
-endif()
+option(OTBR_BORDER_ROUTING_NAT64 "Enable NAT64 support in Border Routing Manager" OFF)
 
 option(OTBR_DBUS "Enable DBus support" OFF)
 if(OTBR_DBUS)
@@ -74,7 +71,7 @@ if (OTBR_SRP_ADVERTISING_PROXY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_SRP_ADVERTISING_PROXY=1)
 endif()
 
-option(OTBR_DNSSD_DISCOVERY_PROXY  "Enable DNS-SD Discovery Proxy support" OFF)
+option(OTBR_DNSSD_DISCOVERY_PROXY   "Enable DNS-SD Discovery Proxy support" OFF)
 if (OTBR_DNSSD_DISCOVERY_PROXY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_DNSSD_DISCOVERY_PROXY=1)
 endif()
--- a/src/nat64/CMakeLists.txt
+++ b/src/nat64/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-#  Copyright (c) 2020-2021, The OpenThread Authors.
+#  Copyright (c) 2022, The OpenThread Authors.
 #  All rights reserved.
 #
 #  Redistribution and use in source and binary forms, with or without
@@ -26,38 +26,10 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
-add_subdirectory(agent)
-if(OTBR_BORDER_AGENT)
-    add_subdirectory(border_agent)
-endif()
-add_subdirectory(common)
-add_subdirectory(ncp)
-add_subdirectory(nat64)
-add_subdirectory(sdp_proxy)
-add_subdirectory(trel_dnssd)
-
-if(OTBR_DBUS)
-    add_subdirectory(dbus)
-endif()
-
-if(OTBR_MDNS)
-    add_subdirectory(mdns)
-endif()
-
-add_subdirectory(utils)
-
-if(OTBR_OPENWRT)
-    add_subdirectory(openwrt)
-endif()
-
-if(OTBR_WEB)
-    add_subdirectory(web)
-endif()
-
-if(OTBR_REST)
-    add_subdirectory(rest)
-endif()
-
-if (OTBR_BACKBONE_ROUTER)
-    add_subdirectory(backbone_router)
-endif()
+add_library(otbr-nat64
+    tayga_helper.cpp
+)
+target_link_libraries(otbr-nat64 PRIVATE
+    otbr-common
+    otbr-utils
+)
--- a/src/nat64/tayga_helper.cpp
+++ b/src/nat64/tayga_helper.cpp
@@ -1,0 +1,136 @@
+/*
+ *    Copyright (c) 2022, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#if OTBR_ENABLE_TAYGA_NAT64
+
+#include <fstream>
+
+#include <openthread/border_router.h>
+
+#include "common/code_utils.hpp"
+#include "common/types.hpp"
+#include "nat64/tayga_helper.hpp"
+#include "utils/system_utils.hpp"
+
+namespace otbr {
+namespace Tayga {
+namespace {
+
+static const char kTaygaConf[]       = "/etc/tayga.conf";
+static const char kTaygaConfTmp[]    = "/etc/tayga_tmp.conf";
+static const char kTaygaRestartCmd[] = OTBR_TAYGA_RESTART_CMD;
+static const char kTaygaStopCmd[]    = OTBR_TAYGA_STOP_CMD;
+static const char kPrefixLineStart[] = "prefix ";
+
+bool UpdatePrefix(const std::string &prefix)
+{
+    bool          is_copied  = false;
+    bool          is_updated = false;
+    std::ifstream file(kTaygaConf);
+    std::ofstream tmp_file(kTaygaConfTmp);
+    std::string   line;
+
+    VerifyOrExit(file && tmp_file);
+
+    while (std::getline(file, line))
+    {
+        if (line.find(kPrefixLineStart, 0) == 0)
+        {
+            VerifyOrExit(line.find(prefix, 0) == std::string::npos, is_updated = false);
+
+            line       = kPrefixLineStart + prefix;
+            is_updated = true;
+        }
+        tmp_file << line << std::endl;
+        VerifyOrExit(tmp_file.good(), is_updated = false);
+    }
+
+    VerifyOrExit(file.eof(), is_updated = false);
+
+    file.close();
+    tmp_file.close();
+    is_copied = true;
+
+    if (is_updated)
+    {
+        VerifyOrExit(remove(kTaygaConf) == 0 && rename(kTaygaConfTmp, kTaygaConf) == 0, is_updated = false);
+    }
+
+exit:
+    if (!is_copied)
+    {
+        file.close();
+        tmp_file.close();
+    }
+
+    if (!is_updated)
+    {
+        remove(kTaygaConfTmp);
+    }
+
+    otbrLogInfo("NAT64 prefix in Tayga configuration file %s updated", is_updated ? "is" : "isn't");
+
+    return is_updated;
+}
+
+} // namespace
+
+void ConfigTaygaNat64Prefix(otInstance *aInstance)
+{
+    otbrError   error      = OTBR_ERROR_NONE;
+    bool        is_updated = false;
+    otIp6Prefix nat64Prefix;
+    Ip6Prefix   prefix;
+
+    VerifyOrExit(otBorderRoutingGetNat64Prefix(aInstance, &nat64Prefix) == OT_ERROR_NONE,
+                 error = OTBR_ERROR_OPENTHREAD);
+
+    prefix.Set(nat64Prefix);
+    is_updated = UpdatePrefix(prefix.ToString());
+
+    VerifyOrExit(is_updated && SystemUtils::ExecuteCommand(kTaygaRestartCmd) == 0, error = OTBR_ERROR_ERRNO);
+
+exit:
+    if (error == OTBR_ERROR_OPENTHREAD)
+    {
+        otbrLogInfo("Failed to get nat64 prefix. Stopping Tayga...");
+        SystemUtils::ExecuteCommand(kTaygaStopCmd);
+    }
+    else if (error == OTBR_ERROR_ERRNO)
+    {
+        otbrLogInfo("Didn't re-configure Tayga: %s", otbrErrorString(error));
+    }
+    else
+    {
+        otbrLogInfo("Configured Tayga with NAT64 prefix: %s", prefix.ToString().c_str());
+    }
+}
+
+} // namespace Tayga
+} // namespace otbr
+#endif // OTBR_ENABLE_TAYGA_NAT64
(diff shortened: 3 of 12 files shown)


src/nat/tayga_helper.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    otError     error = OT_ERROR_NONE;
+    otIp6Prefix nat64Prefix;
+
+    error = otBorderRoutingGetNat64Prefix(aInstance, &nat64Prefix);
+    if (error == OT_ERROR_NONE)
+    {
+        Ip6Prefix prefix;
+
+        SystemUtils::ExecuteCommand("sudo sed -i '/^prefix /d' %s", kTaygaConf);
```
wgtdkp, line 50:
```suggestion
        SystemUtils::ExecuteCommand("sed -i '/^prefix /d' %s", kTaygaConf);
```

And please log if `ExecuteCommand()` fails.
The same for following lines :)

xuyirio: Updated. Thanks!

src/nat/tayga_helper.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    otbrError   error = OTBR_ERROR_NONE;
+    otIp6Prefix nat64Prefix;
+    Ip6Prefix   prefix;
+
+    VerifyOrExit(otBorderRoutingGetNat64Prefix(aInstance, &nat64Prefix) == OT_ERROR_NONE,
+                 error = OTBR_ERROR_OPENTHREAD);
+
+    prefix.Set(nat64Prefix);
+    VerifyOrExit(SystemUtils::ExecuteCommand("sed -i '/^prefix /d' %s", kTaygaConf) == 0, error = OTBR_ERROR_ERRNO);
```
bukepo, line 52: This depends on commands may not be present on the system.

src/nat/tayga_helper.cpp
```diff
@@ -0,0 +1,72 @@
...
+    VerifyOrExit(
+        SystemUtils::ExecuteCommand("sed -i '/^##prefix /a prefix '%s %s", prefix.ToString().c_str(), kTaygaConf) == 0,
+        error = OTBR_ERROR_ERRNO);
+
+    VerifyOrExit(SystemUtils::ExecuteCommand("systemctl restart tayga") == 0, error = OTBR_ERROR_ERRNO);
```
bukepo, line 57: The system may not use systemd to manage services.

src/nat/tayga_helper.cpp
bukepo: Perhaps you can use [libconfig](http://www.hyperrealm.com/libconfig/libconfig_manual.html) instead.

xuyirio: Thanks for point out. This nat64 prefix configuration will be enabled when border routing and nat64 are both supported, for which we expect to be ubuntu and raspbian as of now. And the sed command is expected to work just as in the _nat64 script.
libconfig usually works on structured conf files

bukepo: The scripts are supposed to run on ubuntu and rasbian, but the c++ binaries are not limited to these platforms.

simonlingoogle: > Perhaps you can use [libconfig](http://www.hyperrealm.com/libconfig/libconfig_manual.html) instead.

Using `libconfig` seems too much dependency for just changing one line in a file?

Maybe we can generate the whole `tayga.conf` file from scratch in C++ code without using any system command or library?

src/nat/tayga_helper.cpp
```diff
@@ -0,0 +1,72 @@
+/*
+ *    Copyright (c) 2017, The OpenThread Authors.
```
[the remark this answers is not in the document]
xuyirio: Updated

src/nat/tayga_helper.hpp
```diff
@@ -0,0 +1,52 @@
+/*
+ *    Copyright (c) 2017, The OpenThread Authors.
```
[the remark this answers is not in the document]
xuyirio: Updated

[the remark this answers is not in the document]
xuyirio: Was updated.

src/nat/tayga_helper.cpp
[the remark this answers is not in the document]
xuyirio: Updated to also try "service tayga restart" (e.g. in DOCKER).

src/nat/tayga_helper.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @file
+ *   The file implements helper functions for TAYGA.
+ */
+
+#ifndef OTBR_AGENT_TAYGA_HELPER_HPP_
+#define OTBR_AGENT_TAYGA_HELPER_HPP_
+
+#include <openthread/instance.h>
```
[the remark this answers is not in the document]
xuyirio: Looks that we are doing this. E.g. in src/utils/thread_helper.hpp

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,115 @@
...
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <fstream>
```
simonlingoogle, line 29: `#include <fstream>` should be inside `#if OTBR_ENABLE_NAT64` ?

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    bool          is_updated = false;
+    std::ifstream file(kTaygaConf);
+    std::ofstream tmp_file(kTaygaConfTmp);
+    std::string   line;
+
+    if (!file || !tmp_file)
+    {
+        otbrLogErr("Failed to open Tayga configure files");
+        return is_updated;
```
simonlingoogle, line 55: we only allow one return for each function/method.

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,115 @@
...
+    }
```
simonlingoogle, line 56:
```suggestion
    VerifyOrExit(file && tmp_file, otbrLogErr("Failed to open Tayga configure files"));
```

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    while (getline(file, line))
```
simonlingoogle, line 58: handle retval -1?

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,115 @@
...
+        if (line.find(kPrefixLineStart, 0) == 0)
+        {
+            line       = kPrefixLineStart + prefix;
+            is_updated = true;
+        }
+        tmp_file << line << std::endl;
+    }
+
+    file.close();
+    tmp_file.close();
```
simonlingoogle, line 69: need to make sure both files are closed whenever this function returns.

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    if (is_updated)
+    {
+        remove(kTaygaConf);
```
simonlingoogle, line 73: handle failure?

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,115 @@
...
+        rename(kTaygaConfTmp, kTaygaConf);
```
simonlingoogle, line 74: handle failure?

simonlingoogle, line 74: handle failure?

simonlingoogle, line 74: rename will automatically replace `kTaygaConf` if it exists. So, no need to `remove` it first?

src/nat64/tayga_helper.cpp
xuyirio: Updated.

src/nat64/tayga_helper.cpp
xuyirio: Updated.

src/nat64/tayga_helper.cpp
xuyirio: Updated.

src/nat64/tayga_helper.cpp
xuyirio: Added checking for unexpected stream state.

src/nat64/tayga_helper.cpp
xuyirio: Updated.

src/nat64/tayga_helper.cpp
xuyirio: Added.

src/nat64/tayga_helper.cpp
xuyirio: Added.

xuyirio: rename might fail for existing file name in some system. So call remove first to make it safer.

src/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ endif()
 add_subdirectory(common)
 add_subdirectory(ncp)
 add_subdirectory(sdp_proxy)
+add_subdirectory(nat)
```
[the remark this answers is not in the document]
xuyirio: Updated.

src/nat/tayga_helper.cpp
xuyirio: Updated to copy and rewrite the file.

src/nat/tayga_helper.cpp
```diff
@@ -0,0 +1,74 @@
...
+                 error = OTBR_ERROR_OPENTHREAD);
+
+    prefix.Set(nat64Prefix);
+    VerifyOrExit(SystemUtils::ExecuteCommand("sed -i '/^prefix /d' %s", kTaygaConf) == 0, error = OTBR_ERROR_ERRNO);
+    VerifyOrExit(
+        SystemUtils::ExecuteCommand("sed -i '/^##prefix /a prefix '%s %s", prefix.ToString().c_str(), kTaygaConf) == 0,
+        error = OTBR_ERROR_ERRNO);
+
+    VerifyOrExit(SystemUtils::ExecuteCommand("systemctl restart tayga") == 0 ||
+                     SystemUtils::ExecuteCommand("service tayga restart") == 0,
```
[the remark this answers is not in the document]
xuyirio: Thanks for the suggestion. Added an option OTBR_TAYGA_RESTART_CMD for configuring the command. Its default value is set to "systemctl restart tayga".

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    bool          is_copied  = false;
+    bool          is_updated = false;
+    std::ifstream file(kTaygaConf);
+    std::ofstream tmp_file(kTaygaConfTmp);
+    std::string   line;
+
+    VerifyOrExit(file && tmp_file);
+
+    while (getline(file, line))
```
simonlingoogle, line 58:
```suggestion
    while (std::getline(file, line))
```
nit: reduce confusion

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,128 @@
...
+        if (line.find(kPrefixLineStart, 0) == 0)
+        {
+            line       = kPrefixLineStart + prefix;
+            is_updated = true;
+        }
+        tmp_file << line << std::endl;
+        VerifyOrExit(tmp_file.good(), is_updated = false);
+    }
+
+    VerifyOrExit(file.eof(), is_updated = false);
```
simonlingoogle, line 69: may fail to remove tmp file if exit earlier?

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    file.close();
+    tmp_file.close();
+    is_copied = true;
+
+    if (is_updated)
+    {
+        VerifyOrExit(remove(kTaygaConf) == 0 && rename(kTaygaConfTmp, kTaygaConf) == 0, is_updated = false);
```
simonlingoogle, line 77:
```suggestion
        VerifyOrExit(rename(kTaygaConfTmp, kTaygaConf) == 0, is_updated = false);
```
`remove` not necessary?

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,128 @@
...
+namespace otbr {
+namespace Tayga {
+namespace {
+
+static const char kTaygaConf[]       = "/etc/tayga.conf";
+static const char kTaygaConfTmp[]    = "/etc/tayga_tmp.conf";
+static const char kTaygaRestartCmd[] = OTBR_TAYGA_RESTART_CMD;
+static const char kPrefixLineStart[] = "prefix ";
+
+bool UpdatePrefix(const std::string &prefix)
```
simonlingoogle, line 48:
```suggestion
otbrError UpdatePrefix(const std::string &prefix)
```
nit: we commonly return errors

src/nat64/tayga_helper.cpp
xuyirio: Updated to remove tmp file on exit

src/nat64/tayga_helper.cpp
xuyirio: `rename` might fail for exiting files in some system. So call `remove` to make it safer.

src/nat64/tayga_helper.cpp
xuyirio: updated

src/nat64/tayga_helper.cpp
xuyirio: update. thanks!

src/nat64/tayga_helper.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <openthread/instance.h>
+
+namespace otbr {
+namespace Tayga {
+
+/**
+ * This function configures the nat64 prefix for Tayga.
+ *
+ * @param aInstance A pointer to an OpenThread instance.
+ */
```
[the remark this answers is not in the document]
xuyirio: updated.

xuyirio:
@simonlingoogle @wgtdkp @bukepo @zhanglongxia Could you help take another look at this PR?
Although we are still evaluating 1) using Tayga and integrate with 1P products vs. 2) building a new stateful NAT64 service, I'd like to get the Tayga configured with the appropriate prefix first. This will help us create a prototype quickly and make testing easier. Thanks!

simonlingoogle:
> I'd like to get the Tayga configured with the appropriate prefix first. This will help us create a prototype quickly and make testing easier. Thanks!

+1. 
I think a TAYGA based approach can still be useful for open source solutions, at least in the short term.

etc/cmake/options.cmake
```diff
@@ -69,7 +74,7 @@ if (OTBR_SRP_ADVERTISING_PROXY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_SRP_ADVERTISING_PROXY=1)
 endif()
 
-option(OTBR_DNSSD_DISCOVERY_PROXY   "Enable DNS-SD Discovery Proxy support" OFF)
+option(OTBR_DNSSD_DISCOVERY_PROXY "Enable DNS-SD Discovery Proxy support" OFF)
```
simonlingoogle, line 77: nit: better avoid unrelated change.

src/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ if(OTBR_BORDER_AGENT)
 endif()
 add_subdirectory(common)
 add_subdirectory(ncp)
+add_subdirectory(nat64)
```
simonlingoogle, line 35:
```suggestion
add_subdirectory(tayga_nat64)
```

I think this directory is dedicated for tayga?

src/nat64/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+#  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+#  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+
+add_library(otbr-nat64
```
simonlingoogle, line 29:
```suggestion
add_library(otbr-tayga-nat64
```
nit

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,129 @@
...
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#if OTBR_ENABLE_TAYGA_NAT64
+#include <fstream>
```
simonlingoogle, line 30:
```suggestion
#if OTBR_ENABLE_TAYGA_NAT64

#include <fstream>
```

third_party/openthread/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@
 set(OT_BORDER_AGENT ON CACHE BOOL "enable border agent" FORCE)
 set(OT_BORDER_ROUTER ON CACHE BOOL "enable border router feature" FORCE)
 set(OT_BORDER_ROUTING ${OTBR_BORDER_ROUTING} CACHE BOOL "enable border routing feature" FORCE)
+set(OT_BORDER_ROUTING_NAT64 ${OTBR_TAYGA_NAT64}&&${OTBR_BORDER_ROUTING} CACHE BOOL "enable border routing NAT64 feature" FORCE)
```
simonlingoogle, line 32: We may still set OTBR_TAYGA_NAT64=ON when OTBR_TAYGA_NAT64 is OFF, if we are using other NAT64 implementations?

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    otbrLogInfo("NAT64 prefix in Tayga configuration file %s updated", is_updated ? "is" : "isn't");
+
+    return is_updated;
+}
+
+} // namespace
+
+void ConfigTaygaNat64Prefix(otInstance *aInstance)
```
simonlingoogle, line 101:
```suggestion
void UpdateTaygaNat64Prefix(otInstance *aInstance)
```
nit

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+    otbrError   error      = OTBR_ERROR_NONE;
+    bool        is_updated = false;
+    otIp6Prefix nat64Prefix;
+    Ip6Prefix   prefix;
+
+    VerifyOrExit(otBorderRoutingGetNat64Prefix(aInstance, &nat64Prefix) == OT_ERROR_NONE,
```
simonlingoogle, line 108: when there there is no NAT64 prefix, should we stop tayga?

src/nat64/tayga_helper.cpp
```diff
@@ -0,0 +1,129 @@
...
+                 error = OTBR_ERROR_OPENTHREAD);
+
+    prefix.Set(nat64Prefix);
```
simonlingoogle, line 111: should we avoid restarting tayga if the NAT64 prefix is not actually changed?

src/nat64/tayga_helper.cpp
simonlingoogle: > `rename` might fail for exiting files in some system. So call `remove` to make it safer.

According to [manpages](https://linux.die.net/man/3/rename), the new file is always removed if exists. Do we have any clue why it may fail for existing files in some system?

xuyirio: Based on the description of [rename()](https://www.cplusplus.com/reference/cstdio/rename/), `If newname names an existing file, the function may either fail or override the existing file, depending on the specific system and library implementation`. So I think first removing the existing file is safer?

src/nat64/tayga_helper.cpp
xuyirio: Yes. We check `is_updated` before restarting tayga.

third_party/openthread/CMakeLists.txt
xuyirio: I don't quite understand. Do you mean to set `OT_BORDER_ROUTING_NAT64=ON` if we are using any other NAT64 implementation? When we have other implementations later, we can update the condition here to reflect the change.

src/CMakeLists.txt
xuyirio: Tayga is one kind of implementation for NAT64. Other implementation can also be added into the same directory. If size of the directory becomes a concern later, we can refactor the structure accordingly.

src/nat64/tayga_helper.cpp
xuyirio: Since we might restart tayga, I think `config` is closer description?

src/nat64/tayga_helper.cpp
xuyirio: Updated to stop tayga. 

When `OT_BORDER_ROUTING_NAT64` is ON, a nat64 prefix usually exists when `ConfigTaygaNat64Prefix()` is called (routing manager is initialized). So it should be a rare case.

src/nat64/tayga_helper.cpp
simonlingoogle: Thanks for the reference. LGTM.

src/CMakeLists.txt
simonlingoogle: SGTM.

PR #1168, opened 2021-12-29
[mdns] remove redundant ERR log

simonlingoogle:


src/mdns/mdns_avahi.cpp
```diff
@@ -393,7 +393,6 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
 
     case AVAHI_ENTRY_GROUP_UNCOMMITED:
     case AVAHI_ENTRY_GROUP_REGISTERING:
-        otbrLogErr("Group ready.");
```
superwhd, line 396: Why deleting it instead of using a lower log level?

superwhd (approved):
LGTM with a question

src/mdns/mdns_avahi.cpp
simonlingoogle: There are significant number of this logs. I guess the real question is "Is the log useful in anyway?".

superwhd: Then I think it's fine to remove it.

jwhui (approved):

PR #1170, opened 2021-12-30
[dbus] force quit if DBus fails to connect

simonlingoogle:
This commit forces otbr-agent to quit if DBus fails to connect, so as to help avoid `DBusAgent` running with invalid DBus connection.


simonlingoogle:
close in favor of https://github.com/openthread/ot-br-posix/pull/1175

PR #1178, opened 2022-01-04
[mdnssd] call `DNSServiceProcessResult` at the end of `Process`

simonlingoogle:
This commit makes sure `DNSServiceProcessResult` is always called at the end of `Process`, so as to avoid potential memory issues.


superwhd (approved):
LGTM :+1:

jwhui (approved):

PR #1185, opened 2022-01-07
[certification] return ipv6 in testharness-discovery script

tomaszkob89:
Current implementation returns on Windows platforms by default IPv4 address and with this PR we will start to return IPv6 link local.

### DIFF
--- a/tools/reference_device/testharness-discovery
+++ b/tools/reference_device/testharness-discovery
@@ -53,7 +53,7 @@ def if_nametoindex(name):
 
 
 def get_ipaddr():
-    for line in os.popen(f'ip addr list dev {IFNAME} | grep inet | grep global'):
+    for line in os.popen(f'ip addr list dev {IFNAME} | grep inet6 | grep \'scope link\' '):
         addr = line.strip().split()[1]
         return addr.split('/')[0]
 


simonlingoogle:
Why use IPv6 address instead?

tomaszkob89:
> 

It is less error-prone based on our experience. We have observed issues introduced by the use of IPv4 in the tests. Besides that e.g. Kirale BR utilizes IPv6 from the beginning and we would like to comply. 

@edmont could you add something?

edmont:
Basically the IPv6 link local address remains constant across reboots but the IPv4 address might change depending on the availability of a DHCP server. We have observed cases in which the BR reports some address in the discovery mechanism to which it doesn't respond by the time the harness tries to send some commands.

edmont (approved):

simonlingoogle:
> Basically the IPv6 link local address remains constant across reboots but the IPv4 address might change depending on the availability of a DHCP server. We have observed cases in which the BR reports some address in the discovery mechanism to which it doesn't respond by the time the harness tries to send some commands.

Thanks for clarification. Advertising IPv6 address should be fine as long as Harness supports it.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):
Thanks! 👍🏼

PR #1191, opened 2022-01-12
[continuous-integration] add an individual CI step for MATN test cases

superwhd:
MATN test cases often takes long time. We'd better move them to a separate GitHub action so that we don't have to rerun MATN cases every time a border router action fails.

See also: https://github.com/openthread/openthread/pull/7308


simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #1198, opened 2022-01-14
[dbus] fix dbus reconnection

simonlingoogle:
Found that `dbus_bus_request_name` may fail on Raspberry Pi. 

This commit fixes the issue by re-connecting DBus in such case.

### DIFF
--- a/src/dbus/server/dbus_agent.cpp
+++ b/src/dbus/server/dbus_agent.cpp
@@ -49,47 +49,58 @@ DBusAgent::DBusAgent(otbr::Ncp::ControllerOpenThread &aNcp)
 
 void DBusAgent::Init(void)
 {
-    DBusError   dbusError;
-    otbrError   error = OTBR_ERROR_NONE;
-    int         requestReply;
-    std::string serverName = OTBR_DBUS_SERVER_PREFIX + mInterfaceName;
+    otbrError error = OTBR_ERROR_NONE;
 
-    dbus_error_init(&dbusError);
-    DBusConnection *conn                = nullptr;
-    auto            connection_deadline = std::chrono::steady_clock::now() + std::chrono::seconds(30);
+    auto connection_deadline = std::chrono::steady_clock::now() + std::chrono::seconds(30);
 
-    while (true)
+    mConnection = PrepareDBusConnection();
+
+    while (mConnection == nullptr && std::chrono::steady_clock::now() < connection_deadline)
     {
-        conn = dbus_bus_get(DBUS_BUS_SYSTEM, &dbusError);
-        if (conn != nullptr || std::chrono::steady_clock::now() > connection_deadline)
-        {
-            break;
-        }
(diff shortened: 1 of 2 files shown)


src/dbus/server/dbus_agent.cpp
```diff
@@ -49,47 +49,58 @@ DBusAgent::DBusAgent(otbr::Ncp::ControllerOpenThread &aNcp)
...
-    DBusError   dbusError;
-    otbrError   error = OTBR_ERROR_NONE;
-    int         requestReply;
-    std::string serverName = OTBR_DBUS_SERVER_PREFIX + mInterfaceName;
+    otbrError error = OTBR_ERROR_NONE;
 
-    dbus_error_init(&dbusError);
-    DBusConnection *conn                = nullptr;
-    auto            connection_deadline = std::chrono::steady_clock::now() + std::chrono::seconds(30);
+    auto connection_deadline = std::chrono::steady_clock::now() + std::chrono::seconds(30);
```
wgtdkp, line 54:
```suggestion
    auto connection_deadline = Clock::now() + Seconds(30);
```

consider using the `Clock` type alias in `common/time.hpp` to enforce always using `steady_clock`.
```c++
using Seconds      = std::chrono::seconds;
using Milliseconds = std::chrono::milliseconds;
using Microseconds = std::chrono::microseconds;
using Clock        = std::chrono::steady_clock;
using Timepoint    = Clock::time_point;
```

simonlingoogle: Added `using Clock        = std::chrono::steady_clock;`

But other suggestions are not accepted.

wgtdkp (approved):

jinran-google (approved):
LGTM :+1:

bukepo (approved):

src/dbus/server/dbus_agent.cpp
```diff
@@ -49,47 +51,58 @@ DBusAgent::DBusAgent(otbr::Ncp::ControllerOpenThread &aNcp)
...
-        {
-            break;
-        }
-        otbrLogWarning("Failed to get DBus connection: %s: %s, will retry after 3 seconds", dbusError.name,
-                       dbusError.message);
-        dbus_error_free(&dbusError);
-        sleep(3);
+        otbrLogWarning("Failed to setup DBus connection, will retry after 1 second");
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+        mConnection = PrepareDBusConnection();
```
jwhui, line 62: We will call `PrepareDBusConnection()` again at the top of the while-loop. Is that intended?

simonlingoogle: Good catch! Failed to remove the call below.

simonlingoogle: Another thought is that, thanks to `unique_ptr`, even such mistake won't cause DBus connection leak.

jwhui (approved):

PR #1203, opened 2022-01-19
[script] add `set -euxo pipefail` for startup scripts

superwhd:
Add `set -euxo pipefail` to startup scripts. Then shellcheck complained about some issues in the unchanged shell scripts. I fixed the issues in this PR as well.

### DIFF
--- a/script/otbr-firewall
+++ b/script/otbr-firewall
@@ -44,6 +44,8 @@ OTBR_FORWARD_INGRESS_CHAIN="OTBR_FORWARD_INGRESS"
 . /lib/lsb/init-functions
 . /lib/init/vars.sh
 
+set -euxo pipefail
+
 ipset_destroy_if_exist()
 {
     if ipset list "$1"; then
--- a/script/server
+++ b/script/server
@@ -38,6 +38,7 @@
 
 main()
 {
+    # shellcheck source=/dev/null
     . "$BEFORE_HOOK"
     sudo sysctl --system
     nat64_start || die 'Failed to start NAT64!'
@@ -60,6 +61,7 @@ main()
     else
         die 'Unable to find service manager. Try script/console to start in console mode!'
     fi
+    # shellcheck source=/dev/null
     . "$AFTER_HOOK"
 }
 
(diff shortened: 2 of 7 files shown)


superwhd:
Apparently there are some CI failures caused by the PR, which indicates possible startup script failures. I'll take a deeper look later.

simonlingoogle (approved):
LGTM :+1: 

with a nit

script/_border_routing
```diff
@@ -30,6 +30,8 @@
 #       This script sets up border routing configurations.
 #
 
+set -euxo pipefail
```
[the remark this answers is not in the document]
superwhd: I think `script/setup` is not enough, at least we also need to put `set -euxo pipefail` in service files. `script/setup` may not be called by a user after a reboot.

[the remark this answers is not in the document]
bukepo: We set this up by `. "$(dirname "$0")"/_initrc`.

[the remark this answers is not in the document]
superwhd: I see. I'll update this PR to remove unnecessary changes.

script/otbr-firewall
```diff
@@ -44,6 +44,8 @@ OTBR_FORWARD_INGRESS_CHAIN="OTBR_FORWARD_INGRESS"
 . /lib/lsb/init-functions
 . /lib/init/vars.sh
 
+set -euxo pipefail
```
bukepo, line 47: source _initrc instead?

superwhd: The script `otbr-firewall` will be installed into `/etc` and will be run as a service. In such a case I think it shouldn't rely on the source code.

bukepo (approved):

script/server
```diff
@@ -38,6 +38,7 @@
 
 main()
 {
+    # shellcheck source=/dev/null
```
jwhui, line 41: Just curious - did something change to require these exclusions?

superwhd: In a previous push, `make-pretty` complained about this (even in some unchanged files) so I made such changes. I don't know what will happen if I revert these `# shellcheck` changes.

https://github.com/openthread/ot-br-posix/runs/4862547983?check_suite_focus=true

jwhui (approved):

PR #1209, opened 2022-01-20
[DEBUG] use NOTE log level (for debug only)

simonlingoogle:


bukepo:
The docker is for development and testing, is it necessary to lower the log level here?

simonlingoogle:
> The docker is for development and testing, is it necessary to lower the log level here?

I am actually using this PR just to get the log files.

PR #1212, opened 2022-01-21
[test] use cmake to build ot-rcp

bukepo:
This commit switch to cmake when building ot-rcp during tests.


bukepo:
Depends on https://github.com/openthread/openthread/pull/7344

jwhui (approved):

PR #1214, opened 2022-01-24
[dbus] refine DBus `GetProperty` error logs

simonlingoogle:
This commit refines DBus GetProperty error logs:
- Use INFO log level if the property is valid but handler returns error (used to be WARN)

### DIFF
--- a/src/dbus/server/dbus_object.cpp
+++ b/src/dbus/server/dbus_object.cpp
@@ -134,7 +134,8 @@ void DBusObject::GetPropertyMethodHandler(DBusRequest &aRequest)
     DBusMessageIter iter;
     std::string     interfaceName;
     std::string     propertyName;
-    otError         error = OT_ERROR_NONE;
+    otError         error      = OT_ERROR_NONE;
+    otError         replyError = OT_ERROR_NONE;
 
     VerifyOrExit(reply != nullptr, error = OT_ERROR_NO_BUFS);
     VerifyOrExit(dbus_message_iter_init(aRequest.GetMessage(), &iter), error = OT_ERROR_FAILED);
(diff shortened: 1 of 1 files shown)


src/dbus/server/dbus_object.cpp
```diff
@@ -134,7 +134,8 @@ void DBusObject::GetPropertyMethodHandler(DBusRequest &aRequest)
     DBusMessageIter iter;
     std::string     interfaceName;
     std::string     propertyName;
-    otError         error = OT_ERROR_NONE;
+    otError         error      = OT_ERROR_NONE;
+    otError         replyError = OT_ERROR_NONE;
```
jinran-google, line 138: I think we don't need a separate variable here.

simonlingoogle: We need to differentiate between two cases:

- We failed when trying to call the handler, without actually calling the handler
- We called the handler, but the handler returned an OT error

We can't use error values because we don't know what errors the handler may return.

jinran-google: But the handling of 2 errors are almost identical. The only difference is the log level "info" vs "warning". Why don't make them all "warning"?

simonlingoogle: If the property path is invalid, this is clearly a logic error on the client side that should not happen. So, such case should use WARN because it's incorrect behavior. 

If the property handler replies error, like `ActiveDatasetTlvs`, it's a normal case not a logical error, so we should use INFO ?

jinran-google (approved):
LGTM :+1:

src/dbus/server/dbus_object.cpp
```diff
@@ -144,19 +145,19 @@ void DBusObject::GetPropertyMethodHandler(DBusRequest &aRequest)
         auto propertyIter = mGetPropertyHandlers.find(interfaceName);
 
         otbrLogInfo("GetProperty %s.%s", interfaceName.c_str(), propertyName.c_str());
-        VerifyOrExit(propertyIter != mGetPropertyHandlers.end(), error = OT_ERROR_NOT_FOUND);
+        VerifyOrExit(propertyIter != mGetPropertyHandlers.end(), error = OT_ERROR_INVALID_ARGS);
```
bukepo, line 148: The client wants to get a property, but the property is not found. I think the original not found is more natural. The purpose of this PR is to reduce logs, how about focusing on this first?

simonlingoogle, line 148:
```suggestion
        VerifyOrExit(propertyIter != mGetPropertyHandlers.end(), error = OT_ERROR_NOT_FOUND);
```

src/dbus/server/dbus_object.cpp
```diff
@@ -144,19 +145,19 @@ void DBusObject::GetPropertyMethodHandler(DBusRequest &aRequest)
...
         {
             DBusMessageIter replyIter;
             auto &          interfaceHandlers = propertyIter->second;
             auto            interfaceIter     = interfaceHandlers.find(propertyName);
 
-            VerifyOrExit(interfaceIter != interfaceHandlers.end(), error = OT_ERROR_NOT_FOUND);
+            VerifyOrExit(interfaceIter != interfaceHandlers.end(), error = OT_ERROR_INVALID_ARGS);
```
simonlingoogle, line 154:
```suggestion
            VerifyOrExit(interfaceIter != interfaceHandlers.end(), error = OT_ERROR_NOT_FOUND);
```

src/dbus/server/dbus_object.cpp
simonlingoogle: fine to me to keep using `OT_ERROR_NOT_FOUND`

bukepo (approved):

jwhui (approved):

PR #1217, opened 2022-01-25
[certification] disable auto Thread attaching for Reference Device

simonlingoogle:
This commit enhances `OTBR_NO_AUTO_ATTACH` option to disable auto Thread attaching for Reference Device:
- Use `-DOTBR_NO_AUTO_ATTACH=1` to disable auto  Thread attaching

### DIFF
--- a/script/_otbr
+++ b/script/_otbr
@@ -95,6 +95,7 @@ otbr_install()
 
     if [[ ${REFERENCE_DEVICE} == "1" ]]; then
         otbr_options+=(
+            "-DOTBR_NO_AUTO_ATTACH=1"
             "-DOT_REFERENCE_DEVICE=ON"
             "-DOT_DHCP6_CLIENT=ON"
             "-DOT_DHCP6_SERVER=ON"
--- a/src/agent/CMakeLists.txt
+++ b/src/agent/CMakeLists.txt
@@ -95,6 +95,8 @@ elseif(NOT OTBR_OPENWRT)
         RENAME otbr-agent)
 endif()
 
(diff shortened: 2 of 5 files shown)


edmont:
Hi @simonlingoogle, could you elaborate a bit what's the motivation for this change?

simonlingoogle:
> Hi @simonlingoogle, could you elaborate a bit what's the motivation for this change?

Normally for a SoC reference device (e.g. nRF52840 Dongle), the Thread stack does not automatically start attaching when the device is power up, and the Thread state will be `disabled` after `reset` command, until the `thread start` command is issued. 

However, `otbr-agent` behaves differently. `otbr-agent` will automatically start Thread attaching when it's launched or reset by `reset` command (when `OTBR_NO_AUTO_ATTACH` was not implemented). This brings notable troubles to testing and certification because we usually want to configure Thread parameters after launching or `reset` and then execute `thread start`. 

So, we used `OTBR_NO_AUTO_ATTACH` environment variable in #649 to keep Thread in `disabled` state after executing `reset` command to address testing and certification issues. 
However, `otbr-agent` will still automatically start Thread stack after `systemctl restart otbr-agent`, which is used in THCI powerUp API. More specifically, this cause OTBR SED DUT fails to restore SED poll period in powerUp.

Thus, this PR helps OTBR and SoC reference devices to have a consistent behavior (stay in `disabled` state after launching and reset).

edmont (approved):
Makes sense, thanks.

bukepo (approved):

jwhui (approved):

PR #1223, opened 2022-01-28
[signals] handle `SIGINT` same as `SIGTERM`

simonlingoogle:
This commit handles `SIGINT` so that `otbr-agent` exits gracefully under `Ctrl+C`.


jinran-google (approved):
LGTM :+1:

simonlingoogle:
> I think we intentionally not handles SIGINT so that it can better support gdb. We use Ctrl-C to interrupt the program and start debugging.

Thanks for information. That make sense to me. 
Should be fine to leave SIGINT unhandled anyway. 
Closeing this PR.

PR #1232, opened 2022-02-07
[docker] remove unused arguments from Docker entrypoint

agners:
Both, --disable-default-prefix-route and --disable-default-prefix-slaac
have been used for wpantund which is no longer supported by the Docker
entrypoint script. Remove them to avoid confusion.


wgtdkp (approved):
LGTM :+1: 
Thanks!

simonlingoogle (approved):
LGTM 👍 

Thanks

jwhui (approved):
Thanks! 👍🏼

PR #1234, opened 2022-02-08
[trel-dnssd] parse `xa` TXT value as binary

simonlingoogle:
Depends-On: openthread/openthread#7401

### DIFF
--- a/src/trel_dnssd/trel_dnssd.cpp
+++ b/src/trel_dnssd/trel_dnssd.cpp
@@ -478,20 +478,21 @@ void TrelDnssd::Peer::ReadExtAddrFromTxtData(void)
     {
         if (StringUtils::EqualCaseInsensitive(txtEntry.mName, kTxtRecordExtAddressKey))
         {
-            char extAddrHexBuf[sizeof(mExtAddr) * 2 + 1];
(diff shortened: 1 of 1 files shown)


abtink (approved):
LGTM. 👍

src/trel_dnssd/trel_dnssd.cpp
```diff
@@ -478,14 +478,9 @@ void TrelDnssd::Peer::ReadExtAddrFromTxtData(void)
     {
         if (StringUtils::EqualCaseInsensitive(txtEntry.mName, kTxtRecordExtAddressKey))
         {
-            char extAddrHexBuf[sizeof(mExtAddr) * 2 + 1];
+            VerifyOrExit(txtEntry.mValue.size() == sizeof(mExtAddr));
```
bukepo, line 481: should we add a debug level log here?

simonlingoogle: Make sense. Will add.

simonlingoogle: Added a `INFO` log if we can't dissect Extended Address from TXT data. 
BTW, this error should not happen in practise.

bukepo (approved):

jwhui (approved):

PR #1241, opened 2022-02-10
[trel-dnssd] fix wrong instance name size

gjc13:
The pointer size was misused as the extended address size. They happen to be the same(8 bytes) on 64-bit machines but can cause stack overflow on other platforms.

### DIFF
--- a/src/trel_dnssd/trel_dnssd.cpp
+++ b/src/trel_dnssd/trel_dnssd.cpp
@@ -229,12 +229,12 @@ std::string TrelDnssd::GetTrelInstanceName(void)
 {
     const otExtAddress *extaddr = otLinkGetExtendedAddress(mNcp.GetInstance());
     std::string         name;
-    char                nameBuf[sizeof(extaddr) * 2 + 1];
+    char                nameBuf[sizeof(otExtAddress) * 2 + 1];
 
-    Utils::Bytes2Hex(extaddr->m8, sizeof(extaddr), nameBuf);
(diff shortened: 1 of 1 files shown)


wgtdkp (approved):
LGTM :+1:

gabekassel:
@gjc13 @wgtdkp @simonlingoogle out of curiosity, do you know why this wasn't caught by CI? Any test improvements we should add?

jwhui:
> @gjc13 @wgtdkp @simonlingoogle out of curiosity, do you know why this wasn't caught by CI? Any test improvements we should add?

I think we need to add some checks against 32-bit arch, similar to what we do in OpenThread core, for example:
https://github.com/openthread/openthread/blob/9fcdbed9c526c70f1556d1ed84099c1535c7cd32/.github/workflows/simulation-1.1.yml#L110

jwhui (approved):
Thanks! 👍🏼

simonlingoogle:
> > @gjc13 @wgtdkp @simonlingoogle out of curiosity, do you know why this wasn't caught by CI? Any test improvements we should add?
> 
> I think we need to add some checks against 32-bit arch, similar to what we do in OpenThread core, for example: https://github.com/openthread/openthread/blob/9fcdbed9c526c70f1556d1ed84099c1535c7cd32/.github/workflows/simulation-1.1.yml#L110

Maybe we can build 32bit OTBR docker image use it in Border Router tests.

PR #1245, opened 2022-02-11
[dbus] add debugging log for UpdateMeshcopTxtHandler

wgtdkp:
It should be considered an error case if the `UpdateMeshcopTxtHandler` is not properly set.


superwhd (approved):
LGTM :+1:

By the way, would you like to log more when updating the meshcop TXT?

jwhui (approved):

PR #1248, opened 2022-02-13
[mdns] remove ServiceRegistration before invoking the callback

wgtdkp:
Current implementation invokes the service/host registration callback before removing the
`ServiceRegistraion` entry from the `ServiceRegistraionMap`. This will result in invalid access
to the `ServiceRegistration` entry when it's freed from the callback. This commit changes to
invoke the callback after removing the entry.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -31,6 +31,8 @@
  *   This file includes implementation of mDNS publisher.
  */
 
+#define OTBR_LOG_TAG "MDNS"
+
 #include "mdns/mdns.hpp"
 
 #include <assert.h>
@@ -209,10 +211,23 @@ void Publisher::AddServiceRegistration(ServiceRegistrationPtr &&aServiceReg)
     mServiceRegistrations.emplace(MakeFullServiceName(aServiceReg->mName, aServiceReg->mType), std::move(aServiceReg));
 }
 
-void Publisher::RemoveServiceRegistration(const std::string &aName, const std::string &aType)
+void Publisher::RemoveServiceRegistration(const std::string &aName, const std::string &aType, otbrError aError)
 {
+    auto                   it = mServiceRegistrations.find(MakeFullServiceName(aName, aType));
+    ServiceRegistrationPtr serviceReg;
+
     otbrLogInfo("Removing service %s.%s", aName.c_str(), aType.c_str());
-    mServiceRegistrations.erase(MakeFullServiceName(aName, aType));
+    VerifyOrExit(it != mServiceRegistrations.end());
+
+    // Keep the ServiceRegistration around before calling `Complete`
+    // to invoke the callback. This is for avoiding invalid access
+    // to the ServiceRegistration when it's freed from the callback.
+    serviceReg = std::move(it->second);
+    mServiceRegistrations.erase(it);
+    serviceReg->Complete(aError);
+
+exit:
+    return;
 }
 
 Publisher::ServiceRegistration *Publisher::FindServiceRegistration(const std::string &aName, const std::string &aType)
@@ -236,7 +251,8 @@ Publisher::ResultCallback Publisher::HandleDuplicateServiceRegistration(const st
 
     if (serviceReg->IsOutdated(aHostName, aName, aType, aSubTypeList, aPort, aTxtList))
     {
-        RemoveServiceRegistration(aName, aType);
+        otbrLogInfo("Removing existing service %s.%s: outdated", aName.c_str(), aType.c_str());
+        RemoveServiceRegistration(aName, aType, OTBR_ERROR_ABORTED);
     }
     else if (serviceReg->IsCompleted())
     {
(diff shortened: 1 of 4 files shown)


src/mdns/mdns.cpp
```diff
@@ -209,10 +211,23 @@ void Publisher::AddServiceRegistration(ServiceRegistrationPtr &&aServiceReg)
...
 
-void Publisher::RemoveServiceRegistration(const std::string &aName, const std::string &aType)
+void Publisher::RemoveServiceRegistration(const std::string &aName, const std::string &aType, otbrError aError)
 {
+    auto                   it = mServiceRegistrations.find(MakeFullServiceName(aName, aType));
+    ServiceRegistrationPtr serviceReg;
+
     otbrLogInfo("Removing service %s.%s", aName.c_str(), aType.c_str());
-    mServiceRegistrations.erase(MakeFullServiceName(aName, aType));
+    VerifyOrExit(it != mServiceRegistrations.end());
```
simonlingoogle, line 220: log after it's found?

src/mdns/mdns.cpp
```diff
@@ -236,7 +251,8 @@ Publisher::ResultCallback Publisher::HandleDuplicateServiceRegistration(const st
 
     if (serviceReg->IsOutdated(aHostName, aName, aType, aSubTypeList, aPort, aTxtList))
     {
-        RemoveServiceRegistration(aName, aType);
+        otbrLogInfo("Removing existing service %s.%s: it's outdated", aName.c_str(), aType.c_str());
```
simonlingoogle, line 254:
```suggestion
        otbrLogInfo("Removing existing service %s.%s: outdated", aName.c_str(), aType.c_str());
```
nit

src/mdns/mdns.cpp
```diff
@@ -272,7 +288,7 @@ Publisher::ResultCallback Publisher::HandleDuplicateHostRegistration(const std::
 
     if (hostReg->IsOutdated(aName, aAddress))
     {
-        RemoveHostRegistration(hostReg->mName);
+        RemoveHostRegistration(hostReg->mName, /* aError */ OTBR_ERROR_ABORTED);
```
simonlingoogle, line 291:
```suggestion
        RemoveHostRegistration(hostReg->mName, OTBR_ERROR_ABORTED);
```
nit

src/mdns/mdns.cpp
```diff
@@ -303,9 +319,23 @@ void Publisher::AddHostRegistration(HostRegistrationPtr &&aHostReg)
...
 
-void Publisher::RemoveHostRegistration(const std::string &aName)
+void Publisher::RemoveHostRegistration(const std::string &aName, otbrError aError)
 {
-    mHostRegistrations.erase(MakeFullHostName(aName));
+    auto                it = mHostRegistrations.find(MakeFullHostName(aName));
+    HostRegistrationPtr hostReg;
+
+    otbrLogInfo("Removing host %s", aName.c_str());
+    VerifyOrExit(it != mHostRegistrations.end());
```
simonlingoogle, line 328:
```suggestion
    VerifyOrExit(it != mHostRegistrations.end());
    otbrLogInfo("Removing host %s", aName.c_str());
```

src/mdns/mdns_avahi.cpp
```diff
@@ -485,18 +485,24 @@ void PublisherAvahi::CallHostOrServiceCallback(AvahiEntryGroup *aGroup, otbrErro
...
-        hostReg->Complete(aError);
-        if (aError != OTBR_ERROR_NONE)
+        if (aError == OTBR_ERROR_NONE)
+        {
+            hostReg->Complete(aError);
+        }
+        else
         {
-            RemoveHostRegistration(hostReg->mName);
+            RemoveHostRegistration(hostReg->mName, aError);
```
simonlingoogle, line 505: nit: Add a new method for this logic? e.g. FinalizeHostRegistration(); ?

src/mdns/mdns.cpp
wgtdkp: The log logs whether this method is called.

src/mdns/mdns.cpp
wgtdkp: ditto

src/mdns/mdns_avahi.cpp
wgtdkp: `FinalizeHostRegistration` doesn't indicates whether the service is going to be removed or not.

simonlingoogle: But deciding that we should remove the service is essentially part of the finalization logic?

The same logic pattern seems to occur in both mdns_avahi.cpp and  mdns_mdnssd.cpp. So, maybe better to encapsulate it?

simonlingoogle (approved):
LGTM :+1: 

with a nit. 

Verified that this PR makes `backbone-router` much more reliable on https://github.com/openthread/openthread/pull/7409

superwhd (approved):
LGTM :+1:

src/mdns/mdns_avahi.cpp
wgtdkp: It sounds good, but we have to log specific mdnsResponder or avahi error when failed to register the service. so not easy to make a common function for it.

jwhui (approved):
Thanks! 👍🏼

PR #1256, opened 2022-02-16
[dbus] add method `LeaveNetwork`

jinran-google:
This commit adds a DBus method `LeaveNetwork` to detach from the network and erase all credentials.


src/dbus/server/dbus_thread_object.cpp
```diff
@@ -1522,6 +1524,21 @@ void DBusThreadObject::ActiveDatasetChangeHandler(const otOperationalDatasetTlvs
...
 }
 
+void DBusThreadObject::LeaveNetworkHandler(DBusRequest &aRequest)
+{
+    mNcp->GetThreadHelper()->DetachGracefully([aRequest, this](otError error) mutable {
+        SuccessOrExit(error);
+        SuccessOrExit(error = otInstanceErasePersistentInfo(mNcp->GetThreadHelper()->GetInstance()));
+
+    exit:
+        aRequest.ReplyOtResult(error);
```
bukepo, line 1534: Is this a synchronized call? Will the client receive the reply as exit is called right after this call.

jinran-google: Probably not. Added flush to ensure reply is sent before exit.

jinran-google:
> @jinran-google Should we remove the `FactoryReset` API if it's not used any more?

I agree. We can do it in subsequent PRs.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -1522,6 +1524,22 @@ void DBusThreadObject::ActiveDatasetChangeHandler(const otOperationalDatasetTlvs
...
+        if (error == OT_ERROR_NONE)
```
simonlingoogle, line 1535: what if detach succeeded but erasing settings failed? should we exit in such case?

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -1522,6 +1524,22 @@ void DBusThreadObject::ActiveDatasetChangeHandler(const otOperationalDatasetTlvs
...
 }
 
+void DBusThreadObject::LeaveNetworkHandler(DBusRequest &aRequest)
+{
+    mNcp->GetThreadHelper()->DetachGracefully([aRequest, this](otError error) mutable {
+        SuccessOrExit(error);
+        SuccessOrExit(error = otInstanceErasePersistentInfo(mNcp->GetThreadHelper()->GetInstance()));
+
+    exit:
+        aRequest.ReplyOtResult(error);
```
simonlingoogle, line 1534: Why don't we always flush in ReplyOtResult?

src/dbus/server/dbus_thread_object.cpp
jinran-google: We should not exit if there are errors, because the operation is not complete.

In this case, erasing fails only when state is not disabled. It's not possible here.

src/dbus/server/dbus_thread_object.cpp
jinran-google: It's not necessary. The messages will be sent the next time the main loop is run.

bukepo (approved):

wgtdkp (approved):
LGTM 👍

simonlingoogle (approved):
LGTM 👍 
Thanks

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -1522,6 +1524,22 @@ void DBusThreadObject::ActiveDatasetChangeHandler(const otOperationalDatasetTlvs
     SignalPropertyChanged(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_PROPERTY_ACTIVE_DATASET_TLVS, value);
 }
 
+void DBusThreadObject::LeaveNetworkHandler(DBusRequest &aRequest)
+{
+    mNcp->GetThreadHelper()->DetachGracefully([aRequest, this](otError error) mutable {
+        SuccessOrExit(error);
+        SuccessOrExit(error = otInstanceErasePersistentInfo(mNcp->GetThreadHelper()->GetInstance()));
```
wgtdkp, line 1531: We are saving old local on-link prefixes in settings, see https://github.com/openthread/openthread/pull/8330.

The change here will erase this information when leave current network, but the old local on-link prefixes should persist between network migrations. 

What about erasing only Thread operational datasets? @bukepo @abtink

abtink: I think if we fully leave the Thread network we should erase everything. 
- We don't want to reuse any of the previous info associated with old network (e.g. SRP key, child table, parent info, mac address

If we migrate from one network to another, then may be fine to keep some of the old info.

PR #1260, opened 2022-02-18
[docker] cleanup script/bootstrap and etc/docker/Dockerfile

agners:
Minor improvements in script/bootstrap and realign etc/docker/Dockerfile with what is being done in script/bootstrap.


jwhui (approved):
Thanks! 👍🏼

PR #1261, opened 2022-02-18
[web] set text/html Content-Type for html files

agners:
Explicitly setting Content-Type avoids confusion in cases where proxy or similar
intermediaries are involved.


jwhui:
@agners , thanks for this contribution! The pretty check is failing, can you help resolve? Thanks.

jwhui (approved):
Thanks! 👍🏼

PR #1262, opened 2022-02-19
[web] use relative paths in xhr requests

agners:
This allows to add prefixes to the path part of the URL e.g. when using a reverse proxy.


jwhui (approved):
Thanks! 👍🏼

PR #1263, opened 2022-02-21
[tests] upgrade CI docker base image to `ubuntu:focal`

superwhd:
The main reason I'm upgrading the base image is to leverage the `zeroconf` library. Latest few versions of `zeroconf` require python version >= 3.7 and `ubuntu:focal` uses python 3.8 as default.


simonlingoogle:
@superwhd why is the CI failing?

superwhd:
> @superwhd why is the CI failing?

I don't have much idea yet. Haven't got much time to debug it.

PR #1268, opened 2022-02-23
[border-router-tests] support ARM32 arch

simonlingoogle:


simonlingoogle:
seems not working, close for now

PR #1274, opened 2022-02-27
[github-actions] add i386 BR tests

simonlingoogle:


simonlingoogle:
The test seems still 64bit, so not working. 
Close for now.

PR #1290, opened 2022-03-09
[cmake] add  `OTBR_NOTIFY_UPSTART` option

simonlingoogle:


src/agent/application.cpp
```diff
@@ -121,8 +121,13 @@ otbrError Application::Run(void)
         // See https://www.freedesktop.org/software/systemd/man/sd_notify.html
         sd_notify(0, "READY=1");
     }
+#endif
+
+#if defined(HAVE_LIBSYSTEMD) && OTBR_ENABLE_NOTIFY_UPSTART
     else
 #endif
+
```
bukepo, line 129: Perhaps, we must remove this line.
```suggestion
```

bukepo (approved):

src/agent/application.cpp
simonlingoogle: shouldn't we use `else` when both HAVE_LIBSYSTEMD and OTBR_ENABLE_NOTIFY_UPSTART is ON ?

bukepo (approved):
LGTM:+1:

PR #1291, opened 2022-03-09
[dns64] set DNS64 default value to 0

xuyirio:
This PR sets DNS64 default value to 0 for all platforms. The DNS64 flag currently installs and configures bind9. When it's ON, the DNS-SD server will not work properly.

It also adds `python3` as a Docker dependency since some tests require it. It seems that python3 was silently installed along with bind9.


etc/docker/Dockerfile
```diff
@@ -81,13 +81,17 @@ ENV OTBR_OT_BACKBONE_CI_DEPS curl lcov wget build-essential
 ENV OTBR_NORELEASE_DEPS \
   cpputest-dev
 
+# Required for testing
+ENV OTBR_TEST_DEPS python3
```
simonlingoogle, line 85: seems unrelated change?

xuyirio: Seems python3 is installed along with bind9. So tests using python3 cmd will fail when DNS64=0.

simonlingoogle: Maybe we can add python3 to OTBR_DOCKER_REQS?

xuyirio: Update. Thanks!

simonlingoogle (approved):
LGTM 👍

wgtdkp (approved):

jwhui (approved):

PR #1307, opened 2022-03-23
[border-agent] don't treat `OTBR_ERROR_ABORTED` as an error in logs

superwhd:
otbr-agent generates `WARNING` level logs when there are frequent meshcop service updates: 
```
otbr-agent[2638658]: [WARN]-BA------: Result of publish meshcop service OpenThread BorderRouter._meshcop._udp.local: Aborted
otbr-agent[2638658]: [WARN]-BA------: Result of publish meshcop service OpenThread BorderRouter._meshcop._udp.local: Aborted
```
Such a log line is generated because an ongoing meshcop service registration is canceled for some reason (a newer registration preempts the previous registration, or border agent wants to stop advertising the service). However, this behavior is working as expected and shouldn't be regarded as an error. Hence I lowered the log level to `INFO` in such cases.

### DIFF
--- a/src/border_agent/border_agent.cpp
+++ b/src/border_agent/border_agent.cpp
@@ -309,8 +309,19 @@ void BorderAgent::PublishMeshCopService(void)
 
     mPublisher->PublishService(/* aHostName */ "", mServiceInstanceName, kBorderAgentServiceType,
                                Mdns::Publisher::SubTypeList{}, port, txtList, [this](otbrError aError) {
-                                   otbrLogResult(aError, "Result of publish meshcop service %s.%s.local",
-                                                 mServiceInstanceName.c_str(), kBorderAgentServiceType);
+                                   if (aError == OTBR_ERROR_ABORTED)
+                                   {
+                                       // OTBR_ERROR_ABORTED is thrown when an ongoing service registration is
+                                       // cancelled. This can happen when the meshcop service is being updated
+                                       // frequently. To avoid false alarms, it should not be logged like a real error.
+                                       otbrLogInfo("Cancelled previous publishing meshcop service %s.%s.local",
+                                                   mServiceInstanceName.c_str(), kBorderAgentServiceType);
+                                   }
+                                   else
+                                   {
+                                       otbrLogResult(aError, "Result of publish meshcop service %s.%s.local",
+                                                     mServiceInstanceName.c_str(), kBorderAgentServiceType);
+                                   }
                                    if (aError == OTBR_ERROR_DUPLICATED)
                                    {
                                        // Try to unpublish current service in case we are trying to register


src/border_agent/border_agent.cpp
```diff
@@ -309,8 +309,20 @@ void BorderAgent::PublishMeshCopService(void)
...
+                                       // OTBR_ERROR_ABORTED is thrown when an ongoing service registration is
+                                       // gracefully cancelled. This can happen when the meshcop service is being
+                                       // updated frequently. To avoid false alarms, it should not be logged like a real
+                                       // error.
+                                       otbrLogInfo("Cancel publishing meshcop service %s.%s.local",
+                                                   mServiceInstanceName.c_str(), kBorderAgentServiceType);
+                                   }
+                                   else
+                                   {
+                                       otbrLogResult(aError, "Result of publish meshcop service %s.%s.local",
```
wgtdkp, line 323: What about putting the logic into `otbrLogResult` and dump `NOTE` level messages for all `OTBR_ERROR_ABORTED` errors?

wgtdkp: `ABORT` means the operation is explicitly cancelled by the client, so maybe all such errors are actually not an error.

simonlingoogle: > `ABORT` means the operation is explicitly cancelled by the client, so maybe all such errors are actually not an error.

I think it might be too aggressive. Maybe we can do it later when we have more signals.

src/border_agent/border_agent.cpp
```diff
@@ -309,8 +309,20 @@ void BorderAgent::PublishMeshCopService(void)
...
                                Mdns::Publisher::SubTypeList{}, port, txtList, [this](otbrError aError) {
-                                   otbrLogResult(aError, "Result of publish meshcop service %s.%s.local",
-                                                 mServiceInstanceName.c_str(), kBorderAgentServiceType);
+                                   if (aError == OTBR_ERROR_ABORTED)
+                                   {
+                                       // OTBR_ERROR_ABORTED is thrown when an ongoing service registration is
+                                       // gracefully cancelled. This can happen when the meshcop service is being
+                                       // updated frequently. To avoid false alarms, it should not be logged like a real
+                                       // error.
+                                       otbrLogInfo("Cancel publishing meshcop service %s.%s.local",
```
simonlingoogle, line 318:
```suggestion
                                       otbrLogInfo("Cancel previous publishing meshcop service %s.%s.local",
```
nit

simonlingoogle (approved):
LGTM 👍 

with a nit

src/border_agent/border_agent.cpp
wgtdkp:
```suggestion
                                       otbrLogInfo("Cancelled publishing meshcop service %s.%s.local",
```

src/border_agent/border_agent.cpp
wgtdkp: No strong opinion.

wgtdkp (approved):

jwhui (approved):

PR #1310, opened 2022-03-24
[logging] use capitalized first letter for log messages

superwhd:
I checked all logging lines by `grep -rn otbrLog src/` and `discovery_proxy.cpp` is the only file using lower case for every first letter in log messages.


simonlingoogle (approved):
LGTM 👍 

Thanks!

jwhui (approved):

PR #1314, opened 2022-03-25
[mac] remove beacons payload

edmont:
Thread Specification v1.2.1 removes support for Thread Beacons payload.

### DIFF
--- a/src/openwrt/view/admin_thread/thread_scan.htm
+++ b/src/openwrt/view/admin_thread/thread_scan.htm
@@ -74,12 +74,9 @@
 		<div class="table">
 			<div class="tr table-titles">
 				<div class="th col-1 center"><%:RSSI%></div>
-				<div class="th col-3 center"><%:Network Name%></div>
 				<div class="th col-1 center"><%:Channel%></div>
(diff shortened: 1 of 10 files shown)


src/openwrt/view/admin_thread/thread_scan.htm
```diff
@@ -74,12 +74,9 @@ <h2><%:Join Network: Thread Scan%></h2>
 		<div class="table">
 			<div class="tr table-titles">
 				<div class="th col-1 center"><%:RSSI%></div>
-				<div class="th col-3 center"><%:Network Name%></div>
```
bukepo, line 77: We probably want to use `Discover`, the commissioning process requires `network name` and `extended panid` for computing `pskc`.

edmont: Any insight about which places in the code to modify? I'm not very familiar with this `openwrt` functionality.

jwhui: We can address this in a separate PR.

jwhui (approved):

PR #1325, opened 2022-04-02
[agent] remove `InstanceParams` class

simonlingoogle:
This commit removes the `InstanceParams` to make the code more clean.


src/backbone_router/nd_proxy.hpp
```diff
@@ -150,6 +152,7 @@ class NdProxyManager : public MainloopProcessor, private NonCopyable
     int HandleNetfilterQueue(struct nfq_q_handle *aNfQueueHandler, struct nfgenmsg *aNfMsg, struct nfq_data *aNfData);
 
     otbr::Ncp::ControllerOpenThread &mNcp;
+    std::string                      mBackboneInterfaceName;
```
bukepo, line 155: interface has fixed max length, is it better to use char array?

simonlingoogle: The max interface name size is more than 100 bytes, but a typical interface name is only about 10 bytes. It seems a waste of memory to pre-allocate big buffers. 

Moreover, std::string is much more easier to use.

So, I would prefer using std::string as is. 

Thoughts?

bukepo: From https://github.com/spotify/linux/blob/master/include/linux/if.h#L26, it's only 16 bytes long.

simonlingoogle: > From https://github.com/spotify/linux/blob/master/include/linux/if.h#L26, it's only 16 bytes long.

thanks for the link. 

It's still more convenient to use std::string? 

I don't think there is obvious advantage of using fixed length char arrays.

bukepo (approved):

src/backbone_router/nd_proxy.cpp
```diff
@@ -377,8 +376,8 @@ otbrError NdProxyManager::InitIcmp6RawSocket(void)
     VerifyOrExit(mIcmp6RawSock >= 0, error = OTBR_ERROR_ERRNO);
 
 #if __linux__
-    VerifyOrExit(setsockopt(mIcmp6RawSock, SOL_SOCKET, SO_BINDTODEVICE, InstanceParams::Get().GetBackboneIfName(),
-                            strlen(InstanceParams::Get().GetBackboneIfName())) == 0,
+    VerifyOrExit(setsockopt(mIcmp6RawSock, SOL_SOCKET, SO_BINDTODEVICE, mBackboneInterfaceName.c_str(),
+                            strlen(mBackboneInterfaceName.c_str())) == 0,
```
superwhd, line 380: Use `mBackboneInterfaceName.size()`?

superwhd (approved):
LGTM with a nit 👍

src/backbone_router/nd_proxy.cpp
simonlingoogle: Good catch

jwhui (approved):

PR #1326, opened 2022-04-02
[web-service] ot-client: add needed header for fd_set() call

Stefan-Schmidt:
The select.h header file is needed for fd_set(). Depending on the
compiler and libc combinations this will result in an error when
not included. I saw the error with gcc and musl on our builds.

Signed-off-by: Stefan Schmidt <stefan.schmidt@huawei.com>

### DIFF
--- a/src/web/web-service/ot_client.cpp
+++ b/src/web/web-service/ot_client.cpp
@@ -39,6 +39,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/select.h>
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <unistd.h>


src/web/web-service/ot_client.cpp
```diff
@@ -41,6 +41,7 @@
 #include <string.h>
 #include <sys/socket.h>
 #include <sys/un.h>
+#include <sys/select.h>
```
jwhui, line 44: nit: our coding style requires the include ordering to be alphabetical
```suggestion
#include <sys/select.h>
#include <sys/socket.h>
#include <sys/un.h>
```

Stefan-Schmidt:
Fair point, I did not pay attention to this. Fixed now.

jwhui (approved):
Thanks! 👍🏼

PR #1339, opened 2022-04-12
[platform] de-init application in `otPlatReset` and verify `Publisher` state in public methods

superwhd:
When `otPlatReset` is called on `otbr-agent`, it should de-init the old `otbr::Application` so that the resources will be properly disposed, otherwise there will be dangling resources. Also, `otbr::Application` should de-init the TREL module when itself is de-initing.

One example is that if we call `ot-ctl factoryreset`, the old `otbr::Application` won't be destructed so that the publisher will keep the old mDNS service registrations until the process exits. In such a situation, there will be two different MeshCoP services published by the same `otbr-agent` process. This test case is introduced in https://github.com/openthread/openthread/pull/7592.

Also, this PR adds a verification at `Publisher` public methods to ensure that `Publisher` is at `kReady` state.

### DIFF
--- a/src/agent/main.cpp
+++ b/src/agent/main.cpp
@@ -68,7 +68,8 @@ enum
     OTBR_OPT_RADIO_VERSION,
 };
 
-static jmp_buf sResetJump;
+static jmp_buf            sResetJump;
+static otbr::Application *gApp = nullptr;
 
 void                       __gcov_flush();
 static const struct option kOptions[] = {
@@ -211,6 +212,7 @@ static int realmain(int argc, char *argv[])
     {
         otbr::Application app(interfaceName, backboneInterfaceName, radioUrls);
 
+        gApp = &app;
         app.Init();
 
         ret = app.Run();
@@ -230,7 +232,9 @@ void otPlatReset(otInstance *aInstance)
 
     gPlatResetReason = OT_PLAT_RESET_REASON_SOFTWARE;
 
-    otSysDeinit();
+    VerifyOrDie(gApp != nullptr, "gApp is null");
+    gApp->Deinit();
+    gApp = nullptr;
 
     longjmp(sResetJump, 1);
     assert(false);
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -440,6 +440,8 @@ void PublisherAvahi::Stop(void)
         avahi_client_free(mClient);
         mClient = nullptr;
     }
+
+    mState = Mdns::Publisher::State::kIdle;
 }
 
 void PublisherAvahi::HandleClientState(AvahiClient *aClient, AvahiClientState aState, void *aContext)
@@ -671,8 +673,13 @@ exit:
 
 void PublisherAvahi::UnpublishService(const std::string &aName, const std::string &aType, ResultCallback &&aCallback)
 {
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mState == Publisher::State::kReady, error = OTBR_ERROR_INVALID_STATE);
     RemoveServiceRegistration(aName, aType, OTBR_ERROR_ABORTED);
-    std::move(aCallback)(OTBR_ERROR_NONE);
+
+exit:
+    std::move(aCallback)(error);
 }
 
 void PublisherAvahi::PublishHost(const std::string &         aName,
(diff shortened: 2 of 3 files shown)


src/agent/main.cpp
```diff
@@ -230,7 +232,11 @@ void otPlatReset(otInstance *aInstance)
 
     gPlatResetReason = OT_PLAT_RESET_REASON_SOFTWARE;
 
-    otSysDeinit();
+    if (gApp)
```
simonlingoogle, line 235: Maybe we can use singleton pattern like otbr::Application::Get() ?

Thoughts? @wgtdkp @superwhd

simonlingoogle, line 235: Is it possible that gApp is nullptr ?

src/agent/application.cpp
```diff
@@ -99,6 +101,9 @@ void Application::Init(void)
 
 void Application::Deinit(void)
 {
+#if OTBR_ENABLE_TREL
+    otPlatTrelDisable(mNcp.GetInstance());
```
simonlingoogle, line 105: Is otPlatTrelDisable already called in otSysDeinit -> platformDeinit -> platformTrelDeinit ?

superwhd: When doing `platformTrelDeinit`, Trel will be accessing its `mPublisher`, which is a reference to the border agent's `Publisher`. The border agent's publisher has already been destructed at that timepoint. Therefore, I need to explicitly disable Trel before the publisher is destructed.

src/agent/main.cpp
superwhd: I don't see big benefits of making it a singleton. The `Application` class is not used across a lot of files. Also, we're not  reusing the same `Application` instance when doing a factory reset. Instead, we need to de-init the old instance and create a new instance, and assign the global instance as the new instance.

superwhd: I think it's impossible among the use cases today. However, there's no harm to add a sanity check here.

src/agent/application.cpp
simonlingoogle: This is a little messy, but not because of this PR.
We have the plan to move `mPublisher` out of Application so that mPublisher is always available during the OT instance lifecycle. 

I think we should add a comment and TODO so that we can refactor it later.

simonlingoogle: > The border agent's publisher has already been destructed at that timepoint. Therefore, I need to explicitly disable Trel before the publisher is destructed.

Why is the publisher already distructed? I think the Application is not destructed when calling `Deinit` ?

superwhd: Correction: Publisher is not destructed but all subscribed services are cleared in `Publisher::Stop()`. In `platformTrelDeinit()`, it will unsubscribe the TREL service. However, it cannot find the subscription so that it will crash at `assert(it != mSubscribedServices.end());`.

simonlingoogle: IIUC, the issue is that `UnpublishService` is called after the Publisher is already stopped.

What about quiting `UnpublishService` if Publisher is already stopped.

It seems still logical for TrenDnssd to call on a stopped Publisher.

superwhd: Should we also verify whether Publisher is running on other Publisher methods?

simonlingoogle: The Publisher must be running for other publisher method ?

I think we are already checking that the Publisher is running in other methods, like in `PublisherAvahi::PublishService`:
```
    VerifyOrExit(mState == State::kReady, error = OTBR_ERROR_INVALID_STATE);
    VerifyOrExit(mClient != nullptr, error = OTBR_ERROR_INVALID_STATE);
```

Should be fine to just fix UnpublishXxxx methods for now ?

superwhd: Done

src/mdns/mdns_avahi.cpp
```diff
@@ -808,6 +820,7 @@ void PublisherAvahi::SubscribeService(const std::string &aType, const std::strin
 {
     auto service = MakeUnique<ServiceSubscription>(*this, aType, aInstanceName);
 
+    VerifyOrExit(mState == Publisher::State::kReady);
```
simonlingoogle, line 823: add a log when SubscribeService failed ?

simonlingoogle (approved):
LGTM 👍 

with nits

src/agent/main.cpp
jinran-google: I think it's better to `Die` if such cases that are not supposed to happen happen.

src/agent/main.cpp
```diff
@@ -230,7 +232,11 @@ void otPlatReset(otInstance *aInstance)
 
     gPlatResetReason = OT_PLAT_RESET_REASON_SOFTWARE;
 
-    otSysDeinit();
```
jinran-google, line 233: Why remove the call to `otSysDeinit()`?

superwhd: `Application::Deinit()` calls `ControllerOpenThread::Deinit()` which calls `otSysDeinit()`.

jinran-google (approved):
LGTM 👍

src/mdns/mdns_avahi.cpp
superwhd: Will do in a later PR

bukepo (approved):
Better to make otbr::Application a singleton.

jwhui (approved):

PR #1352, opened 2022-04-27
[dbus] Implement detach client API

gabekassel:
Implement detach d-bus client API
Return appropriate value from ThreadApiDBus::DBusMessageFilter()

### DIFF
--- a/src/dbus/client/thread_api_dbus.cpp
+++ b/src/dbus/client/thread_api_dbus.cpp
@@ -123,6 +123,7 @@ DBusHandlerResult ThreadApiDBus::sDBusMessageFilter(DBusConnection *aConnection,
 
 DBusHandlerResult ThreadApiDBus::DBusMessageFilter(DBusConnection *aConnection, DBusMessage *aMessage)
 {
+    DBusHandlerResult handled = DBUS_HANDLER_RESULT_NOT_YET_HANDLED;
     OTBR_UNUSED_VARIABLE(aConnection);
 
     DBusMessageIter iter, subIter, dictEntryIter, valIter;
@@ -150,9 +151,10 @@ DBusHandlerResult ThreadApiDBus::DBusMessageFilter(DBusConnection *aConnection, 
     {
         f(role);
     }
+    handled = DBUS_HANDLER_RESULT_HANDLED;
 
 exit:
-    return DBUS_HANDLER_RESULT_NOT_YET_HANDLED;
+    return handled;
 }
 
 void ThreadApiDBus::AddDeviceRoleHandler(const DeviceRoleHandler &aHandler)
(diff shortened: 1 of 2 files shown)


gabekassel:
@jwhui thoughts on this generally and anyone who can be assigned to review?

jwhui:
> @jwhui thoughts on this generally and anyone who can be assigned to review?

LGTM. @jinran-google , can you help take a look?

src/dbus/client/thread_api_dbus.cpp
```diff
@@ -150,6 +150,7 @@ DBusHandlerResult ThreadApiDBus::DBusMessageFilter(DBusConnection *aConnection,
     {
         f(role);
     }
+    return DBUS_HANDLER_RESULT_HANDLED;
```
jinran-google, line 153: There should be one return statement per free function or method at the end of the free function or method.

jinran-google, line 153: I think here the filter is an observer and should not interfere with other components that look for these messages, so the return value should be `DBUS_HANDLER_RESULT_NOT_YET_HANDLED`.

jinran-google (commented):
Thanks. LGTM overwall. Some nit suggestions:

src/dbus/client/thread_api_dbus.cpp
abhilasht1: Thanks for review.
Can we fix it by creating a return value variable?
`DBusHandlerResult ThreadApiDBus::DBusMessageFilter(DBusConnection *aConnection, DBusMessage *aMessage)
{
    DBusHandlerResult handled = DBUS_HANDLER_RESULT_NOT_YET_HANDLED;
    .....
    .....
    handled = DBUS_HANDLER_RESULT_HANDLED;

exit:
    return handled;
}`

jwhui: Yes, exactly. Thanks.

abhilasht1: DBusMessageFilter() calls 'device role handlers' on line 151, which mean it handled the dbus message with 'device role change'? So it should return 'DBUS_HANDLER_RESULT_HANDLED'?

`for (const auto &f : mDeviceRoleHandlers)
    {
        f(role);
    }`

gabekassel: updated

jinran-google: I was worried that other filters may not be called, but it should be fine since there are no other filters. LGTM. Thanks.

jinran-google (approved):
LGTM 👍

jwhui (approved):
Thanks! 👍🏼

PR #1355, opened 2022-05-03
[adv-proxy] treat `NOT_FOUND` as success when unpublishing host/service

abtink:
This commit updates `AdvertisingProxy` so that when it is unpublishing
a host or service entry on the underlying `Mdns::Publisher` if the
returned error is `OTBR_ERROR_NOT_FOUND` the error is treated as
success and reported as such to OT core `Srp::Server`. This ensures
that an SRP update request to remove a not previously registered
host/service is not rejected by the platform and therefore by the
`ot::Srp:Server`


mspang (approved):
Tested & verified

simonlingoogle (approved):
LGTM 👍 


Thanks

jwhui (approved):
Thanks! 👍🏼

PR #1363, opened 2022-05-06
[dbus] allow `AttachAllNodesTo` when has active dataset but detached

jinran-google:
This PR changes the behavior of `AttachAllNodesTo` when there is an active dataset but the device isn't attached to it:

Before this PR: if the device is in `detached` or `disabled` state, and it has an active dataset, it returns an error.

After this PR: if the device is in `detached` or `disabled` state, and it has an active dataset, it attaches to the network specified by the active dataset, and then sends a `MGMT_PENDING_SET` command to set the pending dataset to the dataset specified by the argument.

### DIFF
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -557,23 +557,17 @@ void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, Re
     if (role == OT_DEVICE_ROLE_DISABLED || role == OT_DEVICE_ROLE_DETACHED)
     {
         otOperationalDataset existingDataset;
+        bool                 hasActiveDataset;
 
         error = otDatasetGetActive(mInstance, &existingDataset);
         VerifyOrExit(error == OT_ERROR_NONE || error == OT_ERROR_NOT_FOUND);
 
-        if (error == OT_ERROR_NONE)
-        {
-            if (!otIp6IsEnabled(mInstance))
-            {
-                SuccessOrExit(error = otIp6SetEnabled(mInstance, true));
-            }
-            SuccessOrExit(error = otThreadSetEnabled(mInstance, true));
-            mAttachHandler            = aHandler;
-            mAttachPendingDatasetTlvs = datasetTlvs;
-            ExitNow();
-        }
+        hasActiveDataset = (error == OT_ERROR_NONE);
 
-        SuccessOrExit(error = otDatasetSetActiveTlvs(mInstance, &datasetTlvs));
+        if (!hasActiveDataset)
+        {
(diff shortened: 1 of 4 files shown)


zhanglongxia (approved):
LGTM

src/utils/thread_helper.cpp
```diff
@@ -557,6 +552,39 @@ void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, Re
...
+    assert(datasetTlvs.mLength > 0);
+
+    if (role == OT_DEVICE_ROLE_DISABLED || role == OT_DEVICE_ROLE_DETACHED)
+    {
+        otOperationalDataset existingDataset;
+
+        error = otDatasetGetActive(mInstance, &existingDataset);
+        VerifyOrExit(error == OT_ERROR_NONE || error == OT_ERROR_NOT_FOUND);
+
+        if (error == OT_ERROR_NONE)
```
bukepo, line 564: better to use a flag to indicate migrate or not, so that we can avoid duplicate for enabling Thread.

bukepo (approved):

src/utils/thread_helper.cpp
jinran-google: Done.

src/utils/thread_helper.cpp
```diff
@@ -92,8 +92,24 @@ void ThreadHelper::StateChangedCallback(otChangedFlags aFlags)
...
+                }
+                else
+                {
+                    otOperationalDataset emptyDataset = {};
+                    otError              error =
+                        otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, mAttachPendingDatasetTlvs.mTlvs,
+                                                    mAttachPendingDatasetTlvs.mLength, MgmtSetResponseHandler, this);
+                    if (error != OT_ERROR_NONE)
+                    {
+                        mAttachHandler(error);
```
librasungirl, line 108: Maybe little chance, however what's the expecting operations if `otDatasetSendMgmtPendingSet` fails due to either NO_BUF or BUSY error? Do we want the upper layer to retry?

jinran-google: I think in this case, it will return the error, and the caller of this API should be responsible to retry.

librasungirl (approved):

simonlingoogle (approved):
LGTM 👍 

with a nit

src/utils/thread_helper.cpp
```diff
@@ -92,8 +92,24 @@ void ThreadHelper::StateChangedCallback(otChangedFlags aFlags)
...
+                        mAttachHandler            = nullptr;
```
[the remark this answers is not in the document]
jinran-google: Done.

jwhui (approved):

PR #1364, opened 2022-05-07
[tests] add a test case of D-BUS API `Reset`

superwhd:
This PR adds a test case to verify D-BUS APIs are working well after a `Reset` call.

This PR also updates the mDNSResponder to latest in `tests/scripts/bootstrap.sh`.

### DIFF
--- a/tests/scripts/bootstrap.sh
+++ b/tests/scripts/bootstrap.sh
@@ -133,10 +133,13 @@ case "$(uname)" in
         fi
 
         if [ "${OTBR_MDNS-}" == 'mDNSResponder' ]; then
-            SOURCE_NAME=mDNSResponder-878.30.4
+            SOURCE_NAME=mDNSResponder-1310.80.1
             wget https://opensource.apple.com/tarballs/mDNSResponder/$SOURCE_NAME.tar.gz \
                 && tar xvf $SOURCE_NAME.tar.gz \
-                && cd $SOURCE_NAME/mDNSPosix \
+                && cd $SOURCE_NAME/Clients \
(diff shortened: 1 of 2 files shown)


tests/scripts/bootstrap.sh
```diff
@@ -133,10 +133,13 @@ case "$(uname)" in
...
 
         if [ "${OTBR_MDNS-}" == 'mDNSResponder' ]; then
-            SOURCE_NAME=mDNSResponder-878.30.4
+            SOURCE_NAME=mDNSResponder-1310.80.1
             wget https://opensource.apple.com/tarballs/mDNSResponder/$SOURCE_NAME.tar.gz \
                 && tar xvf $SOURCE_NAME.tar.gz \
-                && cd $SOURCE_NAME/mDNSPosix \
+                && cd $SOURCE_NAME/Clients \
+                && sed -i '/#include <ctype.h>/a #include <stdarg.h>' dns-sd.c \
+                && sed -i '/#include <ctype.h>/a #include <sys/param.h>' dns-sd.c \
```
jinran-google, line 141: Is this a known issue of mDNSResponder? Do we have more context?

jinran-google (approved):
LGTM 👍 

A question below:

tests/scripts/bootstrap.sh
superwhd: We need to add these headers to import `MIN`, `va_*` macros, otherwise the build will fail. We've already applied the same fix in `script/bootstrap`.

jwhui (approved):

PR #1367, opened 2022-05-10
[dbus] implement D-BUS API `GetSrpServerInfo`

superwhd:
This PR implements the D-BUS API `GetSrpServerInfo` for getting the information of the SRP server. 

This PR depends on the new SRP server APIs introduced in https://github.com/openthread/openthread/pull/7680 which is not merged yet. Hence it cannot pass the CI for now.

### DIFF
--- a/src/dbus/common/types.hpp
+++ b/src/dbus/common/types.hpp
@@ -41,6 +41,8 @@
 #include <string>
 #include <vector>
 
+#include <openthread/srp_server.h>
+
 namespace otbr {
 namespace DBus {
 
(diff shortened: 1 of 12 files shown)


src/dbus/common/types.hpp
```diff
@@ -36,6 +36,8 @@
 
 #include "dbus/common/error.hpp"
 
+#include <openthread/srp_server.h>
```
simonlingoogle, line 39: OT header should be below C++ standard header?

src/dbus/common/types.hpp
```diff
@@ -517,6 +519,61 @@ struct TxtEntry
...
 };
 
+enum SrpServerState : uint8_t
+{
+    OTBR_SRP_SERVER_STATE_DISABLED = 0, ///< The SRP server is disabled.
+    OTBR_SRP_SERVER_STATE_RUNNING  = 1, ///< The SRP server is running.
+    OTBR_SRP_SERVER_STATE_STOPPED  = 2, ///< The SRP server is stopped.
+};
+
+static_assert(OTBR_SRP_SERVER_STATE_DISABLED == static_cast<uint8_t>(OT_SRP_SERVER_STATE_DISABLED),
```
simonlingoogle, line 529: it works for enum with small number of values. 

What we do if the enum have tens of values?

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -36,7 +36,9 @@
 #include <openthread/link_raw.h>
 #include <openthread/ncp.h>
 #include <openthread/netdata.h>
+#include <openthread/srp_server.h>
 #include <openthread/thread_ftd.h>
+#include <openthread/platform/alarm-milli.h>
```
simonlingoogle, line 41: What is this header used for?

simonlingoogle (approved):
LGTM 👍 

with small questions

src/dbus/common/types.hpp
superwhd: One possible way is that we define `OTBR_*` enums by applying a macro function on `OT_*` enums. In this way we can ensure that every `OTBR_*` enum has the same value as the according `OT_*` enum.
```
enum {
    DEFINE_OTBR_ENUM(A),
    DEFINE_OTBR_ENUM(B),
    ...
};
```

jwhui (approved):

PR #1370, opened 2022-05-11
[meshcop] append last 2 bytes of ext addr to service instance name

superwhd:
This PR appends last 2 bytes of extended address to the MeshCoP service instance name. This is used for reducing service instance name conflicts when there are two BRs of the same model connected to the same wifi.

An example of the new service instance name is `OpenThread BorderRouter #B76D`.

### DIFF
--- a/src/border_agent/border_agent.cpp
+++ b/src/border_agent/border_agent.cpp
@@ -44,6 +44,7 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <iomanip>
 #include <random>
 #include <sstream>
 
@@ -107,7 +108,6 @@ BorderAgent::BorderAgent(otbr::Ncp::ControllerOpenThread &aNcp)
 #if OTBR_ENABLE_TREL
     , mTrelDnssd(aNcp, *mPublisher)
 #endif
-    , mServiceInstanceName(kBorderAgentServiceInstanceName)
 {
 }
 
@@ -125,6 +125,8 @@ void BorderAgent::Init(void)
         });
     });
 #endif
+
+    mServiceInstanceName = BaseServiceInstanceName();
 
     Start();
 }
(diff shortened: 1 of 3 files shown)


src/border_agent/border_agent.cpp
```diff
@@ -388,6 +389,16 @@ bool BorderAgent::IsThreadStarted(void) const
     return role == OT_DEVICE_ROLE_CHILD || role == OT_DEVICE_ROLE_ROUTER || role == OT_DEVICE_ROLE_LEADER;
 }
 
+std::string BorderAgent::BaseServiceInstanceName() const
+{
+    const otExtAddress *extAddress    = otLinkGetExtendedAddress(mNcp.GetInstance());
+    uint16_t            extAddressNum = *reinterpret_cast<const uint16_t *>(&extAddress->m8[6]);
```
wgtdkp, line 395: Why bother casting it to uint16_t but not just output the two uint8_t bytes? :)

wgtdkp (approved):

src/border_agent/border_agent.cpp
```diff
@@ -388,6 +390,16 @@ bool BorderAgent::IsThreadStarted(void) const
     return role == OT_DEVICE_ROLE_CHILD || role == OT_DEVICE_ROLE_ROUTER || role == OT_DEVICE_ROLE_LEADER;
 }
 
+std::string BorderAgent::BaseServiceInstanceName() const
+{
+    const otExtAddress *extAddress = otLinkGetExtendedAddress(mNcp.GetInstance());
+    std::stringstream   ss;
+
+    ss << kBorderAgentServiceInstanceName << " " << std::uppercase << std::hex << std::setfill('0') << std::setw(2)
```
bukepo, line 398: Is it better to add a separator like '#' between the name and the discriminator? Say `OpenThread #ABCD`.

bukepo (approved):

bukepo (approved):
LGTM:+1:

jwhui (approved):

PR #1373, opened 2022-05-13
[openthread] fix build for Thread 1.1

leojrfs:
This commit fixes the build for Thread 1.1 (when `OT_THREAD_VERSION=1.1` is passed) by enabling `OPENTHREAD_CONFIG_TMF_PROXY` for Thread 1.2 only.


third_party/openthread/CMakeLists.txt
```diff
@@ -94,6 +94,11 @@ target_compile_definitions(ot-config INTERFACE
     "-DOPENTHREAD_CONFIG_MAX_STATECHANGE_HANDLERS=3"
     "-DOPENTHREAD_CONFIG_MLE_STEERING_DATA_SET_OOB_ENABLE=1"
     "-DOPENTHREAD_CONFIG_TCP_ENABLE=0"
-    "-DOPENTHREAD_CONFIG_TMF_PROXY_DUA_ENABLE=1"
-    "-DOPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE=1"
 )
+
+if (OT_THREAD_VERSION STREQUAL "1.2")
```
jwhui, line 99: We recently added support for "1.3" as well (see https://github.com/openthread/openthread/pull/7692). How about something like:
```suggestion
if (NOT OT_THREAD_VERSION STREQUAL "1.1")
```

jwhui (approved):
Thanks! 👍🏼

PR #1383, opened 2022-05-20
[agent] remove potential data race on `sShouldTerminate`

simonlingoogle:
This commit remove potential data race on `sShouldTerminate`, which is accessed in both main loop and the signal handler.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -43,8 +43,8 @@
 
 namespace otbr {
 
-bool                 Application::sShouldTerminate = false;
-const struct timeval Application::kPollTimeout     = {10, 0};
+std::atomic_bool     Application::sShouldTerminate(false);
+const struct timeval Application::kPollTimeout = {10, 0};
 
 Application::Application(const std::string &              aInterfaceName,
                          const std::string &              aBackboneInterfaceName,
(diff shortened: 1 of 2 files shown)


src/agent/application.cpp
```diff
@@ -43,8 +43,8 @@
 
 namespace otbr {
 
-bool                 Application::sShouldTerminate = false;
-const struct timeval Application::kPollTimeout     = {10, 0};
+std::atomic_bool     Application::sShouldTerminate{false};
```
jinran-google, line 46: I think the old "= false" initialization is better.

jinran-google (approved):
LGTM overall 👍

src/agent/application.cpp
simonlingoogle: It does not compile.

jinran-google: OK, it doesn't compile without mandatory copy elision, which was introduced in C++17.

Then I think it's better to use the traditional constructor syntax (with parentheses) in this case.

simonlingoogle: > OK, it doesn't compile without mandatory copy elision, which was introduced in C++17.

Thanks for the insight. 


> Then I think it's better to use the traditional constructor syntax (with parentheses) in this case.

Is there a explanation why which is better?

bukepo (approved):

jwhui (approved):

PR #1385, opened 2022-05-20
[github-actions] set `fail-fast: false` for matrix

jwhui:
This commit sets `fail-fast: false` for matrix in GitHub Actions so
that tests in a matrix do not fail all at once. It's more convenient
now that we can re-run failed tests.


simonlingoogle (approved):
LGTM 👍

PR #1386, opened 2022-05-22
[meshcop] publish the full timestamp tlv

wgtdkp:


bukepo (approved):

superwhd (approved):
LGTM :+1:

src/border_agent/border_agent.cpp
```diff
@@ -252,8 +251,14 @@ void BorderAgent::PublishMeshCopService(void)
         }
         else
         {
-            activeTimestamp = htobe64(activeDataset.mActiveTimestamp);
-            txtList.emplace_back("at", reinterpret_cast<uint8_t *>(&activeTimestamp), sizeof(activeTimestamp));
+            uint64_t activeTimestampValue;
+
+            activeTimestampValue = (activeDataset.mActiveTimestamp.mSeconds << 16) |
```
simonlingoogle, line 256: nit: we can elide the bit operations if `otTimestamp` is defined as an union with a `uint64_t` value to get the full timesstamp.

wgtdkp: ditto~

simonlingoogle, line 256: nit: define constant for 16 and 1.

simonlingoogle (approved):
LGTM 👍

librasungirl (approved):

src/border_agent/border_agent.cpp
jwhui: @wgtdkp , thoughts?

bukepo: I think it's fine to use the numbers here. It should be easy to understand.

wgtdkp: I tend to not define constants too far away from where they are being used, especially when there are a single use case. To help readability, I made a short function for converting the timestamp to uint64_t and added documentation for its layout.

jwhui (approved):

PR #1402, opened 2022-05-31
[agent] add command-line arg to disable auto attach

jinran-google:
This commit adds a command line argument to specify if OTBR automatically attempts to attach after starting.

### DIFF
--- a/src/agent/application.hpp
+++ b/src/agent/application.hpp
@@ -83,7 +83,7 @@ public:
      * @param[in] aInterfaceName         Name of the Thread network interface.
      * @param[in] aBackboneInterfaceName Name of the backbone network interface.
      * @param[in] aRadioUrls             The radio URLs (can be IEEE802.15.4 or TREL radio).
-     * @param[in] aEnableAutoAttach      Whether or not automatically attaches to the saved network.
+     * @param[in] aEnableAutoAttach      Whether or not to automatically attach to the saved network.
      *
      */
     explicit Application(const std::string &              aInterfaceName,
--- a/src/agent/application.hpp
+++ b/src/agent/application.hpp
@@ -80,12 +80,16 @@ public:
     /**
      * This constructor initializes the Application instance.
      *
-     * @param[in] aOpenThread  A reference to the OpenThread instance.
+     * @param[in] aInterfaceName         Name of the Thread network interface.
+     * @param[in] aBackboneInterfaceName Name of the backbone network interface.
+     * @param[in] aRadioUrls             The radio URLs (can be IEEE802.15.4 or TREL radio).
+     * @param[in] aEnableAutoAttach      Whether or not automatically attaches to the saved network.
      *
      */
     explicit Application(const std::string &              aInterfaceName,
                          const std::string &              aBackboneInterfaceName,
-                         const std::vector<const char *> &aRadioUrls);
+                         const std::vector<const char *> &aRadioUrls,
+                         bool                             aEnableAutoAttach);
 
     /**
      * This method initializes the Application instance.
--- a/src/agent/main.cpp
+++ b/src/agent/main.cpp
@@ -30,9 +30,12 @@
 
 #include <openthread-br/config.h>
 
+#include <algorithm>
 #include <fstream>
 #include <mutex>
 #include <sstream>
+#include <string>
+#include <vector>
 
 #include <getopt.h>
 #include <stdio.h>
@@ -66,9 +69,9 @@ enum
     OTBR_OPT_INTERFACE_NAME          = 'I',
     OTBR_OPT_VERBOSE                 = 'v',
     OTBR_OPT_VERSION                 = 'V',
-    OTBR_OPT_AUTO_ATTACH             = 'a',
     OTBR_OPT_SHORTMAX                = 128,
     OTBR_OPT_RADIO_VERSION,
+    OTBR_OPT_AUTO_ATTACH,
 };
 
 static jmp_buf            sResetJump;
@@ -83,15 +86,15 @@ static const struct option kOptions[] = {
     {"verbose", no_argument, nullptr, OTBR_OPT_VERBOSE},
     {"version", no_argument, nullptr, OTBR_OPT_VERSION},
     {"radio-version", no_argument, nullptr, OTBR_OPT_RADIO_VERSION},
-    {"auto-attach", required_argument, nullptr, OTBR_OPT_AUTO_ATTACH},
+    {"auto-attach", optional_argument, nullptr, OTBR_OPT_AUTO_ATTACH},
     {0, 0, 0, 0}};
 
 static void PrintHelp(const char *aProgramName)
 {
     fprintf(stderr,
-            "Usage: %s [-I interfaceName] [-B backboneIfName] [-d DEBUG_LEVEL] [-v] [-a enableAutoAttach] RADIO_URL "
+            "Usage: %s [-I interfaceName] [-B backboneIfName] [-d DEBUG_LEVEL] [-v] [--auto-attach[=0/1]] RADIO_URL "
             "[RADIO_URL]\n"
-            "    enableAutoAttach defaults to 1\n",
+            "    --auto-attach defaults to 1\n",
             aProgramName);
     fprintf(stderr, "%s", otSysGetRadioUrlHelpString());
 }
@@ -155,7 +158,7 @@ static int realmain(int argc, char *argv[])
 
     std::set_new_handler(OnAllocateFailed);
 
-    while ((opt = getopt_long(argc, argv, "B:d:hI:Vva:", kOptions, nullptr)) != -1)
+    while ((opt = getopt_long(argc, argv, "B:d:hI:Vv", kOptions, nullptr)) != -1)
     {
         switch (opt)
         {
@@ -191,7 +194,7 @@ static int realmain(int argc, char *argv[])
             break;
 
         case OTBR_OPT_AUTO_ATTACH:
-            enableAutoAttach = atoi(optarg);
+            enableAutoAttach = (optarg == nullptr) || atoi(optarg);
             break;
 
         default:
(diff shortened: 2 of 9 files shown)


src/agent/main.cpp
```diff
@@ -82,11 +83,14 @@ static const struct option kOptions[] = {
     {"verbose", no_argument, nullptr, OTBR_OPT_VERBOSE},
     {"version", no_argument, nullptr, OTBR_OPT_VERSION},
     {"radio-version", no_argument, nullptr, OTBR_OPT_RADIO_VERSION},
+    {"auto-attach", required_argument, nullptr, OTBR_OPT_AUTO_ATTACH},
     {0, 0, 0, 0}};
 
 static void PrintHelp(const char *aProgramName)
 {
-    fprintf(stderr, "Usage: %s [-I interfaceName] [-B backboneIfName] [-d DEBUG_LEVEL] [-v] RADIO_URL [RADIO_URL]\n",
+    fprintf(stderr,
```
bukepo, line 91: Should we declare the default behavior without specifying `--enable-auto-attach`?

src/agent/main.cpp
```diff
@@ -145,11 +149,12 @@ static int realmain(int argc, char *argv[])
     const char *              backboneInterfaceName = "";
     bool                      verbose               = false;
     bool                      printRadioVersion     = false;
+    bool                      enableAutoAttach      = true;
     std::vector<const char *> radioUrls;
 
     std::set_new_handler(OnAllocateFailed);
 
-    while ((opt = getopt_long(argc, argv, "B:d:hI:Vv", kOptions, nullptr)) != -1)
+    while ((opt = getopt_long(argc, argv, "B:d:hI:Vva:", kOptions, nullptr)) != -1)
```
bukepo, line 157:
```suggestion
    while ((opt = getopt_long(argc, argv, ":B:d:hI:Vva:", kOptions, nullptr)) != -1)
```
Make `--enable-auto-attach` optional? https://pubs.opengroup.org/onlinepubs/9699919799/functions/getopt.html

src/ncp/ncp_openthread.hpp
```diff
@@ -198,6 +200,7 @@ class ControllerOpenThread : public MainloopProcessor
     std::vector<std::function<void(void)>>     mResetHandlers;
     TaskRunner                                 mTaskRunner;
     std::vector<ThreadStateChangedCallback>    mThreadStateChangedCallbacks;
+    bool                                       mEnableAutoAttach;
```
bukepo, line 203:
```suggestion
    bool                                       mEnableAutoAttach = false;
```

src/agent/main.cpp
jinran-google: It is optional. Defaults to `true`. It looks like adding a colon doesn't change the behavior.

src/agent/main.cpp
jinran-google: Done.

src/ncp/ncp_openthread.hpp
jinran-google: Done.

src/agent/main.cpp
bukepo: it allows optional argument.

jinran-google: It is already optional without the colon. You can see the test script that checks the default behavior.

simonlingoogle:
I recall that I created a similar PR: https://github.com/openthread/ot-br-posix/pull/1155

src/agent/main.cpp
```diff
@@ -66,6 +66,7 @@ enum
     OTBR_OPT_INTERFACE_NAME          = 'I',
     OTBR_OPT_VERBOSE                 = 'v',
     OTBR_OPT_VERSION                 = 'V',
+    OTBR_OPT_AUTO_ATTACH             = 'a',
```
simonlingoogle, line 69: maybe we don't need a short cut, `--auto-attach` is good enough?

src/ncp/ncp_openthread.cpp
```diff
@@ -245,7 +247,7 @@ bool ControllerOpenThread::IsAutoAttachEnabled(void)
     const char *val = getenv("OTBR_NO_AUTO_ATTACH");
 
     // Auto Thread attaching is enabled if OTBR_NO_AUTO_ATTACH is unset, empty or "0"
-    return (val == nullptr || !strcmp(val, "") || !strcmp(val, "0"));
+    return (val == nullptr || !strcmp(val, "") || !strcmp(val, "0")) && mEnableAutoAttach;
```
simonlingoogle, line 250: We can remove `OTBR_NO_AUTO_ATTACH` logic to always use `--auto-attach` ?

We can add `--auto-attach 0` arguments when exec to spawn the new process.

src/agent/main.cpp
jinran-google: Done.

src/ncp/ncp_openthread.cpp
jinran-google: Done.

bukepo (approved):

src/agent/main.cpp
```diff
@@ -184,6 +193,10 @@ static int realmain(int argc, char *argv[])
             printRadioVersion = true;
             break;
 
+        case OTBR_OPT_AUTO_ATTACH:
+            enableAutoAttach = (optarg == nullptr) || atoi(optarg);
```
simonlingoogle, line 197: What happens if `optarg` is not a valid int? 

I guess we need to catch the error by using `strtol` instead, and notify user of the invalid argument. 

Thoughts?

src/agent/main.cpp
```diff
@@ -248,7 +261,16 @@ int main(int argc, char *argv[])
...
-        execvp(argv[0], argv);
+        std::vector<std::string> args(argv, argv + argc);
+        args.erase(std::remove_if(args.begin(), args.end(),
+                                  [](const std::string &arg) { return arg.rfind("--auto-attach", 0) == 0; }),
+                   args.end());
+        args.push_back("--auto-attach=0");
+        std::vector<char *> args_cstr(args.size() + 1);
+        std::transform(args.begin(), args.end(), args_cstr.begin(),
+                       [](const std::string &arg) { return const_cast<char *>(arg.c_str()); });
+        *args_cstr.rbegin() = nullptr;
```
simonlingoogle, line 272: nit: may add a function to encapsulate the logic?

src/agent/main.cpp
```diff
@@ -248,7 +261,16 @@ int main(int argc, char *argv[])
 #if OPENTHREAD_ENABLE_COVERAGE
         __gcov_flush();
 #endif
-        execvp(argv[0], argv);
+        std::vector<std::string> args(argv, argv + argc);
```
simonlingoogle, line 264:
```suggestion
        std::vector<const char *> args(argv, argv + argc);
```

nit: seems simpler to use `const char *` directly.

simonlingoogle (approved):
LGTM 👍 

with nits

src/agent/main.cpp
simonlingoogle: I guess we need to avoid exception according to https://google.github.io/styleguide/cppguide.html#Exceptions

Thoughts? @bukepo

src/agent/main.cpp
jinran-google: Done.

src/agent/main.cpp
jinran-google: Done.

src/agent/main.cpp
jinran-google: Agreed not to use exceptions. Done with `strtol`.

src/agent/main.cpp
```diff
@@ -244,11 +297,18 @@ int main(int argc, char *argv[])
 {
     if (setjmp(sResetJump))
     {
+        std::vector<const char *> args = AppendAutoAttachDisablingArg(argc, argv);
```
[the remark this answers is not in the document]
simonlingoogle: Why are the strings immutable?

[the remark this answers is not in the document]
jinran-google: The added `--auto-attach=0` is a string literal, so it's immutable.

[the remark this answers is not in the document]
simonlingoogle: > The added `--auto-attach=0` is a string literal, so it's immutable.

How do we know that it's safe to access `--auto-attach=0` string literal in `exec` ? 

Moreover, can we avoid using string literal directly? For example, we can get a `char *` by copying `--auto-attach=0` to a static buffer:
```c++
static char sNoAutoAttachArg[] = "--auto-attach=0";
args.append(&sNoAutoAttachArg);
```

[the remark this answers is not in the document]
jinran-google: The POSIX spec guarantees that
> The argv[] and envp[] arrays of pointers and the strings to which those arrays point shall not be modified by a call to one of the exec functions, except as a consequence of replacing the process image.

https://pubs.opengroup.org/onlinepubs/009604499/functions/exec.html

Using a static buffer SGTM.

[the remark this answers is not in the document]
jinran-google: Done.

simonlingoogle (approved):
LGTM 👍 

Thanks.

src/agent/application.hpp
```diff
@@ -80,12 +80,16 @@ class Application : private NonCopyable
     /**
      * This constructor initializes the Application instance.
      *
-     * @param[in] aOpenThread  A reference to the OpenThread instance.
+     * @param[in] aInterfaceName         Name of the Thread network interface.
+     * @param[in] aBackboneInterfaceName Name of the backbone network interface.
+     * @param[in] aRadioUrls             The radio URLs (can be IEEE802.15.4 or TREL radio).
+     * @param[in] aEnableAutoAttach      Whether or not automatically attaches to the saved network.
```
jwhui, line 86:
```suggestion
     * @param[in] aEnableAutoAttach      Whether or not to automatically attach to the saved network.
```

src/agent/main.cpp
```diff
@@ -82,11 +86,51 @@ static const struct option kOptions[] = {
...
+    aOutResult = result;
+
+exit:
+    return successful;
+}
+
+static constexpr char kAutoAttachDisablingArg[] = "--auto-attach=0";
+static char           sAutoAttachDisablingArgStorage[sizeof(kAutoAttachDisablingArg)];
+
+static std::vector<char *> AppendAutoAttachDisablingArg(int argc, char *argv[])
```
jwhui, line 113: nit:
```suggestion
static constexpr char kAutoAttachDisableArg[] = "--auto-attach=0";
static char           sAutoAttachDisablingArgStorage[sizeof(kAutoAttachDisableArg)];

static std::vector<char *> AppendAutoAttachDisableArg(int argc, char *argv[])
```

jwhui (approved):

PR #1430, opened 2022-06-15
[dbus] return delay timer from `AttachAllNodesTo`

jinran-google:
This commit:
1. Makes the DBus API `AttachAllNodesTo` return the delay between the method returns and the dataset takes effect.
2. Ensures that when `AttachAllNodesTo` returns, the node is attached.

### DIFF
--- a/src/common/code_utils.hpp
+++ b/src/common/code_utils.hpp
@@ -175,4 +175,62 @@ protected:
     NonCopyable(void) = default;
 };
 
+template <typename T> class Optional
+{
+public:
+    constexpr Optional(void) = default;
+
+    Optional(T aValue)
+    {
+        new (&mStorage) T(aValue);
+        mHasValue = true;
+    }
+
+    ~Optional(void)
+    {
+        if (mHasValue)
+        {
+            GetValue().~T();
+        }
+    }
+
+    Optional(const Optional &aOther)
+    {
+        if (aOther.mHasValue)
+        {
+            new (&mStorage) T(aOther.GetValue());
+        }
+        mHasValue = aOther.mHasValue;
+    }
+
+    Optional &operator=(const Optional &aOther)
+    {
+        if (mHasValue)
+        {
+            GetValue().~T();
+        }
+        if (aOther.mHasValue)
+        {
+            new (&mStorage) T(aOther.GetValue());
+        }
+        mHasValue = aOther.mHasValue;
+    }
+
+    constexpr const T *operator->(void)const { return &GetValue(); }
+
+    constexpr const T &operator*(void)const { return GetValue(); }
+
+    constexpr bool HasValue(void) const { return mHasValue; }
+
+private:
+    T &GetValue(void) const
+    {
+        VerifyOrDie(mHasValue, "Optional doesn't contain a value");
+        return *const_cast<T *>(reinterpret_cast<const T *>(&mStorage));
+    }
+
+    alignas(T) unsigned char mStorage[sizeof(T)];
+    bool mHasValue = false;
+};
+
 #endif // OTBR_COMMON_CODE_UTILS_HPP_
--- a/src/common/code_utils.hpp
+++ b/src/common/code_utils.hpp
@@ -180,41 +180,13 @@ template <typename T> class Optional
 public:
     constexpr Optional(void) = default;
 
-    Optional(T aValue)
-    {
-        new (&mStorage) T(aValue);
-        mHasValue = true;
-    }
+    Optional(T aValue) { SetValue(aValue); }
 
-    ~Optional(void)
-    {
-        if (mHasValue)
-        {
-            GetValue().~T();
-        }
-    }
+    ~Optional(void) { ClearValue(); }
 
-    Optional(const Optional &aOther)
-    {
-        if (aOther.mHasValue)
-        {
-            new (&mStorage) T(aOther.GetValue());
-        }
-        mHasValue = aOther.mHasValue;
-    }
+    Optional(const Optional &aOther) { AssignFrom(aOther); }
 
-    Optional &operator=(const Optional &aOther)
-    {
-        if (mHasValue)
-        {
-            GetValue().~T();
-        }
-        if (aOther.mHasValue)
-        {
-            new (&mStorage) T(aOther.GetValue());
-        }
-        mHasValue = aOther.mHasValue;
-    }
+    Optional &operator=(const Optional &aOther) { AssignFrom(aOther); }
 
     constexpr const T *operator->(void)const { return &GetValue(); }
 
@@ -227,6 +199,31 @@ private:
     {
         VerifyOrDie(mHasValue, "Optional doesn't contain a value");
         return *const_cast<T *>(reinterpret_cast<const T *>(&mStorage));
+    }
+
+    void ClearValue(void)
+    {
+        if (mHasValue)
+        {
+            GetValue().~T();
+            mHasValue = false;
+        }
+    }
+
+    void SetValue(const T &aValue)
+    {
+        ClearValue();
+        new (&mStorage) T(aValue);
+        mHasValue = true;
+    }
+
+    void AssignFrom(const Optional &aOther)
+    {
+        ClearValue();
+        if (aOther.mHasValue)
+        {
+            SetValue(aOther.GetValue());
+        }
     }
 
     alignas(T) unsigned char mStorage[sizeof(T)];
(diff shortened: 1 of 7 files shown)


src/common/code_utils.hpp
```diff
@@ -175,4 +175,62 @@ class NonCopyable
...
+    {
+        if (mHasValue)
+        {
+            GetValue().~T();
+        }
+        if (aOther.mHasValue)
+        {
+            new (&mStorage) T(aOther.GetValue());
+        }
+        mHasValue = aOther.mHasValue;
```
simonlingoogle, line 216:
```suggestion
        ClearValue();
        SetValue(aOther.GetValue());
```

seems these methods can be reused in multiple places.

src/common/code_utils.hpp
```diff
@@ -175,4 +175,62 @@ class NonCopyable
...
+    constexpr const T *operator->(void)const { return &GetValue(); }
+
+    constexpr const T &operator*(void)const { return GetValue(); }
+
+    constexpr bool HasValue(void) const { return mHasValue; }
+
+private:
+    T &GetValue(void) const
+    {
+        VerifyOrDie(mHasValue, "Optional doesn't contain a value");
```
simonlingoogle, line 228: why not use assert?
I think alling `GetValue` when there is no value is an logical error, thus `assert` should be most suitable.

src/utils/thread_helper.cpp
```diff
@@ -581,25 +582,27 @@ void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, Re
...
         {
+            mAttachDelayMs            = kDelayTimerMilliseconds;
             mAttachHandler            = aHandler;
             mAttachPendingDatasetTlvs = datasetTlvs;
             ExitNow();
         }
         else
         {
-            aHandler(OT_ERROR_NONE);
+            aHandler(OT_ERROR_NONE, 0);
```
wgtdkp, line 592: Can we make the D-bus API async? For example, the handler is caller only when the BR becomes active (when the API call is truly success).

wgtdkp: Currently, when returned immediately after executing the d-bus command but we doesn't guarantee that the Border Router really did attach successfully. In the case we return `0` delay to the thread deamon, deamon has no idea when Border Router will become active, hence the mobile (GMS Core and iGHA) doesn't know when the Border Router becomes active. For iGHA, it has to wait for some time (how long?) for the Border Router becoming active before it can add the network to iOS, because iOS does check mDNS services when adding new networks.

iGHA doesn't have a good way to decide how long it need to wait before trying to add to iOS, so it has to try and backoff which is not elegant.

src/dbus/server/dbus_request.hpp
```diff
@@ -162,6 +165,19 @@ class DBusRequest
             reply = UniqueDBusMessage(dbus_message_new_error(mMessage, ConvertToDBusErrorName(aError), nullptr));
         }
 
+        if (aResult.HasValue())
+        {
+            DBusMessageIter replyIter;
+            otbrError       error;
+
+            dbus_message_iter_init_append(reply.get(), &replyIter);
+            error = DBusMessageEncode(&replyIter, *aResult);
```
simonlingoogle, line 174: Should we die on such error ?

src/common/code_utils.hpp
jinran-google: Done.

src/common/code_utils.hpp
jinran-google: `assert` depends on the build configuration. I prefer to make it a consistent error checking, so users of this class won't trigger undefined behavior no matter how they use this class. Thoughts?

src/utils/thread_helper.cpp
jinran-google: Makes sense. I made the DBus call wait until it attaches then return.

src/dbus/server/dbus_request.hpp
jinran-google: It should die, I think. But it doesn't die on error of `dbus_message_new_method_return`. Anyway, I made both cases die.

src/common/code_utils.hpp
simonlingoogle: I don't think that's a reason to not use `assert`. 

`assert` should be used for logical errors, while `VerifyOrDie` should be used for fatal errors that can't be recovered.

wgtdkp: > I think it's a good practice to never turn off assert in release builds, but that might be another topic.

If you have this idea in your mind, it means you are wanting something which shouldn't be provided by `assert`. `assert` should be just treated as a comment for helping understand the logic (i.e. what assumptions this method takes to work properly). When deciding whether we should use `assert` or `VerifyOrDie`, ask yourself if simply leaving a comment here is what you want.

wgtdkp:
@jinran-google Could you summary the changes in this commit? For example, 1. what's the problem and how it can be resolved by this change? 2. Are there potential cases this PR doesn't cover?

Looks like you are accepting my suggestion of calling the handler after network attach, but I believe this is applied to cases we don't need to use pending dataset. We better document this behavior change.

src/common/code_utils.hpp
simonlingoogle: I think we are already using `OT_ASSERT` extensively in OpenThread to catch logical errors. 
The default definition of `OT_ASSERT` on Posix platform is `assert`, but it's possible to provide platform-specific assert behaviors if the default `assert` is not exactly what you want.
 
So, I think it's the same practice in ot-br-posix that we should use `assert`. 

Maybe we should also define `OTBR_ASSERT` to replace the standard library `assert`, but I guess it's beyond the scope of this PR.

jinran-google: I agree with @wgtdkp that assertions should not be relied on, because in release builds they are indeed just comments.

For this particular case, I looked up the code base and found that `VerifyOrDie` is for unrecoverable external errors. I think it's also OK to treat this as a comment. Although that would require developers to be more cautious, being consistent on the function purposes is probably more important. Updated.

jinran-google:
> @jinran-google Could you summary the changes in this commit? For example, 1. what's the problem and how it can be resolved by this change? 2. Are there potential cases this PR doesn't cover?
> 
> Looks like you are accepting my suggestion of calling the handler after network attach, but I believe this is applied to only cases we don't need to use pending dataset. We better document this behavior change.

Updated the PR description. The PR is to let the caller know when the requested dataset becomes operative and I don't think there are cases not covered. After this PR, the caller can always expect the dataset being active after the method returns + the delay it returns (except that pending dataset is changed by some other means during the pending delay timer).

wgtdkp:
Thanks @jinran-google 

> 2. Ensures that when AttachAllNodesTo returns, the node is attached.

Is this accurate? I think the node is attached when the method returns only if we don't need to use Pending Dataset, right?

wgtdkp:
> > Thanks @jinran-google
> > > 2. Ensures that when AttachAllNodesTo returns, the node is attached.
> > 
> > 
> > Is this accurate? I think the node is attached when the method returns only if we don't need to use Pending Dataset, right?
> 
> If we are using pending dataset, the node must be in an attached state in the first place to send the MGMT request. It's just that the active dataset isn't the requested one until a couple of minutes later, that's the purpose of the returned delay value.

Okay, it is correct but **what we care is whether the BR is currently active in the designated network, but not the old one.** I think the always-attached guarantee won't be useful if we don't guarantee which network the BR is in. So my suggestion is: making it clear that, if the returned value is 0, the BR is already attached to the designated network; otherwise, it is still in the old network and will be in the designated network after the delay. **We should document this as part of the d-bus API signature**.

simonlingoogle (approved):
LGTM 👍

src/dbus/server/introspect.xml
```diff
@@ -60,9 +60,15 @@
     <!-- AttachAllNodesTo: Request to attach all nodes to the specified Thread network.
       @dataset: The Operational Dataset that contains parameter values of the Thread network
                 to attach to. It must be a full dataset.
+      @delay_ms: The delay between the method returns and the dataset takes effect, in
+                 milliseconds. If this value is 0, then the node is attached to the given network
+                 when this method returns. If this value is not 0, then the node is attached to
+                 its existing network when this method returns, and will attach to the given
+                 network after the delay.
```
wgtdkp, line 67: LGTM, thanks!

wgtdkp (approved):
LGTM 👍

bukepo (approved):

jwhui (approved):

darconeous:
It would be really great if the logic for AttachAllNodesTo could be moved to OpenThread, so that it could be a simple OpenThread call.

PR #1434, opened 2022-06-21
[dbus] add Thread Eui64 property

jinran-google:
This commit adds the Eui64 property of Thread interface. The property
is a unsigned 64bit number whose big-endian encoding is the same as the
EUI-64 value assigned by factory.


simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #1439, opened 2022-06-24
[tests] fix DBus intermittent failures

jinran-google:
This commit adds some delay to DBus client tests to give the network data some time to propagate.


superwhd (approved):
LGTM :+1:

PR #1454, opened 2022-07-11
[dbus] add RCP metrics properties

jinran-google:


src/dbus/common/constants.hpp
```diff
@@ -99,6 +99,18 @@
 #define OTBR_DBUS_PROPERTY_THREAD_VERSION "ThreadVersion"
 #define OTBR_DBUS_PROPERTY_EUI64 "Eui64"
 #define OTBR_DBUS_PROPERTY_MDNS_TELEMETRY_INFO "MdnsTelemetryInfo"
+#define OTBR_DBUS_PROPERTY_RCP_TIMEOUT_COUNT "RcpTimeoutCount"
```
zhanglongxia, line 102: Can we put these counters into two structures and use only two APIs to get these metrics?

jinran-google: We can, but I think using individual counters makes the application more flexible on querying only the needed subset of the properties.

Besides, grouping them into structures makes getting all properties in one call more difficult, because `GetProperties` returns an array of variants, if we have 2 APIs returning 2 structs, the application will see it returns "2 variants of structs", all structs share the same DBus type, but the structs are actually different, so the application needs more work to look at the data inside the structs to determine their actual types.

wgtdkp: +1

jinran-google: On second thought, I think the application calling the DBus API can do it by checking the argument types in the structs. But I still think providing the flexibility of choosing any subset of these properties is good. @bukepo any thoughts?

bukepo: it seems other counters are already structured. Maybe we should keep it consistent?

jinran-google: Consistency makes sense. Done.

wgtdkp (approved):
LGTM 👍

bukepo (approved):

zhanglongxia (approved):
LGTM

jwhui (approved):

PR #1472, opened 2022-07-22
[task-runner] support cancelling delayed tasks

simonlingoogle:


tests/unit/test_task_runner.cpp
```diff
@@ -252,6 +252,83 @@ TEST(TaskRunner, TestDelayedTasksOrder)
...
+
+    tid1 = taskRunner.Post(std::chrono::milliseconds(10), [&]() { str.push_back('a'); });
+    tid2 = taskRunner.Post(std::chrono::milliseconds(20), [&]() { str.push_back('b'); });
+    tid3 = taskRunner.Post(std::chrono::milliseconds(30), [&]() { str.push_back('c'); });
+
+    CHECK(0 < tid1);
+    CHECK(tid1 < tid2);
+    CHECK(tid2 < tid3);
+
+    taskRunner.Cancel(tid2);
```
wgtdkp, line 269: Test also if we can successfully cancel the task after mainloop is started? We can do this by cancelling this task in a fourth task:

```
taskRunner.Post(std::chrono::milliseconds(10), [&]() { taskRunner.Cancel(tid2); });
```

or in a new Thread:
```
std::thread([&]() { sleep(10.0/1000); taskRunner.Cancel(tid2); });
```

src/common/task_runner.cpp
```diff
@@ -75,9 +75,9 @@ void TaskRunner::Post(const Task<void> aTask)
     Post(Milliseconds::zero(), std::move(aTask));
 }
 
-void TaskRunner::Post(Milliseconds aDelay, const Task<void> aTask)
+TaskRunner::TaskId TaskRunner::Post(Milliseconds aDelay, const Task<void> aTask)
 {
-    PushTask(aDelay, std::move(aTask));
+    return PushTask(aDelay, std::move(aTask));
```
superwhd, line 80: Not related to this PR, but it cannot trigger the move constructor because `aTask` is a `const Task<void> &&` after `std::move` instead of a **non-const** rvalue reference.

wgtdkp (approved):
LGTM 👍

tests/unit/test_task_runner.cpp
simonlingoogle: That's a good idea.

simonlingoogle: Test enhanced.

src/common/task_runner.cpp
superwhd: Will do later.

superwhd (approved):
LGTM :+1:

tests/unit/test_task_runner.cpp
```diff
@@ -252,6 +253,60 @@ TEST(TaskRunner, TestDelayedTasksOrder)
...
+    CHECK(0 < tid1);
+    CHECK(tid1 < tid2);
+    CHECK(tid2 < tid3);
+    CHECK(tid3 < tid4);
+    CHECK(tid4 < tid5);
+
+    taskRunner.Cancel(tid2);
+
+    taskRunner.Post(std::chrono::milliseconds(10), [&]() { taskRunner.Cancel(tid3); });
+    std::thread([&]() {
```
wgtdkp, line 277: nit: join the thread at end of this test case for just making it stable.

simonlingoogle: done

jwhui (approved):

PR #1474, opened 2022-07-25
[task-runner] fix `std::move` on const values

superwhd:
There is an issue in `TaskRunner` that it declares `Task<void>` as a const parameter in interfaces (e.g. `void TaskRunner::Post(const Task<void> aTask)`). In the implementation it `std::move`s the `aTask` and pushed it into an STL container. This `std::move` on the const value doesn't lead to the move constructor of `Task<void>` which may have better performance. It leads to the copy constructor instead. To fix this, we need to make the `aTask` a non-const parameter.

### DIFF
--- a/src/common/task_runner.cpp
+++ b/src/common/task_runner.cpp
@@ -70,12 +70,12 @@ TaskRunner::~TaskRunner(void)
     }
 }
 
-void TaskRunner::Post(const Task<void> aTask)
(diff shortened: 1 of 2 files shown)


wgtdkp (approved):
LGTM :+1:

jwhui (approved):

PR #1485, opened 2022-08-02
[tests] upgrade pip before installing cmake

jwhui:
To avoid errors such as this: https://github.com/openthread/ot-br-posix/runs/7634987232?check_suite_focus=true


abtink (approved):
Thanks. 👍

PR #1491, opened 2022-08-09
[advertising-proxy] support publishing multiple host addresses

superwhd:
This PR:

- Allows Advertising Proxy to publish multiple host addresses. 
- Allows Publisher to publish multiple host addresses.

Test is covered in https://github.com/openthread/openthread/pull/8003/.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -60,11 +60,13 @@ void Publisher::PublishService(const std::string &aHostName,
     PublishServiceImpl(aHostName, aName, aType, aSubTypeList, aPort, aTxtList, std::move(aCallback));
 }
 
-void Publisher::PublishHost(const std::string &aName, const std::vector<uint8_t> &aAddress, ResultCallback &&aCallback)
+void Publisher::PublishHost(const std::string &            aName,
+                            const std::vector<Ip6Address> &aAddresses,
+                            ResultCallback &&              aCallback)
 {
     mHostRegistrationBeginTime[aName] = Clock::now();
 
-    PublishHostImpl(aName, aAddress, std::move(aCallback));
+    PublishHostImpl(aName, aAddresses, std::move(aCallback));
 }
 
 void Publisher::OnServiceResolveFailed(const std::string &aType, const std::string &aInstanceName, int32_t aErrorCode)
@@ -236,6 +238,12 @@ Publisher::TxtList Publisher::SortTxtList(TxtList aTxtList)
     return aTxtList;
 }
 
+Publisher::AddressList Publisher::SortAddressList(AddressList aAddressList)
+{
+    std::sort(aAddressList.begin(), aAddressList.end());
+    return aAddressList;
+}
+
 std::string Publisher::MakeFullServiceName(const std::string &aName, const std::string &aType)
 {
     return aName + "." + aType + ".local";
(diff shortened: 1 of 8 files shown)


src/mdns/mdns_avahi.cpp
```diff
@@ -718,38 +718,43 @@ void PublisherAvahi::UnpublishService(const std::string &aName, const std::strin
...
-    address.proto = AVAHI_PROTO_INET6;
-    memcpy(address.data.ipv6.address, aAddress.data(), aAddress.size());
     fullHostName = MakeFullHostName(aName);
+    for (const auto &address : aAddresses)
+    {
+        AvahiAddress avahiAddress;
+
+        avahiAddress.proto = AVAHI_PROTO_INET6;
+        memcpy(avahiAddress.data.ipv6.address, address.m8, sizeof(address.m8));
+        avahiError = avahi_entry_group_add_address(group, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, AVAHI_PUBLISH_NO_REVERSE,
```
simonlingoogle, line 746: Are we ignoring avahi errors intentionally?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -366,26 +366,33 @@ PublisherMDnsSd::DnssdServiceRegistration::~DnssdServiceRegistration(void)
 
 PublisherMDnsSd::DnssdHostRegistration::~DnssdHostRegistration(void)
 {
-    VerifyOrExit(mServiceRef != nullptr && mRecordRef != nullptr);
+    int dnsError;
+    VerifyOrExit(mServiceRef != nullptr);
```
simonlingoogle, line 370:
```suggestion
    int dnsError;

    VerifyOrExit(mServiceRef != nullptr);
```
nit

src/mdns/mdns.cpp
```diff
@@ -236,6 +238,12 @@ Publisher::TxtList Publisher::SortTxtList(TxtList aTxtList)
     return aTxtList;
 }
 
+Publisher::AddressList Publisher::SortAddressList(AddressList aAddressList)
+{
+    std::sort(aAddressList.begin(), aAddressList.end());
+    return aAddressList;
+}
```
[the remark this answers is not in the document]
superwhd: I guess you meant `void Publisher::SortAddressList(AddressList &aAddressList)`.

There are some reasons I made the method signature like this:
- It's more convenient to let it return the sorted list so that we can call `SortAddressList` in the member intializer list.
- This method signature follows the same style as `SortTxtList` and `SortSubTypeList`.
- There's no performance degradation to return the sorted list since we're using move constructors.

src/mdns/mdns_avahi.cpp
superwhd: Good catch! I misplaced the `VerifyOrExit`.

simonlingoogle: From another perspective, I wonder if we should continue committing the group as long as one address is successfully registered. 
Not sure if it could happen at all in practise.

src/mdns/mdns.cpp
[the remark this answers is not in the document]
simonlingoogle: The method prototype seems to imply that the original list is not modified, which is not the case?

src/mdns/mdns_avahi.cpp
superwhd: I took a look at `avahi/avahi-utils/avahi-publish.c` and I think this case is unlikely to happen.

simonlingoogle (approved):
LGTM 👍

wgtdkp (approved):
LGTM 👍

src/mdns/mdns.cpp
```diff
@@ -374,6 +382,7 @@ void Publisher::RemoveHostRegistration(const std::string &aName, otbrError aErro
     hostReg = std::move(it->second);
     mHostRegistrations.erase(it);
     hostReg->Complete(aError);
+    otbrLogInfo("Removing host %s Done", aName.c_str());
```
jwhui, line 385: nit:
```suggestion
    otbrLogInfo("Removed host %s", aName.c_str());
```

jwhui (approved):

PR #1493, opened 2022-08-10
[agent] save region code

bukepo:
This commit adds a feature for otbr-agent to save the region code set by d-bus
API and apply this region code after restart.

Verified by manual test and new integration test.

### DIFF
--- a/script/test
+++ b/script/test
@@ -110,9 +110,10 @@ do_build()
         "-DCMAKE_INSTALL_PREFIX=/usr"
         "-DOT_THREAD_VERSION=1.3"
         "-DOTBR_DBUS=ON"
-        "-DOTBR_WEB=ON"
-        "-DOTBR_UNSECURE_JOIN=ON"
+        "-DOTBR_REGION_FILE=/tmp/otbr-region-file"
         "-DOTBR_TREL=ON"
+        "-DOTBR_UNSECURE_JOIN=ON"
+        "-DOTBR_WEB=ON"
         ${otbr_options[@]+"${otbr_options[@]}"}
     )
 
--- a/src/dbus/server/dbus_thread_object.cpp
+++ b/src/dbus/server/dbus_thread_object.cpp
@@ -247,6 +247,26 @@ otbrError DBusThreadObject::Init(void)
 
     SuccessOrExit(error = Signal(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_SIGNAL_READY, std::make_tuple()));
 
+#ifdef OTBR_REGION_FILE
+    {
+        FILE *regionFile = fopen(OTBR_REGION_FILE, "r");
+
+        VerifyOrExit(regionFile != nullptr);
+
+        char region[sizeof(uint16_t)];
+
+        if (fscanf(regionFile, "%c%c", &region[0], &region[1]) == 2)
+        {
+            uint16_t regionCode = (region[0] << 8 | region[1]);
+
+            otError err = otPlatRadioSetRegion(threadHelper->GetInstance(), regionCode);
+
+            otbrLogInfo("Apply saved region code %c%c: %s", region[0], region[1], otThreadErrorToString(err));
+        }
+
+        fclose(regionFile);
+    }
+#endif
 exit:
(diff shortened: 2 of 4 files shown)


script/test
```diff
@@ -110,9 +110,10 @@ do_build()
         "-DCMAKE_INSTALL_PREFIX=/usr"
         "-DOT_THREAD_VERSION=1.3"
         "-DOTBR_DBUS=ON"
-        "-DOTBR_WEB=ON"
-        "-DOTBR_UNSECURE_JOIN=ON"
+        "-DOTBR_REGION_FILE=/tmp/otbr-region-file"
```
jinran-google, line 113: Would it be better to use a command line argument instead of a build flag?

simonlingoogle, line 113: May not be a good idea to use `/tmp/` for persistent information.

Why not use OT settings?

tests/dbus/test_dbus_client.cpp
```diff
@@ -208,7 +208,20 @@ int main()
...
+        char savedRegion[sizeof("US")];
+
+        printf("reading saved region code: ");
+        FILE *regionFile = fopen(OTBR_REGION_FILE, "r");
+        fscanf(regionFile, "%s", savedRegion);
+        fclose(regionFile);
+        printf("%s\n", savedRegion);
+        TEST_ASSERT(strcmp(savedRegion, "US") == 0);
+    }
+#endif
```
jinran-google, line 224: Restart OTBR and check if region code is restored?

script/test
bukepo: The intended use case is recover the region code after crash. So I think `/tmp` is good enough.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -247,6 +247,26 @@ otbrError DBusThreadObject::Init(void)
 
     SuccessOrExit(error = Signal(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_SIGNAL_READY, std::make_tuple()));
 
+#ifdef OTBR_REGION_FILE
+    {
+        FILE *regionFile = fopen(OTBR_REGION_FILE, "r");
```
simonlingoogle, line 252: use `std::unique_ptr<FILE, decltype(&fclose)>` ?

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -247,6 +247,26 @@ otbrError DBusThreadObject::Init(void)
...
+
+        VerifyOrExit(regionFile != nullptr);
+
+        char region[sizeof(uint16_t)];
```
simonlingoogle, line 256: nit: move variables to the top?

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -247,6 +247,26 @@ otbrError DBusThreadObject::Init(void)
...
+
+        if (fscanf(regionFile, "%c%c", &region[0], &region[1]) == 2)
+        {
+            uint16_t regionCode = (region[0] << 8 | region[1]);
+
+            otError err = otPlatRadioSetRegion(threadHelper->GetInstance(), regionCode);
+
+            otbrLogInfo("Apply saved region code %c%c: %s", region[0], region[1], otThreadErrorToString(err));
```
simonlingoogle, line 264:
```suggestion
            otbrLogNote("Apply saved region code %c%c: %s", region[0], region[1], otThreadErrorToString(err));
```
Use note?

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -1176,6 +1196,14 @@ otError DBusThreadObject::SetRadioRegionHandler(DBusMessageIter &aIter)
 
     error = otPlatRadioSetRegion(threadHelper->GetInstance(), regionCode);
 
+#ifdef OTBR_REGION_FILE
+    {
+        FILE *regionFile = fopen(OTBR_REGION_FILE, "w");
```
simonlingoogle, line 1201: use `std::unique_ptr<FILE, decltype(&fclose)>` ?

PR #1496, opened 2022-08-11
[script] don't use cached version strings if re-installing OTBR

suveshpratapa:
Force re-evaluation of OTBR version strings (both OTBR version for otbr-agent daemon and OT version for the POSIX stack) if not supplied by the user.  This prevents them from using previously cached values from CMakeCache in case the project is being re-built over a previously installed / cached version.

### DIFF
--- a/script/_otbr
+++ b/script/_otbr
@@ -70,6 +70,9 @@ otbr_install()
         "-DOTBR_SRP_ADVERTISING_PROXY=ON"
         "-DOTBR_INFRA_IF_NAME=${INFRA_IF_NAME}"
         "-DOTBR_MDNS=${OTBR_MDNS:=mDNSResponder}"
+        # Force re-evaluation of version strings
+        "-DOTBR_VERSION="
+        "-DOT_PACKAGE_VERSION="
         "${otbr_options[@]}"
     )
 


suveshpratapa:
@jwhui @simonlingoogle

CMakeLists.txt
```diff
@@ -65,7 +65,8 @@ endif()
 
 message(STATUS "OTBR package name: ${OTBR_NAME}")
 
-if(NOT OTBR_VERSION)
+set(OTBR_VERSION "" CACHE STRING "OTBR Package Version")
```
suveshpratapa, line 68: Basically doing the same as OT_PACKAGE_VERSION in third_party/openthread/repo/CMakeLists.txt

suveshpratapa:
Requesting review @jwhui @simonlingoogle ... we want someone that accidentally updates the OTBR install (without cleaning build folders) to have the accurate version string.

simonlingoogle:
Not being an expert on `cmake`, what happens when we build with `-DOTBR_VERSION=1.0` and then without specifying `-DOTBR_VERSION`? @suveshpratapa

suveshpratapa:
> Not being an expert on `cmake`, what happens when we build with `-DOTBR_VERSION=1.0` and then without specifying `-DOTBR_VERSION`? @suveshpratapa

When a version is not specified, we have a calculation to insert the right version with git information included.

In your example, the user is specifying a "1.0" version that would then get cached in CMakeCache.  When the user re-builds without specifying a version:

- Prior to my change, it would use the cached version.
- After my change, it re-evaluates the version the proper way that it would if one didn't provide a version (so includes git information etc.)

simonlingoogle:
> * After my change, it re-evaluates the version the proper way that it would if one didn't provide a version (so includes git information etc.)

IIUC, that's true only when building OTBR using `script/setup`. It does not apply for normal cmake builds (e.g. using ./script/cmake-build). Is that right?

suveshpratapa:
> > * After my change, it re-evaluates the version the proper way that it would if one didn't provide a version (so includes git information etc.)
> 
> IIUC, that's true only when building OTBR using `script/setup`. It does not apply for normal cmake builds (e.g. using ./script/cmake-build). Is that right?

I've not run it with `./script/cmake-build` directly but it is called as part of `./script/_otbr` which gets run as part of `./script/setup`.  If someone is running `./script/cmake-build` directly, it seems they would have to clean up the CMakeCache or specify the version again in the OTBR string arguments to `cmake-build`.

In this scenario, the `./script/setup` process has a wrapper to take care of this (This is the documented way to install the OTBR, which is currently used by a lot of customers as well AFAIK).

simonlingoogle:
> > > * After my change, it re-evaluates the version the proper way that it would if one didn't provide a version (so includes git information etc.)
> > 
> > 
> > IIUC, that's true only when building OTBR using `script/setup`. It does not apply for normal cmake builds (e.g. using ./script/cmake-build). Is that right?
> 
> I've not run it with `./script/cmake-build` directly but it is called as part of `./script/_otbr` which gets run as part of `./script/setup`. If someone is running `./script/cmake-build` directly, it seems they would have to clean up the CMakeCache or specify the version again in the OTBR string arguments to `cmake-build`.
> 
> In this scenario, the `./script/setup` process has a wrapper to take care of this (This is the documented way to install the OTBR, which is currently used by a lot of customers as well AFAIK).

Should we use [Normal Variable](https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable) if we don't like `CACHE` for this variable?

bukepo (approved):
LGTM

suveshpratapa:
> Should we use [Normal Variable](https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable) if we don't like `CACHE` for this variable?

@simonlingoogle I tried this after your suggestion. For example, here's what I did with `OTBR_VERSION` (without any `CACHE` syntax):
```
set(OTBR_VERSION "")
```

However, after doing this, it still calculates the GitHub string even if the user provides their own version.  For reference, I usually do so by:
```
sudo INFRA_IF_NAME=eth0 -DOTBR_VERSION=user_version_string ./script/setup
```

Any thoughts as to why this won't work with the Normal Variable change? I'm also a CMake novice. :)

CMakeLists.txt
```diff
@@ -65,7 +65,8 @@ endif()
...
+if(OTBR_VERSION STREQUAL "")
```
simonlingoogle, line 69:
```suggestion
if (NOT OTBR_VERSION)
```

I think this PR tries to use -DOTBR_VERSION="" to force reevaluating the package version.

But isn't the original code already support that?

@bukepo @suveshpratapa

suveshpratapa: I just wanted it to happen automatically when `./script/setup` is run, as that's the documented method to install or re-install the OTBR. So I made it part of that call chain.

simonlingoogle: Is the change of this PR in `script/_otbr` already enough to fulfil the purpose?
I am trying to understand if the changes to `CMakeLists.txt` is necessary.

suveshpratapa: @simonlingoogle You're right. I wanted to make it be the same as `OT_PACKAGE_VERSION` in `third_party/openthread/CMakeLists.txt`, but it appears it wasn't necessary.  The change in `script/_otbr` is enough.  I reflected the change in the pull request, please take a look.

simonlingoogle (approved):
LGTM 👍

bukepo (approved):

jwhui (approved):

PR #1543, opened 2022-09-22
[nat64] make nat44 a service for native nat64 translator

xuyirio:
This PR updates nat44 as a service for the native nat64 translator, so that `NAT64_DYNAMIC_POOL` can be remembered and restored when otbr-agent restarts.


script/_nat64
```diff
@@ -151,9 +149,15 @@ nat44_install()
...
-        echo "      iptables -t nat -A POSTROUTING -o $IFNAME -j MASQUERADE" | sudo tee -a $NAT44_SERVICE
-    done
+    if [ "$NAT64_SERVICE" = tayga ]; then
+        # Although Tayga itself also configures a NAT44 iptables route, this iptables route is used with Tayga
+        # due to some history reason. It might be removed when native NAT64 service is ready.
+        for IFNAME in $WLAN_IFNAMES; do
+            echo "        iptables -t nat -A POSTROUTING -o $IFNAME -j MASQUERADE" | sudo tee -a $NAT44_SERVICE
+        done
+    else
+        echo "        iptables -t nat -A POSTROUTING -s \"$DYNAMIC_POOL\" -j MASQUERADE" | sudo tee -a $NAT44_SERVICE
```
erjiaqing, line 159: Would suggest adding the following rules for NAT64:

```
iptables -t filter -A FORWARD -o $IFNAME -j ACCEPT
iptables -t filter -A FORWARD -i $IFNAME -j ACCEPT
```

Since some platforms will drop all forwarding packets (by iptables) by default

erjiaqing (commented):
I found my iptables will drop all forwarding packets, would suggest adding two extra rules for NAT64.

erjiaqing (approved):

jwhui (approved):

PR #1548, opened 2022-09-27
[nat64] Use Openthread NAT64 translator by default when building docker images

erjiaqing:


erjiaqing:
dup https://github.com/openthread/ot-br-posix/pull/1539

PR #1577, opened 2022-10-20
[nat64] enable NAT64 on start

erjiaqing:
This PR adds the OTBR_NAT64 flag since we cannot use OT_CONFIG_NAT64_* flags in ot-br-posix.

Note: We will remove tayga support later, and https://github.com/openthread/ot-br-posix/pull/1539/ is merged, so in `script/_otbr` we only enable OTBR_NAT64 when NAT64 is set and NAT64_SERVICE is openthread.

### DIFF
--- a/src/ncp/ncp_openthread.cpp
+++ b/src/ncp/ncp_openthread.cpp
@@ -175,12 +175,11 @@ void ControllerOpenThread::Init(void)
 #if OTBR_ENABLE_SRP_ADVERTISING_PROXY
     otSrpServerSetEnabled(mInstance, /* aEnabled */ true);
 #endif
+#if OTBR_ENABLE_NAT64
+    otNat64SetEnabled(mInstance, /* aEnabled */ true);
+#endif
 
     mThreadHelper = std::unique_ptr<otbr::agent::ThreadHelper>(new otbr::agent::ThreadHelper(mInstance, this));
-
-#if OTBR_ENABLE_NAT64
-    otNat64SetEnabled(mInstance, true);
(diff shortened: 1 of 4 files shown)


src/ncp/ncp_openthread.cpp
```diff
@@ -177,6 +178,10 @@ void ControllerOpenThread::Init(void)
 
     mThreadHelper = std::unique_ptr<otbr::agent::ThreadHelper>(new otbr::agent::ThreadHelper(mInstance, this));
 
+#if OTBR_ENABLE_NAT64
+    otNat64SetEnabled(mInstance, true);
```
xuyirio, line 182: nit: s/`true`/`/* aEnabled */ true`

src/ncp/ncp_openthread.cpp
```diff
@@ -177,6 +178,10 @@ void ControllerOpenThread::Init(void)
...
+#endif
```
xuyirio, line 183: nit: move the block right next to the otSrpServerSetEnabled block since both are making enablement

xuyirio (approved):
Thanks, Song!

bukepo:
what's preventing us from using `OT_CONFIG_NAT64_*` flags?

erjiaqing:
> what's preventing us from using `OT_CONFIG_NAT64_*` flags?

We cannot get OPENTHREAD_CONFIG_NAT64* from the ot-br code.

bukepo (approved):

jwhui (approved):

PR #1587, opened 2022-10-31
[dbus] implement D-BUS API `GetBorderRoutingCounters`

superwhd:
This PR implements the D-BUS API `GetBorderRoutingCounters`. 

Test covered in https://github.com/openthread/openthread/pull/8341.

### DIFF
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -42,6 +42,11 @@ endif()
 
 option(OTBR_BORDER_ROUTING "Enable Border Routing Manager" OFF)
 
+option(OTBR_BORDER_ROUTING_COUNTERS "Enable Border Routing COUNTERS" ON)
+if (OTBR_BORDER_ROUTING AND OTBR_BORDER_ROUTING_COUNTERS)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_BORDER_ROUTING_COUNTERS=1)
+endif()
+
 option(OTBR_DBUS "Enable DBus support" OFF)
 if(OTBR_DBUS)
     pkg_check_modules(DBUS REQUIRED dbus-1)
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -42,7 +42,7 @@ endif()
 
 option(OTBR_BORDER_ROUTING "Enable Border Routing Manager" OFF)
 
-option(OTBR_BORDER_ROUTING_COUNTERS "Enable Border Routing COUNTERS" ON)
+option(OTBR_BORDER_ROUTING_COUNTERS "Enable Border Routing Counters" ON)
 if (OTBR_BORDER_ROUTING AND OTBR_BORDER_ROUTING_COUNTERS)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_BORDER_ROUTING_COUNTERS=1)
 endif()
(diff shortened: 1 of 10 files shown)


src/dbus/common/dbus_message_helper.hpp
```diff
@@ -286,6 +290,12 @@ template <> struct DBusTypeTrait<RadioCoexMetrics>
     static constexpr const char *TYPE_AS_STRING = "(uuuuuuuuuuuuuuuuuub)";
 };
 
+template <> struct DBusTypeTrait<BorderRoutingCounters>
+{
+    // struct of { uint32, uint32, uint32, uint32 }
```
jinran-google, line 295: Per the code it should be `{ { uint32, uint32 }, { uint32, uint32 }, { uint32, uint32 }, { uint32, uint32 } }`.

src/dbus/common/types.hpp
```diff
@@ -619,6 +619,20 @@ struct RadioCoexMetrics
     bool     mStopped;                            ///< Stats collection stopped due to saturation.
 };
 
+struct BorderRoutingCounters
+{
+    struct PacketsAndBytes
+    {
+        uint32_t mPackets; ///< The number of packets.
+        uint32_t mBytes;   ///< The number of bytes.
```
jinran-google, line 627: Is it better to use `int64` to prevent overflow?

jinran-google (approved):
LGTM overall.

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -253,6 +253,8 @@ otbrError DBusThreadObject::Init(void)
                                std::bind(&DBusThreadObject::GetUptimeHandler, this, _1));
     RegisterGetPropertyHandler(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_PROPERTY_RADIO_COEX_METRICS,
                                std::bind(&DBusThreadObject::GetRadioCoexMetrics, this, _1));
+    RegisterGetPropertyHandler(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_PROPERTY_BORDER_ROUTING_COUNTERS,
```
librasungirl, line 256: Should the condition `#if OTBR_ENABLE_BORDER_ROUTING_COUNTERS` be added here or in below implementation?

It seems to me the macro condition is mix-used in either the implementation body or when registering the handler, for instance

https://github.com/openthread/ot-br-posix/blob/6a8db6b66e1b5bb37ef38728bdf27834352367db/src/dbus/server/dbus_thread_object.cpp#L210-L215

Better to unify the style, and I personally prefer a bit to add the macro condition here as it seems useless to register the handler if it turns out to be not enabled.

superwhd: It seems ChannelMonitor is currently the only feature guarded at the declaration. I'd prefer to keep the style consistent with the majority. If we want to migrate to 'guard at declaration' style, we can send a new PR solely for this purpose.

librasungirl: Agree that may unify it in a separate PR.

Any preference on the guard style? @abtink @jwhui

librasungirl (approved):

etc/cmake/options.cmake
```diff
@@ -42,6 +42,11 @@ endif()
 
 option(OTBR_BORDER_ROUTING "Enable Border Routing Manager" OFF)
 
+option(OTBR_BORDER_ROUTING_COUNTERS "Enable Border Routing COUNTERS" ON)
```
jwhui, line 45:
```suggestion
option(OTBR_BORDER_ROUTING_COUNTERS "Enable Border Routing Counters" ON)
```

src/dbus/server/dbus_thread_object.cpp
jwhui: I agree it makes sense to guard each declaration, but can be done in a separate PR.

jwhui (approved):

PR #1598, opened 2022-11-08
[nat64] implement dBus API for NAT64

erjiaqing:
This PR implements the D-BUS API `SetNat64Enabled` and properties `Nat64State`, `Nat64Mappings`, `Nat64TrafficCounters`, `Nat64ErrorCounters`.

Test: openthread/openthread#8392


jinran-google (commented):
Add tests like https://github.com/openthread/ot-br-posix/blob/d9103922af71d5b0a46dd7f224fd7bc2244ca1b4/tests/dbus/test-client#L146 to cover the API?

Would be even better if we can actually verify the assignment of counter values.

superwhd:
FYI another way to implement a test case: https://github.com/openthread/openthread/pull/8341.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -369,6 +369,18 @@ class ThreadApiDBus
...
+     * This method sets the NAT64 switch.
+     *
+     * @param[in] aEnable  A boolean to enable/disable the NAT64.
+     *
+     * @retval ERROR_NONE  Successfully performed the dbus function call
+     * @retval ERROR_DBUS  dbus encode/decode error
+     * @retval ...         OpenThread defined error value otherwise
+     *
+     */
+    ClientError SetNat64Enabled(bool aEnabled);
```
xuyirio, line 382: Except for the test, we don't expect other caller of this API. Is this correct?

My understanding is that setting nat64 from feature flags list will be handled in `ControllerOpenThread::ApplyFeatureFlagList`.

xuyirio (commented):
Thanks, Song!

src/dbus/client/thread_api_dbus.hpp
erjiaqing: This is expected.

`ApplyFeatureFlagList` should call `otNat64SetEnabled` instead of this.

src/dbus/common/dbus_message_helper.hpp
```diff
@@ -299,6 +311,66 @@ template <> struct DBusTypeTrait<BorderRoutingCounters>
...
+
+template <> struct DBusTypeTrait<Nat64AddressMapping>
+{
+    // array of struct of { uint64,
+    //             array of uint8, array of uint8, uint32,
+    //             struct of {
+    //               struct of { uint64, uint64, uint64, uint64 }
+    //               struct of { uint64, uint64, uint64, uint64 }
+    //               struct of { uint64, uint64, uint64, uint64 }
+    //               struct of { uint64, uint64, uint64, uint64 } } }
```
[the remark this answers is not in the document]
erjiaqing: This is intentional.

Since this four structs are members of the outer struct.

erjiaqing:
> Add tests like
> 
> https://github.com/openthread/ot-br-posix/blob/d9103922af71d5b0a46dd7f224fd7bc2244ca1b4/tests/dbus/test-client#L146
> 
> to cover the API?
> Would be even better if we can actually verify the assignment of counter values.

There is a pending PR in openthread using the dbus api: https://github.com/openthread/openthread/actions/runs/3466935338/jobs/5791335591

Also added simple tests in the `test_dbus_client.cpp`

jinran-google (approved):
LGTM :+1:

xuyirio (approved):

jwhui (approved):

PR #1610, opened 2022-11-16
[scripts] fix: check if systemctl exists before using it

### LINKED ISSUE #1607: Add check for having `systemctl` before trying to disable
**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem is.
There is no check for `have systemctl` in the[ _disable_services](https://github.com/openthread/ot-br-posix/blob/main/script/_disable_services) script even though this check exists all around the project. This causes the setup to fail if the machine does not have the `systemctl` command.

**Describe the solution you'd like** A clear and concise description of what you want to happen.
Add the `if have systemctl; then` line before loop for disabling services that could cause problem to otbr.

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.
We are using a patch to skip this loop on our circleci builds that uses docker which does not have `systemctl` command installed.

a-cinar:
There is no check for having `systemctl` in the[ _disable_services](https://github.com/openthread/ot-br-posix/blob/main/script/_disable_services) script even though this check exists all around the project. This causes the setup to fail if the machine does not have the `systemctl` command.

Added necessary condition before using `systemctl`.

Fixes #1607

### DIFF
--- a/script/_disable_services
+++ b/script/_disable_services
@@ -44,13 +44,15 @@ bbb_disable_services()
     SERVICE_LIST="${SERVICE_LIST} nodered.service"
     SERVICE_LIST="${SERVICE_LIST} dnsmasq.service" # Disable well before bind9
 
-    for service in $SERVICE_LIST; do
-        if [ "$(sudo systemctl is-active "$service")" != "inactive" ]; then
-            for action in stop disable; do
-                sudo systemctl "$action" "$service"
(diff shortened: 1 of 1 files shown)


srickardti (approved):

jwhui (approved):
Thanks! 👍🏼

PR #1639, opened 2022-12-06
[nat64] update counter types from uint64_t to int64_t

erjiaqing:
We may want to avoid the use of uint64_t types as int64_t won't overflow for most cases.

This PR update the type of counters for nat64 to int64_t.

Some considerations: almost every existing API is using uint32_t / uint64_t for counters, should we still follow the existing pattern?


src/dbus/common/dbus_message_helper.hpp
```diff
@@ -313,29 +313,29 @@ template <> struct DBusTypeTrait<BorderRoutingCounters>
...
-    // struct of { struct of { uint64, uint64, uint64, uint64 }
-    //             struct of { uint64, uint64, uint64, uint64 }
-    //             struct of { uint64, uint64, uint64, uint64 }
-    //             struct of { uint64, uint64, uint64, uint64 } }
-    static constexpr const char *TYPE_AS_STRING = "((tttt)(tttt)(tttt)(tttt))";
+    // struct of { struct of { int64, int64, int64, int64 }
+    //             struct of { int64, int64, int64, int64 }
+    //             struct of { int64, int64, int64, int64 }
+    //             struct of { int64, int64, int64, int64 } }
+    static constexpr const char *TYPE_AS_STRING = "((xxxx)(xxxx)(xxxx)(xxxx))";
```
abtink, line 338: Same question here. What is the beneift of using `int64` instead of `uint64`?

Also we should consider if this breaks compatibility? (user of dbus).

erjiaqing: The idea is following the style guide of https://google.github.io/styleguide/cppguide.html#Integer_Types

Some benefits can be exposing errors (for integer overflow) eariler.

As these API are recently merged, I would expect the impact of the API change is minimal.

superwhd (approved):
LGTM overall.

Speaking of `int64` vs `uint64`, I think it's fine to keep using `uint64`. I think subtraction overflow is not a problem in `openthread` or `ot-br-posix` because we don't do subtractions for counters in these repos.

src/dbus/common/dbus_message_helper.hpp
abtink: I read the style guide. It recommends using precise-width integer types over generic `int` and `unsigned it` which we recommend and do it OT as well and I fully agree with it.  👍 

Not sure if there is enough advantage to use `int64` over `uint64` for counters.
- It may lead to confusion, e.g., people may wonder what does it mean and why the counter value can be negative. 
- We can use it to report overflow/undeflow error but not sure how useful and how practical it can be.  Realistically `uint64` counter should not overflow that easily (even if it is increased by `1000` every `usec` it should take more than 500 years).

erjiaqing: For this PR, I think it is OK to preserve the unsigned types. Given the fact that this values will only be advanced or set to 0. And we won't do other arithmetics on these values.

One idea for using signed values (probably for future, new APIs) is the users is very likely to do arithmetic on these values. Previously, I think it will be more convenient if the users can save this explict type convension, however, seems the compiler won't complain about the convension beetween uint64 and int64.

Based on the style guide, "non negative" should not be implied by the type, and if we are always using signed int unless for some specific cases, I don't think it will confuse people?

erjiaqing:
Close the PR based on discussion. Further discussion required for the types for similar fields in the future.

PR #1644, opened 2022-12-08
[feature flags] Add OTBR_OVERRIDE_RUNTIME_FEATURE_FLAGS_ON_START_UP

erjiaqing:
This PR addes the option to override the default values specified in feature flag proto file.

Also this PR will bring up all feature managed by feature flags when feature flag is not enabled.


xuyirio (approved):
Thanks, Song!

erjiaqing:
Confirmed that the failure can be fixed by https://github.com/openthread/openthread/pull/8520

jwhui (approved):

PR #1657, opened 2022-12-14
[rest] use correct Content-Type header for RESTful API

agners:
Set the content type to `text/plain` by default, and `application/json` where needed.

### DIFF
--- a/src/rest/response.cpp
+++ b/src/rest/response.cpp
@@ -30,6 +30,8 @@
 
 #include <stdio.h>
 
+#define OT_REST_RESPONSE_CONTENT_TYPE_HEADER "Content-Type"
+
 #define OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_ORIGIN "*"
 #define OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_HEADERS                                                              \
     "Access-Control-Allow-Headers, Origin,Accept, X-Requested-With, Content-Type, Access-Control-Request-Method, " \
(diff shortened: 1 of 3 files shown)


src/rest/response.hpp
```diff
@@ -61,7 +64,22 @@ class Response
      *
      *
      */
-    Response(void);
+    Response();
```
jwhui, line 67: nit: unintended change?

src/rest/response.hpp
```diff
@@ -61,7 +64,22 @@ class Response
...
+
+    /**
+     * This method set the response Content-Type header.
```
jwhui, line 70: nit:
```suggestion
     * This method sets the response Content-Type header.
```

src/rest/response.hpp
```diff
@@ -61,7 +64,22 @@ class Response
...
+     *
+     * @param[in] aContentType  A string to be set as response Content-Type header.
+     *
+     */
+    void SetContentType(const std::string &aContentType);
+
+    /**
+     * This method return a string of the response Content-Type.
```
jwhui, line 78: nit:
```suggestion
     * This method returns a string of the response Content-Type.
```

jwhui (commented):
Thanks for contributing this! Commented on some minor nits.

bukepo (approved):

src/rest/response.cpp
```diff
@@ -91,6 +85,16 @@ void Response::SetCallback(void)
...
 }
 
+void Response::SetContentType(const std::string &aContentType)
+{
+    mHeader[OT_REST_RESPONSE_CONTENT_TYPE_HEADER] = aContentType;
+}
+
+std::string Response::GetContentType() const
+{
+    return mHeader.find(OT_REST_RESPONSE_CONTENT_TYPE_HEADER)->second;
```
wgtdkp, line 95: Why not access by `[]`?
```suggestion
    return mHeader[OT_REST_RESPONSE_CONTENT_TYPE_HEADER];
```

wgtdkp (approved):
Thanks! @agners 
LGTM! :+1:

wgtdkp:
> Set the content type to `text/plain` by default, and `application/json` where needed.

Curious, are there cases where we need to use `text/plain`?

src/rest/response.cpp
agners: That does not work:

```
/home/sag/projects/openthread/ot-br-posix/src/rest/response.cpp: In member function ‘std::string otbr::rest::Response::GetContentType() const’:
/home/sag/projects/openthread/ot-br-posix/src/rest/response.cpp:93:34: error: passing ‘const std::map<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >’ as ‘this’ argument discards qualifiers [-fpermissive]
   93 |     return mHeader["Content-Type"];
      |                                  ^
```

agners: More detailed explanation: https://stackoverflow.com/a/262872/184207

Maybe `at()` is the better option here as we know it must exist?

agners:
> > Set the content type to `text/plain` by default, and `application/json` where needed.
> 
> Curious, are there cases where we need to use `text/plain`?

All `/node/*` (e.g. `/node/rloc` etc) endpoints return a single value only, not JSON encoded. They represent the majority of endpoints.

wgtdkp:
> /node/rloc

@agners But a single integer is till a valid JSON value. right? https://www.json.org/json-en.html

agners:
> > /node/rloc
> 
> @agners But a single integer is till a valid JSON value. right? https://www.json.org/json-en.html

Hm, right, single values are valid JSON. Strings need to be in quotes, and that seems to case.

I did not really had a practical reason for the content-type change, somehow in my mind just values was not valid JSON :sweat_smile:  

Where should we go from here? There is currently no need for an API to change content-type, so this makes the PR moot. I can separate the switch to the map data structure as it makes the code nicer and submit a new PR for that.

wgtdkp:
> Where should we go from here? There is currently no need for an API to change content-type, so this makes the PR moot. I can separate the switch to the map data structure as it makes the code nicer and submit a new PR for that.

Let's keep all the RESTful APIs returning the same content-type. A separate PR which changes only the data structure sounds good to me! thanks!

agners:
Data structure update PR #1665.

PR #1659, opened 2022-12-14
[web/rest] listen on IPv6 addresses as well

agners:
Make the Web and REST API listen on IPv4 and IPv6 by default.

### DIFF
--- a/src/rest/rest_web_server.cpp
+++ b/src/rest/rest_web_server.cpp
@@ -118,20 +118,24 @@ void RestWebServer::InitializeListenFd(void)
     otbrError   error = OTBR_ERROR_NONE;
     std::string errorMessage;
     int32_t     ret;
-    int32_t     err    = errno;
-    int32_t     optval = 1;
+    int32_t     err = errno;
+    int32_t     yes = 1;
+    int32_t     no  = 0;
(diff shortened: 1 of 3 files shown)


src/rest/rest_web_server.cpp
```diff
@@ -121,17 +121,17 @@ void RestWebServer::InitializeListenFd(void)
...
 
-    mAddress.sin_family      = AF_INET;
-    mAddress.sin_addr.s_addr = INADDR_ANY;
-    mAddress.sin_port        = htons(kPortNumber);
+    mAddress.sin6_family = AF_INET6;
+    mAddress.sin6_addr   = in6addr_any;
+    mAddress.sin6_port   = htons(kPortNumber);
 
-    mListenFd = SocketWithCloseExec(AF_INET, SOCK_STREAM, 0, kSocketNonBlock);
+    mListenFd = SocketWithCloseExec(AF_INET6, SOCK_STREAM, 0, kSocketNonBlock);
```
jwhui, line 128: Should we explicitly set `IPV6_V6ONLY` to zero?

jwhui (approved):
Thanks! 👍🏼

PR #1678, opened 2022-12-22
[logging] detailed logging over feature experiment

sunytt:
This PR enables detailed logging over feature experiment.

Background: On some test devices the default log level is NOTE, which may not be sufficient for debugging when a bug is produced. So we would like to be able to set a detailed Thread log level in an automated fashion, which will be applied at the beginning of the thread process.

A feature flag 'enable_detailed_logging‘ of bool type is supported to enable default detailed logging level 'INFO'.
An additional field 'detailed_logging_level‘ is used to specify the log level.


.github/workflows/build.yml
```diff
@@ -74,7 +74,7 @@ jobs:
       BUILD_TARGET: check
       OTBR_BUILD_TYPE: ${{ matrix.build_type }}
       OTBR_MDNS: ${{ matrix.mdns }}
-      OTBR_OPTIONS: "-DOTBR_SRP_ADVERTISING_PROXY=ON -DOTBR_BORDER_ROUTING=ON -DOTBR_NAT64=1 -DOTBR_SRP_SERVER_AUTO_ENABLE=OFF"
+      OTBR_OPTIONS: "-DOTBR_SRP_ADVERTISING_PROXY=ON -DOTBR_BORDER_ROUTING=ON -DOTBR_NAT64=1 -DOTBR_SRP_SERVER_AUTO_ENABLE=OFF -DOTBR_ENABLE_DETAILED_LOGGING_FLAG=ON"
```
superwhd, line 77:
```suggestion
      OTBR_OPTIONS: "-DOTBR_SRP_ADVERTISING_PROXY=ON -DOTBR_BORDER_ROUTING=ON -DOTBR_NAT64=1 -DOTBR_SRP_SERVER_AUTO_ENABLE=OFF -DOTBR_DETAILED_LOGGING=ON"
```

src/common/logging.cpp
```diff
@@ -49,9 +49,14 @@
...
 #include "common/time.hpp"
 
-static otbrLogLevel sLevel            = OTBR_LOG_INFO;
-static const char   sLevelString[][8] = {
-      "[EMERG]", "[ALERT]", "[CRIT]", "[ERR ]", "[WARN]", "[NOTE]", "[INFO]", "[DEBG]",
+static otbrLogLevel sLevel = OTBR_LOG_INFO;
+
+#if OTBR_ENABLE_DETAILED_LOGGING_FLAG
+static otbrLogLevel sDefaultLevel = OTBR_LOG_INFO;
+#endif
```
superwhd, line 56: Is it possible to move this part below `sLevelString` so that the diff won't involve format changes.

src/common/logging.cpp
```diff
@@ -60,6 +65,14 @@ otbrLogLevel otbrLogGetLevel(void)
     return sLevel;
 }
 
+#if OTBR_ENABLE_DETAILED_LOGGING_FLAG
+/** Get the initial log level */
```
superwhd, line 69:
```suggestion
/** Get the default log level */
```

src/ncp/ncp_openthread.cpp
```diff
@@ -209,6 +219,18 @@ otError ControllerOpenThread::ApplyFeatureFlagList(const FeatureFlagList &aFeatu
...
 
+#if OTBR_ENABLE_DETAILED_LOGGING_FLAG
+    if (aFeatureFlagList.enable_detailed_logging())
+    {
+        error = SetOtbrAndOtLogLevel(OTBR_LOG_INFO);
+    }
+    else
+    {
+        otbrLogLevel defaultLogLevel = otbrLogGetDefaultLevel();
+        error                        = SetOtbrAndOtLogLevel(defaultLogLevel);
```
superwhd, line 230:
```suggestion
        error = SetOtbrAndOtLogLevel(otbrLogGetDefaultLevel());
```

src/ncp/ncp_openthread.cpp
```diff
@@ -209,6 +219,18 @@ otError ControllerOpenThread::ApplyFeatureFlagList(const FeatureFlagList &aFeatu
     otNat64SetEnabled(mInstance, aFeatureFlagList.enable_nat64());
 #endif
 
+#if OTBR_ENABLE_DETAILED_LOGGING_FLAG
```
superwhd, line 222: I wonder when we want to set this macro to 0?

superwhd (commented):
Not specific to this PR: we may add a border router test case to verify that OTBR is in the desired state desired according to feature flags. For example, we can check `ot-ctl log level`  in the test case after enabling detailed logging via feature flag.

src/ncp/ncp_openthread.cpp
morningboata: Like to check, does it make sense to remove OTBR_ENABLE_DETAILED_LOGGING_FLAG if we always want to enable it, and it has small memory/cpu usage?

morningboata (commented):
+1 to add a unit test for applying flag.

src/common/logging.cpp
```diff
@@ -49,9 +49,14 @@
 #include "common/code_utils.hpp"
 #include "common/time.hpp"
 
-static otbrLogLevel sLevel            = OTBR_LOG_INFO;
-static const char   sLevelString[][8] = {
-      "[EMERG]", "[ALERT]", "[CRIT]", "[ERR ]", "[WARN]", "[NOTE]", "[INFO]", "[DEBG]",
+static otbrLogLevel sLevel = OTBR_LOG_INFO;
+
+#if OTBR_ENABLE_DETAILED_LOGGING_FLAG
+static otbrLogLevel sDefaultLevel = OTBR_LOG_INFO;
```
superwhd, line 55: Should we set the default log level as the value passed in by command line argument `-d`? I'm kind of concerned there may be too many definitions of log level: 'detailed' log level, 'default' log level, 'start up' log level.

sunytt: It's set to the command line argument value at otbrLogInit() https://github.com/openthread/ot-br-posix/pull/1678/files#diff-ee208097f13a01b500e564c1b15f89e2184eaeb0b61f71f7bcbe684bc75709f8R94
This sDefaultLevel is used when the detailed-logging-flag changes from true to false at run time.
Another idea is to ignore the above case, and not change log level when flag changes from true to false, as this scenario should be quite rare, and a restart of otbr-agent should get things right.
Please suggest.

src/ncp/ncp_openthread.cpp
sunytt: I guess this feature is only wanted when feature flags are supported, maybe I can remove OTBR_ENABLE_DETAILED_LOGGING_FLAG and use OTBR_ENABLE_FEATURE_FLAGS instead?

src/ncp/ncp_openthread.cpp
```diff
@@ -209,6 +219,18 @@ otError ControllerOpenThread::ApplyFeatureFlagList(const FeatureFlagList &aFeatu
...
+    if (aFeatureFlagList.enable_detailed_logging())
+    {
+        error = SetOtbrAndOtLogLevel(OTBR_LOG_INFO);
```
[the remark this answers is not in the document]
sunytt: I think INFO as it might be sufficient, DEBG might be too verbose.

src/common/logging.cpp
superwhd: Thanks. Setting default log level as the command line argument aligns with my though. I missed that line in my previous review.

src/ncp/ncp_openthread.cpp
superwhd: I think you don't need to add guard for `OTBR_ENABLE_FEATURE_FLAGS` since the whole function has already been guarded by it.

sunytt: Will remove the flag

morningboata: Suggest to still use OTBR_ENABLE_DETAILED_LOGGING_FLAG since it is more readable,  OTBR_ENABLE_FEATURE_FLAGS is  not a good name for your new feature. Or remove all OTBR_ENABLE_DETAILED_LOGGING_FLAG if it is not needed.

src/ncp/ncp_openthread.cpp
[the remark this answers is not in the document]
morningboata: I see. For this LOG level, maybe you can define something like int detail_log_level field in the proto with default value in addition to enable_detailed_logging? In such way you don't need to change OTBR code later (only server side integration is needed in phase 2).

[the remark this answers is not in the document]
morningboata: I think enable_detailed_logging is needed: only when it is true, use detail_log_level to apply the detailed log level.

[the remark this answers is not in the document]
sunytt: Is it a requirement from server side?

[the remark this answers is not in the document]
morningboata: detail_log_level will be applied only when enable_detailed_logging is true. Feeling both flags are needed, for clarity and for server side integration requirement.

.github/workflows/build.yml
sunytt: Flag removed.

src/common/logging.cpp
sunytt: Done

src/common/logging.cpp
sunytt: done

src/ncp/ncp_openthread.cpp
sunytt: done

src/ncp/ncp_openthread.cpp
sunytt: Flag removed.

src/ncp/ncp_openthread.cpp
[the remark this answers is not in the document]
sunytt: Added detailed_log_level field.

sunytt:
> Not specific to this PR: we may add a border router test case to verify that OTBR is in the desired state desired according to feature flags. For example, we can check `ot-ctl log level` in the test case after enabling detailed logging via feature flag.

Good point 👍  Will look into how to enable feature flags and add a test as follow up.

src/proto/feature_flag.proto
```diff
@@ -41,4 +41,8 @@ package otbr;
 message FeatureFlagList {
   // Whether to enable the NAT64 feature.
   optional bool enable_nat64 = 1 [default = false];
+  // Whether to enable detailed logging.
+  optional bool enable_detailed_logging = 2 [default = false];
+  // Set specific detailed logging level, default = OTBR_LOG_INFO
```
morningboata, line 46: Can you add detail comment about which number corresponds to which LOG level?

src/common/logging.cpp
```diff
@@ -54,12 +54,20 @@ static const char   sLevelString[][8] = {
       "[EMERG]", "[ALERT]", "[CRIT]", "[ERR ]", "[WARN]", "[NOTE]", "[INFO]", "[DEBG]",
 };
 
+static otbrLogLevel sDefaultLevel = OTBR_LOG_INFO;
```
morningboata, line 57: Do you know what is the current initial log level for OTBR on device? and align the value to the current initial level?

sunytt: The default value is INFO, and then it can be changed by command line argument to otbr-agent. We align this value with the otbr-agent initial value at line 88 of the current file: https://github.com/openthread/ot-br-posix/pull/1678/files#diff-ee208097f13a01b500e564c1b15f89e2184eaeb0b61f71f7bcbe684bc75709f8R88

src/proto/feature_flag.proto
sunytt: Done

morningboata: If the OTBR's initial log level is INFO, when you enable detailed logging, do you want to use same INFO level? Or you want to use detailed debug level?

morningboata (approved):

src/proto/feature_flag.proto
sunytt: Normally the user build images have log level NOTE, setting log level to INFO on these devices would help with debugging. INFO level might already be sufficient, if we figured DEBG is more desirable we can use the 'detailed_logging_level' field to change it to DEBG.

src/proto/feature_flag.proto
```diff
@@ -41,4 +41,17 @@ package otbr;
...
+  // Mapping from detailed_logging_level number to otbrLogLevel:
+  //   0 - OTBR_LOG_EMERG
+  //   1 - OTBR_LOG_ALERT
+  //   2 - OTBR_LOG_CRIT
+  //   3 - OTBR_LOG_ERR
+  //   4 - OTBR_LOG_WARNING
+  //   5 - OTBR_LOG_NOTICE
+  //   6 - OTBR_LOG_INFO
+  //   7 - OTBR_LOG_DEBUG
```
superwhd, line 55: Why do we want to introduce these new log levels?

sunytt: We would like to be able to specify detailed logging level from feature flags. There doesn't seem to be a way to directly use the otbrLogLevel definition in proto files or the generated cc/h file, so I added this translation.

superwhd: I wrongly thought these were different from OTBR's enums. 

On the other hand, why do we still need to manually map the integer to enum `otbrLogLevel`? Why not just assign the integer to the enum?

sunytt: "assign the integer to the enum" do you mean something like casting the int to the enum? I prefer the mapping as it's more transparent, and it separates the definition in otbr and proto. The proto int number may be updated from feature flag server side. Say if the otbrLogLevel numbers changed, e.g. we somehow decided to add a otbr log level in the middle, then casting the int will require us to modify the proto file and also the feature flags server side.

superwhd: Proto enum should be a more clear way to address such enum value changes.

src/proto/feature_flag.proto
```diff
@@ -3,6 +3,18 @@ option optimize_for = LITE_RUNTIME;
 
 package otbr;
 
+enum otbrProtoLogLevel
```
superwhd, line 6:
```suggestion
enum LogLevel
```
- Use CamelCase.
- I think it may not be necessary to emphasize `otbr` and `proto` in the name.

src/proto/feature_flag.proto
```diff
@@ -3,6 +3,18 @@ option optimize_for = LITE_RUNTIME;
...
+{
+    PROTO_LOG_EMERG   = 0;  ///< System is unusable
```
superwhd, line 8: https://developers.google.com/protocol-buffers/docs/style#enums

The zero value enum should have the suffix `UNSPECIFIED`.

src/proto/feature_flag.proto
sunytt: Agree 'otbr' is not necessary, I kept 'proto' to indicate this is for the proto file.

src/proto/feature_flag.proto
sunytt: 👍

src/ncp/ncp_openthread.cpp
```diff
@@ -124,6 +124,45 @@ otbrLogLevel ControllerOpenThread::ConvertToOtbrLogLevel(otLogLevel aLogLevel)
     return otbrLogLevel;
 }
 
+#if OTBR_ENABLE_FEATURE_FLAGS
+/* Converts otbrProtoLogLevel to otbrLogLevel */
```
superwhd, line 128:
```suggestion
/* Converts ProtoLogLevel to otbrLogLevel */
```

superwhd (approved):
LGTM overall

src/common/logging.cpp
```diff
@@ -54,12 +54,20 @@ static const char   sLevelString[][8] = {
...
+
```
abtink, line 58: I want to echo same things as other reviewers mentioned about the `sDefaultLevel` and how it is used.
I feel it may be better to remove it and logic will be smipler:
- If the `aFeatureFlagList.enable_detailed_logging()` is used, then we control log level.
- If it is not set, we let the log level be whatever specified during build time.
  - OT core (and I think otbr) provide cmake `OT_LOG_LEVEL` option to specify this
  - I think this config may be set differently when otbr is integrated into products

src/ncp/ncp_openthread.cpp
```diff
@@ -209,6 +256,15 @@ otError ControllerOpenThread::ApplyFeatureFlagList(const FeatureFlagList &aFeatu
     otNat64SetEnabled(mInstance, aFeatureFlagList.enable_nat64());
 #endif
 
+    if (aFeatureFlagList.enable_detailed_logging())
+    {
+        error = SetOtbrAndOtLogLevel(ConvertProtoToOtbrLogLevel(aFeatureFlagList.detailed_logging_level()));
+    }
+    else
+    {
+        error = SetOtbrAndOtLogLevel(otbrLogGetDefaultLevel());
```
abtink, line 265: If `OT_LOG_LEVEL` is explicitly set to some value (on a product build), here we may change it at run-time to some other value?

src/common/logging.cpp
sunytt: The `sDefaultLevel` is introduced to handle the corner case where `aFeatureFlagList.enable_detailed_logging()` is changed from `true` to `false` at run time. In this case we may want to set log level back to what was used in the beginning of otbr start.
When otbr is integrated into products the compile time log level can be override by otbr command line arguments at run time (https://github.com/openthread/ot-br-posix/blob/main/src/agent/main.cpp#L213), so I set `sDefaultLevel` to be same as otbr's final choice at otbrLogInit: https://github.com/openthread/ot-br-posix/pull/1678/files#diff-ee208097f13a01b500e564c1b15f89e2184eaeb0b61f71f7bcbe684bc75709f8R88

src/ncp/ncp_openthread.cpp
sunytt: It looks like otbr doesn't use OT_LOG_LEVEL, but starts from OTBR_LOG_INFO, then takes in otbr command line argument `-d`, and set that value to otbr and ot log level.
Here I'm trying to set back to the above value.

src/common/logging.cpp
abtink: Got it. Thanks. Sounds good.

src/ncp/ncp_openthread.cpp
abtink: Got it. Make sense.

I assumed that the `-d` was specifying `otbr` log level only.
- Looking at the functions in `common/logging.hpp`, they all seem to be dealing with logging within `otbr` itself.
- So I thought that the logging in `otbr` and OT core were treated differently and may even use two different levels.

But looks like in `ControllerOpenThread::Init(void)` we call `otLoggingSetLevel()` and set the OT log level to the same (converted) level:
```
void ControllerOpenThread::Init(void)
{
    otbrError  error = OTBR_ERROR_NONE;
    otLogLevel level = ConvertToOtLogLevel(otbrLogGetLevel());

    VerifyOrExit(otLoggingSetLevel(level) == OT_ERROR_NONE, error = OTBR_ERROR_OPENTHREAD);
```

abtink (approved):
LGTM

jwhui (approved):

PR #1682, opened 2022-12-24
[rest] add Operational Dataset API

agners:
This PR extends the REST API to support reading and writing Active and Pending Operational Datasets. It largely implements what has been discussed in https://github.com/openthread/openthread/discussions/8567.

It adds the following endpoints and methods:

* /node/dataset/active [`GET`, `OPTIONS`, `POST`, `PUT`]
* /node/dataset/pending [`GET`, `OPTIONS`, `POST`, `PUT`]

The `GET` method returns current datasets, `POST` allows to create a new dataset (with defaults as initialized by the OpenThread implementation just like `ot-ctl dataset init new`). Any fields provided with `POST` will overwrite the defaults, hence a call like

```
curl -X POST http://localhost:8081/node/dataset/pending -d '{ "Delay": 30000, "ActiveTimestamp": 10 }'
```

Is equivalent to:

```
ot-ctl dataset init new
ot-ctl dataset delay 30000
ot-ctl dataset activetimestamp 10
ot-ctl dataset commit pending
```

`PUT` requests do not initialize a new dataset (which makes the call idempotent as defined in HTTP). `PUT` simply loads the current dataset and updates it with the new values provided by the JSON body. When using the TLV format PUT is a simple write to the active/pending dataset.

Datasets can be read and written using a straight forward JSON format or alternatively using the Dataset TLV format. This is implemented in a RESTful manor by using the same endpoints but `Content-Type` to distinguish the two data formats. The `application/json` type is used for JSON and `plain/text` for Dataset TLV format.

The API is documented by a OpenAPI Specification (Swagger). The YAML description file is stored at `src/rest/openapi.yaml`. Swagger UI allows to test all the variants described above as well. To start Swagger UI locally the following command can be used:

```
docker run -p 8082:8080 -e SWAGGER_JSON=/openapi.yaml -v $(pwd)/src/rest/openapi.yaml:/openapi.yaml swaggerapi/swagger-ui
```

The following things are open/tbd at this point:

- No method to change state of the OTBR (probably best done in a separate PR)
- Automatic/implicit `ActiveTimestamp` support (or should this be left to the client)
- Generate Swagger UI API in GitHub actions/deploy GitHub pages?
- Additional testing/unit tests?

The PR also removes the `/node/active-dataset-tlvs` endpoint as it is obsolete with this much more capable endpoints.

### DIFF
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -40,16 +40,6 @@ namespace otbr {
 namespace rest {
 namespace Json {
 
-std::string Bytes2HexString(const uint8_t *aBytes, uint8_t aLength)
-{
-    char hex[2 * aLength + 1];
-
-    otbr::Utils::Bytes2Hex(aBytes, aLength, hex);
-    hex[2 * aLength] = '\0';
-
-    return std::string(hex);
-}
-
 static cJSON *Bytes2HexJson(const uint8_t *aBytes, uint8_t aLength)
 {
     char hex[2 * aLength + 1];
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -27,6 +27,7 @@
  */
 
 #include "rest/json.hpp"
+#include <sstream>
 
 #include "common/code_utils.hpp"
 #include "common/types.hpp"
@@ -112,6 +113,43 @@ static cJSON *IpAddr2Json(const otIp6Address &aAddress)
     Ip6Address addr(aAddress.mFields.m8);
 
     return cJSON_CreateString(addr.ToString().c_str());
+}
+
+static cJSON *IpPrefix2Json(const otIp6NetworkPrefix &aAddress)
+{
+    std::stringstream ss;
+    otIp6Address      address = {};
+
+    address.mFields.mComponents.mNetworkPrefix = aAddress;
+    Ip6Address addr(address.mFields.m8);
+
+    ss << addr.ToString() << "/" << OT_IP6_PREFIX_BITSIZE;
+
+    return cJSON_CreateString(ss.str().c_str());
+}
+
+static cJSON *SecurityPolicy2Json(const otSecurityPolicy &aSecurityPolicy)
+{
+    cJSON *securityPolicy = cJSON_CreateObject();
+
+    cJSON_AddItemToObject(securityPolicy, "RotationTime", cJSON_CreateNumber(aSecurityPolicy.mRotationTime));
+    cJSON_AddItemToObject(securityPolicy, "ObtainNetworkKey",
+                          cJSON_CreateBool(aSecurityPolicy.mObtainNetworkKeyEnabled));
+    cJSON_AddItemToObject(securityPolicy, "NativeCommissioning",
+                          cJSON_CreateBool(aSecurityPolicy.mNativeCommissioningEnabled));
+    cJSON_AddItemToObject(securityPolicy, "Routers", cJSON_CreateBool(aSecurityPolicy.mRoutersEnabled));
+    cJSON_AddItemToObject(securityPolicy, "ExternalCommissioning",
+                          cJSON_CreateBool(aSecurityPolicy.mExternalCommissioningEnabled));
+    cJSON_AddItemToObject(securityPolicy, "CommercialCommissioning",
+                          cJSON_CreateBool(aSecurityPolicy.mCommercialCommissioningEnabled));
+    cJSON_AddItemToObject(securityPolicy, "AutonomousEnrollment",
+                          cJSON_CreateBool(aSecurityPolicy.mAutonomousEnrollmentEnabled));
+    cJSON_AddItemToObject(securityPolicy, "NetworkKeyProvisioning",
+                          cJSON_CreateBool(aSecurityPolicy.mNetworkKeyProvisioningEnabled));
+    cJSON_AddItemToObject(securityPolicy, "TobleLink", cJSON_CreateBool(aSecurityPolicy.mTobleLinkEnabled));
+    cJSON_AddItemToObject(securityPolicy, "NonCcmRouters", cJSON_CreateBool(aSecurityPolicy.mNonCcmRoutersEnabled));
+
+    return securityPolicy;
 }
 
 static cJSON *ChildTableEntry2Json(const otNetworkDiagChildEntry &aChildEntry)
@@ -478,6 +516,66 @@ std::string Error2JsonString(HttpStatusCode aErrorCode, std::string aErrorMessag
     ret = Json2String(error);
 
     cJSON_Delete(error);
+
+    return ret;
+}
+
+std::string Dataset2JsonString(const otOperationalDataset &aDataset)
+{
+    cJSON      *node = cJSON_CreateObject();
+    std::string ret;
+
+    if (aDataset.mComponents.mIsActiveTimestampPresent)
+    {
+        cJSON_AddItemToObject(node, "ActiveTimestamp", cJSON_CreateNumber(aDataset.mActiveTimestamp.mSeconds));
+    }
+    if (aDataset.mComponents.mIsPendingTimestampPresent)
+    {
+        cJSON_AddItemToObject(node, "PendingTimestamp", cJSON_CreateNumber(aDataset.mPendingTimestamp.mSeconds));
+    }
+    if (aDataset.mComponents.mIsNetworkKeyPresent)
+    {
+        cJSON_AddItemToObject(node, "NetworkKey", Bytes2HexJson(aDataset.mNetworkKey.m8, OT_NETWORK_KEY_SIZE));
+    }
+    if (aDataset.mComponents.mIsNetworkNamePresent)
+    {
+        cJSON_AddItemToObject(node, "NetworkName", cJSON_CreateString(aDataset.mNetworkName.m8));
+    }
+    if (aDataset.mComponents.mIsExtendedPanIdPresent)
+    {
+        cJSON_AddItemToObject(node, "ExtPanId", Bytes2HexJson(aDataset.mExtendedPanId.m8, OT_EXT_PAN_ID_SIZE));
+    }
+    if (aDataset.mComponents.mIsMeshLocalPrefixPresent)
+    {
+        cJSON_AddItemToObject(node, "MeshLocalPrefix", IpPrefix2Json(aDataset.mMeshLocalPrefix));
+    }
+    if (aDataset.mComponents.mIsDelayPresent)
+    {
+        cJSON_AddItemToObject(node, "Delay", cJSON_CreateNumber(aDataset.mDelay));
+    }
+    if (aDataset.mComponents.mIsPanIdPresent)
+    {
+        cJSON_AddItemToObject(node, "PanId", cJSON_CreateNumber(aDataset.mPanId));
+    }
+    if (aDataset.mComponents.mIsChannelPresent)
+    {
+        cJSON_AddItemToObject(node, "Channel", cJSON_CreateNumber(aDataset.mChannel));
+    }
+    if (aDataset.mComponents.mIsPskcPresent)
+    {
+        cJSON_AddItemToObject(node, "PSKc", Bytes2HexJson(aDataset.mPskc.m8, OT_PSKC_MAX_SIZE));
+    }
+    if (aDataset.mComponents.mIsSecurityPolicyPresent)
+    {
+        cJSON_AddItemToObject(node, "SecurityPolicy", SecurityPolicy2Json(aDataset.mSecurityPolicy));
+    }
+    if (aDataset.mComponents.mIsChannelMaskPresent)
+    {
+        cJSON_AddItemToObject(node, "ChannelMask", cJSON_CreateNumber(aDataset.mChannelMask));
+    }
+
+    ret = Json2String(node);
+    cJSON_Delete(node);
 
     return ret;
 }
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -676,6 +676,7 @@ bool JsonString2Dataset(const std::string &aJsonDataset, otOperationalDataset &a
     bool        ret = true;
 
     VerifyOrExit((jsonDataset = cJSON_Parse(aJsonDataset.c_str())) != nullptr, ret = false);
+    VerifyOrExit(cJSON_IsObject(jsonDataset), ret = false);
 
     value = cJSON_GetObjectItemCaseSensitive(jsonDataset, "ActiveTimestamp");
     if (cJSON_IsObject(value))
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -40,6 +40,16 @@ namespace otbr {
 namespace rest {
 namespace Json {
 
+std::string Bytes2HexString(const uint8_t *aBytes, uint8_t aLength)
+{
+    char hex[2 * aLength + 1];
+
+    otbr::Utils::Bytes2Hex(aBytes, aLength, hex);
+    hex[2 * aLength] = '\0';
+
+    return std::string(hex);
+}
+
 static cJSON *Bytes2HexJson(const uint8_t *aBytes, uint8_t aLength)
 {
     char hex[2 * aLength + 1];
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -128,6 +128,21 @@ static cJSON *IpPrefix2Json(const otIp6NetworkPrefix &aAddress)
     return cJSON_CreateString(ss.str().c_str());
 }
 
+otbrError Json2IpPrefix(const cJSON *aJson, otIp6NetworkPrefix &aIpPrefix)
+{
+    otbrError          error = OTBR_ERROR_NONE;
+    std::istringstream ipPrefixStr(std::string(aJson->valuestring));
+    std::string        tmp;
+    Ip6Address         addr;
+
+    VerifyOrExit(std::getline(ipPrefixStr, tmp, '/'), error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit((error = addr.FromString(tmp.c_str(), addr)) == OTBR_ERROR_NONE);
+
+    memcpy(aIpPrefix.m8, addr.m8, OT_IP6_PREFIX_SIZE);
+exit:
+    return error;
+}
+
 static cJSON *SecurityPolicy2Json(const otSecurityPolicy &aSecurityPolicy)
 {
     cJSON *securityPolicy = cJSON_CreateObject();
@@ -150,6 +165,38 @@ static cJSON *SecurityPolicy2Json(const otSecurityPolicy &aSecurityPolicy)
     cJSON_AddItemToObject(securityPolicy, "NonCcmRouters", cJSON_CreateBool(aSecurityPolicy.mNonCcmRoutersEnabled));
 
     return securityPolicy;
+}
+
+bool Json2SecurityPolicy(const cJSON *jsonSecurityPolicy, otSecurityPolicy &aSecurityPolicy)
+{
+    cJSON *value;
+
+    value = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "RotationTime");
+    if (cJSON_IsNumber(value))
+    {
+        aSecurityPolicy.mRotationTime = static_cast<uint16_t>(value->valueint);
+    }
+
+    value                                    = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "ObtainNetworkKey");
+    aSecurityPolicy.mObtainNetworkKeyEnabled = cJSON_IsTrue(value);
+    value = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "NativeCommissioning");
+    aSecurityPolicy.mNativeCommissioningEnabled = cJSON_IsTrue(value);
+    value                                       = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "Routers");
+    aSecurityPolicy.mRoutersEnabled             = cJSON_IsTrue(value);
+    value = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "ExternalCommissioning");
+    aSecurityPolicy.mExternalCommissioningEnabled = cJSON_IsTrue(value);
+    value = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "CommercialCommissioning");
+    aSecurityPolicy.mCommercialCommissioningEnabled = cJSON_IsTrue(value);
+    value = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "AutonomousEnrollment");
+    aSecurityPolicy.mAutonomousEnrollmentEnabled = cJSON_IsTrue(value);
+    value = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "NetworkKeyProvisioning");
+    aSecurityPolicy.mNetworkKeyProvisioningEnabled = cJSON_IsTrue(value);
+    value                                          = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "TobleLink");
+    aSecurityPolicy.mTobleLinkEnabled              = cJSON_IsTrue(value);
+    value                                 = cJSON_GetObjectItemCaseSensitive(jsonSecurityPolicy, "NonCcmRouters");
+    aSecurityPolicy.mNonCcmRoutersEnabled = cJSON_IsTrue(value);
+
+    return true;
 }
 
 static cJSON *ChildTableEntry2Json(const otNetworkDiagChildEntry &aChildEntry)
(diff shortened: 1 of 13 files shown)


src/rest/json.cpp
```diff
@@ -39,6 +40,16 @@ namespace otbr {
 namespace rest {
 namespace Json {
 
+std::string Bytes2HexString(const uint8_t *aBytes, uint8_t aLength)
+{
+    char hex[2 * aLength + 1];
```
bukepo, line 45: not c++ standard?
```suggestion
    char* hex = new char[2 * aLength + 1];
```

agners: I am not a C++ expert, but looking at [Array declaration on cppreference.com](https://en.cppreference.com/w/cpp/language/array) it doesn't look like anything would forbid this type of declaration?

In any case, the function just below uses the same type of declaration. If we should get rid of this type of declaration we probably should do it there too.

In general json.cpp probably could use some modernization using new C++ constructs.

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+      responses:
+        "200":
+          description: Returns currently active opertional dataset
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Dataset"
+            text/plain:
+              schema:
+                $ref: "#/components/schemas/TlvDataset"
```
wgtdkp, line 178: nit:
```suggestion
                $ref: "#/components/schemas/DatasetTlv"
```

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+
+        Default parameters are choosen for parameters not set in the request body.
+      requestBody:
+        description: |-
+          Operational dataset object that will be stored as active operational dataset. JSON keys which are not set will be initialized with
+          default or random values.
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Dataset"
```
wgtdkp, line 197: What about supporting "#/components/schemas/DatasetTlv" as well. Thread Active Operational Dataset is typically shared in form of the TLV byte array/HEX string, and it will be useful to commissioning the BR device directly with the dataset TLVs.

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+      responses:
+        "202":
+          description: Successfully created the active operational dataset.
```
wgtdkp, line 200: Should the response code be "201"? https://restfulapi.net/http-status-codes/

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+      tags:
+        - node
+      summary: Update current active opertional dataset
+      description: |-
+        Updates the current active operational dataset on the current node. Only allowed if the Thread node is inactive. Create a
+        pending dataset if you would like to change the current active opertional dataset if the Thread node is active.
+      requestBody:
+        description: |-
+          Operational dataset object that will be stored as active operational dataset. Supports request body Content-Type `text/plain`
+          (dataset in TLV format as hex string) or `application/json` (dataset in JSOn format, JSON keys which are not set will be
```
wgtdkp, line 213:
```suggestion
          (dataset in TLV format as hex string) or `application/json` (dataset in JSON format, JSON keys which are not set will be
```

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+        Default parameters are choosen for parameters not set in the request body.
+      requestBody:
+        description: |-
+          Operational dataset object that will be stored as active operational dataset. JSON keys which are not set will be initialized with
+          default or random values.
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Dataset"
+      responses:
```
wgtdkp, line 198: Use "409" for error of trying to create active dataset when the Thread node is active?

| | |
| --- | --- |
| 409 Conflict | The request could not be completed due to a conflict with the current state of the resource. |

https://restfulapi.net/http-status-codes/

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+          (dataset in TLV format as hex string) or `application/json` (dataset in JSOn format, JSON keys which are not set will be
+          initialized with default or random values).
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Dataset"
+          plain/text:
+            schema:
+              $ref: "#/components/schemas/TlvDataset"
+      responses:
```
wgtdkp, line 222: Ditto, use "409" for `InvalidState error`?

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+    put:
+      tags:
+        - node
+      summary: Update current pending opertional dataset
+      description: |-
+        Updates the current pending operational dataset on the current node.
+      requestBody:
+        description: |-
+          Operational dataset object that will be stored as pending operational dataset. Supports request body Content-Type `text/plain`
+          (dataset in TLV format as hex string) or `application/json` (dataset in JSOn format, JSON keys which are not set will be
```
wgtdkp, line 280:
```suggestion
          (dataset in TLV format as hex string) or `application/json` (dataset in JSON format, JSON keys which are not set will be
```

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+          description: No pending operational dataset
+    post:
+      tags:
+        - node
+      summary: Create new pending opertional dataset
+      description: |-
+        Creates new pending operational dataset on the current node. If the delay timer is specified, the new pending operational dataset
+        will become active after the time elapsed. The pending dataset will become empty.
+
+        Default parameters are choosen for parameters not set in the request body. By default, the delay timer is *not* set.
```
wgtdkp, line 254: Does it mean the delay timer **MUST** be included in the the request body?

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+        ActiveTimestamp:
+          type: integer
+          description: Active timestamp
+          format: uint64
+          example: 10
+          default: 1
+        PendingTimestamp:
+          type: integer
+          description: Pending timestamp
+          format: int32
```
wgtdkp, line 337:
```suggestion
          format: uint64
```

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+          example: 08277229F21FB7342D705D3CEFDC042A
+          default: random
+        NetworkName:
+          type: string
+          description: Network name, 16 bytes long
+          example: OpenThread-e445
+          default: OpenThread-<PanId>
+        ExtPanId:
+          type: string
+          description: Extended PAN ID, 8 bytes long
```
wgtdkp, line 352:
```suggestion
          description: Extended PAN ID, 8 bytes long, formatted as a hexadecimal string
```

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
+openapi: 3.0.3
```
wgtdkp, line 1: This is a feature we wanted long ago, thanks @agners! 👍

src/rest/request.cpp
```diff
@@ -87,6 +98,16 @@ std::string Request::GetUrl(void) const
...
 }
 
+std::string Request::GetHeaderValue(const std::string aHeaderField) const
+{
+    auto it = mHeaders.find(StringUtils::ToLowercase(aHeaderField));
+
+    if (it == mHeaders.end())
+        return "";
+    else
+        return it->second;
```
wgtdkp, line 108: nit:
```suggestion
    return (it == mHeaders.end()) ? "" : return it->second;
```

src/rest/openapi.yaml
```diff
@@ -0,0 +1,455 @@
...
+        Note: To be considered as a valid active opertional dataset the ActiveTimestamp needs to be newer than the current active operational
+        dataset.
+      requestBody:
+        description: |-
+          Operational dataset object that will be stored as pending operational dataset. JSON keys which are not set will be initialized with
+          default or random values.
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Dataset"
```
wgtdkp, line 265: Ditto, support datasetTlv?

src/rest/resource.cpp
```diff
@@ -499,75 +505,179 @@ void Resource::Rloc(const Request &aRequest, Response &aResponse) const
...
+    if (error == OTBR_ERROR_NONE)
+    {
+        errorCode = GetHttpStatus(HttpStatusCode::kStatusOk);
+        aResponse.SetResponsCode(errorCode);
+    }
+    else if (error == OTBR_ERROR_NOT_FOUND)
+    {
+        errorCode = GetHttpStatus(HttpStatusCode::kNoContent);
+        aResponse.SetResponsCode(errorCode);
+    }
```
wgtdkp, line 558: Try adding a `SetResponsCode(otbrError aError)` method to `Response` so that we can handle common `otbrError` to Response status code conversion in a single place.

src/rest/resource.cpp
```diff
@@ -499,75 +505,179 @@ void Resource::Rloc(const Request &aRequest, Response &aResponse) const
...
-        ExitNow();
+        // Only PUT allowed for text/plain
+        VerifyOrExit(!create, error = OTBR_ERROR_INVALID_ARGS);
+        ret = Json::Hex2BytesJsonString(aRequest.GetBody(), datasetTlvs.mTlvs, OT_OPERATIONAL_DATASET_MAX_LENGTH);
+        VerifyOrExit(ret >= 0, error = OTBR_ERROR_INVALID_ARGS);
+        datasetTlvs.mLength = ret;
+
+        if (aDatasetType == DatasetType::kActive)
+        {
+            // TODO: Reject if Thread is running
```
wgtdkp, line 588: Check the state and use `OTBR_ERROR_INVALID_STATE` to produce a HTTP 409 error?

src/rest/resource.hpp
```diff
@@ -102,6 +104,16 @@ class Resource
     void ErrorHandler(Response &aResponse, HttpStatusCode aErrorCode) const;
 
 private:
+    /**
+     * This enumeration represents the Dataset type (active or pending).
+     *
+     */
+    enum DatasetType : uint8_t
```
wgtdkp, line 111:
```suggestion
    enum class DatasetType : uint8_t
```

src/rest/types.hpp
```diff
@@ -60,6 +66,7 @@ enum class HttpStatusCode : std::uint16_t
 {
     kStatusOk                  = 200,
     kStatusAccepted            = 202,
+    kNoContent                 = 204,
     kStatusBadRequest          = 400,
     kStatusResourceNotFound    = 404,
     kStatusMethodNotAllowed    = 405,
```
wgtdkp, line 72: I am happy if you remove all "Status" from those enum constants :)

src/rest/openapi.yaml
agners: Yeah maybe. I was assuming that changes get applied asynchronously, but maybe that is not true? Or at least not for the actual dataset store?

src/rest/openapi.yaml
agners: You are welcome. It is really useful for testing, I quickly discovered some bugs/issues/discrepancies once I had the description written :smile:

src/rest/openapi.yaml
agners: No, but if its not set the pending operational dataset will just not get applied. Maybe extend that?

```suggestion
        Default parameters are chosen for parameters not set in the request body. By default, the delay timer is *not* set, which means
        the pending dataset will get stored but not automatically applied as active dataset.
```

agners: No, but if its not set the pending operational dataset will just not get applied. Maybe extend that?

```suggestion
        Default parameters are chosen for parameters not set in the request body. By default, the delay timer is *not* set, which means
        the pending dataset will get stored but not automatically applied as active dataset.
```

src/rest/json.cpp
```diff
@@ -39,6 +40,16 @@ namespace otbr {
...
+
+    otbr::Utils::Bytes2Hex(aBytes, aLength, hex);
```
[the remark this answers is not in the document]
agners: Hm, looking closer at that implementation it's very weird to me that it uses a `std::string` to do it's job. Also the extra copy seems very unnecessary. I'll move `std::string Bytes2Hex(const uint8_t *aBytes, const uint16_t aBytesLength);` and optimize `size_t Bytes2Hex(const uint8_t *aBytes, const uint16_t aBytesLength, char *aHex)`.

src/rest/json.cpp
```diff
@@ -482,6 +577,224 @@ std::string Error2JsonString(HttpStatusCode aErrorCode, std::string aErrorMessag
...
 }
 
+std::string Dataset2JsonString(const otOperationalDataset &aDataset)
+{
+    cJSON      *node = cJSON_CreateObject();
+    std::string ret;
+
+    if (aDataset.mComponents.mIsActiveTimestampPresent)
+    {
+        cJSON_AddItemToObject(node, "ActiveTimestamp", cJSON_CreateNumber(aDataset.mActiveTimestamp.mSeconds));
```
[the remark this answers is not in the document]
agners: I went for seconds only for the sake of simplicity. My thinking was that we can still extend the code to accept a JSON object as `ActiveTimestamp` if needed later. It is pretty easy to check which datatype is passed to `ActiveTimestamp`.

That said, it would break the API if we change the format of the JSON structure we return, so maybe a separate field is better still.

That said, I am also ok to extend the API to take the full timestamp as JSON object. I guess the JSON would end up looking like this:


```json
"ActiveTimestamp": {
    "Seconds": 123,
    "Ticks": 123,
    "Authoritative": false
}
```

src/rest/json.cpp
```diff
@@ -482,6 +577,224 @@ std::string Error2JsonString(HttpStatusCode aErrorCode, std::string aErrorMessag
...
+    cJSON *value;
+    cJSON *jsonDataset;
+    bool   ret = true;
+
+    VerifyOrExit((jsonDataset = cJSON_Parse(aJsonDataset.c_str())) != nullptr, ret = false);
+
+    value = cJSON_GetObjectItemCaseSensitive(jsonDataset, "ActiveTimestamp");
+    if (cJSON_IsNumber(value))
+    {
+        aDataset.mActiveTimestamp.mSeconds             = value->valueint;
```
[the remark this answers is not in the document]
agners: See suggestion above.

src/rest/openapi.yaml
agners: I decided against it as it would implement the same behavior as the `PUT` method. 

Since setting the DatasetTlv is idempotent, I felt `PUT` is more suitable. I guess its not really violating RESTful definitions if we would do the same in `POST`, but why have the same functionality in two methods?

src/rest/types.hpp
agners: Not against it, but I would prefer to do it in a follow up as it leads to code changes throughout `resource.cpp`.

src/rest/resource.cpp
agners: Doing the response status code conversion in `Response` seems the right place indeed, as this is a HTTP thing really. However, we do use it in `void Resource::ErrorHandler(Response &aResponse, HttpStatusCode aErrorCode) const` as well, which generates a JSON body with the same string. This is more an implementation specific handling of the HTTP body, which probably should stay in `Resource`?

If we look at `Response` as an "OTBR opinionated" Response implementation, then maybe we can move ErrorHandler to `Response` as well (e.g. `SetErrorResponse(HttpStatusCode aCode)`?)

It kinda goes into refactoring real which I'd prefer to do in a follow up.

src/rest/openapi.yaml
agners: It can be done with the `PUT` method.

src/rest/json.cpp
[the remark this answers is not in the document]
agners: I've added such an implementation.

src/rest/openapi.yaml
wgtdkp: But as said in https://github.com/openthread/ot-br-posix/pull/1682#issue-1510058174, POST will create a new dataset but PUT won't. So for commissioning a new BR device, we need to use the POST method and it's more common that a TLV format dataset will be provided for commissioning the BR device. That's why I think TLV format for POST can be useful.

src/rest/openapi.yaml
wgtdkp: I think the dataset is created when the method returns. It's just not propagated to the network but propagation is not part of this API (we need to start the Thread state separately to propagate dataset).

src/rest/types.hpp
wgtdkp: SGTM, thanks!

src/rest/json.cpp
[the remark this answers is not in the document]
wgtdkp: Thanks!

agners:
Let me know if I can do something to help get this merged.

src/rest/openapi.yaml
wgtdkp: @agners What do you think about this?

agners: This use case is covered, for the TLV format, we accept `PUT` to create a new active dataset, see:
https://github.com/agners/ot-br-posix/blob/add-dataset-rest-api/src/rest/resource.cpp#L590

Ideally we should only accept `PUT`, as we do not create a sub-item (in a RESTful API `POST` is meant to create a new item underneath the path we are using, this is not the case here). However, for JSON format I need a way to differentiate between creating a new dataset and updating an existing. Since creating a new dataset is not idempotent (since some values are generated random), I feel `POST` is a good choice for that particular use case, even though it doesn't create a sub-item.

src/rest/openapi.yaml
agners: So far I avoided adding too much "business logic" to the REST API, adding an implicit default for the REST API only feels wrong to me. However, I am fine enforcing a delay timer value.

src/rest/openapi.yaml
wgtdkp: > Ideally we should only accept PUT, as we do not create a sub-item (in a RESTful API POST is meant to create a new item underneath the path we are using, this is not the case here). 

Make sense.

> However, for JSON format I need a way to differentiate between creating a new dataset and updating an existing.

Why JSON format will be different? I don't think it's a good idea to have any semantic implicities depending on the request payload format. And, a Dataset TLV binary array may include partial dataset as well.

On the other hand, I think the behavior of PUT can be defined as below to make it `idempotent`:
1. create a new random dataset if no existing dataset
2. update the (existing) dataset with the request pyaload

This is `idempotent`, because we are in the same state if the same PUT requests have been processed for multiple times.

agners: Hm, merging dataset in TLV format is not something we have code for, from what I can tell. In my mind, when using dataset format one almost always handles complete configuration (e.g. from Android API or somewhere else).

Also, if we only use `PUT`, how can we avoid an update (e.g. if one of the fields should no longer be set/cleared)?

> 1. create a new random dataset if no existing dataset

Google defines Idempotence as no matter how many times you execute them, you achieve the same result. But if random data is generated, that would not be idempotent IMHO (the reason I went for `POST` operation for the dataset JSON API).

wgtdkp: > Hm, merging dataset in TLV format is not something we have code for, from what I can tell. In my mind, when using dataset format one almost always handles complete configuration (e.g. from Android API or somewhere else).

Almost true, but the RESTful API isn't necessary to make assumptions on Android API.

> Also, if we only use PUT, how can we avoid an update (e.g. if one of the fields should no longer be set/cleared)?

Sorry, I didn't get your point, could you elaborate?

> Google defines Idempotence as "no matter how many times you execute them [operations], you achieve the same result". But if random data is generated, that would not be idempotent IMHO (the reason I went for POST operation for the dataset JSON API).

It's okay, because only the first the PUT request will cause generation of random dataset, the second PUT with the same payload will simply do nothing because we have an existing dataset now.

agners: > > Also, if we only use PUT, how can we avoid an update (e.g. if one of the fields should no longer be set/cleared)?
> 
> Sorry, I didn't get your point, could you elaborate?
> 

Currently, when using Dataset TLV, the user controls all the fields which are being set (e.g. if no active operational data set is present, and the user provides a operational dataset without, say, `ExtPanId`, after the operation the active operational dataset will be without `ExtPanId` (this might be a bad example as this is then not a valid configuration? My point is the dataset TLV is applied as is, nothing more, nothing less). It won't be possible to "clear" fields, as we only have the `PUT` method and always use the merge operation.

If we adopt the behavior of `application/json`, it will mean that we generate a dataset and merge the provided one always. The user can not 100% be certain that only the fields of the provided dataset TLV get applied.

Maybe this is the behavior we want, I am just pointing out differences.

> > Google defines Idempotence as "no matter how many times you execute them [operations], you achieve the same result". But if random data is generated, that would not be idempotent IMHO (the reason I went for POST operation for the dataset JSON API).
> 
> It's okay, because only the first the PUT request will cause generation of random dataset, the second PUT with the same payload will simply do nothing because we have an existing dataset now.

Hm, I see, so you'd argue that makes it idempotent. In my mind because randomness is involved, the operation is not idempotent by definition. But I can also live with your interpretation :smile: 

If we are being pedantic, we violate HTTP/REST in some way or another anyways (when using `POST`, because we are not creating a subordinate resource of the resource identified by the Request-URI, when using `PUT` because randomness can be seen as not idempotent operation).

I see your point in making the JSON (`application/json`) and Dataset TLV (`text/plain`) format behave the same way. So following that approach we'd end up with:

- Only `PUT` operation is supported
- If the target dataset doesn't exist, one will get created automatically and the provided fields will be merged (dataset and dataset TLV, `201 Created` is returned)
- If the target dataset exists the provided fields will be merged (dataset and dataset TLV, `200 OK` is returned)
- `PUT` on active dataset is rejected if Thread is running `409 Status Conflict`
- `PUT` on pending dataset is rejected if no delay attribute is provided (dataset and dataset TLV) `400 Bad Request`

wgtdkp: > Maybe this is the behavior we want, I am just pointing out differences.

Thanks! Yeah, setting partial operational dataset is discouraged. If you want to support this, you should support the JSON format as well.

> Hm, I see, so you'd argue that makes it idempotent. In my mind because randomness is involved, the operation is not idempotent by definition

You are probably referring to "deterministic" rather than "idempotent"?
> Idempotence ([UK](https://en.wikipedia.org/wiki/British_English): [/ˌɪdɛmˈpoʊtəns/](https://en.wikipedia.org/wiki/Help:IPA/English),[[1]](https://en.wikipedia.org/wiki/Idempotence#cite_note-1) [US](https://en.wikipedia.org/wiki/American_English): [/ˈaɪdəm-/](https://en.wikipedia.org/wiki/Help:IPA/English))[[2]](https://en.wikipedia.org/wiki/Idempotence#cite_note-2) is the property of certain [operations](https://en.wikipedia.org/wiki/Operation_(mathematics)) in [mathematics](https://en.wikipedia.org/wiki/Mathematics) and [computer science](https://en.wikipedia.org/wiki/Computer_science) whereby they can be applied multiple times without changing the result beyond the initial application.

This is the wifi definition of `idempotent`, I don't think my PUT operation is violating this :)

wgtdkp: > - Only PUT operation is supported
> - If the target dataset doesn't exist, one will get created automatically and the provided fields will be merged (dataset and dataset TLV, 201 Created is returned)
> - If the target dataset exists the provided fields will be merged (dataset and dataset TLV, 200 OK is returned)
> - PUT on active dataset is rejected if Thread is running 409 Status Conflict
> - PUT on pending dataset is rejected if no delay attribute is provided (dataset and dataset TLV) 400 Bad Request

All looks good except returning 201 - this code should only be used for response of a POST operation? https://restfulapi.net/http-status-201-created:
> HTTP Status 201 indicates that as a result of HTTP POST request, one or more new resources have been successfully created on the server.

wgtdkp: SGTM!

src/rest/openapi.yaml
wgtdkp: > However, I am fine enforcing that a delay timer is present.

This works for me

agners: I've discovered one downside with that approach when using TLVs: The client needs to be able to add a new field to the TLV.

Before, we could set a pending operational dataset using TLV  (e.g. from Android API), and then set the delay timer using a JSON request.

Now, the client needs to be able to extend the TLV we get from Android API and extend that with a delay timer :cold_sweat: 

Adding a default timer would be a work around. But I actually kinda liked the behavior before where we were able to modify the pending dataset and then set the delay timer in one last request.

wgtdkp: > Before, we could set a pending operational dataset using TLV (e.g. from Android API), and then set the delay timer using a JSON request.
>
> Now, the client needs to be able to parse the TLV we get from Android API and extend that with a delay timer cold_sweat

What's the "Android API" you are referring to? a Pending Operational Dataset without a Delay Timer shouldn't be considered valid I believe.

wgtdkp: Is it https://developers.home.google.com/thread?

agners: > What's the "Android API" you are referring to? a Pending Operational Dataset without a Delay Timer shouldn't be considered valid I believe.

Yes, the API you've linked. Well, if it is an active operational dataset, no delay timer is normal right?

Maybe it is possible to get a dataset and modify it using the Android API, I am not very familiar with it.

Let's assume you'd use this REST API to synchronize the credentials of two OTBR's. One of the OTBR is already running, so you are reading its active operational dataset. Now you need to add a timer delay to it, and set it as pending. You have to have the knowledge how to change the TLVs...

wgtdkp: > Yes, the API you've linked. Well, if it is an active operational dataset, no delay timer is normal right?

Yes, Active Operational Dataset doesn't have delay timer.

One path is we add new Android/GMS API to convert Active Operational Dataset to Pending Dataset, but this will generally take a few month to release.

Another solution (and I prefer) is to define Pending Operational Dataset as a composition of `Active Operational Dataset`, `Pending Timestamp` and `Delay Timer`. For example:

```
// PendingOperational Dataset
{
    "ActiveDataset": "000102030ADEFBCDD...",
    "PendingTimestamp": "1",
    "DelayTimer": "200",
}

// And format with JSON ActiveDataset
{
    "ActiveDataset": {
        "Channel": "13",
        "NetworkName": "MyThreadNet",
        // ......
    },
    "PendingTimestamp": "1",
    "DelayTimer": "200",
}

```

We don't need to define strict 1-to-1 mapping to the openthread API for REST - I think we can define the REST API in a way which it's most easy-to-use for typical REST clients such as Apps or service backends. Those client don't need to know the underlying openthread API or structs, so we are free to define the REST data structure.

wgtdkp: The REST Operational Dataset definition is even better because it expresses the relationship that "A Pending Dataset includes an Active Dataset"

agners: I've implemented what we discussed yesterday (`PUT` only, same semantics for JSON and TLV). It requires a change in OpenThread as well: https://github.com/openthread/openthread/pull/8669

I plan to tackle Pending Operational Dataset as a composition of Active Operational Dataset, Pending Timestamp and Delay Timer as well.

jsaiko:
Is there anything I can do to help get this merged?  I have to use Agners fork in order to work with Homeassistant.

agners:
> Is there anything I can do to help get this merged? I have to use Agners fork in order to work with Homeassistant.

To implement the latest feedback a new API in the core of `openthread` is required. There was some discussion was on-going for a while, but we came to an agreement, see https://github.com/openthread/openthread/pull/8669#issuecomment-1421263535.

This needs to be implemented, so we can use it here. I probably get around to do it next week. Unless you happen to have spare cycles :smile:

bdunn44:
+1 for this! Can't use the OTBR functionality in HA without it as far as I can tell. If this was Python or Java I'd contribute it myself but fear submitting my first CPP code here may do more harm than good 🙂

wgtdkp (approved):
LGTM :+1: 
@agners Thanks for your contribution! This is the most significant improvement to the REST API after its initial commit :1st_place_medal: (Especially the yaml API spec!)

src/rest/response.cpp
```diff
@@ -30,12 +30,11 @@
 
 #include <stdio.h>
 
-#define OT_REST_RESPONSE_CONTENT_TYPE_JSON "application/json"
 #define OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_ORIGIN "*"
 #define OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_HEADERS                                                              \
     "Access-Control-Allow-Headers, Origin,Accept, X-Requested-With, Content-Type, Access-Control-Request-Method, " \
     "Access-Control-Request-Headers"
-#define OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_METHOD "GET"
+#define OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_METHOD "GET, OPTIONS, PUT"
```
bukepo, line 37: does it make sense to support `DELETE` for leaving the current network?

src/rest/response.cpp
```diff
@@ -49,7 +48,7 @@ Response::Response(void)
     mProtocol = "HTTP/1.1";
 
     // Pre-defined headers
-    mHeaders["Content-Type"]                 = OT_REST_RESPONSE_CONTENT_TYPE_JSON;
+    mHeaders[OT_REST_CONTENT_TYPE_HEADER]    = OT_REST_CONTENT_TYPE_JSON;
     mHeaders["Access-Control-Allow-Origin"]  = OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_ORIGIN;
     mHeaders["Access-Control-Allow-Methods"] = OT_REST_RESPONSE_ACCESS_CONTROL_ALLOW_METHOD;
```
bukepo, line 53: Not all resources support these `PUT`, should we set this by resource?

bukepo (approved):
We may resolve those comments in future PRs.

src/rest/response.cpp
agners: Yeah seems sensible. I'd prefer to do this in a follow up PR.

src/rest/response.cpp
agners: It doesn't matter much as this is only about which methods are allowed. Since they are not implemented, it doesn't really matter weather they are allowed.

wgtdkp:
@agners There are coding format issues on this PR. You may want to run `./script/make-pretty clang` on your code.

agners:
> @agners There are coding format issues on this PR. You may want to run `./script/make-pretty clang` on your code.

Ugh, yeah used to go through at one point, but doesn't it seem. I don't have Clang 14 locally anymore, I'll check if I can downgrade to run this.

src/rest/json.cpp
```diff
@@ -482,6 +608,306 @@ std::string Error2JsonString(HttpStatusCode aErrorCode, std::string aErrorMessag
...
+    return node;
+}
+
+std::string ActiveDataset2JsonString(const otOperationalDataset &aActiveDataset)
+{
+    cJSON      *node;
+    std::string ret;
+
+    node = ActiveDataset2Json(aActiveDataset);
+    ret = Json2String(node);
```
wgtdkp, line 665:
```suggestion
    ret  = Json2String(node);
```

src/rest/json.cpp
```diff
@@ -482,6 +608,306 @@ std::string Error2JsonString(HttpStatusCode aErrorCode, std::string aErrorMessag
...
+    if (cJSON_IsObject(value))
+    {
+        VerifyOrExit(JsonActiveDataset2Dataset(value, aDataset), ret = false);
+    }
+    else if (cJSON_IsString(value))
+    {
+        otOperationalDatasetTlvs datasetTlvs;
+        int                      len;
+
+        len = Hex2BytesJsonString(std::string(value->valuestring), datasetTlvs.mTlvs, OT_OPERATIONAL_DATASET_MAX_LENGTH);
```
wgtdkp, line 867:
```suggestion
        len =
            Hex2BytesJsonString(std::string(value->valuestring), datasetTlvs.mTlvs, OT_OPERATIONAL_DATASET_MAX_LENGTH);
```

src/utils/hex.cpp
```diff
@@ -96,35 +96,49 @@ size_t Bytes2Hex(const uint8_t *aBytes, const uint16_t aBytesLength, char *aHex)
...
         uint8_t byte = longValue & 0xff;
         snprintf(byteHex, sizeof(byteHex), "%02X", byte);
-        hexString += byteHex;
+        strcat(aHex, byteHex);
         longValue = longValue >> 8;
     }
-    strcpy(aHex, hexString.c_str());
-    return strlen(aHex);
+
+        return strlen(aHex);
```
wgtdkp, line 141:
```suggestion
    return strlen(aHex);
```

wgtdkp (approved):

agners:
@wgtdkp thanks for tackling this! :pray:

jwhui (approved):

PR #1690, opened 2023-01-03
[border-agent] refactor `BorderAgent::PublishMeshCopService()`

superwhd:


src/border_agent/border_agent.hpp
```diff
@@ -154,6 +153,7 @@ class BorderAgent : private NonCopyable
         uint32_t ToUint32(void) const;
     };
```
wgtdkp, line 154: What about moving those classes to .cpp file as well?

wgtdkp (approved):
LGTM :+1:

src/border_agent/border_agent.hpp
superwhd: Done

jwhui (approved):

PR #1700, opened 2023-01-09
[bootstrap] fix to work on IPv6 only machines

### LINKED ISSUE #1697: bootstrap script fails on IPv6 only machine
**Describe the bug** A clear and concise description of what the bug is.

If a machine is run in an IPv6 only network, then the bootstrap script fails at the MDNS wget step, because the step is explicitly configured for IPv4 only (with the -4 flag).  This is just a wget download, so I can't think of any reason why it would be set to IPv4 only. I removed the flag and the script completed fine.

(Note: The target file is from github.com, which is an IPv4 only service, but an IPv6 network will usually be using something like DNS64+NAT64 to connect. This is not much different from a private IPv4 network which can't connect directly either, and has to use NAT44.)

**To Reproduce** Information to reproduce the behavior, including:

1. Git commit id = adba2d91d9caaf4eccec2c0b07f06e25202fe26b
2. IEEE 802.15.4 hardware platform = MakerDiary nRF52840 USB Dongle (Radio Co-Processor); host is Raspberry Pi 4.
3. Build steps = first step, i.e. ./script/bootstrap
4. Network topology = n/a

Steps:
1. Run your host in an IPv6 only network (you may want to run a separate VLAN for this, as many things don't work in IPv6 only). You need to have an interop technology like DNS64 + NAT64.
2. Clone the repository to the local machine
3. Run ./script/bootstrap
4. What breaks:

* Script fails at MDNS wget due to IPv4 only limit

**Expected behavior** A clear and concise description of what you expected to happen.

* There is no need for wget to be restricted to IPv4 as it is simply downloading. The script should download on either network (provided the network supports it, e.g. a IPv4 private network needs a NAT44 service to connect to GitHub, and IPv6 needs NAT64 + DNS64).

**Console/log output** If applicable, add console/log output to help explain your problem.

Output where the bootstrap script fails:

```
+ wget -4 --no-check-certificate https://github.com/apple-oss-distributions/mDNSResponder/archive/refs/tags/mDNSResponder-1310.80.1.tar.gz
--2023-01-06 20:04:48--  https://github.com/apple-oss-distributions/mDNSResponder/archive/refs/tags/mDNSResponder-1310.80.1.tar.gz
Resolving github.com (github.com)... 20.248.137.48
Connecting to github.com (github.com)|20.248.137.48|:443... failed: No route to host.
```

Note that github.com resolves to a NAT64 address on my network (and connections work fine):

```
pi@pi2:~/ot-br-posix $ nslookup github.com
Server:		fd7c:e25e:67e8::1
Address:	fd7c:e25e:67e8::1#53

Non-authoritative answer:
Name:	github.com
Address: 20.248.137.48
Name:	github.com
Address: 64:ff9b::14f8:8930
```

**Additional context** Add any other context about the problem here.

Work around (and fix):
* Remove the `-4` from the `wget` line and it downloads correctly.

sgryphon:
No reason to limit wget to IPv4 only, which breaks IPv6 only machines.

Fixes #1697

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -58,7 +58,7 @@ install_packages_apt()
     sudo apt-get install --no-install-recommends -y libavahi-client3 libavahi-common-dev libavahi-client-dev avahi-daemon
     (MDNS_RESPONDER_SOURCE_NAME=mDNSResponder-1310.80.1 \
         && cd /tmp \
-        && wget -4 --no-check-certificate https://github.com/apple-oss-distributions/mDNSResponder/archive/refs/tags/$MDNS_RESPONDER_SOURCE_NAME.tar.gz \
+        && wget --no-check-certificate https://github.com/apple-oss-distributions/mDNSResponder/archive/refs/tags/$MDNS_RESPONDER_SOURCE_NAME.tar.gz \
         && mkdir -p $MDNS_RESPONDER_SOURCE_NAME \
         && tar xvf $MDNS_RESPONDER_SOURCE_NAME.tar.gz -C $MDNS_RESPONDER_SOURCE_NAME --strip-components=1 \
         && cd /tmp/$MDNS_RESPONDER_SOURCE_NAME/Clients \


jwhui (approved):
Thanks! 👍🏼

PR #1701, opened 2023-01-09
[test] move the config file to the correct path for testing

zhanglongxia:
The config file uses a relative path. When the test script calls the otbr-agent in the test folder, it causes the otbr-agent failed to find the config file. This commit moves the config file to the correct path before running the tests and enables the power calibration.


abtink (approved):
👍

PR #1713, opened 2023-01-16
[rest] remove superfluous space in HTTP status line

agners:
A space is already added in Response::Serialize(). According to the HTTP specification there should only be a single space between the protcol version and the status code.


wgtdkp (approved):
Thanks!

PR #1722, opened 2023-01-21
[docker] perform a proper shutdown

fgervais:
Catch the docker sigterm and handle a clean shutdown.

At least those advantages come to mind:

1. Allows for a faster shutdown (about 1.5s as seen in the example below) because we don't have to wait until docker times out (10s) and kills everything.

```
container-otbr-autostart-otbr-1   | Jan 21 14:28:04 nuc2 otbr-agent[93]: 00:00:15.619 [I] BorderRouter--: RouterAdvert: Added RIO for fd04:2240:0:20::/64 (lifetime=1800)
container-otbr-autostart-otbr-1   | Jan 21 14:28:04 nuc2 otbr-agent[93]: 00:00:15.619 [I] BorderRouter--: Sent Router Advertisement on infra netif 2
container-otbr-autostart-otbr-1   | Jan 21 14:28:04 nuc2 otbr-agent[93]: 00:00:15.619 [I] BorderRouter--: Start evaluating routing policy, scheduled in 16000 milliseconds
container-otbr-autostart-otbr-1   | Jan 21 14:28:04 nuc2 otbr-agent[93]: 00:00:15.619 [I] BorderRouter--: Received Router Advertisement from fe80:0:0:0:baae:edff:fe71:259d on infra netif 2
^CGracefully stopping... (press Ctrl+C again to force)
[+] Running 2/2
 ⠿ Container container-otbr-autostart-setup-1  Stopped                                                                                                                               0.0s
 ⠿ Container container-otbr-autostart-otbr-1   Stopped                                                                                                                               1.5s
``` 

2. If running the container with `--network host`, it won't leave iptables rules behind.

### DIFF
--- a/etc/docker/docker_entrypoint.sh
+++ b/etc/docker/docker_entrypoint.sh
@@ -65,6 +65,15 @@ function parse_args()
     done
 }
 
+function shutdown()
+{
+    echo "Shutting down"
+    /app/script/server shutdown
+    exit 0
+}
+
+trap shutdown TERM INT
+
 parse_args "$@"
 
 [ -n "$RADIO_URL" ] || RADIO_URL="spinel+hdlc+uart:///dev/ttyUSB0"
@@ -96,4 +105,5 @@ while [[ ! -f /var/log/syslog ]]; do
     sleep 1
 done
 
-tail -f /var/log/syslog
+tail -f /var/log/syslog &
+wait $!
--- a/script/server
+++ b/script/server
@@ -65,4 +65,33 @@ main()
     . "$AFTER_HOOK"
 }
 
-main
+shutdown()
+{
+    nat64_stop || echo 'Failed to start NAT64!'
+    dns64_stop || echo 'Failed to start DNS64!'
+    firewall_stop || echo 'Failed to start firewall'
+    if have systemctl; then
+        systemctl is-active rsyslog && sudo systemctl stop rsyslog || echo 'Failed to stop rsyslog!'
+        systemctl is-active dbus && sudo systemctl stop dbus || echo 'Failed to stop dbus!'
+        systemctl is-active avahi-daemon && sudo systemctl stop avahi-daemon || echo 'Failed to stop avahi!'
+        without WEB_GUI || systemctl is-active otbr-web && sudo systemctl stop otbr-web || echo 'Failed to stop otbr-web!'
+        systemctl is-active otbr-agent && sudo systemctl stop otbr-agent || echo 'Failed to stop otbr-agent!'
+    elif have service; then
+        sudo service rsyslog status && sudo service rsyslog stop || echo 'Failed to stop rsyslog!'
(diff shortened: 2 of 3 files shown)


fgervais:
I'm not sure I understood your request. Did you mean something like in this last commit?

script/server
```diff
@@ -65,4 +65,32 @@ main()
...
+        echo 'Unable to find service manager. Try script/console to stop in console mode!'
+    fi
+}
+
+if [ "${1-}" == "shutdown" ]
+then
+    shutdown
+else
+    main
+fi
```
bukepo, line 96: prefer main function as the single entry point for a script. How about rename the current main() to startup() and wrap these lines as the new main()?
see more: https://google.github.io/styleguide/shellguide.html#main
```suggestion
main()
{
    if [[ "${1-}" == "shutdown" ]]; then
        shift
        shutdown "$@"
    else
        startup "$@"
    fi
}
```

bukepo:
> I'm not sure I understood your request. Did you mean something like in this last commit?

yes, just to verify the proper shutdown works as expected.

fgervais:
I rebased on top of `main` to include https://github.com/openthread/ot-br-posix/pull/1719 which this PR depends on.

script/server
fgervais: I think it's good now.

etc/docker/docker_entrypoint.sh
```diff
@@ -96,4 +105,5 @@ while [[ ! -f /var/log/syslog ]]; do
     sleep 1
 done
 
-tail -f /var/log/syslog
+tail -f /var/log/syslog &
+wait $!
```
bukepo, line 109: is this change necessary?

fgervais: Yes, the script needs to be the one catching the sigterm from docker to execute the shutdown procedure. Without this, the child process will get it instead.

bukepo: thanks!

script/server
```diff
@@ -65,4 +65,37 @@ main()
     . "$AFTER_HOOK"
 }
 
-main
+shutdown()
+{
+    nat64_stop || echo 'Failed to start NAT64!'
+    dns64_stop || echo 'Failed to start DNS64!'
```
bukepo, line 71:
```suggestion
    nat64_stop || echo 'Failed to stop NAT64!'
    dns64_stop || echo 'Failed to stop DNS64!'
```

script/server
```diff
@@ -65,4 +65,37 @@ main()
...
+    firewall_stop || echo 'Failed to start firewall'
```
bukepo, line 72:
```suggestion
    firewall_stop || echo 'Failed to stop firewall'
```

bukepo (approved):

jwhui (approved):

PR #1742, opened 2023-02-06
[mdns] publish a service using the default hostname as the instance name

superwhd:
This PR has 2 enhancements for mDNS publisher which may be useful for supporting SRPL.

- Fix a potential bug: A service subscription callback may modify the container `mDiscoveredCallbacks`. This can invalidate an iterator and results in a crash.
- Supports publishing a service using the hostname as the service instance name.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -177,6 +177,8 @@ uint64_t Publisher::AddSubscriptionCallbacks(Publisher::DiscoveredServiceInstanc
 
 void Publisher::OnServiceResolved(const std::string &aType, const DiscoveredInstanceInfo &aInstanceInfo)
 {
+    std::vector<uint64_t> subscriberIds;
+
     otbrLogInfo("Service %s is resolved successfully: %s %s host %s addresses %zu", aType.c_str(),
                 aInstanceInfo.mRemoved ? "remove" : "add", aInstanceInfo.mName.c_str(), aInstanceInfo.mHostName.c_str(),
                 aInstanceInfo.mAddresses.size());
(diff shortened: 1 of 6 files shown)


src/mdns/mdns.hpp
```diff
@@ -189,7 +189,8 @@ class Publisher : private NonCopyable
      *                          provided, this service resides on local host and it is the implementation
      *                          to provide specific host name. Otherwise, the caller MUST publish the host
      *                          with method PublishHost.
-     * @param[in] aName         The name of this service.
+     * @param[in] aName         The name of this service. If an empty string is provided, the service's name will be the
```
wgtdkp, line 192: Why this is useful and necessary?

superwhd: SRP Replication's [RFC draft](https://www.ietf.org/archive/id/draft-lemon-srp-replication-02.html#name-advertising-ones-own-replic) asks to use the hostname as the service instance name.

src/mdns/mdns.cpp
```diff
@@ -193,11 +195,21 @@ void Publisher::OnServiceResolved(const std::string &aType, const DiscoveredInst
     UpdateMdnsResponseCounters(mTelemetryInfo.mServiceResolutions, OTBR_ERROR_NONE);
     UpdateServiceInstanceResolutionEmaLatency(aInstanceInfo.mName, aType, OTBR_ERROR_NONE);
 
+    subscriberIds.reserve(mDiscoveredCallbacks.size());
     for (const auto &subCallback : mDiscoveredCallbacks)
     {
-        if (subCallback.second.first != nullptr)
+        subscriberIds.push_back(subCallback.first);
```
wgtdkp, line 201: nit: document why the copy is needed.

src/mdns/mdns.hpp
```diff
@@ -561,6 +563,19 @@ class Publisher : private NonCopyable
...
+    {
+        for (const auto &serviceReg : mServiceRegistrations)
+        {
+            if (serviceReg.second->mType == aType)
+            {
+                return serviceReg.second.get();
+            }
+        }
+        return nullptr;
+    }
```
wgtdkp, line 578: 1. When is this method used?
2. Does it have to be `public`?
3. Move the implementation to `mdns.cpp`.

superwhd: Removed this part for the PR. I may add it back in a future one.

wgtdkp (approved):

jwhui (approved):

PR #1745, opened 2023-02-07
[tests] update router reset wait time

jwhui:
To resolve: https://github.com/openthread/ot-br-posix/actions/runs/4112181186/jobs/7106327761


sunytt (approved):
LGTM, thx

An improvement for future could be to add definition for the delay numbers.

PR #1756, opened 2023-02-17
[docker] accelerate building an OTBR docker image by cache

superwhd:
This PR changes the order of the steps for building the OTBR docker image. In this way we can leverage more on the docker build cache and save time when building an image. 

When we want to build an image after only modifying C++ code, we won't need to rerun the bootstrap step.

Local runs comparison:
- Before:
  - Build without cache: 7m32.835s
  - Build with cache: 6m40.570s
- After:
  - Build without cache: 6m32.783s
  - Build with cache: 2m13.498s

### DIFF
--- a/etc/docker/Dockerfile
+++ b/etc/docker/Dockerfile
@@ -63,9 +63,6 @@ ENV DOCKER 1
 
 RUN env
 
-COPY . /app
-WORKDIR /app
-
 # Required during build or run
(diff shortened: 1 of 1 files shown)


bukepo (approved):

Irving-cl (approved):
LGTM

jwhui (approved):

PR #1761, opened 2023-02-21
[trel] disable trel if feature flag is used

sunytt:
This PR sets trel to disabled if feature flag is used.

Added a test in border_router.yml with both trel and feature flag compiled.

Updated border_router.yml tests to use the OTBR_TREL flag.

### DIFF
--- a/.github/workflows/border_router.yml
+++ b/.github/workflows/border_router.yml
@@ -50,56 +50,56 @@ jobs:
       matrix:
         include:
           - name: "Border Router (mDNSResponder)"
-            otbr_options: "-DOTBR_TREL=OFF -DOT_MLR=ON -DOTBR_COVERAGE=ON -DOT_SRP_SERVER=ON -DOT_ECDSA=ON -DOT_SERVICE=ON -DOTBR_DUA_ROUTING=ON -DOT_DUA=ON"
+            otbr_options: "-DOT_DUA=ON -DOT_ECDSA=ON -DOT_MLR=ON -DOT_SERVICE=ON -DOT_SRP_SERVER=ON -DOTBR_COVERAGE=ON -DOTBR_DUA_ROUTING=ON -DOTBR_TREL=OFF"
             border_routing: 1
             nat64: 0
             otbr_mdns: "mDNSResponder"
             cert_scripts: ./tests/scripts/thread-cert/border_router/*.py
             packet_verification: 1
           - name: "Border Router (Avahi)"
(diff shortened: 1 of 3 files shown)


.github/workflows/border_router.yml
```diff
@@ -50,54 +50,61 @@ jobs:
       matrix:
         include:
           - name: "Border Router (mDNSResponder)"
-            otbr_options: "-DOT_TREL=OFF -DOT_MLR=ON -DOTBR_COVERAGE=ON -DOT_SRP_SERVER=ON -DOT_ECDSA=ON -DOT_SERVICE=ON -DOTBR_DUA_ROUTING=ON -DOT_DUA=ON"
+            otbr_options: "-DOTBR_TREL=OFF -DOT_MLR=ON -DOTBR_COVERAGE=ON -DOT_SRP_SERVER=ON -DOT_ECDSA=ON -DOT_SERVICE=ON -DOTBR_DUA_ROUTING=ON -DOT_DUA=ON"
```
superwhd, line 53: Shall we group OTBR options and OT options respectively?

src/ncp/ncp_openthread.cpp
```diff
@@ -215,6 +215,11 @@ void ControllerOpenThread::Init(void)
         VerifyOrExit(result == OT_ERROR_NONE, error = OTBR_ERROR_OPENTHREAD);
     }
 
+#if OTBR_ENABLE_FEATURE_FLAGS && OTBR_ENABLE_TREL
+    // Disable trel when feature flag is enabled, let feature flag handle enable/disable trel
+    otTrelSetEnabled(mInstance, false);
+#endif
```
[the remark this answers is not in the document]
sunytt: Yes, by default trel is enabled if feature_flags is disabled.
This is to keep the default behavior so if other teams rely on it then we won't break their code.

superwhd (approved):
LGTM :+1:

abtink (approved):
👍

src/ncp/ncp_openthread.cpp
```diff
@@ -215,6 +215,11 @@ void ControllerOpenThread::Init(void)
         VerifyOrExit(result == OT_ERROR_NONE, error = OTBR_ERROR_OPENTHREAD);
     }
 
+#if OTBR_ENABLE_FEATURE_FLAGS && OTBR_ENABLE_TREL
+    // Disable trel when feature flag is enabled, let feature flag handle enable/disable trel
+    otTrelSetEnabled(mInstance, false);
```
abtink, line 220: Is it possible to get here with feature flags wanting the TREL to be enabled already?

sunytt: Good point, the feature flags are updated by thread daemon after otbr-agent signals OTBR_DBUS_SIGNAL_READY, so here we don't get the updated value. But the default feature flag value can be 'true' or 'false' so I think we should use that instead. Will update.

jwhui (approved):

PR #1771, opened 2023-02-28
[script] allow OTBR options with space

agners:
Use parenthesis syntax to assign OTBR_OPTIONS to the options array. This allows to pass strings with spaces (in quotes) as OTBR options, e.g. "-DOTBR_VENDOR_NAME=My Vendor".


agners:
This doesn't seem to work as intended. Strings passed with quotes still don't get applied with spaces.

PR #1774, opened 2023-03-01
[rest] explicitly set Connection header to close

agners:
By default, HTTP 1.1 connections should stay open after a transaction. However, that is not how the current REST server implementation behaves: After each transaction the HTTP connection is being closed by the server.

Set the Connection header to "close" to tell the client about this behavior.


wgtdkp (approved):
LGTM :+1:
thanks!

jwhui (approved):
Thanks! 👍🏼

PR #1780, opened 2023-03-08
Keep avahi resolver until service instance is removed for trel

sunytt:
In the current avahi mdns design, after service browse we resolve the discovered instance once, when instance is resolved the resolver is removed.

Trel uses dnssd browse to discover peers, when a discovered peer changes ipv6 address, trel would not know the change and still try to send packets to the old address.

This PR lets trel service subscription to keep the avahi service resolver until the instance is removed, so any updates to the instance will be noticed.

A followup to this PR is to add test in openthread.

### DIFF
--- a/src/mdns/mdns.hpp
+++ b/src/mdns/mdns.hpp
@@ -258,9 +258,15 @@ public:
      *
      * @param[in] aType          The service type.
      * @param[in] aInstanceName  The service instance to subscribe, or empty to subscribe the service.
+     * @param[in] aKeepAlive     Whether the subscription should be kept alive to receive updates to the service
+     *                           instance. Note this only applies to avahi, as for avahi browse operation,
+     *                           instance updates will not trigger add/rm events.
+     *                           For dnssd browse operation, updates of instances will trigger add/rmv events.
      *
      */
-    virtual void SubscribeService(const std::string &aType, const std::string &aInstanceName) = 0;
+    virtual void SubscribeService(const std::string &aType,
+                                  const std::string &aInstanceName,
+                                  const bool         aKeepAlive) = 0;
 
     /**
      * This method unsubscribes a given service or service instance.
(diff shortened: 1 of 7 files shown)


src/mdns/mdns.hpp
```diff
@@ -258,9 +258,15 @@ class Publisher : private NonCopyable
...
+     * @param[in] aKeepAlive     Whether the subscription should be kept alive to receive updates to the service
+     *                           instance. Note this only applies to avahi, as for avahi browse operation,
+     *                           instance updates will not trigger add/rm events.
+     *                           For dnssd browse operation, updates of instances will trigger add/rmv events.
      *
      */
-    virtual void SubscribeService(const std::string &aType, const std::string &aInstanceName) = 0;
+    virtual void SubscribeService(const std::string &aType,
+                                  const std::string &aInstanceName,
+                                  const bool         aKeepAlive) = 0;
```
superwhd, line 269: I'm thinking of if we are able to make 'keepalive' as the default behavior for all mDNS publishers.

As far as I know we have two use cases of subscribing a service:
1. Continuous browsing & resolving: TREL and SRPL. When a service updates its address, we want to get notified. 
 - mDNSResponder will send a remove event and an add event. The 'remove' event is optional from TREL/SRPL's perspective.
 - If we keep the Avahi resolver alive, Avahi will send an add event.

      The behaviors should be fine when we enable 'keep alive' for Avahi.

2. Discovery Proxy. In this case Discovery Proxy will unsubscribe the service after the service is resolved. The behaviors of Avahi and mDNSResponder are the same which is not impacted by 'keep alive'.

sunytt: Agree that we can make 'keepalive' as default behavior for all. When the caller of PublisherAvahi::SubscribeService doesn't want to get updates for the instances it should unsubscribe.

Will close this PR and open a new one.

PR #1794, opened 2023-03-20
[script] only install a single mDNS publisher

bukepo:


script/server
```diff
@@ -86,6 +89,8 @@ shutdown()
     else
         echo 'Unable to find service manager. Try script/console to stop in console mode!'
     fi
+
+    ! lsof -i udp:5353 || echo 'Failed to stop mDNS publisher'
```
superwhd, line 93: What's the purpose of having an `!`?

PR #1810, opened 2023-03-31
[mdns] publish 'id' as Extended MAC Address

wgtdkp:
An unique ID is required for adding the Thread network credentials
of this BR to Android or iOS system. Products can use the dbus
"UpdateVendorMeshCopTxtEntries" method to set a vendor-specific ID,
or otherwise, enable the "OTBR_PUBLISH_ID_AS_EXT_ADDRESS" feature
to set ID to Extended MAC Address by padding zeros at the tail.

### DIFF
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -127,3 +127,8 @@ option(OTBR_DNS_UPSTREAM_QUERY "Allow sending DNS queries to upstream" OFF)
 if (OTBR_DNS_UPSTREAM_QUERY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_DNS_UPSTREAM_QUERY=1)
 endif()
+
+option(OTBR_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS "Publish the MeshCoP mDNS 'id' TXT entry as the Extended MAC Address, enable this feature only when 'id' is not set via dbus API" OFF)
+if (OTBR_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS=1)
+endif()
--- a/src/border_agent/border_agent.cpp
+++ b/src/border_agent/border_agent.cpp
@@ -378,6 +378,19 @@ void BorderAgent::PublishMeshCopService(void)
     // "xa" stands for Extended MAC Address (64-bit) of the Thread Interface of the Border Agent.
     txtList.emplace_back("xa", extAddr->m8, sizeof(extAddr->m8));
 
+    // An unique ID is required for adding the Thread network credentials of this BR to Android
+    // or iOS system. Products can use the dbus "UpdateVendorMeshCopTxtEntries" method to set a
+    // vendor-specific ID, or otherwise, enable the feature here to set ID to Extended MAC Address
+    // by padding zeros at the tail.
+#if OTBR_ENABLE_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS
+    {
+        uint8_t id[16] = {0};
+
+        memcpy(id, extAddr->m8, sizeof(extAddr->m8));
+        txtList.emplace_back("id", id, sizeof(id));
+    }
+#endif
+
     state       = GetStateBitmap(*instance);
     stateUint32 = htobe32(state.ToUint32());
     txtList.emplace_back("sb", reinterpret_cast<uint8_t *>(&stateUint32), sizeof(stateUint32));


src/border_agent/border_agent.cpp
```diff
@@ -378,6 +378,19 @@ void BorderAgent::PublishMeshCopService(void)
...
 
+    // An unique ID is required for adding the Thread network credentials of this BR to Android
+    // or iOS system. Products can use the dbus "UpdateVendorMeshCopTxtEntries" method to set a
+    // vendor-specific ID, or otherwise, enable the feature here set ID to Extended MAC Address
+    // by padding zeros at the tail.
+#if OTBR_ENABLE_OTBR_PUBLISH_ID_AS_EXT_ADDRESS
+    {
+        uint8_t id[16] = {0};
+
+        memcpy(id, extAddr->m8, sizeof(extAddr->m8));
```
bukepo, line 389: is it better to put into the last 8 bytes? It might be easier for us to represent as `0x....`.

wgtdkp: Agree. But unfortunately there is legacy code in Android/GMS that zeros are padding at the tail, so here to follow it to avoid potential compatibility issues.

etc/cmake/options.cmake
```diff
@@ -127,3 +127,8 @@ option(OTBR_DNS_UPSTREAM_QUERY "Allow sending DNS queries to upstream" OFF)
 if (OTBR_DNS_UPSTREAM_QUERY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_DNS_UPSTREAM_QUERY=1)
 endif()
+
+option(OTBR_PUBLISH_ID_AS_EXT_ADDRESS "Publish the 'id' TXT entry as the Extended MAC Address, enable this feature only when 'id' is not set via dbus API" OFF)
+if (OTBR_PUBLISH_ID_AS_EXT_ADDRESS)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_OTBR_PUBLISH_ID_AS_EXT_ADDRESS=1)
+endif()
```
superwhd, line 134: I feel the second `OTBR` in `OTBR_ENABLE_OTBR_PUBLISH_ID_AS_EXT_ADDRESS` is redundant. Also it may worth specifying in the name that this option only applies to `meshcop` service.

wgtdkp: Thanks @superwhd!

> Also it may worth specifying in the name that this option only applies to meshcop service.

I am updating the help text to say "Publish the MeshCoP mDNS 'id' TXT entry as the Extended MAC Address...", should this help? Or do you have a name suggestion?

superwhd: I'd suggest mentioning `meshcop` in the cmake option and the macro's names. e.g. `OTBR_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS` and `OTBR_ENABLE_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS`.

wgtdkp: But it's `Border Agent/Router ID` rather than `MeshCoP ID`

superwhd: I agree on including `Border Agent` in the names. My point is to clarify the scope that the option applies to. There could be more than one ID concepts among all BR features.

wgtdkp: Okay, updated to `OTBR_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS`!

superwhd (approved):
LGTM :+1:

bukepo (approved):

bukepo (approved):

src/border_agent/border_agent.cpp
```diff
@@ -378,6 +378,19 @@ void BorderAgent::PublishMeshCopService(void)
     // "xa" stands for Extended MAC Address (64-bit) of the Thread Interface of the Border Agent.
     txtList.emplace_back("xa", extAddr->m8, sizeof(extAddr->m8));
 
+    // An unique ID is required for adding the Thread network credentials of this BR to Android
```
jwhui, line 381:
```suggestion
    // A unique ID is required for adding the Thread network credentials of this BR to Android
```

etc/cmake/options.cmake
```diff
@@ -127,3 +127,8 @@ option(OTBR_DNS_UPSTREAM_QUERY "Allow sending DNS queries to upstream" OFF)
 if (OTBR_DNS_UPSTREAM_QUERY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_DNS_UPSTREAM_QUERY=1)
 endif()
+
+option(OTBR_PUBLISH_MESHCOP_ID_AS_EXT_ADDRESS "Publish the MeshCoP mDNS 'id' TXT entry as the Extended MAC Address, enable this feature only when 'id' is not set via dbus API" OFF)
```
jwhui, line 131: Can we default to ON and allow override via API and/or DBus?

wgtdkp: I am okay to default ON.

But to allow override via API or d-bus, there can be race conditions that different IDs are advertised for the same device: if the API is called after the default MeshCoP mDNS service has been advertised. It's even worse to advertise a wrong ID than missing an ID for a short while. Thoughts?

bukepo: currently the mobile platform will fallback to `xa`, if so, it will have race condition whether or not the `id` is published.

jwhui: My concern here is that a default OTBR build will not include the `id` field, which can cause issues for those using mobile OS APIs. Can we have a build switch that configures how the `id` field is set?

wgtdkp: Even the Extended MAC Address version here is not a preferred generation of the ID, making it default ON is suggesting a non-preferred solution. If we want a default ID implementation, we should follow https://threadgroup.atlassian.net/browse/SPEC-1020 to randomly generate it and save in the persistent settings. This was not added in ot-br-posix because we think vendor BRs already have a kind of ID in their ecosystem. But it does seem to be the case?

What about highlighting the Thread BR requirements in the mobile OS API guide doc?

wgtdkp:
Closing this PR as it's superseded by https://github.com/openthread/ot-br-posix/pull/1832

PR #1814, opened 2023-04-05
[dbus] fix `Scan` crashing

jdswensen:
I ran into an issue while trying to use the `dbus` `Scan` API. For my platform (imx6ull, built with yocto), I was able to repeatably cause `otbr` to error by running:

```
dbus-send --reply-timeout=30000 --print-reply --system --type="method_call" --dest=io.openthread.BorderRouter.wpan0 /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.Scan
```

Callstack:

```
[ 00 ] __libc_do_syscall
[ 01 ] __pthread_kill_implementation ( pthread_kill.c:43 )
[ 02 ] raise ( raise.c:26 )
[ 03 ] abort ( abort.c:79 )
[ 04 ] 0x76ee4f63
[ 05 ] free_mem
[ 06 ] 0x76ed56a5
[ 07 ] 0x76eedbbe
[ 08 ] 0x76eec106
[ 09 ] 0x76ed5927
[ 10 ] 0x76ef27e2
[ 11 ] 0x76eedbbe
[ 12 ] 0x76eecbe2
[ 13 ] 0x76eec106
[ 14 ] 0x76eec106
[ 15 ] 0x76ec0d15
[ 16 ] 0x76eedbbe
[ 17 ] 0x76ec0bdf
[ 18 ] 0x76eedbbe
[ 19 ] otbr::DBus::DBusMessageEncode(DBusMessageIter*, bool) ( dbus_message_helper.cpp:114 )
[ 20 ] otbr::DBus::DBusMessageEncode(DBusMessageIter*, otbr::DBus::ActiveScanResult const&) ( dbus_message_helper_openthread.cpp:96 )
[ 21 ] otbr::DBus::DBusThreadObject::ReplyScanResult(otbr::DBus::DBusRequest&, otError, std::vector<otActiveScanResult, std::allocator<otActiveScanResult> > const&) ( dbus_message_helper.hpp:475 )
[ 22 ] otbr::agent::ThreadHelper::ActiveScanHandler(otActiveScanResult*) ( std_function.h:590 )
[ 23 ] ot::Mac::Mac::PerformActiveScan() ( mac.cpp:298 )
[ 24 ] otSysMainloopProcess ( system.cpp:393 )
[ 25 ] otbr::Ncp::ControllerOpenThread::Process(otSysMainloopContext const&) ( ncp_openthread.cpp:237 )
[ 26 ] otbr::MainloopManager::Process(otSysMainloopContext const&) ( mainloop_manager.cpp:57 )
[ 27 ] otbr::Application::Run() ( application.cpp:166 )
[ 28 ] realmain ( main.cpp:282 )
[ 29 ] main ( main.cpp:321 )
[ 30 ] __libc_start_call_main ( libc_start_call_main.h:58 )
[ 31 ] __libc_start_main_impl ( libc-start.c:389 )
[ 32 ] _start
[ 33 ] 0x76f1189f
```

There are three intertwined problems with the current `dbus` `Scan` API 

**`ActiveScanResult` is uninitialized when forming a reply message**.
When cross-compiling for a Cortex-A7 platform, the boolean values were not being set causing the uninitialized values to be used during message encoding. This resulted in the boolean values being invalid data and causing `dbus` error messages when attempting to send the message.

```arguments to dbus_message_iter_append_basic() were incorrect, assertion "*bool_p == 0 || *bool_p == 1"```

I tested the same thing on my `x86_64` host machine at one point, but it appears that the compiler chose to initialize the values to zero, so there were no issues.

**The introspect contract and the scan reply message do not match**.
The introspect document lists five variables, but the actual return has 12 variables. For example, after fixing the boolean return:
```
      struct {
         uint64 10547911602365048480
         string ""
         uint64 0
         array [
         ]
         uint16 57578
         uint16 0
         byte 11
         byte 199
         byte 206
         byte 0
         boolean false
         boolean false
      }
```

However, the definition in the introspect is:
```
    <!-- Scan: Perform a Thread network scan.
      @scan_result: array of scan results.

      The result struture definition is:
      <literallayout>
        struct {
          uint64 ext_address
          uint16 panid
          uint16 channel
          uint16 rssi
          uint8 lqi
        }
      </literallayout>
    -->
    <method name="Scan">
      <arg name="scan_result" type="a(tqqqy)" direction="out"/>
    </method>
```

I have assumed that the introspect document is the source of truth and removed the excess data from both the encoding step and the data structure itself. The data is not present (XPAN is 0), so it looks like it isn't being populated by anything anyway.

**The introspect contract data types do not match the data being extracted from the data structure**.
For example, the `rssi` is an 8-bit value in the `ActiveScanResult` data structure, but it is a 16-bit value in the `dbus` return definition. This causes `dbus` to error with:

```Array or variant type requires that type uint16 be written, but byte was written```

Given that the older definitions (`channel`, `rssi`) were correct, I chose to fix this by reverting the definitions in the introspect document.

Please let me know if anything needs to be changed!

### DIFF
--- a/src/dbus/server/introspect.xml
+++ b/src/dbus/server/introspect.xml
@@ -9,15 +9,22 @@
       <literallayout>
         struct {
           uint64 ext_address
+          string network_name
+          uint64 ext_panid
+          uint8[] steering_data
           uint16 panid
-          uint16 channel
-          uint16 rssi
+          uint16 joiner_udp_port
+          uint8 channel
+          int16 rssi
           uint8 lqi
+          uint8 version
+          bool is_native
+          bool is_joinable
         }
       </literallayout>
     -->
     <method name="Scan">
-      <arg name="scan_result" type="a(tqqqy)" direction="out"/>
+      <arg name="scan_result" type="a(tstayqqynyybb)" direction="out"/>
     </method>
 
     <!-- Energy Scan: Perform a Thread energy scan.
--- a/src/dbus/common/dbus_message_helper.hpp
+++ b/src/dbus/common/dbus_message_helper.hpp
@@ -212,8 +212,8 @@ template <> struct DBusTypeTrait<ChildInfo>
 template <> struct DBusTypeTrait<ActiveScanResult>
 {
     // struct of { uint64, string, uint64, array<uint8>, uint16, uint16, uint8,
-    //             uint8, uint8, uint8, bool, bool }
-    static constexpr const char *TYPE_AS_STRING = "(tstayqqyyyybb)";
+    //             int16, uint8, uint8, bool, bool }
+    static constexpr const char *TYPE_AS_STRING = "(tstayqqynyybb)";
 };
 
 template <> struct DBusTypeTrait<EnergyScanResult>
--- a/src/dbus/common/dbus_message_helper_openthread.cpp
+++ b/src/dbus/common/dbus_message_helper_openthread.cpp
@@ -54,6 +54,10 @@ otbrError DBusMessageExtract(DBusMessageIter *aIter, ActiveScanResult &aScanResu
     DBusMessageIter sub;
     otbrError       error = OTBR_ERROR_NONE;
 
+    // Local variable to help convert an RSSI value.
+    // Dbus doesn't have the concept of a signed byte
+    int16_t rssi = 0;
+
     VerifyOrExit(dbus_message_iter_get_arg_type(aIter) == DBUS_TYPE_STRUCT, error = OTBR_ERROR_DBUS);
     dbus_message_iter_recurse(aIter, &sub);
 
@@ -64,11 +68,15 @@ otbrError DBusMessageExtract(DBusMessageIter *aIter, ActiveScanResult &aScanResu
     SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mPanId));
     SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mJoinerUdpPort));
     SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mChannel));
-    SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mRssi));
+    SuccessOrExit(error = DBusMessageExtract<int16_t>(&sub, rssi));
     SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mLqi));
     SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mVersion));
     SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mIsNative));
     SuccessOrExit(error = DBusMessageExtract(&sub, aScanResult.mDiscover));
+
+    // Double check the value is within int8 bounds and cast back
+    VerifyOrExit((rssi <= INT8_MAX) && (rssi >= INT8_MIN), error = OTBR_ERROR_PARSE);
+    aScanResult.mRssi = static_cast<int8_t>(rssi);
 
     dbus_message_iter_next(aIter);
     error = OTBR_ERROR_NONE;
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: src/dbus/server/introspect.xml [hunk 1]
```diff
@@ -10,14 +10,14 @@
         struct {
           uint64 ext_address
           uint16 panid
-          uint16 channel
-          uint16 rssi
-          uint8 lqi
+          uint8  channel
+          int8   rssi
```
bukepo, line 14: d-bus doesn't support signed byte. Should we use int16 instead?
```suggestion
          int16  rssi
```

src/dbus/common/types.hpp
```diff
@@ -57,17 +57,10 @@ enum DeviceRole
...
-    std::vector<uint8_t> mSteeringData;  ///< Steering Data
     uint16_t             mPanId;         ///< IEEE 802.15.4 PAN ID
-    uint16_t             mJoinerUdpPort; ///< Joiner UDP Port
     uint8_t              mChannel;       ///< IEEE 802.15.4 Channel
     int8_t               mRssi;          ///< RSSI (dBm)
     uint8_t              mLqi;           ///< LQI
-    uint8_t              mVersion;       ///< Version
-    bool                 mIsNative;      ///< Native Commissioner flag
-    bool                 mDiscover;      ///< Result from MLE Discovery
 };
```
superwhd, line 64: I wonder if we should remove these fields from the struct. It seems they are still present in `otActiveScanResult`. @bukepo do you have any idea whether we should remove them?

bukepo: I agree with you. We should keep these fields until they are not available in OpenThread anymore.

jdswensen: Gotcha, I can add all those values back to the structure. Do you want to keep the introspect as it is or would you like to return the entire structure?

src/dbus/server/introspect.xml
jdswensen: I can see three different paths for changing it:

1. Change the type in `ActiveScanResult`. `DBusMessageEncode` eventually passes the data its trying to use [through a template](https://github.com/openthread/ot-br-posix/blob/0253f26ce53b1db6a4f395c3273fde353fd3901b/src/dbus/common/dbus_message_helper.hpp#L389-L393) to get the type. The easiest fix would be to change the type to force the encoder to encode it as an `int16`. This would make the types inconsistent between [`otActiveScanResult`](https://github.com/openthread/openthread/blob/6a21b5f9ae29dd165ea2a359943264910048957e/include/openthread/link.h#L376-L398) and `ActiveScanResult` though.
2.  Define a specific trait template for the `ActiveScanResult` struct that would end up encoding the value as a `int16`. It would keep the data structures consistent, but introduce complexity. I could also see it causing confusion.
3. Instead of an entire struct, define a specific trait for this one value (newtype pattern). I think it has the same problems as 2 though.

If the type is left as `int8` in the return, it could also be documented that it should be cast as a signed value on the client end.

Let me know what you think. I'm indifferent on the outcome as I can make the platform I'm supporting work in whichever way is deemed appropriate. I just need it to not crash. 😄

src/dbus/common/types.hpp
superwhd: I'd prefer update the introspect to return the entire structure.

src/dbus/server/introspect.xml
jdswensen: Fixed!

src/dbus/common/types.hpp
jdswensen: Fixed!

jdswensen:
Finally got around to addressing the feedback (sorry for the delay). Please let me know if either of you would like additional changes. 😄

superwhd (approved):
LGTM 👍

superwhd (approved):
LGTM 👍

superwhd (approved):
LGTM 👍

jdswensen:
@bukepo, would you have time to give this another look? 😃

bukepo (approved):

jwhui (approved):
Thanks! 👍🏼

jwhui:
@jdswensen , if you would like to keep the commits separate after merge, can you add the PR number to the commit header? For example:
```
[dbus] init ActiveScanResult to zero (#1814)
```
Alternatively, I can squash-merge and GitHub will automatically add the PR number to the commit header.

Let me know what you prefer. Thanks!

jdswensen:
@jwhui, commits have been updated and PR has been rebased onto the latest.

PR #1816, opened 2023-04-05
update channel mask in ThreadHelper attach

AlanLCollins:
when using ThreadHelper::Attach, the active dataset is not updated with the network credentials. It remains with default channel mask

### DIFF
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -391,6 +391,7 @@ void ThreadHelper::Attach(const std::string          &aNetworkName,
 
     channel = RandomChannelFromChannelMask(channelMask);
     SuccessOrExit(otLinkSetChannel(mInstance, channel));
+    SuccessOrExit(otLinkSetSupportedChannelMask(mInstance,channelMask));
 
     SuccessOrExit(error = otThreadSetPskc(mInstance, &pskc));
 


src/utils/thread_helper.cpp
```diff
@@ -391,6 +391,7 @@ void ThreadHelper::Attach(const std::string          &aNetworkName,
 
     channel = RandomChannelFromChannelMask(channelMask);
     SuccessOrExit(otLinkSetChannel(mInstance, channel));
+    SuccessOrExit(otLinkSetSupportedChannelMask(mInstance,channelMask));
```
bukepo, line 394: The supported channel mask is about the channels the radio supports according to regulatory compliance. I don't think it needs to be updated here.

AlanLCollins: Yes, that was my initial analysis, I was hesitant to use this approach.  
What is your recommendation to update the channel mask in the active dataset that will be created after using this helper function?

jwhui: @AlanLCollins , are you trying to set the preferred channel mask?
https://github.com/openthread/ot-br-posix/blob/ec6123d7e80efe2f8af8168fb885f9822cd263fa/src/utils/thread_helper.cpp#L384
Or are you really looking to configure the supported channel mask?

Also, what is your use case for using this `Attach` method? I would think it's better to set the Active Operational Dataset directly.

AlanLCollins: My intention is to make sure all the thread credentials used in this attach method, get reflected into the active dataset after the device has successfully started thread. So If I query the active dataset TLV, the tag 0x35 would return the channel mask used to create/join the network. The preferred channel mask does not have that impact.

The use case for using this attach method is to have a sync_function that performs attachment without multiple async calls to dbus to commit new dataset + thread start.

wgtdkp: I believe the `Attach()` method is deprecated by [`AttachAllNodesTo()`](https://github.com/openthread/ot-br-posix/blob/main/src/utils/thread_helper.cpp#L540).

AlanLCollins:
Thanks @wgtdkp , I will use the method you shared.

src/utils/thread_helper.cpp
spalas1996: @wgtdkp @bukepo @wgtdkp Can you confirm if Attach() method is indded getting deprecated by AttachAllNodesTo()?
From the description I see the purpose of the 2 methods is very different
`/**
     * This method attaches the device to the Thread network.
     *
     * @note The joiner start and the attach proccesses are exclusive
     *
     * @param[in] aNetworkName  The network name.
     * @param[in] aPanId        The pan id, UINT16_MAX for random.
     * @param[in] aExtPanId     The extended pan id, UINT64_MAX for random.
     * @param[in] aNetworkKey   The network key, empty for random.
     * @param[in] aPSKc         The pre-shared commissioner key, empty for random.
     * @param[in] aChannelMask  A bitmask for valid channels, will random select one.
     * @param[in] aHandler      The attach result handler.
     *
     */
    void Attach(const std::string &         aNetworkName,
                uint16_t                    aPanId,
                uint64_t                    aExtPanId,
                const std::vector<uint8_t> &aNetworkKey,
                const std::vector<uint8_t> &aPSKc,
                uint32_t                    aChannelMask,
                AttachHandler               aHandler);`
`    /**
     * This method makes all nodes in the current network attach to the network specified by the dataset TLVs.
     *
     * @param[in] aDatasetTlvs  The dataset TLVs.
     * @param[in] aHandler      The result handler.
     *
     */
    void AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, AttachHandler aHandler);`

wgtdkp: `Attach()` is a dangerous operation because it forces reconnecting this BR device to the new Thread network (or create the new network) and orphan any end devices which were connected to this BR. I would never suggest using this method is any serious production environment because this would never be what the user wants.

That said, `Attach()` is still useful for testing or demo, it the behavior is what you need.

PR #1862, opened 2023-05-11
[rest] allow to specify REST listen port

agners:
Allow to bind to a specific port to listen to for REST requests. Note that the built-in web interface always assumes the REST to be available on port 8081!

### DIFF
--- a/src/agent/main.cpp
+++ b/src/agent/main.cpp
@@ -61,6 +61,9 @@
 static const char kSyslogIdent[]          = "otbr-agent";
 static const char kDefaultInterfaceName[] = "wpan0";
 
+// Port number used by Rest server.
+static const uint32_t kPortNumber = 8081;
+
 enum
 {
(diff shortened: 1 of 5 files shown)


src/agent/main.cpp
```diff
@@ -192,6 +194,7 @@ static int realmain(int argc, char *argv[])
     bool                      printRadioVersion = false;
     bool                      enableAutoAttach  = true;
     const char               *restListenAddress = "";
+    int                       restListenPort    = 8081;
```
wgtdkp, line 197: Please move `kPortNumber` and use it here.

wgtdkp (approved):
LGTM :+1:

wgtdkp:
> Allow to bind to a specific port to listen to for REST requests. Note that the built-in web interface always assumes the REST to be available on port 8081!

The JS code should access the REST backend directly at all... The interface is more like a prototype

jwhui (approved):

PR #1864, opened 2023-05-11
[rest] return border agent ID

agners:
Add BaId to the node JSON object and add a dedicated endpoint
/node/ba-id to retrive the border agent ID.

Also return node information even when the border router is
detached.


wgtdkp (approved):
LGTM 👍 
Thanks!

jwhui (approved):

PR #1866, opened 2023-05-11
[rest] support Thread state change

agners:
Allow to enable and disable the Thread protocol operation.

### DIFF
--- a/src/common/api_strings.cpp
+++ b/src/common/api_strings.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright (c) 2021, The OpenThread Authors.
+ *    Copyright (c) 2023, The OpenThread Authors.
  *    All rights reserved.
  *
  *    Redistribution and use in source and binary forms, with or without
--- a/src/common/api_strings.cpp
+++ b/src/common/api_strings.cpp
@@ -1,0 +1,55 @@
+/*
+ *    Copyright (c) 2021, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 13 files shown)


src/rest/openapi.yaml
```diff
@@ -107,6 +107,24 @@ paths:
...
+        hasn't been started yet, it will get started automatically.
+      responses:
+        "200":
+          description: Successful operation.
+      requestBody:
+        description: New Thread state
+        content:
+          application/json:
+            schema:
+              type: string
```
wgtdkp, line 125: The GET method returns an integer, but the PUT here gives a string. What about changing the GET format to string as well (I prefer string against integer to make the REST API more readable)

src/rest/resource.cpp
```diff
@@ -312,17 +312,66 @@ void Resource::GetDataState(Response &aResponse) const
...
+        ErrorHandler(aResponse, HttpStatusCode::kStatusConflict);
+    }
+    if (error == OTBR_ERROR_INVALID_ARGS)
+    {
+        ErrorHandler(aResponse, HttpStatusCode::kStatusBadRequest);
+    }
+    else if (error != OTBR_ERROR_NONE)
+    {
+        ErrorHandler(aResponse, HttpStatusCode::kStatusInternalServerError);
+    }
```
wgtdkp, line 352: What about writing a short function to convert otbrError/otError to the response code? For example:
```
int ErrorToResponseCode(otbrError aError) {
    // Converts an otbr error code to a HTTP response code
    // ...
}
```

Then we can have simple error handling code here and everywhere else:
```
exit:
    ErrorHandler(aResponse, ErrorToResponseCode(aError));
```

This will help create consistent response code for every REST APIs

agners: There is another downside of this: It makes it harder to reason what codes exactly a particular resource can return (which is quite important, IMHO). That said, it does simplify code quite ab it so :man_shrugging: 

Thoughts? Maybe better in a follow up PR?

src/rest/openapi.yaml
wgtdkp: The REST API is currently version `0.3.0`. The zero major version means the API is not finalized and we are not responsible for compatibility (see https://semver.org/#spec-item-4). Though you should update the version to `0.4.0` regardless of the major version.

In a later time if you feel confident on the REST API, feel free to propose updating the version to `1.0.0`!

src/rest/resource.cpp
wgtdkp: Fine to leave it as is

wgtdkp (approved):
LGTM :+1:

src/common/api_strings.cpp
```diff
@@ -0,0 +1,55 @@
+/*
+ *    Copyright (c) 2021, The OpenThread Authors.
```
jwhui, line 2:
```suggestion
 *    Copyright (c) 2023, The OpenThread Authors.
```

src/common/api_strings.hpp
```diff
@@ -0,0 +1,50 @@
+/*
+ *    Copyright (c) 2021, The OpenThread Authors.
```
jwhui, line 2:
```suggestion
 *    Copyright (c) 2023, The OpenThread Authors.
```

src/rest/json.hpp
```diff
@@ -103,6 +103,16 @@ std::string CString2JsonString(const char *aCString);
  */
 std::string String2JsonString(const std::string &aString);
 
+/**
+ * This method parses a Json string and checks its datatype and returns a string if it is a string.
+ *
+ * @param[in]  aJsonString  A Json string.
+ * @param[out] aString      The string.
+ *
+ * @returns A boolean if the Json string indeed was a string.
```
jwhui, line 112:
```suggestion
 * @returns A boolean indicating whether the Json string was indeed a string.
```

src/rest/resource.cpp
```diff
@@ -339,33 +337,77 @@ void Resource::ExtendedAddr(const Request &aRequest, Response &aResponse) const
...
     std::string errorCode;
+    std::string body;
 
-    if (aRequest.GetMethod() == HttpMethod::kGet)
+    VerifyOrExit(Json::JsonString2String(aRequest.GetBody(), body), error = OTBR_ERROR_INVALID_ARGS);
+    if (body == "enable")
     {
-        GetDataState(aResponse);
+        if (!otIp6IsEnabled(mInstance))
+            SuccessOrExit(otIp6SetEnabled(mInstance, true));
```
jwhui, line 361: nit: prefer parens
```suggestion
        if (!otIp6IsEnabled(mInstance))
        {
            SuccessOrExit(otIp6SetEnabled(mInstance, true));
        }
```

jwhui (approved):
Thanks! 👍🏼

PR #1884, opened 2023-05-29
[docker] support updating nat64 CIDR during runtime

erjiaqing:
Current firewall rules uses a static CIDR for NAT64. This will block all traffic if we change the NAT64 CIDR。

This PR uses fwmark to identify the traffic from wpan0 interface for NAT44.

This is mainly for testing purpose.

Test: https://github.com/openthread/openthread/actions/runs/5114000045/jobs/9193809926


xuyirio (approved):
Could you add a bit more description to this PR?

jwhui (approved):

PR #1892, opened 2023-06-05
[rest] initialize dataset

agners:
When there is already a dataset, and setting properties using the PUT method and an incomplete JSON dataset, some fields of the dataset variable might not be initialized. This leads to uninitialized data being used as active dataset.

Make sure we initialize the dataset variable. This makes sure that all components of the dataset are considered not set.


wgtdkp (approved):
LGTM :+1:

jwhui (approved):
Thanks! 👍🏼

PR #1893, opened 2023-06-05
[rest] support deleting the dataset

agners:
Add REST API to support deleting the active or pending operational dataset. Deleting the active operational dataset requires the Thread network to be disabled (just like modifying the active operational dataset). Subsequent use of the PUT method allows to build entirly new datasets with values generated by the stack (through otDatasetCreateNewNetwork).

### DIFF
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -239,6 +239,18 @@ paths:
           description: Invalid request body.
         "409":
           description: Writing active operational dataset rejected because Thread network is active.
+    delete:
+      tags:
+        - node
+      summary: Deletes the active operational dataset
+      description: |-
+        Deletes the the active operational dataset on the current node. Only allowed if the Thread node
+        is inactive.
+      responses:
+        "200":
+          description: Successfully deleted the active operational dataset.
+        "409":
+          description: Deleting active operational dataset rejected because Thread network is active.
   /node/dataset/pending:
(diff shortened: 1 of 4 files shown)


agners:
The macOS issue is kinda weird, I rebased  on an earlier commit which built fine according to the main branch. It seems some external resource changed which broke the `macOS / build-check` here?

agners:
> There are also other persistent data that you probably want to remove as well when joining a new network. OpenThread provides `otInstanceErasePersistentInfo()` to erase all those data. What about we provide an equivalent REST API?

Hm, `otInstanceErasePersistentInfo()` probably would need another REST resource path. Maybe a `DELETE` on `/node`?

What type of data is that erasing beside the dataset? I was assuming that all network related data is being reset by clearing the dataset.

jwhui:
> What type of data is that erasing beside the dataset? I was assuming that all network related data is being reset by clearing the dataset.

All of the non-volatile settings are defined here: https://github.com/openthread/openthread/blob/main/src/core/common/settings.hpp

Thread Specification requires certain identifying information to be reset on a factory reset.

If we are truly want to factory reset the Thread configuration, then we should erase all non-volatile settings.

I wonder if we should provide a way to delete all non-volatile settings rather than just the Operational Dataset itself.

agners:
Replaced by #1908.

PR #1904, opened 2023-06-14
[rest] deactive IPv6 interface when disabling the OTBR

agners:
Deactivate the IPv6 interface when disabling the OTBR. This follows the Attach/Deatach semantics of the D-Bus API.


bukepo (approved):

jwhui (approved):
Thanks! 👍🏼

PR #1908, opened 2023-06-15
[rest] support erasing all persistent info

agners:
Add REST API to support erasing all persistent information, effectively factory resetting the OTBR. The implementation follows the semantic of the D-Bus API and automatically disables the Thread network.

After erasing all persistent information the dataset is cleared too. So this allows to build a new Thread network with subsequent use of the PUT method to the /node/dataset/active endpoint.

### DIFF
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -43,6 +43,15 @@ paths:
             application/json:
               schema:
                 type: object
+    delete:
+      tags:
+        - node
+      summary: Erase all persistent information, essentially factory reset the Border Router.
+      responses:
+        "200":
+          description: Successful operation
+        "409":
+          description: Thread interface is in wrong state.
   /node/ba-id:
     get:
       tags:
--- a/src/rest/resource.cpp
+++ b/src/rest/resource.cpp
@@ -257,16 +257,46 @@ exit:
     }
 }
 
+void Resource::DeleteNodeInfo(Response &aResponse) const
+{
+    otbrError   error = OTBR_ERROR_NONE;
+    std::string errorCode;
+
+    VerifyOrExit(mNcp->GetThreadHelper()->Detach() == OT_ERROR_NONE, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(otInstanceErasePersistentInfo(mInstance) == OT_ERROR_NONE, error = OTBR_ERROR_REST);
+    mNcp->Reset();
+
+exit:
+    if (error == OTBR_ERROR_NONE)
+    {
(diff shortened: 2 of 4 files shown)


agners:
This replaces #1893 and implements what has been discussed in that PR.

agners:
I think the test failure "Border Router / Border Router MATN (mDNSResponder)" is not caused by this PR, could someone trigger another test run of that test?

wgtdkp:
> I think the test failure "Border Router / Border Router MATN (mDNSResponder)" is not caused by this PR, could someone trigger another test run of that test?

There are conflicts in third_party/openthread/repo

src/rest/openapi.yaml
```diff
@@ -43,6 +43,15 @@ paths:
             application/json:
               schema:
                 type: object
+    delete:
+      tags:
+        - node
```
wgtdkp, line 48: It doesn't look accurate that "delete node" means removing the persistent info. I think the first impression may be "delete this device from the Thread network".

What about using  `node/persistent-info`?

agners: Hm, in my mind persistent information is like an implementation detail. I don't see us having a `GET` or `PUT` method there.

In a way, logically, we reset/kill the whole (Thread) node, so to me `/node` sound not really wrong... Thoughts?

agners:
> There are conflicts in third_party/openthread/repo

The issue is resolved now.

src/rest/openapi.yaml
wgtdkp: > In a way, logically, we reset/kill the whole (Thread) node, so to me /node sound not really wrong... Thoughts?

My concern is that the method name is "delete", but not "reset", so it could be misleading.

If you are looking for a more generic name, what about "node/network-settings"?

agners: > My concern is that the method name is "delete", but not "reset", so it could be misleading.

Granted, `DELETE` is not an ideal method name for this operation, but it is the best fit from what HTTP offers.

> If you are looking for a more generic name, what about "node/network-settings"?

I think I'd rather prefer `node/persistent-info` then.

The thing which bothers me using an endpoint below `/node` is that we delete/reset all the settings of a node, e.g. the `GET` of `/node` or `/node/ext-address` will be different. This, to me, makes the `DELETE` to the parent endpoint `/node` a very good fit.

In the end, its just an endpoint, it doesn't matter too much. I'd still prefer `/node`, but I can live with either endpoint :smile: 

@jwhui do you have thoughts/an opinion on what to choose here?

wgtdkp (approved):
LGTM :+1:

PR #1910, opened 2023-06-16
[mdns] pass copy of `mType` for `OnServiceResolved`

sunytt:
https://github.com/openthread/ot-br-posix/blob/main/src/mdns/mdns_avahi.cpp#L1261
`PublisherAvahi::ServiceResolver::HandleResolveHostResult` passes the reference of `ServiceResolver::mType` to `OnServiceResolved()` function which calls each registered service discovery callback on the resolved service. The ServiceResolver may be removed by a callback, when there are more callbacks called afterwards, there will be a crash as the `mType` is gone.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -175,7 +175,7 @@ uint64_t Publisher::AddSubscriptionCallbacks(Publisher::DiscoveredServiceInstanc
     return subscriberId;
 }
 
-void Publisher::OnServiceResolved(const std::string aType, const DiscoveredInstanceInfo &aInstanceInfo)
+void Publisher::OnServiceResolved(std::string aType, const DiscoveredInstanceInfo &aInstanceInfo)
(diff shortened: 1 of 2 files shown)


src/mdns/mdns.cpp
```diff
@@ -175,7 +175,7 @@ uint64_t Publisher::AddSubscriptionCallbacks(Publisher::DiscoveredServiceInstanc
     return subscriberId;
 }
 
-void Publisher::OnServiceResolved(const std::string &aType, const DiscoveredInstanceInfo &aInstanceInfo)
+void Publisher::OnServiceResolved(const std::string aType, const DiscoveredInstanceInfo &aInstanceInfo)
```
jwhui, line 178:
```suggestion
void Publisher::OnServiceResolved(std::string aType, const DiscoveredInstanceInfo &aInstanceInfo)
```

PR #1911, opened 2023-06-16
[mdns] all `On{Item}Resolved()` callbacks to use value parameters

abtink:
This commit updates all base class `Publisher::On{Item}Resolved()` callbacks to use value parameters instead of `const` reference. This change ensures that the objects passed to these callbacks will not be freed until the callbacks have completed and prevent any potential problem. In particular user may unsubscribe (invoke `UnsubscribeService()` or `UnsubscribeHost()`) from the callback.

----

This is follow up https://github.com/openthread/ot-br-posix/pull/1910.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -81,14 +81,14 @@ void Publisher::PublishHost(const std::string             &aName,
     }
 }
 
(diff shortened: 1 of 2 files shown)


sunytt (approved):
LGTM 👍

wgtdkp (approved):
LGTM 👍

jwhui (approved):

PR #1919, opened 2023-06-21
[mdns] update mDNSResponder to 1790.80.10

agners:


agners:
It seems that some tests fail despite not using mDNSResponder. The reason is that the tests use mDNSResponder (`mdnsd` and `dns-sd` CLI utility) to find services. It seems that this new version causes problems at times, e.g. it seems to discover the OTBR twice (e.g. from the [Border Router TREL (Avahi)](https://github.com/openthread/ot-br-posix/actions/runs/5333902284/jobs/9667010528?pr=1919#logs) test):

```
FAIL: test (__main__.PublishMeshCopService)
  File "/home/runner/work/ot-br-posix/ot-br-posix/third_party/openthread/repo/tests/scripts/thread-cert/border_router/test_publish_meshcop_service.py", line 117, in test
    self.assertEqual(len(service_instances), 2)
AssertionError: 4 != 2
```

It could be just timing, or maybe it is really a bug in that new mDNSResponder version.

I am currently not setup to run these tests/didn't investigate on how to run them. I probably can look into these issues at one point, but it doesn't seem to be a quick/easy thing from what I can tell.

If someone else has a good idea I am happy to hear it.

agners:
With the patches from https://github.com/apple-oss-distributions/mDNSResponder/pull/2 and https://github.com/apple-oss-distributions/mDNSResponder/pull/6 the tests seem to run through nicely (the first time there were some errors, but they were unrelated. With this second run it seems everything is green now.

I'll update the branch again to trigger another run, if that returns green again I think it shows some certainty that this the update is solid.

agners:
Any thoughts on this PR?  It does address issue we have seen where `mdnsd` suddenly used 100% CPU. We haven't observed that issue with this version so far.

superwhd (approved):
LGTM, Thanks!

wgtdkp (approved):
LGTM :+1:

jwhui (approved):
Thanks! 👍🏼

PR #1929, opened 2023-07-03
[utils] update SuccessOrExit for error tracking

wgtdkp:
This commit updates the `SuccessOrExit` macro to accept an action before goto "exit". It allows us to add a log when error happens. For example, we can write:

```
SuccessOrExit((error = writeDatasetToSettings()), otbrLogErr("Failed to write dataset to settings"));
```

So that we know exactly which step a method fails at.

------

This commit also prints the error/status code for macro `SuccessOrDie`.

### DIFF
--- a/src/common/code_utils.hpp
+++ b/src/common/code_utils.hpp
@@ -72,13 +72,14 @@
  *  @param[in] aStatus  A scalar status to be evaluated against zero (0).
  *
  */
-#define SuccessOrExit(aStatus) \
-    do                         \
-    {                          \
-        if ((aStatus) != 0)    \
-        {                      \
-            goto exit;         \
-        }                      \
+#define SuccessOrExit(aStatus, ...) \
+    do                              \
+    {                               \
+        if ((aStatus) != 0)         \
+        {                           \
+            __VA_ARGS__;            \
+            goto exit;              \
+        }                           \
     } while (false)
 
 /**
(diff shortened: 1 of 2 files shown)


src/common/code_utils.hpp
```diff
@@ -89,14 +90,14 @@
...
-        {                                                                    \
-            otbrLogEmerg("FAILED %s:%d - %s", __FILE__, __LINE__, aMessage); \
-            exit(-1);                                                        \
-        }                                                                    \
+#define SuccessOrDie(aStatus, aMessage)                                                   \
+    do                                                                                    \
+    {                                                                                     \
+        if ((aStatus) != 0)                                                               \
+        {                                                                                 \
+            otbrLogEmerg("FAILED %s:%d - %s: %d", __FILE__, __LINE__, aMessage, aStatus); \
```
zhanglongxia, line 98: How about putting the status value in front of the message string? The length of the message string is uncertain. Putting the status value at the beginning makes the developer easier to find the status value.

wgtdkp: SGTM, updated!

zhanglongxia (approved):
LGTM

bukepo (approved):

tests/scripts/openwrt
```diff
@@ -43,7 +43,7 @@ readonly CONTAINER_NAME
 
 do_prepare()
 {
-    docker create --name "${CONTAINER_NAME}" --rm -v openwrt-bin:/home/build/openwrt/bin -it openwrt/sdk
+    docker create --name "${CONTAINER_NAME}" --rm -v openwrt-bin:/home/build/openwrt/bin -it openwrt/sdk:22.03.3
```
bukepo, line 46: Is this change intentional?

wgtdkp: Yes, current openwrt CI is broken because the openwrt/sdk:latest has incompatible changes that otbr can't build it the docker image (e.g. no `sudo`, no `/home/build/openwrt` directory). It's probably more stable to stick with the latest "good" build to unblock the CI on the main branch (for example, https://github.com/openthread/ot-br-posix/pull/1930).

I will ask the commits being merged separately.

wgtdkp:
@jwhui When merging this PR, please merge the two commits separately. The second one is for fixing the broken openwrt CI tests

jwhui (approved):

PR #1940, opened 2023-07-14
[rest] add query parsing

CodingRays:
This adds the capability to parse url query strings to the rest api.

It also contains a small refactoring of the parser and request class which moved state that is only required for parsing out of the request class. The only exception to this is the body field which is still built up in the request class.

### DIFF
--- a/src/rest/parser.hpp
+++ b/src/rest/parser.hpp
@@ -80,8 +80,25 @@ public:
     void Process(const char *aBuf, size_t aLength);
 
 private:
+    class State {
+    public:
+        Request    *mRequest;
+        std::string mUrl;
+        std::string mNextHeaderField;
+    };
+
+    static int OnUrl(http_parser *parser, const char *at, size_t len);
+    static int OnBody(http_parser *parser, const char *at, size_t len);
+    static int OnMessageComplete(http_parser *parser);
+    static int OnMessageBegin(http_parser *parser);
+    static int OnHeaderComplete(http_parser *parser);
+    static int OnHandlerData(http_parser *, const char *, size_t);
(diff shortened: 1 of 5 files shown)


src/rest/request.cpp
```diff
@@ -105,6 +89,13 @@ std::string Request::GetHeaderValue(const std::string aHeaderField) const
     return (it == mHeaders.end()) ? "" : it->second;
 }
 
+std::string Request::GetQueryValue(const std::string aQueryName) const
+{
+    auto it = mHeaders.find(aQueryName);
+
+    return (it == mHeaders.end()) ? "" : it->second;
```
wgtdkp, line 96: Shouldn't we use `mQueryParameters` here?

wgtdkp:
It looks like there are formatting error with your PR: https://github.com/openthread/ot-br-posix/actions/runs/5553833818/jobs/10171787035?pr=1940

Please run command `./script/make-pretty clang-format` to fix it locally.

src/rest/parser.hpp
```diff
@@ -80,8 +80,25 @@ class Parser
     void Process(const char *aBuf, size_t aLength);
 
 private:
+    class State {
+    public:
```
wgtdkp, line 84: nit:
```suggestion
    struct State {
```

Let's juts use `struct` for data classes

src/rest/request.cpp
CodingRays: Yes we definitely should.
I've tested the parsing code but not the api. Looking at the automated tests its gonna be hard to test without having a endpoint that actually uses the query parameters. We have code under works that uses them but only for timeouts so that would not be testable either.

wgtdkp (approved):

PR #1947, opened 2023-07-20
[dbus] add `PendingDatasetTlvs`

jdswensen:
Add dbus call to fetch the pending dataset.


src/dbus/server/introspect.xml
```diff
@@ -536,6 +536,11 @@
       <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="false"/>
     </property>
 
+    <!-- PendingDatasetTlvs: The Thread active dataset tlv in binary form. -->
+    <property name="PendingDatasetTlvs" type="ay" access="readwrite">
+      <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="false"/>
+    </property>
+
```
jdswensen, line 543: Just noticed this, I'll change after CI is done in case there is anything I need to address.
```suggestion
    <!-- PendingDatasetTlvs: The Thread pending dataset tlv in binary form. -->
    <property name="PendingDatasetTlvs" type="ay" access="read">
      <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="false"/>
    </property>

```

jdswensen:
@bukepo would you have time to review this or recommend someone that may have time? 😄

jwhui (approved):
Thanks! 👍🏼

PR #1953, opened 2023-07-26
[logging] use program name for log id

wgtdkp:
On Android, we use the "ot-daemon" name for the runnable program
and want the syslog id be updated to "ot-daemon" as well. Instead of
hard coding the syslog id, this commit derive the id from the program
name so it works for both linux and Android.

For example, logs on Android will be like:
```
ot-daemon[2152867]: [NOTE]-AGENT---: Running 0.3.0-090e5b55-dirty
ot-daemon[2152867]: [NOTE]-AGENT---: Thread version: 1.3.0
ot-daemon[2152867]: [NOTE]-AGENT---: Thread interface: wpan0
```

### DIFF
--- a/src/common/logging.cpp
+++ b/src/common/logging.cpp
@@ -78,12 +78,17 @@ void otbrLogSetLevel(otbrLogLevel aLevel)
 }
 
 /** Initialize logging */
-void otbrLogInit(const char *aIdent, otbrLogLevel aLevel, bool aPrintStderr)
+void otbrLogInit(const char *aProgramName, otbrLogLevel aLevel, bool aPrintStderr)
 {
-    assert(aIdent);
+    const char *ident;
+
+    assert(aProgramName != nullptr);
     assert(aLevel >= OTBR_LOG_EMERG && aLevel <= OTBR_LOG_DEBUG);
 
-    openlog(aIdent, (LOG_CONS | LOG_PID) | (aPrintStderr ? LOG_PERROR : 0), OTBR_SYSLOG_FACILITY_ID);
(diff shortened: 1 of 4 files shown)


zhanglongxia (approved):
LGTM

superwhd (approved):
LGTM 👍

bukepo (approved):

src/common/logging.cpp
```diff
@@ -78,12 +78,17 @@ void otbrLogSetLevel(otbrLogLevel aLevel)
...
 {
-    assert(aIdent);
+    const char *ident;
+
+    assert(aProgramName != nullptr);
     assert(aLevel >= OTBR_LOG_EMERG && aLevel <= OTBR_LOG_DEBUG);
 
-    openlog(aIdent, (LOG_CONS | LOG_PID) | (aPrintStderr ? LOG_PERROR : 0), OTBR_SYSLOG_FACILITY_ID);
+    ident = strrchr(aProgramName, '/');
+    ident = (ident != nullptr) ? ident + 1 : aProgramName;
```
abtink, line 89: Is it possible to have more than one `/` in the `aProgramName`? 
We can consider skipping over them all until we see no more.

abtink (approved):
LGTM.

src/common/logging.cpp
wgtdkp: Hi abtink, it's exactly what [strrchr](https://cplusplus.com/reference/cstring/strrchr/) does!

jwhui (approved):

PR #1962, opened 2023-08-04
Dockerfile: EXPOSE REST API port 8080

ajlennon:
As this is needed by the WebUI topology page which interacts with the OTBR REST agent on port 8080


scyto:
I think you will find the rest port is 8081 not 8080 (thats how it is by default when I use macvlan)
i still like the idea of you including it as an EXPOSE

ajlennon:
Thanks :) I've been fiddling around as I've been trying to get Traefik to forward requests so I can have everything running over a single port...

jwhui:
Closing stale PR.

PR #1992, opened 2023-09-04
[mdns] update 'TxtEntry'  to handle boolean attribute

abtink:
This commit updates `TxtEntry`, `EncodeTxtData()`, and `DecodeTxtData()` to support boolean attributes, which are encoded as `key` string without the `=` character. It also ensures that `EncodeTxtData()` will generate valid TXT data (containing a single zero byte) when given an empty `TxtList`. Additionally, this commit updates `AdvertisingProxy::MakeTxtList()` to use `DecodeTxtData()`, and adds a test to validate the encoder and decoder behavior.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -99,18 +99,32 @@ otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, std::vector<uint8_t>
 {
     otbrError error = OTBR_ERROR_NONE;
 
-    for (const auto &txtEntry : aTxtList)
+    aTxtData.clear();
+
+    for (const TxtEntry &txtEntry : aTxtList)
     {
-        const auto  &name        = txtEntry.mName;
-        const auto  &value       = txtEntry.mValue;
-        const size_t entryLength = name.length() + 1 + value.size();
+        size_t entryLength = txtEntry.mKey.length();
+
+        if (!txtEntry.mIsBooleanAttribute)
+        {
+            entryLength += txtEntry.mValue.size() + sizeof(uint8_t); // for `=` char.
+        }
 
         VerifyOrExit(entryLength <= kMaxTextEntrySize, error = OTBR_ERROR_INVALID_ARGS);
 
         aTxtData.push_back(static_cast<uint8_t>(entryLength));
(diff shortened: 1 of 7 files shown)


wgtdkp (approved):
Thanks @abtink! Looks great!

src/mdns/mdns.cpp
```diff
@@ -99,18 +99,32 @@ otbrError Publisher::EncodeTxtData(const TxtList &aTxtList, std::vector<uint8_t>
...
+        aTxtData.insert(aTxtData.end(), txtEntry.mKey.begin(), txtEntry.mKey.end());
+
+        if (!txtEntry.mIsBooleanAttribute)
+        {
+            aTxtData.push_back('=');
+            aTxtData.insert(aTxtData.end(), txtEntry.mValue.begin(), txtEntry.mValue.end());
+        }
+    }
+
+    if (aTxtData.size() == 0)
```
superwhd, line 125: nit
```suggestion
    if (aTxtData.empty())
```

tests/mdns/main.cpp
```diff
@@ -410,10 +410,65 @@ otbrError TestStopService(void)
     return ret;
 }
 
+otbrError ChekcTxtDataEncoderDecoder(void)
```
superwhd, line 413:
```suggestion
otbrError CheckTxtDataEncoderDecoder(void)
```

tests/mdns/main.cpp
```diff
@@ -410,10 +410,65 @@ otbrError TestStopService(void)
...
+
+exit:
+    return error;
+}
+
 int main(int argc, char *argv[])
 {
     int ret = 0;
 
+    if (ChekcTxtDataEncoderDecoder() != OTBR_ERROR_NONE)
```
superwhd, line 467:
```suggestion
    if (CheckTxtDataEncoderDecoder() != OTBR_ERROR_NONE)
```

superwhd (approved):
LGTM 👍

src/mdns/mdns.cpp
abtink: Thanks! Updated in new push.

jwhui (approved):

PR #1996, opened 2023-09-06
[mdns] change `PublishService()` to get encoded TXT data as input

abtink:
This commit updates `Mdns::Publisher::PublishService()` to take encoded TXT data as its input, harmonizing the `Publish` API with the `Subscribe` API, which also uses encoded TXT data. This makes it easier to use with advertising proxy and TREL, avoiding the extra step of decoding already encoded TXT data to call `PublishService()`.

---

- ~This PR currently contains the commit from PR https://github.com/openthread/ot-br-posix/pull/1992.~ 
- ~Please check/review the second commit only.~


superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2006, opened 2023-09-08
[tests] fix service type names used in `mdns/main` tests

abtink:
This commit updates the test to use "_meshcop._udp" as the service type (removing ending `.` char) as the service type should not end with dot character.

It also updates the documentation of `Mdns::Publisher` methods to indicate that the service type strings must not end with dot character.

### DIFF
--- a/src/mdns/mdns.hpp
+++ b/src/mdns/mdns.hpp
@@ -211,7 +211,7 @@ public:
      *                          with method PublishHost.
      * @param[in] aName         The name of this service. If an empty string is provided, the service's name will be the
      *                          same as the platform's hostname.
-     * @param[in] aType         The type of this service.
+     * @param[in] aType         The type of this service, e.g., "_srv._udp" (MUST NOT end with dot).
      * @param[in] aSubTypeList  A list of service subtypes.
      * @param[in] aPort         The port number of this service.
      * @param[in] aTxtData      The encoded TXT data for this service.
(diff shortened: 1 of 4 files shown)


superwhd (commented):
I wonder the thoughts behind this change. Is it required by RFC or it simplifies the implementation?

abtink:
> I wonder the thoughts behind this change. Is it required by RFC or it simplifies the implementation?

This is not a change of the API but a clarification of existing assumption and fix of the example test code in `mdns/main.cpp`. 

The existing implementation of `Publisher` indirectly assumed no "dot" in the service type string. This is clear from how the full name is constructed in https://github.com/openthread/ot-br-posix/blob/4e75260bad63a5c9a80cfff8d599f1e817c036e8/src/mdns/mdns.cpp#L287-L290

If the caller include `.` in the `aType` the generated full name would be incorrect and will have an extra `.`. 

Also this is how the `Publisher` APIs are used every other place other than the test `mdns/main` example (i.e. dot is not include after `aHostName` or `aServiceType`). For example 
https://github.com/openthread/ot-br-posix/blob/4e75260bad63a5c9a80cfff8d599f1e817c036e8/src/border_agent/border_agent.cpp#L74
https://github.com/openthread/ot-br-posix/blob/4e75260bad63a5c9a80cfff8d599f1e817c036e8/src/trel_dnssd/trel_dnssd.cpp#L51
or how `SplitFullServiceName()` and `SplitFullDnsName()` work.

Regarding how service type is represnted in RFCs,  the examples of service type generally do not include the `.`.

superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2008, opened 2023-09-10
[mdns-mdnssd] simplify service and host registrations

abtink:
This commit updates and enhances the `Mdns::PublisherMDnsSd` class, which implements the `Mdns::Publisher` interface using MDNSResponder as its underlying DNS-SD provider.

The following changes are made related to service registration:

- Registration simplified: The `PublishServiceImpl()` method is updated to allocate a `DnssdServiceRegistration` object first and add it to the list of registrations before asking it to register itself. This simplifies the code path for invoking the `ResultCallback` on success or failure (one path to handle all cases) and also makes it clear that the `DNSServiceRef` object is owned and managed by the `DnssdServiceRegistration` instance.

- The `HandleRegisterResult` callback passed to the MDNSResponder `DNSServiceRegister()` function is now defined and handled by the `DnssdServiceRegistration` class instead of by the `PublisherMDnsSd` class. This simplifies the code, as we no longer need to iterate through the full list of `Registration` objects to find the one corresponding to a related `mServiceRef`. It also fits well with the notion of the `mServiceRef` being owned by a corresponding `DnssdServiceRegistration` object (the destructor will call the `DNSServiceRefDeallocate()` function to deallocate the `mServiceRef`, which will stop the MDNSResponder from invoking the callback).

This commit also makes similar changes to the `PublishHostImpl()` and `DnssdHostRegistration methods, but there are some additional changes specific to host registration:

- The `DnssdHostRegistration` class now tracks whether each host IPv6 address is registered or not.

- The `DnssdHostRegistration` destructor is also changed so that it will always remove all outstanding `DNSRecordRef` objects for all host IPv6 addresses. This updates and fixes the existing behavior where the records were removed only if `IsCompleted()` returned `true`, which requires all addresses to be registered, and could lead to some of the address records not being removed if the host registration was replaced or deleted quickly after its registration.

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -270,6 +270,19 @@ exit:
     return;
 }
 
+DNSServiceErrorType PublisherMDnsSd::CreateSharedHostsRef(void)
+{
+    DNSServiceErrorType dnsError = kDNSServiceErr_NoError;
+
+    VerifyOrExit(mHostsRef == nullptr);
+
+    dnsError = DNSServiceCreateConnection(&mHostsRef);
+    otbrLogDebug("Created new shared DNSServiceRef: %p", mHostsRef);
+
+exit:
+    return dnsError;
+}
+
 void PublisherMDnsSd::DeallocateHostsRef(void)
 {
     VerifyOrExit(mHostsRef != nullptr);
@@ -289,15 +302,7 @@ void PublisherMDnsSd::Update(MainloopContext &aMainloop)
     {
         auto &serviceReg = static_cast<DnssdServiceRegistration &>(*kv.second);
 
-        assert(serviceReg.GetServiceRef() != nullptr);
-
-        int fd = DNSServiceRefSockFD(serviceReg.GetServiceRef());
-
-        if (fd != -1)
-        {
-            FD_SET(fd, &aMainloop.mReadFdSet);
-            aMainloop.mMaxFd = std::max(aMainloop.mMaxFd, fd);
-        }
+        serviceReg.Update(aMainloop);
     }
 
     if (mHostsRef != nullptr)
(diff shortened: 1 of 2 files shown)


abtink:
I will rebase this PR on top of some the later PRs/commits (in particular https://github.com/openthread/ot-br-posix/pull/2012).

abtink:
BTW, I think we should do similar changes as in this PR and https://github.com/openthread/ot-br-posix/pull/2012 for Avahi implemention (will take care of these in future PRs).

src/mdns/mdns_mdnssd.cpp
```diff
@@ -408,131 +409,224 @@ void PublisherMDnsSd::HandleServiceRefDeallocating(const DNSServiceRef &aService
...
-        // We are sure that the service registrations must be instances of `DnssdServiceRegistration`.
-        auto &serviceReg = static_cast<DnssdServiceRegistration &>(*kv.second);
+        DNSRecordRef recordRef = nullptr;
 
-        if (serviceReg.GetServiceRef() == aServiceRef)
-        {
-            result = kv.second.get();
-            break;
-        }
+        dnsError = GetPublisher().CreateSharedHostsRef();
```
superwhd, line 529: I wonder why `mHostsRef` needs to be created for every address? IIUC it's never cleared.

superwhd: Never mind. I just realized `CreateSharedHostsRef` returns no error when the ref is already created.

superwhd (approved):
LGTM 👍

src/mdns/mdns_mdnssd.cpp
wgtdkp: en, so should it be renamed to `GetSharedHostsRef()`? `GetXxxx()` seems more like a singleton

wgtdkp (approved):
Thanks for making this significant improvement! @abtink :+1:

src/mdns/mdns_mdnssd.cpp
abtink: I personally try to use `Get{Item}()` for simple getters, and find/use a different verb for anything that does maybe more work than a simple getter.

jwhui (approved):

Issue #2007, opened 2023-09-09
[mdns-mdnssd] potential access of freed `mServiceRef` from `Stop()`


abtink:
Notice this as I was checking the logs and code related to MDNSResponder `Mdsn::Publisher` implementaion.

- We use a common `mHostsRef` for all host registrations.
- The `Stop()` is as follows:
https://github.com/openthread/ot-br-posix/blob/8a280e3516641f707324bbc83059736d02f68955/src/mdns/mdns_mdnssd.cpp#L240-L265
- It swaps the `mHostRegistrations` with a empty local map which means the entries in the map are retained and freed at the end of this method (when the local `hostRegistrations` map goes out of scope) and the destructor `~DnssdHostRegistration()` is called for all entries
- But `Stop()` will already call `DNSServiceRefDeallocate(mHostsRef);` before that so the the ref is already freed.

------

Digging at history of code, noticed the model of using `swap()` was added in #1221 and before this the maps were `clear()`.
@superwhd can you please share background/reason for this? Thanks.

superwhd:
In my impression I did this to ensure the `mServiceRegistrations` and `mHostRegistrations` are cleaned up only after `DNSServiceRefDeallocate`. 

Probably `DNSServiceRefDeallocate(mHostsRef)` triggers something that touches `mServiceRegistrations/mHostRegistrations` (it may crash if the containers are not empty) but I cannot find an evidence today.

abtink:
> In my impression I did this to ensure the `mServiceRegistrations` and `mHostRegistrations`'s elements are destructed only after `DNSServiceRefDeallocate`.

But that is the problem, the destructor of `DnssdHostRegisteration` will use the already freed and unallocated `mHostsRef`.

The documentation mentioned this:
https://github.com/apple-oss-distributions/mDNSResponder/blob/806254210edec4cab01f794f6fd28658aa6ba59d/mDNSShared/dns_sd.h#L403-L416
```c++
   /* ...
     *
     * 4. Don't Double-Deallocate
     * Calling DNSServiceRefDeallocate(OpRef) for a particular operation's DNSServiceRef terminates
     * just that operation. Calling DNSServiceRefDeallocate(SharedRef) for the main shared DNSServiceRef
     * (the parent DNSServiceRef, originally created by DNSServiceCreateConnection(&SharedRef))
     * automatically terminates the shared connection *and* all operations that were still using it.
[...]

PR #2012, opened 2023-09-11
[mdns-mdnssd] handle deallocated `ServiceRef`s as processing run loop

abtink:
This commit updates `PublisherMDnsSd::Process()`, which processes ready `DNSServiceRef`s from the main run loop. We now track the list of ready service refs in a member variable, `mServiceRefsToProcess`.

As we iterate over the list of `mServiceRefsToProcess`, the call to `DNSServiceProcessResult()` can itself invoke callbacks into `PublisherMDnsSd` and OT, which in turn, may change the state of Publisher and potentially trigger a previously valid `ServiceRef` in the list to be deallocated.

This commit adds `HandleServiceRefDeallocating()`, which is called whenever a `ServiceRef` is being deallocated. From this method, we update the entry in the `mServiceRefsToProcess` list to `nullptr` to avoid calling `DNSServiceProcessResult()` on an already freed `ServiceRef`.

----

Related to https://github.com/openthread/ot-br-posix/issues/2007#issuecomment-1714262548

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -349,10 +349,10 @@ void PublisherMDnsSd::Process(const MainloopContext &aMainloop)
         // state of `Publisher` and potentially trigger a previously
         // valid `ServiceRef` in the list to be deallocated. We use
         // `HandleServiceRefDeallocating()` which is called whenever a
-        // `ServcieRef` is being deallocated and from this we update
+        // `ServiceRef` is being deallocated and from this we update
         // the entry in `mServiceRefsToProcess` list to `nullptr` so to
         // avoid calling `DNSServiceProcessResult()` on an already
-        // freed `ServcieRef`.
+        // freed `ServiceRef`.
 
         if (serviceRef == nullptr)
         {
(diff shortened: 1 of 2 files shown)


superwhd (approved):
LGTM 👍

jwhui (approved):

src/mdns/mdns_mdnssd.cpp
```diff
@@ -324,23 +325,41 @@ void PublisherMDnsSd::Process(const MainloopContext &aMainloop)
...
+        // As we go through the list of `mServiceRefsToProcess` the call
+        // to `DNSServiceProcessResult()` can itself invoke callbacks
+        // into `PublisherMDnsSd` and OT, which in turn, may change the
+        // state of `Publisher` and potentially trigger a previously
+        // valid `ServiceRef` in the list to be deallocated. We use
+        // `HandleServiceRefDeallocating()` which is called whenever a
+        // `ServcieRef` is being deallocated and from this we update
+        // the entry in `mServiceRefsToProcess` list to `nullptr` so to
+        // avoid calling `DNSServiceProcessResult()` on an already
+        // freed `ServcieRef`.
```
jwhui, line 355:
```suggestion
        // `ServiceRef` is being deallocated and from this we update
        // the entry in `mServiceRefsToProcess` list to `nullptr` so to
        // avoid calling `DNSServiceProcessResult()` on an already
        // freed `ServiceRef`.
```

abtink: Thanks for fixing it :)

PR #2013, opened 2023-09-12
[mdns-mdnssd] fix access of freed `mHostsRef` in `Stop()`

abtink:
This commit updates the `Stop()` method to behave differently based on a `StopMode` input. If called from `DNSServiceProcessResult()` call where we get a `kDNSServiceErr_ServiceNotRunning` error, we need to restart the `Publisher` and should immediately de-allocate all `ServiceRefs`, including `mHostsRef`. Otherwise, during a `kNormalStop`, we first clear the `m{Host/Service}Registrations` lists before de-allocating the `mHostsRef`. This ensures that `mHostsRef` and all its related child `DNSRecordRef`s remain valid when `DnssdHostRegistration` destructor is called, which accesses and uses them to update the published AAAA (IPv6 address) records.

----

- ~This PR currently contains the commit from https://github.com/openthread/ot-br-posix/pull/2012.~
- ~Please check and review the second commit on this PR. Thanks.~ 
- Should help address https://github.com/openthread/ot-br-posix/issues/2007

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -222,7 +222,7 @@ PublisherMDnsSd::PublisherMDnsSd(StateCallback aCallback)
 
 PublisherMDnsSd::~PublisherMDnsSd(void)
 {
-    Stop();
+    Stop(kNormalStop);
 }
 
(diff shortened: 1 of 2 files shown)


superwhd (approved):
LGTM 👍  Good catch!

PR #2017, opened 2023-09-14
[agent] add application integration for VendorServer instead of NCP

morningboata:
This allows VendorServer to visit application and all parts of application through Get API. This addresses the issues like: BorderAgent is not accessible from VendorServer.

### DIFF
--- a/src/agent/application.hpp
+++ b/src/agent/application.hpp
@@ -164,7 +164,7 @@ public:
      *
      * @retval OpenThread UBus agent.
      */
-    ubus::UBusAgent mUbusAgent &GetUBusAgent(void)
+    ubus::UBusAgent &GetUBusAgent(void)
     {
         return mUBusAgent;
     }
--- a/src/agent/application.hpp
+++ b/src/agent/application.hpp
@@ -140,7 +140,10 @@ public:
      *
      * @retval OpenThread border agent.
      */
-    BorderAgent &GetBorderAgent(void) { return mBorderAgent; }
+    BorderAgent &GetBorderAgent(void)
+    {
+        return mBorderAgent;
+    }
 #endif
 
 #if OTBR_ENABLE_BACKBONE_ROUTER
@@ -149,7 +152,10 @@ public:
      *
      * @retval OpenThread backbone agent.
      */
-    BackboneRouter::BackboneAgent &GetBackboneAgent(void) { return mBackboneAgent; }
+    BackboneRouter::BackboneAgent &GetBackboneAgent(void)
+    {
+        return mBackboneAgent;
+    }
 #endif
 
 #if OTBR_ENABLE_OPENWRT
@@ -158,7 +164,10 @@ public:
      *
      * @retval OpenThread UBus agent.
      */
-    ubus::UBusAgent mUbusAgent &GetUBusAgent(void) { return mUBusAgent; }
+    ubus::UBusAgent mUbusAgent &GetUBusAgent(void)
+    {
+        return mUBusAgent;
+    }
 #endif
 
 #if OTBR_ENABLE_REST_SERVER
@@ -167,7 +176,10 @@ public:
      *
      * @retval OpenThread rest web server.
      */
-    rest::RestWebServer &GetRestWebServer(void) { return mRestWebServer; }
+    rest::RestWebServer &GetRestWebServer(void)
+    {
+        return mRestWebServer;
+    }
 #endif
 
 #if OTBR_ENABLE_DBUS_SERVER
@@ -176,7 +188,10 @@ public:
      *
      * @retval OpenThread DBus agent.
      */
-    DBus::DBusAgent &GetDBusAgent(void) { return mDBusAgent; }
+    DBus::DBusAgent &GetDBusAgent(void)
+    {
+        return mDBusAgent;
+    }
 #endif
 
 private:
--- a/src/agent/application.hpp
+++ b/src/agent/application.hpp
@@ -130,7 +130,7 @@ public:
     /**
      * Get the OpenThread controller object the application is using.
      *
-     * @returns OpenThread controller object.
+     * @returns The OpenThread controller object.
      */
     Ncp::ControllerOpenThread &GetNcp(void) { return mNcp; }
 
@@ -138,7 +138,7 @@ public:
     /**
      * Get the border agent the application is using.
      *
-     * @returns border agent.
+     * @returns The border agent.
      */
     BorderAgent &GetBorderAgent(void)
     {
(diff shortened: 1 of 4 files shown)


src/agent/application.hpp
```diff
@@ -119,6 +127,73 @@ class Application : private NonCopyable
      */
     otbrError Run(void);
 
+    /**
+     * Returns the OpenThread controller object the application is using.
+     *
+     * @retval OpenThread controller object.
```
superwhd, line 133: My understanding is `@retval` should be used with constants like enums or booleans. For this case `@returns` may fit better.

src/agent/vendor.hpp
```diff
@@ -54,11 +57,11 @@ class VendorServer
      *
      * Custom vendor servers should implement this method to return an object of the derived class.
      *
-     * @param[in]  aNcp  The OpenThread controller object.
+     * @param[in]  application  The OpenThread application.
```
superwhd, line 60:
```suggestion
     * @param[in]  application  The OTBR application.
```

src/agent/vendor.hpp
```diff
@@ -36,9 +36,12 @@
 
 #include "openthread-br/config.h"
 
-#include "ncp/ncp_openthread.hpp"
+#include "agent/application.hpp"
 
 namespace otbr {
+
+class Application;
```
superwhd, line 43: Is the forward declaration needed? You've already included the `application.hpp`.

src/agent/application.hpp
```diff
@@ -64,6 +64,14 @@
 
 namespace otbr {
 
+#if OTBR_ENABLE_VENDOR_SERVER
+namespace vendor {
+
+class VendorServer;
```
superwhd, line 70: What's this forward declaration used for?

src/agent/vendor.hpp
morningboata: Done.

src/agent/application.hpp
morningboata: Done.

src/agent/vendor.hpp
morningboata: I used this to fix compile issue.

src/agent/application.hpp
morningboata: I used this to fix compile issue.

src/agent/vendor.hpp
```diff
@@ -54,11 +57,11 @@ class VendorServer
      *
      * Custom vendor servers should implement this method to return an object of the derived class.
      *
-     * @param[in]  aNcp  The OpenThread controller object.
+     * @param[in]  application  The OTBR application.
      *
      * @returns  New derived VendorServer instance.
      */
-    static std::shared_ptr<VendorServer> newInstance(otbr::Ncp::ControllerOpenThread &aNcp);
+    static std::shared_ptr<VendorServer> newInstance(otbr::Application &application);
```
superwhd, line 64:
```suggestion
    static std::shared_ptr<VendorServer> newInstance(Application &application);
```

morningboata: Done.

superwhd (approved):
LGTM 👍

src/agent/application.hpp
```diff
@@ -119,6 +127,73 @@ class Application : private NonCopyable
      */
     otbrError Run(void);
 
+    /**
+     * Returns the OpenThread controller object the application is using.
+     *
+     * @returns OpenThread controller object.
```
wgtdkp, line 133: The `@returns` tag looks redundant. @jwhui Do we always need it?

wgtdkp (approved):
LGTM :+1:

src/agent/application.hpp
jwhui: For the core OpenThread project, we currently rely on the brief and tag to populate the OpenThread reference documentation. Example: https://openthread.io/reference/group/plat-alarm#otplatalarmmicrogetnow

Maybe we could omit the brief and only the "Details" table remains. Thoughts?

/cc/ @Vyrastas

wgtdkp: > Maybe we could omit the brief and only the "Details" table remains. Thoughts?

I would prefer keeping the brief but omitting the "Details", because the brief is used in the summary section. For example, https://openthread.io/reference/group/plat-alarm#summary

Vyrastas: If we're adding comments, ideally they should be formatted and tagged appropriately as if we expect to generate docs for them in the future.

From @jwhui 's example (otPlatAlarmMicroGetNow), there's no brief tag, the first sentence is what is taken for the summary, the rest of the description is included in the function detail.  I.e. https://github.com/openthread/openthread/blob/df128246ac9fac5f984e7b7f9a1317321185c97d/include/openthread/platform/alarm-micro.h#L74

So for something like this, I would suggest comments like this:

```
     * Get the OpenThread controller object the application is using.
     *
     * @returns The OpenThread controller object.
```

morningboata: Thanks for the suggestion. Sounds good to me. I will update the comments in this PR accordingly.

Vyrastas (approved):
LGTM for docs/comments

src/agent/vendor.hpp
```diff
@@ -54,11 +57,11 @@ class VendorServer
      *
      * Custom vendor servers should implement this method to return an object of the derived class.
      *
-     * @param[in]  aNcp  The OpenThread controller object.
+     * @param[in]  application  The OTBR application.
      *
      * @returns  New derived VendorServer instance.
      */
-    static std::shared_ptr<VendorServer> newInstance(otbr::Ncp::ControllerOpenThread &aNcp);
+    static std::shared_ptr<VendorServer> newInstance(Application &application);
```
abtink, line 64:
```suggestion
    static std::shared_ptr<VendorServer> newInstance(Application &aApplication);
```

src/agent/vendor.hpp
```diff
@@ -54,11 +57,11 @@ class VendorServer
      *
      * Custom vendor servers should implement this method to return an object of the derived class.
      *
-     * @param[in]  aNcp  The OpenThread controller object.
+     * @param[in]  application  The OTBR application.
```
abtink, line 60:
```suggestion
     * @param[in]  aApplication  The OTBR application.
```

abtink (approved):
Looks good. 👍 

Couple of small style change suggestions below.

src/agent/vendor.hpp
morningboata: Done.

src/agent/vendor.hpp
morningboata: Done.

jwhui (approved):

PR #2019, opened 2023-09-15
[mdns-avahi] handle freed `AvahiWatch/Timeout` entries as processing run loop

abtink:
Avahi provides a mechanism to integrate it to a `select()` style run loop where an implementation can provide an `AvahiWatch` and an `AvahiTimeout` structure along with a set of related APIs (as function pointers given to Avahi client) for it to allocate new watch/timeout entries or update or free them. A watch entry provides a set of events to watch for and expects its given callback to be invoked when the events occur. A timeout provides an abstraction for a timer.

This commit updates `AvahiPoller` (which integrates Avahi to run loop as a `MainloopProcessor`) to protect against situations where entries get freed or updated as Avahi callbacks are invoked from `AvahiPoller::Process()`.

When we invoke the callback for an `AvahiWatch` or `AvahiTimeout` the Avahi module can call any of APIs we provided to it. For example, it can update or free any existing `AvahiWatch/Timeout` entry, which in turn, modifies the `mWatches` or `mTimers` list being tracked by the `AvahiPoller`.

This commit updates the `AvahiPoller` class to correctly handle such situations. Before invoking the callback, we update the entry's state and after returning from callback we restart the iteration over the watch/timer list to find the next entry to report, as the list may have changed during execution of the callback.

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -53,30 +53,42 @@
 #include "common/logging.hpp"
 #include "common/time.hpp"
 
+namespace otbr {
+namespace Mdns {
+
+class AvahiPoller;
+
+} // namespace Mdns
+} // namespace otbr
+
 struct AvahiWatch
 {
-    int                mFd;       ///< The file descriptor to watch.
-    AvahiWatchEvent    mEvents;   ///< The interested events.
-    int                mHappened; ///< The events happened.
-    AvahiWatchCallback mCallback; ///< The function to be called when interested events happened on mFd.
-    void              *mContext;  ///< A pointer to application-specific context.
-    void              *mPoller;   ///< The poller created this watch.
+    typedef otbr::Mdns::AvahiPoller AvahiPoller;
+
+    int                mFd;           ///< The file descriptor to watch.
+    AvahiWatchEvent    mEvents;       ///< The interested events.
+    int                mHappened;     ///< The events happened.
+    AvahiWatchCallback mCallback;     ///< The function to be called to report events happened on `mFd`.
+    void              *mContext;      ///< A pointer to application-specific context to use with `mCallback`.
+    bool               mShouldReport; ///< Whether or not we need to report events (invoking callback).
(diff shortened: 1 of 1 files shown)


src/mdns/mdns_avahi.cpp
```diff
@@ -364,56 +380,90 @@ void AvahiPoller::Update(MainloopContext &aMainloop)
...
 
-        avahiTimeout->mCallback(avahiTimeout, avahiTimeout->mContext);
+        for (AvahiTimeout *timer : mTimers)
+        {
+            if (timer->mShouldReport)
+            {
+                shouldReport         = true;
+                timer->mShouldReport = false;
+                timer->mCallback(timer, timer->mContext);
+            }
```
superwhd, line 465: Should there be a break?

src/mdns/mdns_avahi.cpp
```diff
@@ -364,56 +380,90 @@ void AvahiPoller::Update(MainloopContext &aMainloop)
...
+        }
```
superwhd, line 466: Probably we can put the iteration of watches and timers into the same `while (shouldReport)`. I'm not sure which version is more readable though.

src/mdns/mdns_avahi.cpp
abtink: Yes. Indeed.  I missed this. Thanks.

abtink: Fixed in new push. Thanks for noticing this.

src/mdns/mdns_avahi.cpp
abtink: I think better to keep them separate to be more efficient during run time. If we combine and use one loop for both watches and timers, every iteration of the outer `while (shouldReport)` loop in order to get to timers we nee to first go over all watches again in the list (even if all watches are already handled).

superwhd (approved):
LGTM 👍

jwhui (approved):

jwhui (approved):

PR #2026, opened 2023-09-19
[tests] remove the assert on `preferredChannelMask` value

abtink:
---

Should help address CI test failures in https://github.com/openthread/ot-br-posix/pull/2025

We have this line here in the test_dbus_client code:

```c++
    TEST_ASSERT(api->GetPreferredChannelMask(preferredChannelMask) == ClientError::ERROR_NONE);
    TEST_ASSERT(preferredChannelMask == 0x7fff800);
```

The value we expect is different from what we get as `134180864` or `0x7ff7000` in the CI runs (which I guess was changed as part of OpenThread PR https://github.com/openthread/openthread/pull/9391)

This PR removes the assert check so to unblock #2025. We can re-add the check later (if we want) once the OT commit is updated on `ot-br-posix` and have the commit from https://github.com/openthread/openthread/pull/9391.

### DIFF
--- a/tests/dbus/test_dbus_client.cpp
+++ b/tests/dbus/test_dbus_client.cpp
@@ -338,7 +338,6 @@ int main()
     TEST_ASSERT(region == "US");
 
     TEST_ASSERT(api->GetPreferredChannelMask(preferredChannelMask) == ClientError::ERROR_NONE);
-    TEST_ASSERT(preferredChannelMask == 0x7fff800);
 
     api->EnergyScan(scanDuration, [&stepDone](const std::vector<EnergyScanResult> &aResult) {
         TEST_ASSERT(!aResult.empty());


tests/dbus/test_dbus_client.cpp
```diff
@@ -338,7 +338,6 @@ int main()
     TEST_ASSERT(region == "US");
 
     TEST_ASSERT(api->GetPreferredChannelMask(preferredChannelMask) == ClientError::ERROR_NONE);
-    TEST_ASSERT(preferredChannelMask == 0x7fff800);
```
zhanglongxia, line 341: Could we change it to 

```
TEST_ASSERT(preferredChannelMask == 0x7ff7000);
```

Here is the configuration of the prefered channel mask: https://github.com/zhanglongxia/openthread/blob/master/src/posix/platform/openthread.conf.example#L35

abtink: The problem with this is that then I guess this PR itself would not pass CI tests (as the OT commit changing it on OT side is being merged from #2025).

zhanglongxia (approved):
LGTM

tests/dbus/test_dbus_client.cpp
zhanglongxia: Agree, we should merge this CL first to make the OpenThread be able to be synchronized to ot-br-posix in https://github.com/openthread/ot-br-posix/pull/2025, then we can submit the changes to check the value of `preferredChannelMask`.

jwhui (approved):

PR #2039, opened 2023-09-26
[mdns] handle list change from `DiscoverCallback` (host or service)

abtink:
This commit updates how the `DiscoverCallback` are tracked in `Mdns::Publisher` and how the callbacks are invoked. In particular, the `mDiscoverCallbacks` list itself can get updated as the callbacks are invoked. The new code ensures that we can handle such a situation from both `OnServiceResolved()` or `OnHostResolved()`.

----

The `OnServiceResolved()` was updated in PR #1742 to handle if the list gets updated from callback. Similar situation can happen in `OnHostResolved()` and from host callbacks. This PR changes the model to handle both situations.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -174,31 +174,24 @@ exit:
 
 void Publisher::RemoveSubscriptionCallbacks(uint64_t aSubscriberId)
 {
-    size_t erased;
-
-    OTBR_UNUSED_VARIABLE(erased);
-
-    assert(aSubscriberId > 0);
-
-    erased = mDiscoveredCallbacks.erase(aSubscriberId);
-
-    assert(erased == 1);
+    mDiscoverCallbacks.remove_if(
+        [aSubscriberId](const DiscoverCallback &aCallback) { return (aCallback.mId == aSubscriberId); });
 }
 
 uint64_t Publisher::AddSubscriptionCallbacks(Publisher::DiscoveredServiceInstanceCallback aInstanceCallback,
                                              Publisher::DiscoveredHostCallback            aHostCallback)
 {
-    uint64_t subscriberId = mNextSubscriberId++;
+    uint64_t id = mNextSubscriberId++;
 
-    assert(subscriberId > 0);
+    assert(id > 0);
+    mDiscoverCallbacks.emplace_back(id, aInstanceCallback, aHostCallback);
 
-    mDiscoveredCallbacks.emplace(subscriberId, std::make_pair(std::move(aInstanceCallback), std::move(aHostCallback)));
-    return subscriberId;
+    return id;
 }
 
 void Publisher::OnServiceResolved(std::string aType, DiscoveredInstanceInfo aInstanceInfo)
 {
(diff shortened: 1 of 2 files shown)


src/mdns/mdns.cpp
```diff
@@ -174,31 +174,24 @@ otbrError Publisher::DecodeTxtData(Publisher::TxtList &aTxtList, const uint8_t *
...
-
-    OTBR_UNUSED_VARIABLE(erased);
-
-    assert(aSubscriberId > 0);
-
-    erased = mDiscoveredCallbacks.erase(aSubscriberId);
-
-    assert(erased == 1);
+    mDiscoverCallbacks.remove_if(
+        [aSubscriberId](DiscoverCallback &aCallback) { return (aCallback.mId == aSubscriberId); });
```
superwhd, line 178:
```suggestion
        [aSubscriberId](const DiscoverCallback &aCallback) { return (aCallback.mId == aSubscriberId); });
```
nit

src/mdns/mdns.hpp
```diff
@@ -575,9 +576,28 @@ class Publisher : private NonCopyable
...
     HostRegistrationMap    mHostRegistrations;
 
+    struct DiscoverCallback
+    {
+        DiscoverCallback(uint64_t                          aId,
+                         DiscoveredServiceInstanceCallback aServiceCallback,
+                         DiscoveredHostCallback            aHostCallback)
+            : mId(aId)
+            , mServiceCallback(aServiceCallback)
+            , mHostCallback(aHostCallback)
```
superwhd, line 586:
```suggestion
            , mServiceCallback(std::move(aServiceCallback))
            , mHostCallback(std::move(aHostCallback))
```

src/mdns/mdns.cpp
```diff
@@ -216,22 +209,33 @@ void Publisher::OnServiceResolved(std::string aType, DiscoveredInstanceInfo aIns
...
+    // and restart the iteration over the `mDiscoverCallbacks` list
+    // to find the next one to signal, since the list may have changed.
+
+    for (DiscoverCallback &callback : mDiscoverCallbacks)
     {
-        subscriberIds.push_back(subCallback.first);
+        if (callback.mServiceCallback != nullptr)
+        {
+            callback.mShouldInvoke = true;
+            checkToInvoke          = true;
```
superwhd, line 223: Maybe we can initialize `checkToInvoke` as true at beginning then we can save this line.

superwhd (approved):
LGTM overall. Thanks!

wgtdkp (approved):
LGTM :+1:

src/mdns/mdns.cpp
abtink: The idea here is that we want to skip the second `while ()` loop if none of the callbacks need to be invoked.

src/mdns/mdns.cpp
abtink: Updated in new push. Thanks.

jwhui (approved):

PR #2041, opened 2023-09-26
[mdns] report the network interface index for discovered host info

abtink:
This commit updates the `Mdns::Publisher::DiscoveredHostInfo` struct to include the `mNetifIndex` field, which provides the network interface index for a discovered host. This information is already provided by the `DiscoveredInstanceInfo` struct for a discovered service instance.

Both Avahi and MDNSResponder implementations are updated to provide the `mNetifIndex` field.

### DIFF
--- a/src/mdns/mdns.hpp
+++ b/src/mdns/mdns.hpp
@@ -143,9 +143,10 @@ public:
      */
     struct DiscoveredHostInfo
     {
-        std::string mHostName;  ///< Full host name.
-        AddressList mAddresses; ///< IP6 addresses.
-        uint32_t    mTtl = 0;   ///< Host TTL.
+        std::string mHostName;       ///< Full host name.
+        AddressList mAddresses;      ///< IP6 addresses.
+        uint32_t    mNetifIndex = 0; ///< Network interface.
+        uint32_t    mTtl        = 0; ///< Host TTL.
     };
 
     /**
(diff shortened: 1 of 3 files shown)


superwhd (approved):
LGTM 👍

wgtdkp (approved):
LGTM :+1:

src/mdns/mdns.hpp
```diff
@@ -143,9 +143,10 @@ class Publisher : private NonCopyable
      */
     struct DiscoveredHostInfo
     {
-        std::string mHostName;  ///< Full host name.
-        AddressList mAddresses; ///< IP6 addresses.
-        uint32_t    mTtl = 0;   ///< Host TTL.
+        std::string mHostName;       ///< Full host name.
+        AddressList mAddresses;      ///< IP6 addresses.
+        uint32_t    mNetifIndex = 0; ///< Network interface.
```
bukepo, line 148: It seems POSIX defines network interface index to `unsigned int`. Should we use that type instead?

https://pubs.opengroup.org/onlinepubs/9699919799/functions/if_nametoindex.html

abtink: I am not sure. `mDNSResponder` uses `uint32_t` in its API, and `Avahi` has `typedef int AvahiIfIndex`.

Independent of whether we want to change this type or not, I would suggest such a change to be done in its own PR and change all existing definitions of `mNetifIndex` in `Publisher` module. This PR is adding the missing in `mNetIfIdex` in `DiscoveredHostInfo` to make it similar to `DiscoveredServiceInfo`.

bukepo (approved):

jwhui (approved):

PR #2065, opened 2023-10-16
Remove 3 otPlatRadio related API in telemetry retrieval part

morningboata:
It helps to solve the spinel frame timeout issue during Android dump. We can add them back after we fix it.

### DIFF
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -910,12 +910,7 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
         wpanStats->set_mac_cca_fail_rate(static_cast<float>(ccaFailureRate) / 0xffff);
     }
 
-    {
-        int8_t radioTxPower;
-
-        SuccessOrExit(error = otPlatRadioGetTransmitPower(mInstance, &radioTxPower));
-        wpanStats->set_radio_tx_power(radioTxPower);
-    }
+    // TODO: Set radio_tx_power by otPlatRadioGetTransmitPower.
 
     {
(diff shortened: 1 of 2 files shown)


src/utils/thread_helper.cpp
```diff
@@ -1356,31 +1349,7 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
...
-            coexMetrics->set_count_rx_grant_wait(otRadioCoexMetrics.mNumRxGrantWait);
-            coexMetrics->set_count_rx_grant_wait_activated(otRadioCoexMetrics.mNumRxGrantWaitActivated);
-            coexMetrics->set_count_rx_grant_wait_timeout(otRadioCoexMetrics.mNumRxGrantWaitTimeout);
-            coexMetrics->set_count_rx_grant_deactivated_during_request(
-                otRadioCoexMetrics.mNumRxGrantDeactivatedDuringRequest);
-            coexMetrics->set_count_rx_grant_none(otRadioCoexMetrics.mNumRxGrantNone);
-            coexMetrics->set_rx_average_request_to_grant_time_us(otRadioCoexMetrics.mAvgRxRequestToGrantTime);
-        }
-        // End of CoexMetrics section.
+        // TODO: set CoexMetrics by otPlatRadioGetCoexMetrics.
```
abtink, line 1352: Should we instead add build config for these and enclose the related code with `#if <config>` check to allow them to be disabled/removed when needed.

This seems to be better way than deleting existing code.

morningboata: Thanks for the suggestion. I will rethink the strategy for this CL and come back later.

PR #2072, opened 2023-10-20
[utils] fix buffer overflow in `Pskc::SetSalt`

alexanderkent:
This commit addresses a potential buffer overflow vulnerability in the `Pskc::SetSalt` method.

Root Issue:
The `Pskc::SetSalt` method copies data into `mSalt` without sufficiently checking that the combined length of the data being copied (`saltPrefix`, `aExtPanId`, `aNetworkName`) doesn't exceed the size of `mSalt`. This leads to a stack buffer overflow, which may cause stack corruption, resulting in undefined behavior and potential security risks.

Fix:
1. Added a check to ensure that the total length of data to be copied into `mSalt` doesn't exceed its size.
2. Added a warning log if the network name is too long and needs to be truncated.
3. Utilized `std::min` to copy only as much of the network name as can fit into the remaining space in `mSalt`.

This ensures that no buffer overflow occurs, thereby preventing stack corruption and potential security vulnerabilities.

### DIFF
--- a/src/utils/pskc.cpp
+++ b/src/utils/pskc.cpp
@@ -41,15 +41,15 @@ namespace Psk {
 
 void Pskc::SetSalt(const uint8_t *aExtPanId, const char *aNetworkName)
 {
-    const char *saltPrefix      = "Thread";
-    int         cur             = 0;
-    int         ret             = kPskcStatus_Ok;
-    int         remainingSpace  = 0;
-    int         networkNameLen  = 0;
+    const char *saltPrefix     = "Thread";
+    int         cur            = 0;
+    int         ret            = kPskcStatus_Ok;
+    int         remainingSpace = 0;
+    int         networkNameLen = 0;
 
-    if(strlen(saltPrefix) + OT_EXTENDED_PAN_ID_LENGTH + strlen(aNetworkName) >= sizeof(mSalt)){
-      otbrLogWarning("Network name too long; will be truncated.");
-    }
+    if (strlen(saltPrefix) + OT_EXTENDED_PAN_ID_LENGTH + strlen(aNetworkName) >= sizeof(mSalt))
+    {
+        otbrLogWarning("Network name too long; will be truncated.");
 
     memset(mSalt, 0, sizeof(mSalt));
     memcpy(mSalt, saltPrefix, strlen(saltPrefix));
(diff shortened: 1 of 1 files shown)


src/utils/pskc.cpp
```diff
@@ -41,9 +41,15 @@ namespace Psk {
...
-    int         ret        = kPskcStatus_Ok;
+    const char *saltPrefix      = "Thread";
+    int         cur             = 0;
+    int         ret             = kPskcStatus_Ok;
+    int         remainingSpace  = 0;
+    int         networkNameLen  = 0;
+
+    if(strlen(saltPrefix) + OT_EXTENDED_PAN_ID_LENGTH + strlen(aNetworkName) >= sizeof(mSalt)){
+      otbrLogWarning("Network name too long; will be truncated.");
+    }
```
jwhui, line 52: To resolve pretty-check: https://github.com/openthread/ot-br-posix/actions/runs/6590257499/job/17915946140?pr=2072
```suggestion
    const char *saltPrefix     = "Thread";
    int         cur            = 0;
    int         ret            = kPskcStatus_Ok;
    int         remainingSpace = 0;
    int         networkNameLen = 0;

    if (strlen(saltPrefix) + OT_EXTENDED_PAN_ID_LENGTH + strlen(aNetworkName) >= sizeof(mSalt))
    {
        otbrLogWarning("Network name too long; will be truncated.");
```

src/utils/pskc.cpp
```diff
@@ -53,8 +59,13 @@ void Pskc::SetSalt(const uint8_t *aExtPanId, const char *aNetworkName)
     cur += OT_EXTENDED_PAN_ID_LENGTH;
 
     VerifyOrExit(strlen(aNetworkName) > 0, ret = kPskcStatus_InvalidArgument);
-    memcpy(mSalt + cur, aNetworkName, strlen(aNetworkName));
-    cur += strlen(aNetworkName);
+
+    remainingSpace = sizeof(mSalt) - cur;
+    if (remainingSpace > 0) {
```
jwhui, line 64: To resolve pretty-check: https://github.com/openthread/ot-br-posix/actions/runs/6590257499/job/17915946140?pr=2072
```suggestion
    if (remainingSpace > 0)
    {
```

jwhui (commented):
Thanks for this contribution! Just some minor code-style suggestions to resolve the `pretty-check` failure.

jwhui (approved):

PR #2076, opened 2023-10-26
[thread] refine `RetrieveTelemetryData` to populate metrics with best effort

morningboata:
Best effort means, for a given telemetry, if its retrieval has error, it is left unpopulated and the process continues to retrieve the remaining telemetries instead of the immediately return. The error code OT_ERRROR_FAILED will be returned if there is one or more error(s) happened in the process.
This change helps to retrieve telemetries with best effort when needed.


src/utils/thread_helper.cpp
```diff
@@ -968,10 +968,15 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
 
         wpanTopoFull->set_rloc16(rloc16);
 
-        otRouterInfo info;
+        {
+            otRouterInfo info;
 
-        VerifyOrExit(otThreadGetRouterInfo(mInstance, rloc16, &info) == OT_ERROR_NONE, error = OT_ERROR_INVALID_STATE);
-        wpanTopoFull->set_router_id(info.mRouterId);
+            // Store router info only when router is in correct status.
```
superwhd, line 974: The comment may not be accurate. It may return error when invalid arguments are given or the router cannot be found. I'd suggest not to describe the possible errors here.

src/utils/thread_helper.cpp
```diff
@@ -1016,19 +1025,23 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
             uint8_t              len = sizeof(data);
             std::vector<uint8_t> networkData;
 
-            SuccessOrExit(error = otNetDataGet(mInstance, /*stable=*/false, data, &len));
-            networkData = std::vector<uint8_t>(&data[0], &data[len]);
-            wpanTopoFull->set_network_data(std::string(networkData.begin(), networkData.end()));
+            if (otNetDataGet(mInstance, /*stable=*/false, data, &len))
```
superwhd, line 1028: Seems `otNetDataGet` returns an error instead of a boolean.

src/utils/thread_helper.cpp
```diff
@@ -1016,19 +1025,23 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
...
 
         {
             uint8_t              data[kNetworkDataMaxSize];
             uint8_t              len = sizeof(data);
             std::vector<uint8_t> networkData;
 
-            SuccessOrExit(error = otNetDataGet(mInstance, /*stable=*/true, data, &len));
-            networkData = std::vector<uint8_t>(&data[0], &data[len]);
-            wpanTopoFull->set_stable_network_data(std::string(networkData.begin(), networkData.end()));
+            if (otNetDataGet(mInstance, /*stable=*/true, data, &len))
```
superwhd, line 1040: ditto

src/utils/thread_helper.hpp
```diff
@@ -258,7 +258,9 @@ class ThreadHelper
 
 #if OTBR_ENABLE_TELEMETRY_DATA_API
     /**
-     * This method populates the telemetry data and returns the error code if error happens.
+     * This method populates the telemetry data with best effort and returns the error code if error
```
superwhd, line 261: The comment or the method's behavior needs to be updated. Currently the method returns 'no error' when the result is partial.

src/utils/thread_helper.hpp
```diff
@@ -258,7 +258,9 @@ class ThreadHelper
...
+     * happens. Best effort means the telemetry will be populated if it is available; otherwise it
+     * is left unpopulated.
```
superwhd, line 263: From the telemetry analysis's perspective, do we still want the data if the data is incomplete?

src/utils/thread_helper.cpp
morningboata: Done.

src/utils/thread_helper.cpp
morningboata: Done.

src/utils/thread_helper.cpp
morningboata: Done.

src/utils/thread_helper.hpp
morningboata: A last error will be returned. For best effort caller, it can always upload the telemetry as as. For non-best effort caller, the caller can utilize this code to decide whether to upload the telemetry. In my opinion, there are huge amount of telemetries includes, and we want the data as long as the wanted data section is available.

src/utils/thread_helper.cpp
```diff
@@ -913,8 +914,14 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
...
-        SuccessOrExit(error = otPlatRadioGetTransmitPower(mInstance, &radioTxPower));
-        wpanStats->set_radio_tx_power(radioTxPower);
+        if ((latestError = otPlatRadioGetTransmitPower(mInstance, &radioTxPower)) == OT_ERROR_NONE)
+        {
+            wpanStats->set_radio_tx_power(radioTxPower);
+        }
+        else
+        {
+            error = latestError;
+        }
```
superwhd, line 924: What about we only have one `error` variable and we set it to 'failed' error when any error occurs? This may be easier to understand. The 'lastError` variable name could be ambiguous sometimes.

Also, there could be multiple errors happening in this method and it's not so meaningful to return one precise error when there are many errors. Thus I suggest to use 'fail' as the returned error type.

morningboata: SG. I return the error code OT_ERRROR_FAILED now if there is one or more error(s) happened in the process.

src/utils/thread_helper.hpp
```diff
@@ -258,12 +258,15 @@ class ThreadHelper
...
+     * This method populates the telemetry data with best effort. The best effort means, for a given
+     * telemetry, if its retrieval has error, it is left unpopulated and the process continues to
+     * retrieve the remaining telemetries instead of the immediately return. The error code
+     * OT_ERRROR_FAILED will be returned if there is one or more error(s) happened in the process.
      *
      * @param[in] aPublisher     The Mdns::Publisher to provide MDNS telemetry if it is not `nullptr`.
      * @param[in] telemetryData  The telemetry data to be populated.
      *
-     * @returns The error code if error happens during the population of the telemetry data.
+     * @returns The error code OT_ERRROR_FAILED if there is one or more error(s) happened in the process
```
superwhd, line 269: nit: Maybe less ambiguous using `@retval`.

morningboata: Done.

abtink (approved):
LGTM. 👍

jwhui (approved):

PR #2083, opened 2023-11-02
[ot-client] fix buffer overflow in `OpenThreadClient::Execute`

alexanderkent:
This PR addresses a potential buffer overflow in the OpenThreadClient::Execute method. The original implementation did not properly check the size of the data being written to mBuffer, leading to potential buffer overflows.

Changes:
* Modified the vsnprintf function call to write to &mBuffer[1] with a size limit of sizeof(mBuffer) - 2, leaving room for newline characters at both ends.
* Added a check to ensure that the total size of the data (including the two newline characters) does not exceed sizeof(mBuffer).
* Improved error handling for the socket write operation (count != ret) instead of (count < ret) as the function seems to expect the entire command to be sent in one go (and logs an error otherwise).

Testing:
* Manually tested forming a network with various params
* Added unit test for truncated network name in PSKc


src/web/web-service/ot_client.cpp
```diff
@@ -147,31 +147,31 @@ char *OpenThreadClient::Execute(const char *aFormat, ...)
...
 
     va_start(args, aFormat);
-    ret = vsnprintf(&mBuffer[1], sizeof(mBuffer) - 1, aFormat, args);
+    ret = vsnprintf(&mBuffer[1], sizeof(mBuffer) - 2, aFormat, args);
     va_end(args);
 
     if (ret < 0)
     {
         otbrLogErr("Failed to generate command: %s", strerror(errno));
+        goto exit;
```
jwhui, line 156:
```suggestion
        ExitNow();
```

src/web/web-service/ot_client.cpp
```diff
@@ -147,31 +147,31 @@ char *OpenThreadClient::Execute(const char *aFormat, ...)
...
     }
-
-    mBuffer[0] = '\n';
-    ret++;
-
-    if (ret == sizeof(mBuffer))
+    if (static_cast<size_t>(ret) >= sizeof(mBuffer) - 2)
     {
         otbrLogErr("Command exceeds maximum limit: %d", kBufferSize);
+        goto exit;
```
jwhui, line 161:
```suggestion
        ExitNow();
```

src/web/web-service/ot_client.cpp
```diff
@@ -147,31 +147,31 @@ char *OpenThreadClient::Execute(const char *aFormat, ...)
...
+    ret += 2;
 
     count = write(mSocket, mBuffer, ret);
 
-    if (count < ret)
+    if (count != ret)
     {
         mBuffer[ret] = '\0';
         otbrLogErr("Failed to send command: %s", mBuffer);
+        goto exit;
```
jwhui, line 174:
```suggestion
        ExitNow();
```

jwhui (approved):
Thanks! 👍🏼

PR #2102, opened 2023-11-21
[logging] add command line arg to disable syslog

greens:
Adds an optional flag to otbr-agent to allow disabling of logging to syslog.

Current behavior has logs sent to both standard output as well as the file indicated by syslog. With this flag enabled, logs would only be sent to standard output.

### DIFF
--- a/src/agent/main.cpp
+++ b/src/agent/main.cpp
@@ -70,6 +70,7 @@ enum
     OTBR_OPT_HELP                    = 'h',
     OTBR_OPT_INTERFACE_NAME          = 'I',
     OTBR_OPT_VERBOSE                 = 'v',
+    OTBR_OPT_SYSLOG_DISABLE          = 's',
     OTBR_OPT_VERSION                 = 'V',
     OTBR_OPT_SHORTMAX                = 128,
     OTBR_OPT_RADIO_VERSION,
@@ -90,6 +91,7 @@ static const struct option kOptions[] = {
     {"help", no_argument, nullptr, OTBR_OPT_HELP},
     {"thread-ifname", required_argument, nullptr, OTBR_OPT_INTERFACE_NAME},
     {"verbose", no_argument, nullptr, OTBR_OPT_VERBOSE},
+    {"syslog-disable", no_argument, nullptr, OTBR_OPT_SYSLOG_DISABLE},
     {"version", no_argument, nullptr, OTBR_OPT_VERSION},
     {"radio-version", no_argument, nullptr, OTBR_OPT_RADIO_VERSION},
     {"auto-attach", optional_argument, nullptr, OTBR_OPT_AUTO_ATTACH},
@@ -136,9 +138,10 @@ static std::vector<char *> AppendAutoAttachDisableArg(int argc, char *argv[])
 static void PrintHelp(const char *aProgramName)
 {
     fprintf(stderr,
-            "Usage: %s [-I interfaceName] [-B backboneIfName] [-d DEBUG_LEVEL] [-v] [--auto-attach[=0/1]] RADIO_URL "
-            "[RADIO_URL]\n"
-            "    --auto-attach defaults to 1\n",
+            "Usage: %s [-I interfaceName] [-B backboneIfName] [-d DEBUG_LEVEL] [-v] [-s] [--auto-attach[=0/1]] "
+            "RADIO_URL [RADIO_URL]\n"
+            "    --auto-attach defaults to 1\n"
+            "    -s disables syslog and prints to standard out\n",
             aProgramName);
     fprintf(stderr, "%s", otSysGetRadioUrlHelpString());
 }
@@ -195,6 +198,7 @@ static int realmain(int argc, char *argv[])
     int                       ret               = EXIT_SUCCESS;
     const char               *interfaceName     = kDefaultInterfaceName;
     bool                      verbose           = false;
+    bool                      syslogDisable     = false;
     bool                      printRadioVersion = false;
     bool                      enableAutoAttach  = true;
     const char               *restListenAddress = "";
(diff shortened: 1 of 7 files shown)


jdswensen:
Hey @wgtdkp, would you have some time to review this?

The background on this change is that some systems may choose to capture/rotate logs in a method that takes control of `stdout`/`stderr` rather than via syslog. Information coming out of `ot-br-posix` is useful and still needs to be captured, however it can be rather chatty and has potential to clutter the system logs. On resource constrained systems with low size and rotation settings, this behavior results in system logs rotating more quickly than is useful.

greens:
I believe these build failures are unrelated to my changes.

src/common/logging.hpp
```diff
@@ -81,17 +81,18 @@ void otbrLogSetLevel(otbrLogLevel aLevel);
  * @param[in] enable  True to log to/via syslog.
```
wgtdkp, line 81: nit: please update the @param documentation to be consistent

src/common/logging.cpp
```diff
@@ -53,6 +53,7 @@ static otbrLogLevel sLevel            = OTBR_LOG_INFO;
 static const char   sLevelString[][8] = {
       "[EMERG]", "[ALERT]", "[CRIT]", "[ERR ]", "[WARN]", "[NOTE]", "[INFO]", "[DEBG]",
 };
+static bool syslogDisabled = false;
```
wgtdkp, line 56: To be consistent with the code convention:
```suggestion
static bool sSyslogDisabled = false;
```

Add `s` prefix to standard for `static` variable.

wgtdkp (approved):
LGTM :+1: 

Thanks!

src/agent/main.cpp
superwhd, line 1: A comment for calling `getopt_long` at line 209: Should we also add `s` into the opt string (currently "B:d:hI:Vv")?

src/agent/main.cpp
```diff
@@ -88,6 +89,7 @@ static const struct option kOptions[] = {
     {"help", no_argument, nullptr, OTBR_OPT_HELP},
     {"thread-ifname", required_argument, nullptr, OTBR_OPT_INTERFACE_NAME},
     {"verbose", no_argument, nullptr, OTBR_OPT_VERBOSE},
+    {"syslog-disable", no_argument, nullptr, OTBR_OPT_SYSLOG_DISABLE},
```
superwhd, line 92: nit: I'm inclined to making it a positive flag (i.e. enabling syslog) instead of a negative one. In this way it makes more sense to abbreviate it as `-s`.

src/common/logging.cpp
```diff
@@ -77,8 +78,14 @@ void otbrLogSetLevel(otbrLogLevel aLevel)
     sLevel = aLevel;
 }
 
+/** Enable/disable logging with syslog */
+void otbrLogDisableSyslog(bool aDisabled)
```
superwhd, line 82: This function is probably hard for code viewers to understand because it's the opposite direction of enabling according the flag. Probably you can rename it as `otbrLogSyslogSetEnabled` instead.

src/common/logging.cpp
```diff
@@ -150,7 +169,15 @@ void otbrLogv(otbrLogLevel aLevel, const char *aFormat, va_list aArgList)
 
 void otbrLogvNoFilter(otbrLogLevel aLevel, const char *aFormat, va_list aArgList)
 {
-    vsyslog(static_cast<int>(aLevel), aFormat, aArgList);
+    if (syslogDisabled)
+    {
+        vprintf(aFormat, aArgList);
+        printf("\n");
```
superwhd, line 175: I think in this way it implies printing logs to standard output when syslog is disabled. I'd suggest to document the behavior somewhere.

Also I wonder if logging to standard output should be controlled by an independent flag from whether logging to syslog? It's potentially useful to disable all logging outputs.

src/agent/main.cpp
greens: I felt the same, but didn't want to change existing functionality. Folks would have to go add `-s` to their invocations if they were relying on syslog presently. If you're okay with it being a "breaking" change I can do it, but figured it was best to avoid.

superwhd: I also agree to keep the existing behavior. But is it possible to make the flag default to be true?

That is, you need to pass in something like `-s false` to disable syslog. If you don't pass in `-s` option it will enable syslog.

I don't have a strong preference on this. For the simplicity of implementation I'm fine with the `syslog-disable` flag.

greens: I think in the interests of back compat as well as avoiding a (I feel) similarly clunky flag, that we ought to stick with this implementation.

superwhd:
Hi @greens, could you please address my comments?

src/common/logging.cpp
```diff
@@ -77,8 +78,14 @@ void otbrLogSetLevel(otbrLogLevel aLevel)
     sLevel = aLevel;
 }
 
+/** Enable/disable logging with syslog */
+void otbrLogSyslogSetEnabled(bool aDisabled)
```
superwhd, line 82:
```suggestion
void otbrLogSyslogSetEnabled(bool aEnabled)
```

Better to flip the meaning of this parameter to avoid confusion.

src/common/logging.cpp
superwhd: At least I think we should document the behavior (when syslog is off, write to stdout) in the comments otherwise later we'll have to read the code to understand what it does.

greens: Tried to do that here: https://github.com/openthread/ot-br-posix/blob/1e0a48f75320ca8056557827bb3962eada4eb8f1/src/agent/main.cpp#L142

I can put a comment around this as well.

superwhd (approved):
LGTM assuming the comment is resolved.

superwhd (approved):
LGTM 👍

jwhui (approved):

Issue #2116, opened 2023-12-05
[trel-dnssd] trel platform selects first peer address instead of wider scope one


abtink:
The trel platform implementation seem to select the first address (at index `[0]`) when invoking callback to reprot a discovered TREL peer info:
https://github.com/openthread/ot-br-posix/blob/4b826a19dcf1d2398c8112b2c92ae90aa3853167/src/trel_dnssd/trel_dnssd.cpp#L306-L310

The `platform/trel.h` documenation [mentions](
https://github.com/openthread/openthread/blob/19af7118b064bafefa9dffa8c1cfcfd6e3fad459/include/openthread/platform/trel.h#L74-L87):
```
...
 * events are not guaranteed, however. When a TREL service instance is discovered, a new ongoing DNS-SD query for an
 * AAAA record should be started on the hostname indicated in the SRV record of the discovered instance. If multiple
 * host IPv6 addressees are discovered for a peer, one with highest scope among all addresses MUST be reported (if
 * there are multiple address at same scope, one must be selected randomly).
 *
```

This behavior seems to be missing in the current implementation.

---

Thanks To @gabekassel for noticing that different peers seems to be using a mixture of ULA/GUA addresses:
```
Dec  4 18:21:49.145 eero local5.debug otbr-agent[666778]: [D] Cli-----------: Input: trel peers
Dec  4 18:21:49.145 eero local5.debug otbr-agent[666778]: [D] Cli-----------: Output: | No  | Ext MAC Address  | Ext PAN Id       | IPv6 Socket Address                              |
Dec  4 18:21:49.145 eero local5.debug otbr-agent[666778]: [D] Cli-----------: Output: +-----+------------------+------------------+--------------------------------------------------+
Dec  4 18:21:49.145 eero local5.debug otbr-agent[666778]: [D] Cli-----------: Output: |   1 | f2295d25707238d4 | dcbc1020cee5f6af | [2603:7000:c00:a315:8:3e3:857f:e5d4]:59248       |
Dec  4 18:21:49.145 eero local5.debug otbr-agent[666778]: [D] Cli-----------: Output: |   2 | be860a24c6976a2c | dcbc1020cee5f6af | [fdbc:1020:cee5:f6af:829:8a3c:7f40:396a]:57947   |
[...]

PR #2118, opened 2023-12-05
[trel-dnssd] prioritize wider-scope IPv6 address of TREL peers

abtink:
This commit updates `TrelDnssd` to select the numerically smallest IPv6 address when a peer advertises multiple IPv6 addresses. This prioritizes globally-unique addresses (GUA) over unique-local addresses (ULA) with `fc00::/7` prefix, followed by link-local addresses (`fe80::/10`).

----

Should address https://github.com/openthread/ot-br-posix/issues/2116.


librasungirl (approved):
👍

superwhd (approved):
LGTM regarding the logic of selecting address.

However, I think this alone may not be able to resolve the issue. Today in OTBR the mDNS resolvers won't return multiple addresses in one callback. When a host has multiple addresses, mDNS resolvers will trigger a callback for every address. 

To fully fix the issue, we may also need to enhance the mDNS resolvers to make it return all resolved addresses at one time.

PR #2135, opened 2023-12-14
[dbus] add OTBR_PACKAGE_VERSION

jdswensen:
Fetch the `otbr` package version via `dbus`


bukepo (approved):

src/dbus/server/dbus_thread_object.cpp
```diff
@@ -1499,6 +1501,17 @@ void DBusThreadObject::RegisterGetPropertyHandler(const std::string         &aIn
     mGetPropertyHandlers[aPropertyName] = aHandler;
 }
 
+otError DBusThreadObject::GetOtbrVersionHandler(DBusMessageIter &aIter)
+{
+    otError     error   = OT_ERROR_NONE;
+    std::string version = OTBR_PACKAGE_VERSION;
+
+    VerifyOrExit(DBusMessageEncodeToVariant(&aIter, version) == OTBR_ERROR_NONE, error = OT_ERROR_FAILED);
```
superwhd, line 1509: nit: I think this can be simplified by `SuccessOrExit`

superwhd (approved):
LGTM with a nit

src/dbus/server/dbus_thread_object.cpp
jdswensen: Change made!

Are there situations in which `SuccessOrExit` is preferred over `VerifyOrExit`? Or is there a push to move towards using `SuccessOrExit` in general when `OTBR_ERROR_NONE` is expected?

superwhd: I think `SuccessOrExit` is preferred when the checked value is an `OTBR_ERROR_*`.

jwhui (approved):
Thanks! 👍🏼

PR #2142, opened 2023-12-20
[telemetry] add protobuf for upstream DNS metrics

sherysheng:
Add the counters of queries, responses, failures handled by upstream DNS server

### DIFF
--- a/src/proto/thread_telemetry.proto
+++ b/src/proto/thread_telemetry.proto
@@ -357,6 +357,15 @@ message TelemetryData {
 
     // The number of other responses
     optional uint32 other_count = 6;
+
+    // The number of queries handled by Upstream DNS server.
+    optional uint32 upstream_dns_queries = 7;
+
+    // The number of responses handled by Upstream DNS server.
+    optional uint32 upstream_dns_responses = 8;
+
+    // The number of upstream DNS failures.
+    optional uint32 upstream_dns_failures = 9;
   }
 
   message DnsServerInfo {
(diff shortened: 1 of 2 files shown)


src/dbus/server/dbus_thread_object.cpp
```diff
@@ -1394,6 +1394,10 @@ otError DBusThreadObject::GetDnssdCountersHandler(DBusMessageIter &aIter)
 
     dnssdCounters.mResolvedBySrp = otDnssdCounters.mResolvedBySrp;
```
morningboata, line 1396: If you want to change the DBus interface, you need to modify the DBus interface definition at src/dbus/server/introspect.xml, DnssdCounters type. Double check if the client can talk correctly.
If you don't want to change the DBus interface, avoid the content serialization changes.

sherysheng: I need to add three more metrics here, seems that I have to change the struct ?

    <!-- DnssdCounters: The DNS-SD counters
    <literallayout>
        struct {
          uint32 success
          uint32 server_failure
          uint32 format_error
          uint32 name_error
          uint32 not_implemented
          uint32 other
          uint32 resolved_by_srp

erjiaqing: I would suggest not do to so, or in separate PRs.

DBus clients will not work if this is changed unless the users updated their code.

sherysheng: Thanks! Got it

morningboata (approved):

sherysheng:
Depends-on https://github.com/openthread/openthread/pull/9729

erjiaqing (approved):
Thanks!

Irving-cl (approved):
LGTM 👍

abtink (approved):
👍

jwhui (approved):

PR #2148, opened 2023-12-27
[mdns] fixes of mDNS subscriptions

superwhd:
This PR covers following things:
- Previously we didn't handle the removal of addresses properly when resolving a host/service. This causes an issue that when there's an ongoing browse of a specific service type (`_trel._udp`, for example), the callback of `OnServiceResolved` will also give stale addresses of the resolved host.
- In mDNSResponder implementation, we used a flag `kDNSServiceFlagsTimeout` which prevents from doing an ongoing service instance resolution. I removed this flag so that it does an ongoing resolution. The new behavior aligns better with the `Mdns::Publisher`'s API definition.
- In mDNSResponder implementation, don't remove the service instance resolution after an address is resolved. This allows the implementation to return multiple addresses.
- Added a unit test for mDNS subscribers.

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -197,6 +197,20 @@ void Publisher::OnServiceResolved(std::string aType, DiscoveredInstanceInfo aIns
                 aInstanceInfo.mRemoved ? "remove" : "add", aInstanceInfo.mName.c_str(), aInstanceInfo.mHostName.c_str(),
                 aInstanceInfo.mAddresses.size());
 
+    if (!aInstanceInfo.mRemoved)
+    {
+        std::string addressesString;
+        for (const auto &address : aInstanceInfo.mAddresses)
+        {
+            addressesString += address.ToString() + ",";
+        }
+        if (addressesString.size())
+        {
+            addressesString.pop_back();
+        }
+        otbrLogInfo("addresses: [ %s ]", addressesString.c_str());
+    }
+
     DnsUtils::CheckServiceNameSanity(aType);
 
     assert(aInstanceInfo.mNetifIndex > 0);
@@ -616,6 +630,20 @@ void Publisher::UpdateHostResolutionEmaLatency(const std::string &aHostName, otb
         uint32_t latency = std::chrono::duration_cast<Milliseconds>(Clock::now() - it->second).count();
         UpdateEmaLatency(mTelemetryInfo.mHostResolutionEmaLatency, latency, aError);
         mHostResolutionBeginTime.erase(it);
+    }
+}
+
+void Publisher::AddAddress(AddressList &aAddressList, const Ip6Address &aAddress)
+{
+    aAddressList.push_back(aAddress);
+}
+
+void Publisher::RemoveAddress(AddressList &aAddressList, const Ip6Address &aAddress)
+{
+    auto it = std::find(aAddressList.begin(), aAddressList.end(), aAddress);
+    if (it != aAddressList.end())
+    {
+        aAddressList.erase(it);
     }
 }
 
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -200,6 +200,7 @@ void Publisher::OnServiceResolved(std::string aType, DiscoveredInstanceInfo aIns
     if (!aInstanceInfo.mRemoved)
     {
         std::string addressesString;
+
         for (const auto &address : aInstanceInfo.mAddresses)
         {
             addressesString += address.ToString() + ",";
@@ -641,6 +642,7 @@ void Publisher::AddAddress(AddressList &aAddressList, const Ip6Address &aAddress
 void Publisher::RemoveAddress(AddressList &aAddressList, const Ip6Address &aAddress)
 {
     auto it = std::find(aAddressList.begin(), aAddressList.end(), aAddress);
+
     if (it != aAddressList.end())
     {
         aAddressList.erase(it);
--- a/src/mdns/mdns.hpp
+++ b/src/mdns/mdns.hpp
@@ -136,6 +136,9 @@ public:
         uint16_t    mWeight   = 0;       ///< Service weight.
         TxtData     mTxtData;            ///< TXT RDATA bytes.
         uint32_t    mTtl = 0;            ///< Service TTL.
+
+        void AddAddress(const Ip6Address &aAddress) { Publisher::AddAddress(mAddresses, aAddress); }
+        void RemoveAddress(const Ip6Address &aAddress) { Publisher::RemoveAddress(mAddresses, aAddress); }
     };
 
     /**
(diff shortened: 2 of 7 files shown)


src/mdns/mdns_avahi.cpp
```diff
@@ -1309,8 +1310,18 @@ void PublisherAvahi::ServiceResolver::HandleResolveHostResult(AvahiRecordBrowser
...
+        mInstanceInfo.mAddresses.push_back(std::move(address));
+    }
+    else
+    {
+        auto it = std::find(mInstanceInfo.mAddresses.begin(), mInstanceInfo.mAddresses.end(), address);
+        if (it != mInstanceInfo.mAddresses.end())
+        {
+            mInstanceInfo.mAddresses.erase(it);
+        }
+    }
```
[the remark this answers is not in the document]
superwhd: I'm not sure what you meant here. I've added this logic in `ServiceResolver::HandleResolveHostResult` and `HostSubscription::HandleResolveResult`. Are there any other places we should add it?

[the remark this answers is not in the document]
abtink: You are right. Sorry. I missed the `HostSubscription` one. All looks good. 👍

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1056,7 +1043,7 @@ otbrError PublisherMDnsSd::ServiceInstanceResolution::GetAddrInfo(uint32_t aInte
 
     otbrLogInfo("DNSServiceGetAddrInfo %s inf %d", mInstanceInfo.mHostName.c_str(), aInterfaceIndex);
 
-    dnsError = DNSServiceGetAddrInfo(&mServiceRef, kDNSServiceFlagsTimeout, aInterfaceIndex,
+    dnsError = DNSServiceGetAddrInfo(&mServiceRef, /* flags */ 0, aInterfaceIndex,
```
sunytt, line 1046: Looks like the subscription will not stop now? Should we pass the mServiceRef to DNSServiceRefDeallocate() when unsubscribing service?

sunytt (commented):
Thanks, maybe not related to this PR, just adding as a note: in `PublisherMDnsSd::SubscribeService` it's possible to use empty aInstancName to start a browse, but in `PublisherMDnsSd::UnsubscribeService` when aInstanceName is empty, no subscribed service will be erased.

src/mdns/mdns_mdnssd.cpp
superwhd: When unsubscribing a service, the `ServiceInstanceResolution` will be destructed and its base class `ServiceRef` will be destructed. The destructor of `ServiceRef` calls `DNSServiceRefDeallocate()`.

sunytt: I see, then the PublisherMDnsSd::UnsubscribeService with empty aInstanceName does seem related to this PR, as it is not destructing the SubscribedServices thus leaking the ServiceRefs?

superwhd: Explained via offline chat. The short answer is, the `DNSServiceRefDeallocate()` is triggered on the destruction of relevant objects like `ServiceSubscription` and `ServiceInstanceResolution`.

src/mdns/mdns.hpp
```diff
@@ -148,6 +151,9 @@ class Publisher : private NonCopyable
         AddressList mAddresses;      ///< IP6 addresses.
         uint32_t    mNetifIndex = 0; ///< Network interface.
         uint32_t    mTtl        = 0; ///< Host TTL.
+
+        void AddAddress(Ip6Address aAddress) { Publisher::AddAddress(mAddresses, std::move(aAddress)); }
+        void RemoveAddress(const Ip6Address &aAddress) { Publisher::RemoveAddress(mAddresses, aAddress); }
```
abtink, line 156: I noticed `AddAddress()` expects a `Ip6Address aAddress` input but `RemoveAddress()` a `const` reference as `const Ip6Address &aAddress`.

Any reason for this? Can they both get `const Ip6Address &aAddress`?

abtink (commented):
Thanks @superwhd! Looks great overall. 👍 
One question below.

src/mdns/mdns.hpp
superwhd: `AddAddress` takes the parameter as value so it can leverage the move constructor of `Ip6Address` to save copying. Although as of today, `Ip6Address`'s move constructor does the same as its copy constructor, it'll save some copying if some day `Ip6Address` stores some data on the heap.

abtink: Thanks.  Since `Ip6Address` today does not have a move constructor, it is always copied (it seems to follow value semantics as a `union` of 16 bytes). 

So if I understand correctly, with the use of `Ip6Address aAddress` as input, we may end up doing 3 copies of `Ip6Address`:
- When `AddAddress()` is first called, we can have one copy for `Ip6Address aAddress`. 
- Then when we call the `static` `Publisher::AddAddress()` we can have another copy.
- And finally when we `push_back` into the list.
(compiler may optimize some of this, but it does not have to).

I think if we use `const Ip6Address &aAddress`, we end up doing one copy (when we `push_back`)?
If we add move semantics later for `Ip6Address` (which I think probably unlikely?), we can optimize the code then by providing a new flavor of `AddAddress(Ip6Address &&aAddress)` working with `&&` reference?

superwhd: Thanks for pointing it out! I think you're right that `std::move` doesn't fit well here. I've updated the code to pass in const reference.

sunytt (approved):
LGTM 👍  Thanks!

abtink (approved):
LGTM. Thanks. 👍

src/mdns/mdns.cpp
```diff
@@ -197,6 +197,20 @@ void Publisher::OnServiceResolved(std::string aType, DiscoveredInstanceInfo aIns
                 aInstanceInfo.mRemoved ? "remove" : "add", aInstanceInfo.mName.c_str(), aInstanceInfo.mHostName.c_str(),
                 aInstanceInfo.mAddresses.size());
 
+    if (!aInstanceInfo.mRemoved)
+    {
+        std::string addressesString;
+        for (const auto &address : aInstanceInfo.mAddresses)
```
jwhui, line 203: nit: empty line after declaration
```suggestion
        std::string addressesString;

        for (const auto &address : aInstanceInfo.mAddresses)
```

src/mdns/mdns.cpp
```diff
@@ -619,5 +633,19 @@ void Publisher::UpdateHostResolutionEmaLatency(const std::string &aHostName, otb
...
 
+void Publisher::AddAddress(AddressList &aAddressList, const Ip6Address &aAddress)
+{
+    aAddressList.push_back(aAddress);
+}
+
+void Publisher::RemoveAddress(AddressList &aAddressList, const Ip6Address &aAddress)
+{
+    auto it = std::find(aAddressList.begin(), aAddressList.end(), aAddress);
+    if (it != aAddressList.end())
```
jwhui, line 644: nit: empty line after declaration
```suggestion
    auto it = std::find(aAddressList.begin(), aAddressList.end(), aAddress);

    if (it != aAddressList.end())
```

jwhui (approved):

PR #2173, opened 2024-01-27
[android] exit ot-daemon when factoryreset

wgtdkp:
Existing implementation will re-exec the ot-daemon process itself by leveraging the execvp() system call. This specifically doesn't work on Android for reasons:
1. ot-daemon on Android is managed / controlled by the system_server service which creates TUN FD and manages the enabled states that are essential for the ot-daemon to run. So ot-daemon can't re-start on on its own.
2. execvp() will result in lost of runtime callbacks registered from system_server and put the system into an inconsistent state.

It simplifies the system if the lifecycle of ot-daemon is execlusively managed by the system_server. This also allows us to do optimizations such as tearing down the ot-daemon server to save memory and power when Thread is stopped or disabled.

This commit fixes the the sepolicy denied issues (ot-daemon is not allowed to exec itself) when "factoryreset" is executed and follows the design abstracted above by exiting the program. The system_server will be signaled the exiting of ot-daemon and restart the daemon with essential resources and registering callbacks.

Note that even with the fix in this commit, "factoryreset" is not used in prod but only with "ot-ctl" for development and certification.

### DIFF
--- a/src/agent/main.cpp
+++ b/src/agent/main.cpp
@@ -78,7 +78,9 @@ enum
     OTBR_OPT_REST_LISTEN_PORT,
 };
 
-static jmp_buf            sResetJump;
+#ifndef __ANDROID__
+static jmp_buf sResetJump;
+#endif
 static otbr::Application *gApp = nullptr;
 
 void                       __gcov_flush();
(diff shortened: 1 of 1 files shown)


superwhd (approved):
LGTM 👍

bukepo (approved):

jwhui (approved):

PR #2176, opened 2024-01-30
[trel] ignore anycast address as trel peer

mia1yang:


librasungirl (approved):

superwhd (approved):
LGTM 👍

src/trel_dnssd/trel_dnssd.cpp
```diff
@@ -297,6 +297,12 @@ void TrelDnssd::OnTrelServiceInstanceAdded(const Mdns::Publisher::DiscoveredInst
     {
         otbrLogDebug("Peer address: %s", addr.ToString().c_str());
 
+        // Skip anycast (Refer to https://datatracker.ietf.org/doc/html/rfc2373#section-2.6.1)
+        if (addr.m64[1] == 0)
```
abtink, line 301: >```
>   |                         n bits                 |   128-n bits   |
>   +------------------------------------------------+----------------+
>   |                   subnet prefix                | 00000000000000 |
>   +------------------------------------------------+----------------+
>```

We are assuming the subnet prefix `n` will be `<= 64` (so the `m64[1]` will zero). Is `n>64` possible?

abtink (approved):
LGTM.
One question below:

src/trel_dnssd/trel_dnssd.cpp
librasungirl: Generally on-link prefix allocated are of len 64 I think. we may have this assumption for now and take care of it if it happens somehow?

bukepo (approved):

src/trel_dnssd/trel_dnssd.cpp
```diff
@@ -297,6 +297,12 @@ void TrelDnssd::OnTrelServiceInstanceAdded(const Mdns::Publisher::DiscoveredInst
...
+        {
+            continue;
+        }
+
         // If there are multiple addresses, we prefer the address
```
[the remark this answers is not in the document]
abtink: yes

jwhui (approved):

PR #2198, opened 2024-02-22
[openwrt] remove usage of `OT_NETWORK_DIAGNOSTIC_TYPELIST_MAX_ENTRIES`

jwhui:
To address changes in https://github.com/openthread/openthread/pull/9866


abtink (approved):
Thanks for fixing this. 👍 
Did not expect/know that this constant was used in other modules.

PR #2200, opened 2024-02-25
[meshcop] support runtime custom vendor and product name

wgtdkp:
Existing implementation will publish the _meshcop._udp service
immediately when the program is started, even when there is no
provisioned Thread network. This works as expected as the Thread spec
defines.
    
The problem is that for non-Android linux platforms, the vendor,
product name are customized via build time flags but customization needs
to be runtime for Android because all Android device shares the same
Thread stack in the mainline module.
    
To support Android, we make updates to allow delaying the advertisement
of _meshcop._udp service after the customized vendor/product name has
been provided via the new "OTBR_STOP_BORDER_AGENT_ON_INIT" build flag
and the "BorderAgent::SetMeshcopServiceValues()" method.
    
Check https://android-review.git.corp.google.com/c/platform/packages/modules/Connectivity/+/2993930
for how this change is integrated with Android.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -104,7 +104,13 @@ void Application::Init(void)
     mPublisher->Start();
 #endif
 #if OTBR_ENABLE_BORDER_AGENT
+// This is for delaying publishing the MeshCoP service until the correct
+// vendor name and OUI etc. are correctly set by BorderAgent::SetMeshCopServiceValues()
+#if OTBR_STOP_BORDER_AGENT_ON_INIT
+    mBorderAgent.SetEnabled(false);
+#else
     mBorderAgent.SetEnabled(true);
+#endif
 #endif
 #if OTBR_ENABLE_BACKBONE_ROUTER
     mBackboneAgent.Init();
--- a/src/border_agent/border_agent.cpp
+++ b/src/border_agent/border_agent.cpp
@@ -69,11 +69,7 @@
 
 namespace otbr {
 
-static const char kVendorName[]             = OTBR_VENDOR_NAME;
-static const char kProductName[]            = OTBR_PRODUCT_NAME;
-static const char kBorderAgentServiceType[] = "_meshcop._udp"; ///< Border agent service type of mDNS
-static const char kBorderAgentServiceInstanceName[] =
-    OTBR_MESHCOP_SERVICE_INSTANCE_NAME; ///< Border agent service name of mDNS
+static const char    kBorderAgentServiceType[]    = "_meshcop._udp"; ///< Border agent service type of mDNS
 static constexpr int kBorderAgentServiceDummyPort = 49152;
 
 /**
@@ -143,7 +139,31 @@ BorderAgent::BorderAgent(otbr::Ncp::ControllerOpenThread &aNcp, Mdns::Publisher 
     : mNcp(aNcp)
     , mPublisher(aPublisher)
     , mIsEnabled(false)
+    , mVendorName(OTBR_VENDOR_NAME)
+    , mProductName(OTBR_PRODUCT_NAME)
+    , mBaseServiceInstanceName(OTBR_MESHCOP_SERVICE_INSTANCE_NAME)
 {
+}
+
+otbrError BorderAgent::SetMeshCopServiceValues(const std::string          &aServiceInstanceName,
+                                               const std::string          &aProductName,
+                                               const std::string          &aVendorName,
+                                               const std::vector<uint8_t> &aVendorOui)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(aProductName.size() <= kMaxProductNameLength, error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aVendorName.size() <= kMaxVendorNameLength, error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aVendorOui.empty() || aVendorOui.size() == kVendorOuiLength, error = OTBR_ERROR_INVALID_ARGS);
+
+    mProductName = aProductName;
+    mVendorName  = aVendorName;
+    mVendorOui   = aVendorOui;
+
+    mBaseServiceInstanceName = aServiceInstanceName;
+
+exit:
+    return error;
 }
 
 void BorderAgent::SetEnabled(bool aIsEnabled)
@@ -179,7 +199,7 @@ void BorderAgent::Start(void)
     });
 #endif
 
-    mServiceInstanceName = BaseServiceInstanceName();
+    mServiceInstanceName = GetServiceInstanceNameWithExtAddr(mBaseServiceInstanceName);
     UpdateMeshCopService();
 }
 
@@ -360,8 +380,18 @@ void BorderAgent::PublishMeshCopService(void)
     }
 #endif
 
-    txtList.emplace_back("vn", kVendorName);
-    txtList.emplace_back("mn", kProductName);
+    if (!mVendorOui.empty())
+    {
+        txtList.emplace_back("vo", mVendorOui.data(), mVendorOui.size());
+    }
+    if (!mVendorName.empty())
+    {
+        txtList.emplace_back("vn", mVendorName.c_str());
+    }
+    if (!mProductName.empty())
+    {
+        txtList.emplace_back("mn", mProductName.c_str());
+    }
     txtList.emplace_back("nn", networkName);
     txtList.emplace_back("xp", extPanId->m8, sizeof(extPanId->m8));
     txtList.emplace_back("tv", mNcp.GetThreadVersion());
@@ -465,6 +495,8 @@ void BorderAgent::HandleUpdateVendorMeshCoPTxtEntries(std::map<std::string, std:
 
 void BorderAgent::HandleThreadStateChanged(otChangedFlags aFlags)
 {
+    VerifyOrExit(IsEnabled());
+
     if (aFlags & OT_CHANGED_THREAD_ROLE)
     {
         otbrLogInfo("Thread is %s", (IsThreadStarted() ? "up" : "down"));
@@ -475,6 +507,9 @@ void BorderAgent::HandleThreadStateChanged(otChangedFlags aFlags)
     {
         UpdateMeshCopService();
     }
+
+exit:
+    return;
 }
 
 bool BorderAgent::IsThreadStarted(void) const
@@ -484,12 +519,12 @@ bool BorderAgent::IsThreadStarted(void) const
     return role == OT_DEVICE_ROLE_CHILD || role == OT_DEVICE_ROLE_ROUTER || role == OT_DEVICE_ROLE_LEADER;
 }
 
-std::string BorderAgent::BaseServiceInstanceName() const
+std::string BorderAgent::GetServiceInstanceNameWithExtAddr(const std::string &aServiceInstanceName) const
 {
     const otExtAddress *extAddress = otLinkGetExtendedAddress(mNcp.GetInstance());
     std::stringstream   ss;
 
-    ss << kBorderAgentServiceInstanceName << " #";
+    ss << aServiceInstanceName << " #";
     ss << std::uppercase << std::hex << std::setfill('0');
     ss << std::setw(2) << static_cast<int>(extAddress->m8[6]);
     ss << std::setw(2) << static_cast<int>(extAddress->m8[7]);
@@ -504,7 +539,7 @@ std::string BorderAgent::GetAlternativeServiceInstanceName() const
     uint16_t                                rand = uniform_dist(engine);
     std::stringstream                       ss;
 
-    ss << BaseServiceInstanceName() << " (" << rand << ")";
+    ss << GetServiceInstanceNameWithExtAddr(mBaseServiceInstanceName) << " (" << rand << ")";
     return ss.str();
 }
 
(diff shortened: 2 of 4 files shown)


src/border_agent/border_agent.cpp
```diff
@@ -143,7 +141,39 @@ BorderAgent::BorderAgent(otbr::Ncp::ControllerOpenThread &aNcp, Mdns::Publisher
     : mNcp(aNcp)
     , mPublisher(aPublisher)
     , mIsEnabled(false)
+    , mVendorName(OTBR_VENDOR_NAME)
+    , mProductName(OTBR_PRODUCT_NAME)
 {
+    setVendorAndProductName(OTBR_VENDOR_NAME, OTBR_PRODUCT_NAME);
+}
+
+void BorderAgent::setVendorAndProductName(const std::string &aVendorName, const std::string &aProductName)
```
superwhd, line 150: Should we also trigger `UpdateMeshCopService`?

src/border_agent/border_agent.cpp
superwhd, line 1: I think you also need to update line 393~394 using the `mVendorName` and `mProductName`.

src/agent/application.cpp
```diff
@@ -91,30 +91,39 @@ Application::Application(const std::string               &aInterfaceName,
...
     OTBR_UNUSED_VARIABLE(aRestListenAddress);
     OTBR_UNUSED_VARIABLE(aRestListenPort);
 }
 
-void Application::Init(void)
+void Application::onThreadEnabledChanged(bool aThreadEnabled)
+{
+    if (mThreadEnabled != aThreadEnabled)
+    {
+        mThreadEnabled = aThreadEnabled;
```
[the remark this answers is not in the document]
wgtdkp: Publisher is Thread-independent.

Added BackboneAgent handler here.

src/border_agent/border_agent.cpp
```diff
@@ -143,7 +141,39 @@ BorderAgent::BorderAgent(otbr::Ncp::ControllerOpenThread &aNcp, Mdns::Publisher
...
+    {
+        mProductName = aProductName.substr(0, 24);
+        otbrLogWarning("Product name is truncated to %s", mProductName.c_str());
+    }
+    else if (!aProductName.empty())
+    {
+        mProductName = aProductName;
+    }
+
+#ifdef OTBR_MESHCOP_SERVICE_INSTANCE_NAME
```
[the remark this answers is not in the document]
wgtdkp: Cmake users are supposed to use the `OTBR_MESHCOP_SERVICE_INSTANCE_NAME` build option to customize the instance name. See https://github.com/openthread/ot-br-posix/blob/58577e5eb19a71c0842d8b5f62a4a30620fc2070/CMakeLists.txt#L40

src/border_agent/border_agent.cpp
wgtdkp: This method is supposed to be called only before Thread is enabled. See the documentation of this API

src/border_agent/border_agent.cpp
wgtdkp: Good catch, updated

src/border_agent/border_agent.hpp
```diff
@@ -94,6 +96,24 @@ class BorderAgent : private NonCopyable
...
+    /**
+     * Sets the predefined product name, vendor name and vendor OUI that are advertised in the MeshCoP service.
+     *
+     * This method should be called before this BorderAgent is enabled by SetEnabled.
+     *
+     * @param[in] aProductName  The product name; must not exceed length of kMaxProductNameLength
+     *                          and an empty string will be ignored
+     * @param[in] aVendorName   The vendor name; must not exceed length of kMaxVendorNameLength
+     *                          and an empty string will be ignored
+     * @param[in] aVendorOui    The vendor OUI; must has length of 3 bytes or be empty and ignored
```
superwhd, line 108:
```suggestion
     * @param[in] aVendorOui    The vendor OUI; must have length of 3 bytes or be empty and ignored
```

src/border_agent/border_agent.cpp
```diff
@@ -504,7 +543,7 @@ std::string BorderAgent::GetAlternativeServiceInstanceName() const
     uint16_t                                rand = uniform_dist(engine);
     std::stringstream                       ss;
 
-    ss << BaseServiceInstanceName() << " (" << rand << ")";
+    ss << mServiceInstanceName << " (" << rand << ")";
```
superwhd, line 546: Note that `GetAlternativeServiceInstanceName()` could be called multiple times when generating the alternative name. This could result in appending the random suffix multiple times to the instance name.
```
// on conflict:
mServiceInstanceName = GetAlternativeServiceInstanceName();
```

src/agent/main.cpp
```diff
@@ -253,6 +256,17 @@ static int realmain(int argc, char *argv[])
                 enableAutoAttach = parseResult;
             }
             break;
+        case OTBR_OPT_THREAD_ENABLED:
+            if (optarg == nullptr)
+            {
+                threadEnabled = true;
```
superwhd, line 262: Shall we name it as `enableThread` to be consistent with `enableAutoAttach`?

src/agent/application.hpp
```diff
@@ -109,15 +109,23 @@ class Application : private NonCopyable
...
      *
      */
     void Deinit(void);
 
+    /**
+     * Called when the Thread enabled state is changed.
+     *
+     * @param[in] aThreadEnabled  A boolean indicates whether Thread is enabled or not
+     */
+    void onThreadEnabledChanged(bool aThreadEnabled);
```
superwhd, line 127:
```suggestion
    void OnThreadEnabledChanged(bool aThreadEnabled);
```

src/border_agent/border_agent.cpp
```diff
@@ -484,12 +516,19 @@ bool BorderAgent::IsThreadStarted(void) const
...
 
-std::string BorderAgent::BaseServiceInstanceName() const
+std::string BorderAgent::MakeServiceInstanceName() const
 {
     const otExtAddress *extAddress = otLinkGetExtendedAddress(mNcp.GetInstance());
     std::stringstream   ss;
+    std::string         instanceName = mVendorName + " " + mProductName;
+
+    // The service instance name can always be overriden by `OTBR_MESHCOP_SERVICE_INSTANCE_NAME`
+    if (strcmp(OTBR_MESHCOP_SERVICE_INSTANCE_NAME, DEFAULT_MESHCOP_SERVICE_INSTANCE_NAME) != 0)
```
superwhd, line 526: Do we really need `DEFAULT_MESHCOP_SERVICE_INSTANCE_NAME`? What about we just set the `instanceName` as `OTBR_MESHCOP_SERVICE_INSTANCE_NAME`?

src/border_agent/border_agent.cpp
```diff
@@ -143,7 +139,50 @@ BorderAgent::BorderAgent(otbr::Ncp::ControllerOpenThread &aNcp, Mdns::Publisher
...
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(aProductName.size() <= kMaxProductNameLength, error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aVendorName.size() <= kMaxVendorNameLength, error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(aVendorOui.empty() || aVendorOui.size() == kVendorOuiLength, error = OTBR_ERROR_INVALID_ARGS);
+
+    mProductName = aProductName;
+    mVendorName  = aVendorName;
+    mVendorOui   = aVendorOui;
+
```
superwhd, line 162: Why not reuse `SetPredefinedVendorTxtValues`?

src/border_agent/border_agent.cpp
```diff
@@ -484,12 +519,13 @@ bool BorderAgent::IsThreadStarted(void) const
     return role == OT_DEVICE_ROLE_CHILD || role == OT_DEVICE_ROLE_ROUTER || role == OT_DEVICE_ROLE_LEADER;
 }
 
-std::string BorderAgent::BaseServiceInstanceName() const
+std::string BorderAgent::GetServiceInstanceNameWithExtAddr(const std::string &aServiceInstanceName) const
 {
     const otExtAddress *extAddress = otLinkGetExtendedAddress(mNcp.GetInstance());
     std::stringstream   ss;
+    std::string         instanceName = mVendorName + " " + mProductName;
```
superwhd, line 526: This seems unnecessary.

src/border_agent/border_agent.cpp
wgtdkp: SetPredefinedVendorTxtValues is removed, it's a rebase error. Sorry for that

superwhd (approved):
LGTM 👍

bukepo (approved):
Not sure whether these info can be retrieved from Android platform.

src/agent/application.hpp
```diff
@@ -109,15 +109,23 @@ class Application : private NonCopyable
...
      *
      */
     void Deinit(void);
 
+    /**
+     * Called when the Thread enabled state is changed.
+     *
+     * @param[in] aThreadEnabled  A boolean indicates whether Thread is enabled or not
+     */
+    void OnThreadEnabledChanged(bool aThreadEnabled);
```
abtink, line 127: `OnThreadEnabledChanged()` is named as if signaling an event with `aThreadEnabled` input. 
- The  `aThreadEnabled` name itself may a bit confusing.  
- E.g., we have `otThreadSetEnabled()` which I think is not related to this.

If I understand the intention, the idea here is for next layer code (e.g. Android) to use this after it has set vendor/product name so we can start advertising "meshcop" service in AIL (i.e. we are ready to enable the "meshcop" agent).

How about renaming this to make it clear? Maybe `SetMeshcopAgentEnabled()`?

src/agent/application.cpp
```diff
@@ -91,29 +91,45 @@ Application::Application(const std::string               &aInterfaceName,
...
-    mBackboneAgent.Init();
+    mBorderAgent.SetEnabled(mThreadEnabled);
 #endif
 #if OTBR_ENABLE_SRP_ADVERTISING_PROXY
-    mAdvertisingProxy.SetEnabled(true);
+    mAdvertisingProxy.SetEnabled(mThreadEnabled);
 #endif
 #if OTBR_ENABLE_DNSSD_DISCOVERY_PROXY
-    mDiscoveryProxy.SetEnabled(true);
+    mDiscoveryProxy.SetEnabled(mThreadEnabled);
```
abtink, line 117: Is it safe/okay to possibly delay the call to `SetEnabled()` for `mAdvertisingProxy`, `mDiscoveryProxy` and `mBackboneAgent`?

Can we have the OT core stack trying to interact with these platform modules and cause issue if they are not enabled?

abtink: Thinking more on this: how about we limit this only for `mBorderAgent.SetEnabled()` (which I think the main use-case here)? And still enable other components from `Init()`?

abtink:
> The problem is that for non-Android linux platforms, the vendor,
> product name are customized via build time flags but customization needs
> to be runtime for Android because all Android device shares the same
> Thread stack in the mainline module.

An idea for a possibly simpler way.
- We can change `BorderAgent` so that it waits until `SetMeshcopServiceValues()` is called before it will publish the "meshcop" service (i.e. add checks in `PublishMeshCopService()` for this).
- We can have this behavior be defined under a build config flag (so we can enable it for Andriod but not other platform)
- This way we don't need to add a new method in `Application` to delay the call to `BorderAgent::SetEnabled()`. The `SetMeshcopServiceValues()` call itself can act as the trigger.

Thoughts?

src/agent/application.hpp
wgtdkp: Yeah, the Android level Thread enabled state is a bit different with `otThreadSetEnabled()`: When Thread is disabled in Android, apps are not allowed to call majority of the Android Thread APIs (i.e. apps are returned error code ERROR_THREAD_DISABLED), while some of the APIs which don't operate over radio are still allowed.

wgtdkp:
> > The problem is that for non-Android linux platforms, the vendor,
> > product name are customized via build time flags but customization needs
> > to be runtime for Android because all Android device shares the same
> > Thread stack in the mainline module.
> 
> An idea for a possibly simpler way.
> 
> * We can change `BorderAgent` so that it waits until `SetMeshcopServiceValues()` is called before it will publish the "meshcop" service (i.e. add checks in `PublishMeshCopService()` for this).
> * We can have this behavior be defined under a build config flag (so we can enable it for Andriod but not other platform)
> * This way we don't need to add a new method in `Application` to delay the call to `BorderAgent::SetEnabled()`. The `SetMeshcopServiceValues()` call itself can act as the trigger.
> 
> Thoughts?

@abtink Thanks for thoughtful review! Though my initial intention is for supporting runtime enablement of more components, I changed this CL to focus on BorderAgent for simplicity (I added a build flag OTBR_START_BORDER_AGENT_WHEN_INIT). PTAL

abtink (approved):
LGTM. Thanks @wgtdkp  for changing the code. 👍

src/border_agent/border_agent.hpp
```diff
@@ -94,6 +94,27 @@ class BorderAgent : private NonCopyable
 
     ~BorderAgent(void) = default;
 
+    /**
+     * Overrides MeshCoP service (i.e. _meshcop._udp) instance name, product name , vendor name and vendor OUI.
+     *
+     * This method should be called before this BorderAgent is enabled by SetEnabled.
```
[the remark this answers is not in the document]
wgtdkp: Updated to "must"

jwhui (approved):

Issue #2207, opened 2024-02-29
otbr intermittently crashes when a thread sed tries to do a srv & txt query


mkardous-silabs:
**Describe the bug**
OTBR intermittently throws an assert when the thread sed tries to do a SRV & TXT query because the `SplitFullServiceName` function returns an error in `OnServiceDiscovered`.

**To Reproduce** Information to reproduce the behavior, including:

1. 42f98b27bc
2. Silabs EFR32MG24 BRD4187C as a Thread SED
3. Default build steps to build the otbr
4. OTBR as parent of a single SED device - no other devices on the thread network

**Expected behavior** 
OTBR should never crash. queryName seems to be "randomly" mangled which causes the assert to happen.

**Console/log output**
_GDB assert log_
```
otbr-agent: /home/ubuntu/ot-br-posix/src/sdp_proxy/discovery_proxy.cpp:200: void otbr::Dnssd::DiscoveryProxy::OnServiceDiscovered(const string&, const otbr::Mdns::Publisher::DiscoveredInstanceInfo&): Assertion `splitError == OTBR_ERROR_NONE' failed.

Thread 1 "otbr-agent" received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=281474841829408, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
44	./nptl/pthread_kill.c: No such file or directory.
```
_Stack Trace_

```
#0  __pthread_kill_implementation (threadid=281474841829408, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x0000fffff7aef254 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x0000fffff7aaa67c in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x0000fffff7a97130 in __GI_abort () at ./stdlib/abort.c:79
#4  0x0000fffff7aa3fd0 in __assert_fail_base (fmt=0xfffff7bbc550 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n",
    assertion=assertion@entry=0xaaaaaaf71378 "splitError == OTBR_ERROR_NONE",
[...]

PR #2208, opened 2024-02-29
[discovery-proxy] remove `assert` checks on query name parse errors

abtink:
Remove the `assert` checks on query name parse errors and instead skip over such entries. Query names are external input (received as DNS query message from other devices) and discovery proxy code should not assume that external input will follow the proper query name format.

---

- Suggested fix for https://github.com/openthread/ot-br-posix/issues/2207

### DIFF
--- a/src/sdp_proxy/discovery_proxy.cpp
+++ b/src/sdp_proxy/discovery_proxy.cpp
@@ -214,11 +214,9 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string                      
         {
         case OT_DNSSD_QUERY_TYPE_BROWSE:
             splitError = SplitFullServiceName(queryName, serviceName, domain);
-            assert(splitError == OTBR_ERROR_NONE);
             break;
         case OT_DNSSD_QUERY_TYPE_RESOLVE:
             splitError = SplitFullServiceInstanceName(queryName, instanceName, serviceName, domain);
-            assert(splitError == OTBR_ERROR_NONE);
             break;
         default:
             splitError = OTBR_ERROR_NOT_FOUND;
@@ -226,6 +224,7 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string                      
         }
         if (splitError != OTBR_ERROR_NONE)
         {
+            // Incoming service/instance was not what current query wanted to see, move on.
             continue;
         }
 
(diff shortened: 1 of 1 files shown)


wgtdkp (approved):
👍

src/sdp_proxy/discovery_proxy.cpp
```diff
@@ -284,8 +282,13 @@ void DiscoveryProxy::OnHostDiscovered(const std::string
         {
             continue;
         }
+
         splitError = SplitFullHostName(queryName, hostName, domain);
-        assert(splitError == OTBR_ERROR_NONE);
+
+        if (splitError != OTBR_ERROR_NONE)
+        {
+            continue;
```
wgtdkp, line 290: nit: it may worth a debug or info log for triage

abtink: Can be a good idea but may also extra logs (if there are many queries with invalid format).

I leave it to you or other folks to add logs in future PRs.

tcarmelveilleux: The issue is that if the client is doing a resolve, but we get a service published at the time, the path of:

```
       case OT_DNSSD_QUERY_TYPE_BROWSE:
            splitError = SplitFullServiceName(queryName, serviceName, domain);
```

will get hit, but that will (and should) always fail since the request was for an instance, not a service.

src/sdp_proxy/discovery_proxy.cpp
```diff
@@ -214,11 +214,9 @@ void DiscoveryProxy::OnServiceDiscovered(const std::string
...
         case OT_DNSSD_QUERY_TYPE_BROWSE:
             splitError = SplitFullServiceName(queryName, serviceName, domain);
-            assert(splitError == OTBR_ERROR_NONE);
             break;
         case OT_DNSSD_QUERY_TYPE_RESOLVE:
             splitError = SplitFullServiceInstanceName(queryName, instanceName, serviceName, domain);
-            assert(splitError == OTBR_ERROR_NONE);
             break;
         default:
             splitError = OTBR_ERROR_NOT_FOUND;
```
tcarmelveilleux, line 222: Please add a comment in the if below (hidden):
```
        if (splitError != OTBR_ERROR_NONE)
        {
            // Incoming service/instance was not what current query wanted to see, move on.
            continue;
        }
```

tcarmelveilleux (approved):

src/sdp_proxy/discovery_proxy.cpp
abtink: Added in new push.

jwhui (approved):

PR #2206, opened 2024-02-29
[mdns] allow `UnsubscribeService/Host()` to remove non-present entries

abtink:
This commit updates `UnsubscribeService/Host()` to allow them to be called with service or host name that may not be present in the list. This situation may happen if duplicate queries are received for the same names.

---

- Suggested fix for https://github.com/openthread/ot-br-posix/issues/2203.

PR #2209, opened 2024-02-29
[mdns-avahi] allow `UnsubscribeService/Host()` to remove non-present entries

abtink:
This commit adds a similar change to #2206 but for Avahi example. It updates `UnsubscribeService/Host()` to allow them to be called with service or host name that may not be present in the list. This situation may happen if duplicate queries are received for the same names.


superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2210, opened 2024-02-29
[mdns-avahi] handle potential unknown Avahi states gracefully

abtink:
This change avoids `assert(false)` on `AvahiEntryGroupState` and `AvahiClientState` `switch` statements. This prevents potential crashes if Avahi introduces new states in future versions.

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -570,10 +570,6 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
     case AVAHI_ENTRY_GROUP_UNCOMMITED:
     case AVAHI_ENTRY_GROUP_REGISTERING:
         break;
-
-    default:
-        assert(false);
-        break;
     }
 }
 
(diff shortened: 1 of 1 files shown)


bukepo (commented):
We already listed all enumeration values. I think the assert() here was to detect logic error or corruption. Do you think we should keep it here? Otherwise, the wrong state would be ignored.

src/mdns/mdns_avahi.cpp
```diff
@@ -570,10 +570,6 @@ void PublisherAvahi::HandleGroupState(AvahiEntryGroup *aGroup, AvahiEntryGroupSt
     case AVAHI_ENTRY_GROUP_UNCOMMITED:
     case AVAHI_ENTRY_GROUP_REGISTERING:
         break;
-
-    default:
-        assert(false);
```
superwhd, line 575: Should we log this error code? Otherwise we may hardly notice and take care of new error codes.

abtink:
> We already listed all enumeration values. I think the assert() here was to detect logic error or corruption. Do you think we should keep it here? Otherwise, the wrong state would be ignored.

`assert()` seems to be a drastic action on this. Particularly if new states are added in later Avahi versions we don't want our code to crash.

> Should we log this error code? Otherwise we may hardly notice and take care of new error codes.

Best way to detect this is to use complier check (code checkers) which validate that all enum values are covered in `switch()` statement (generate warning otherwise). If we add a `default` this wont be checked. Better to detect at build time rather than from logs at run-time.

I know such a build-time check is already enabled/used in OT core stack (and we treat all warnings as error). It may also be used by `ot-br-posix` already (if not, would be good to enable it).

jwhui (approved):

PR #2247, opened 2024-04-11
[border-agent] config related OpenThread core build flag according to OTBR_EPSKC option

mia1yang:
Test: 
when OTBR_EPSKC ON, CLI is enabled: 
> ba ephemeralkey
inactive
Done

when OTBR_EPSKC OFF, CLI is disabled:
> ba ephemeralkey
Error 35: InvalidCommand


third_party/openthread/CMakeLists.txt
```diff
@@ -76,6 +76,12 @@ if (OTBR_SRP_ADVERTISING_PROXY)
     set(OT_EXTERNAL_HEAP ON CACHE STRING "enable external heap" FORCE)
 endif()
 
+if (OTBR_ENABLE_EPSKC)
```
superwhd, line 79:
```suggestion
if (OTBR_EPSKC)
```

On the other hand, what about making a CMake option like `OT_EPSKC` in OT? In this way that you can use a `set` call to implement the logic.

mia1yang: Thanks! Accept.

librasungirl (approved):

third_party/openthread/CMakeLists.txt
```diff
@@ -114,3 +114,9 @@ if (NOT OT_THREAD_VERSION STREQUAL "1.1")
         "-DOPENTHREAD_CONFIG_TMF_PROXY_MLR_ENABLE=1"
     )
 endif()
+
+if (OTBR_EPSKC)
+    target_compile_definitions(ot-config INTERFACE "-DOPENTHREAD_CONFIG_BORDER_AGENT_EPHEMERAL_KEY_ENABLE=1" )
+else()
+    target_compile_definitions(ot-config INTERFACE "-DOPENTHREAD_CONFIG_BORDER_AGENT_EPHEMERAL_KEY_ENABLE=0" )
+endif()
```
bukepo, line 122: How about we force enabling it just like other features?

third_party/openthread/CMakeLists.txt
superwhd: > On the other hand, what about making a CMake option like OT_EPSKC in OT? In this way that you can use a set call to implement the logic.

Seems this is not done yet. Could you take a look?

third_party/openthread/CMakeLists.txt
mia1yang: How about we have the option here and set OTBR_EPSKC to ON by default? Then we have a method to turn on and off. What's more, keep consistency for OTBR and OT ePSKc feature enable/disable at the same time.

third_party/openthread/CMakeLists.txt
abtink: Took care of it in https://github.com/openthread/openthread/pull/10034.

mia1yang: Thanks for adding the above CL

etc/cmake/options.cmake
```diff
@@ -108,7 +108,7 @@ if(OTBR_TREL)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_TREL=1)
 endif()
 
-option(OTBR_EPSKC "Enable ePSKc Support" OFF)
+option(OTBR_EPSKC "Enable ePSKc Support" ON)
```
librasungirl, line 111: nit for consistency

```suggestion
option(OTBR_EPSKC "Enable ephemeral PSKc" ON)
```

librasungirl (approved):

etc/cmake/options.cmake
mia1yang: Thanks! Accept.

librasungirl (approved):

third_party/openthread/CMakeLists.txt
```diff
@@ -76,6 +76,10 @@ if (OTBR_SRP_ADVERTISING_PROXY)
     set(OT_EXTERNAL_HEAP ON CACHE STRING "enable external heap" FORCE)
 endif()
 
+if (OTBR_EPSKC)
+    set(OT_BORDER_AGENT_EPSKC ON CACHE STRING "enable border agent ephemeral PSKc" FORCE)
+endif()
```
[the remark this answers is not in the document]
bukepo: Let's try to make features runtime configurable instead of build time.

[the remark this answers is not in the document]
superwhd: +1 on the suggestion

> Let's try to make features runtime configurable instead of build time.

I don't see the necessity of doing it in this PR. Could you elaborate?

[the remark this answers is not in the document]
bukepo: the runtime configuration api doesn't need to be in this PR.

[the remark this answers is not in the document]
mia1yang: Accepted. Thanks~~

superwhd (approved):
LGTM 👍

third_party/openthread/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ set(OT_ANYCAST_LOCATOR ON CACHE STRING "enable anycast locator" FORCE)
 set(OT_BACKBONE_ROUTER ${OTBR_BACKBONE_ROUTER} CACHE STRING "Enable Backbone Router feature in OpenThread" FORCE)
 set(OT_BACKBONE_ROUTER_DUA_NDPROXYING ${OTBR_DUA_ROUTING} CACHE STRING "Configure DUA ND Proxy feature in OpenThread" FORCE)
 set(OT_BORDER_AGENT ON CACHE STRING "enable border agent" FORCE)
+set(OT_BORDER_AGENT_EPSKC ON CACHE STRING "enable border agent ephemeral PSKc" FORCE)
```
abtink, line 34: Not sure if we want to change the source files in `third_party` folder (from a PR in `ot-br-posix`).
I think it should be upstreamed from OT repo?

jwhui: This CMake file lives in `ot-br-posix`. The `openthread` submodule actually lives in `third_party/openthread/repo`. :)

abtink: I see. Got it. Thanks @jwhui

abtink (approved):

jwhui (approved):

PR #2254, opened 2024-04-16
Add reserved fields in thread telemetry

sherysheng:
Three upstream DNS metrics were added in wrong Counters in proto file. Add reserved field to avoid misusing.

### DIFF
--- a/src/proto/thread_telemetry.proto
+++ b/src/proto/thread_telemetry.proto
@@ -304,6 +304,9 @@ message TelemetryData {
 
     // The number of other responses
     optional uint32 other_count = 6;
+
+    // TODO: Remove reserved fields, use upstream_dns metrics in DnsServerResponseCounters.
+    reserved 7 to 9;
   }
 
   enum SrpServerState {


src/proto/thread_telemetry.proto
```diff
@@ -304,6 +304,11 @@ message TelemetryData {
 
     // The number of other responses
     optional uint32 other_count = 6;
+
+    // TODO: Remove below reserved fields after b/335091747 is fixed.
+    reserved 7;
```
morningboata, line 309: You can write as "reserved 7 to 9;" to simplify the syntax.

src/proto/thread_telemetry.proto
```diff
@@ -304,6 +304,10 @@ message TelemetryData {
 
     // The number of other responses
     optional uint32 other_count = 6;
+
+    // TODO: Remove below reserved fields after
+    // upstream_dns metrics correctly applied to all platforms.
```
superwhd, line 309: nit: Could they fit in the same line?

superwhd (approved):
LGTM 👍

sherysheng:
Currently these three fields are not using by other products, abandon this PR to avoid confusion

PR #2259, opened 2024-04-22
[border-agent] add `_meshcop-e` service for ePSKc mode

mia1yang:
Meshcop-e service is published when receiving DNS query for PTR records and BA is in the "ba ephemeralkey" active mode with the same instance name with Meshcop service and different UDP port number for the DTLS connection using ePSKc.
Test:
1. Active ePSKc mode using "ba ephemeralkey" commands, verified Meshcop-e package and service
2. DTLS connection could successfully set using ePSKc mode with Candidate Commissioner.
3. DTLS connection using ePSKc by inputing wrong ePSK in Candidate Commissioner, Meshcop-e service has been unpublished and no negative impacted with Meshcop service.
4. Meshcop-e service unpublished after timeout.

### DIFF
--- a/src/border_agent/border_agent.cpp
+++ b/src/border_agent/border_agent.cpp
@@ -69,8 +69,9 @@
 
 namespace otbr {
 
-static const char    kBorderAgentServiceType[]    = "_meshcop._udp"; ///< Border agent service type of mDNS
-static constexpr int kBorderAgentServiceDummyPort = 49152;
+static const char    kBorderAgentServiceType[]      = "_meshcop._udp";   ///< Border agent service type of mDNS
+static const char    kBorderAgentEpskcServiceType[] = "_meshcop-e._udp"; ///< Border agent ePSKc service
+static constexpr int kBorderAgentServiceDummyPort   = 49152;
 
 /**
  * Locators
@@ -202,12 +203,75 @@ void BorderAgent::Start(void)
 
     mServiceInstanceName = GetServiceInstanceNameWithExtAddr(mBaseServiceInstanceName);
     UpdateMeshCopService();
+
+    otBorderAgentSetEphemeralKeyCallback(mNcp.GetInstance(), BorderAgent::HandleEpskcStateChanged, this);
 }
 
 void BorderAgent::Stop(void)
 {
     otbrLogInfo("Stop Thread Border Agent");
     UnpublishMeshCopService();
+}
+
+void BorderAgent::HandleEpskcStateChanged(void *aContext)
+{
+    BorderAgent *borderAgent = static_cast<BorderAgent *>(aContext);
+
+    if (otBorderAgentIsEphemeralKeyActive(borderAgent->mNcp.GetInstance()))
+    {
+        borderAgent->PublishEpskcService();
+    }
+    else
+    {
+        borderAgent->UnpublishEpskcService();
+    }
+}
+
+void BorderAgent::PublishEpskcService()
+{
+    otInstance *instance = mNcp.GetInstance();
+    int         port     = otBorderAgentGetUdpPort(instance);
+
+    otbrLogInfo("Publish meshcop-e service %s.%s.local. port %d", mServiceInstanceName.c_str(),
+                kBorderAgentEpskcServiceType, port);
+
+    mPublisher.PublishService(/* aHostName */ "", mServiceInstanceName, kBorderAgentEpskcServiceType,
+                              Mdns::Publisher::SubTypeList{}, port, /* aTxtData */ {}, [this](otbrError aError) {
+                                  if (aError == OTBR_ERROR_ABORTED)
+                                  {
+                                      // OTBR_ERROR_ABORTED is thrown when an ongoing service registration is
+                                      // cancelled. This can happen when the meshcop-e service is being updated
+                                      // frequently. To avoid false alarms, it should not be logged like a real error.
+                                      otbrLogInfo("Cancelled previous publishing meshcop-e service %s.%s.local",
+                                                  mServiceInstanceName.c_str(), kBorderAgentEpskcServiceType);
+                                  }
+                                  else
+                                  {
+                                      otbrLogResult(aError, "Result of publish meshcop-e service %s.%s.local",
+                                                    mServiceInstanceName.c_str(), kBorderAgentEpskcServiceType);
+                                  }
+
+                                  if (aError == OTBR_ERROR_DUPLICATED)
+                                  {
+                                      // Try to unpublish current service in case we are trying to register
+                                      // multiple new services simultaneously when the original service name
+                                      // is conflicted.
+                                      // Potential risk that instance name is not the same with meshcop service.
+                                      UnpublishEpskcService();
+                                      mServiceInstanceName = GetAlternativeServiceInstanceName();
+                                      PublishEpskcService();
+                                  }
+                              });
+}
+
+void BorderAgent::UnpublishEpskcService()
+{
+    otbrLogInfo("Unpublish meshcop-e service %s.%s.local", mServiceInstanceName.c_str(), kBorderAgentEpskcServiceType);
+
+    mPublisher.UnpublishService(mServiceInstanceName, kBorderAgentEpskcServiceType, [this](otbrError aError) {
+        otbrLogResult(aError, "Result of unpublish meshcop-e service %s.%s.local", mServiceInstanceName.c_str(),
+                      kBorderAgentEpskcServiceType);
+    });
 }
 
 void BorderAgent::HandleMdnsState(Mdns::Publisher::State aState)
--- a/src/border_agent/border_agent.hpp
+++ b/src/border_agent/border_agent.hpp
@@ -149,6 +149,10 @@ private:
     std::string GetServiceInstanceNameWithExtAddr(const std::string &aServiceInstanceName) const;
     std::string GetAlternativeServiceInstanceName() const;
 
+    static void HandleEpskcStateChanged(void *aContext);
+    void        PublishEpskcService(void);
+    void        UnpublishEpskcService(void);
+
     otbr::Ncp::ControllerOpenThread &mNcp;
     Mdns::Publisher                 &mPublisher;
     bool                             mIsEnabled;


src/border_agent/border_agent.cpp
```diff
@@ -210,6 +215,63 @@ void BorderAgent::Stop(void)
...
+                                      // frequently. To avoid false alarms, it should not be logged like a real error.
+                                      otbrLogInfo("Cancelled previous publishing meshcop-e service %s.%s.local",
+                                                  mServiceInstanceName.c_str(), kBorderAgentEpskcServiceType);
+                                  }
+                                  else
+                                  {
+                                      otbrLogResult(aError, "Result of publish meshcop-e service %s.%s.local",
+                                                    mServiceInstanceName.c_str(), kBorderAgentEpskcServiceType);
+                                  }
+                                  if (aError == OTBR_ERROR_DUPLICATED)
```
wgtdkp, line 252: Nit: this `if` is not logically relevant to the `if {} else {}` above it, so better adding a new line to split the two parts clearerly.
```suggestion
                                  
                                  if (aError == OTBR_ERROR_DUPLICATED)
```

wgtdkp (approved):
LGTM :+1:

src/border_agent/border_agent.cpp
mia1yang: Thanks, accept.

src/border_agent/border_agent.cpp
```diff
@@ -210,6 +215,63 @@ void BorderAgent::Stop(void)
...
+    Mdns::Publisher::TxtData txtData;
+
+    otbrLogInfo("Publish meshcop-e service %s.%s.local. port %d", mServiceInstanceName.c_str(), kBorderAgentEpskcServiceType, port);
+
+    mPublisher.PublishService(/* aHostName */ "", mServiceInstanceName, kBorderAgentEpskcServiceType,
+                              Mdns::Publisher::SubTypeList{}, port, txtData, [this](otbrError aError) {
+                                  if (aError == OTBR_ERROR_ABORTED)
+                                  {
+                                      // OTBR_ERROR_ABORTED is thrown when an ongoing service registration is
+                                      // cancelled. This can happen when the meshcop service is being updated
```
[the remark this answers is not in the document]
mia1yang: Thanks! Accept.

src/border_agent/border_agent.cpp
```diff
@@ -210,6 +215,68 @@ void BorderAgent::Stop(void)
...
+    else
+    {
+        borderAgent->UnpublishEpskcService();
+    }
+}
+
+void BorderAgent::PublishEpskcService()
+{
+    otInstance              *instance = mNcp.GetInstance();
+    int                      port     = otBorderAgentGetUdpPort(instance);
```
superwhd, line 237: In spec the section `DNS-SD Service MeshCoP-ePSKc`

 ```The service MUST reside at a different UDP port number than the regular MeshCoP (_meshcop._udp) service.```

src/border_agent/border_agent.cpp
```diff
@@ -210,6 +215,68 @@ void BorderAgent::Stop(void)
...
+                                                    mServiceInstanceName.c_str(), kBorderAgentEpskcServiceType);
+                                  }
+
+                                  if (aError == OTBR_ERROR_DUPLICATED)
+                                  {
+                                      // Try to unpublish current service in case we are trying to register
+                                      // multiple new services simultaneously when the original service name
+                                      // is conflicted.
+                                      UnpublishEpskcService();
+                                      mServiceInstanceName = GetAlternativeServiceInstanceName();
```
superwhd, line 265: What you're doing here could potentially make the instance names of `_meshcop._udp` and `_meshcop-e._udp` different (at a very very low probability).

In the spec it says:
```
The <Instance> section of the DNS-SD Service Instance Name SHALL be chosen the same as the <Instance> section of the MeshCoP service.
```
 It's just a `SHALL` though. So just FYI.

src/border_agent/border_agent.cpp
```diff
@@ -210,6 +215,68 @@ void BorderAgent::Stop(void)
     UnpublishMeshCopService();
 }
 
+#if OTBR_ENABLE_EPSKC
+void BorderAgent::HandleEpskcStateChanged(void *aContext)
+{
+    BorderAgent *borderAgent = static_cast<BorderAgent *>(aContext);
+    if (otBorderAgentIsEphemeralKeyActive(borderAgent->mNcp.GetInstance()))
```
superwhd, line 224:
```suggestion
    BorderAgent *borderAgent = static_cast<BorderAgent *>(aContext);

    if (otBorderAgentIsEphemeralKeyActive(borderAgent->mNcp.GetInstance()))
```

src/border_agent/border_agent.cpp
```diff
@@ -202,6 +203,10 @@ void BorderAgent::Start(void)
 
     mServiceInstanceName = GetServiceInstanceNameWithExtAddr(mBaseServiceInstanceName);
     UpdateMeshCopService();
+
+#if OTBR_ENABLE_EPSKC
+    otBorderAgentSetEphemeralKeyCallback(mNcp.GetInstance(), BorderAgent::HandleEpskcStateChanged, this);
+#endif
```
superwhd, line 211: To avoid duplicate callbacks, I'd suggest to put this in constructor.

src/border_agent/border_agent.cpp
```diff
@@ -210,6 +217,68 @@ void BorderAgent::Stop(void)
...
+{
+    otInstance              *instance = mNcp.GetInstance();
+    int                      port     = otBorderAgentGetUdpPort(instance);
+    Mdns::Publisher::TxtData txtData;
+
+    otbrLogInfo("Publish meshcop-e service %s.%s.local. port %d", mServiceInstanceName.c_str(),
+                kBorderAgentEpskcServiceType, port);
+
+    mPublisher.PublishService(/* aHostName */ "", mServiceInstanceName, kBorderAgentEpskcServiceType,
+                              Mdns::Publisher::SubTypeList{}, port, txtData, [this](otbrError aError) {
```
superwhd, line 244: `txtData` is empty when passed in. What about just using a short-lived temp object?

src/border_agent/border_agent.cpp
wgtdkp: Yes, `otBorderAgentGetUdpPort()` returns the port of `_meshcop-e._udp` service when the device is in ePSKc mode. See https://github.com/openthread/openthread/blob/4737231d64d262eb850e0da729d4b2df130033e1/include/openthread/border_agent.h#L198

@abtink But I think it's better to have a new API to return the ePSKc port independently. Otherwise, if there is a update to the `_mehscop._udp` service during the ePSKc mode, the wrong port will be used there. Thoughts?

abtink: Border Agent may be using PSK (traditional commissioning model) or set up to use e-PSK, but cannot do not both at the same time. So it can accept connections on one port (at a time) depending on its mode. This port can be retrived using `otBorderAgentGetUdpPort()`.

src/border_agent/border_agent.cpp
mia1yang: Accepted. Thanks~

src/border_agent/border_agent.cpp
mia1yang: Accepted. Thanks~

src/border_agent/border_agent.cpp
mia1yang: Accepted. Thanks~

src/border_agent/border_agent.cpp
```diff
@@ -210,6 +217,68 @@ void BorderAgent::Stop(void)
...
+void BorderAgent::PublishEpskcService()
+{
+    otInstance *instance = mNcp.GetInstance();
+    int         port     = otBorderAgentGetUdpPort(instance);
+
+    otbrLogInfo("Publish meshcop-e service %s.%s.local. port %d", mServiceInstanceName.c_str(),
+                kBorderAgentEpskcServiceType, port);
+
+    mPublisher.PublishService(/* aHostName */ "", mServiceInstanceName, kBorderAgentEpskcServiceType,
+                              Mdns::Publisher::SubTypeList{}, port, {}, [this](otbrError aError) {
```
superwhd, line 244: nit: It's hard to infer what it is from a `{}`.  Please add an inline comment (or maybe add its type) indicating the meaning of the TXT data argument.

src/border_agent/border_agent.cpp
```diff
@@ -71,6 +71,9 @@ namespace otbr {
 
 static const char    kBorderAgentServiceType[]    = "_meshcop._udp"; ///< Border agent service type of mDNS
 static constexpr int kBorderAgentServiceDummyPort = 49152;
+#if OTBR_ENABLE_EPSKC
```
bukepo, line 74: Would you mind omitting this guard? We'd better to control the on/off at runtime. We can keep it as a the default value of whether the feature is on or off.

src/border_agent/border_agent.cpp
mia1yang: Accepted; Thanks!

src/border_agent/border_agent.cpp
mia1yang: Per discussion, will enable feature flag to control ePSKc (tracking b/339768305);

src/border_agent/border_agent.cpp
superwhd: I'm fine with keeping the code as-is in this PR.

Some ideas for better handling this in a future PR:
1. There's some common code shared in publishing meshcop and meshcop-e. Maybe we can make a callback function used by both services.
2. In the callback function, when there's a name conflict, we can call something like `UpdateEpskcService` and `UpdateMeshCopService` at the same time.

src/border_agent/border_agent.cpp
mia1yang: Thanks for the clarification,  the concern of using the same UDP ports would not happen in current behavior.

src/border_agent/border_agent.cpp
mia1yang: Thanks for these comments and suggestions. As we discussed, the possibility is kind of low and the modification may potentially impact the original Mesh-cop service; So we plan to keep it and keep close monitoring. Will trigger another fix if there is any.

mia1yang: Add a comment to mark down the risk.

src/border_agent/border_agent.cpp
bukepo: Let's not guard the feature compiling with this macro, unless you want to add an additional test verifying that when `OTBR_ENABLE_EPSKC` is 0, at the source can still be compiled.

librasungirl (approved):

superwhd (approved):
LGTM 👍

bukepo (approved):

src/border_agent/border_agent.cpp
```diff
@@ -146,6 +147,8 @@ BorderAgent::BorderAgent(otbr::Ncp::ControllerOpenThread &aNcp, Mdns::Publisher
     , mBaseServiceInstanceName(OTBR_MESHCOP_SERVICE_INSTANCE_NAME)
 {
     mNcp.AddThreadStateChangedCallback([this](otChangedFlags aFlags) { HandleThreadStateChanged(aFlags); });
+
+    otBorderAgentSetEphemeralKeyCallback(mNcp.GetInstance(), BorderAgent::HandleEpskcStateChanged, this);
```
bukepo, line 151: At this time, mNcp is not initialized yet. It's causing null pointer issue.

src/border_agent/border_agent.cpp
mia1yang: Modified and accepted. Thanks~

src/border_agent/border_agent.cpp
mia1yang: Move to OTBR start, as mNcp.GetInstance met error when OTBR has not started.

src/border_agent/border_agent.cpp
librasungirl: A reminder here to obsolete OTBR_EPSKC option along with introducing runtime on/off control. @mia1yang

abtink (approved):
LGTM.

bukepo (approved):

jwhui (approved):

PR #2265, opened 2024-04-26
[bbr] add a flag to enable BBR on init

sunytt:
Flag `OTBR_ENABLE_BACKBONE_ROUTER_ON_INIT` is created for enabling BBR on init. It's set to OTBR_ENABLE_BACKBONE_ROUTER by default and this works when infra link is statically configured. If infra link is dynamically configured then this flag should be set to 0, the component that sets infra link should enable/disable BBR feature with the otBackboneRouterSetEnabled API.


superwhd (approved):
LGTM 👍

PR #2283, opened 2024-05-11
[ncp] add initial support of NCP

Irving-cl:
This PR adds initial support of NCP in otbr-agent. When otbr-agent
starts, it will detect the Co-processor type. If the Co-processor is
NCP, otbr-agent will still run. But at this moment, it will provide
a dbus API to get the device role of the NCP, which is 'Disabled'
because no other functions are supported now.

This PR implements `NcpSpinel` which is the main class to communicate
with the NCP. Different from `RadioSpinel` in openthread, `NcpSpinel`
is totally async. There are no blocking operations like `WaitResponse`
in `RadioSpinel`.

Currently when the Co-processor type is NCP, most modules in otbr-agent
will be initialized and run. These may be added gradually later. And
the dbus server also initiates differently when Co-processor type is NCP.
When it's NCP, the dbus server registers `AsyncGetPropertyHandler`
instead of `GetPropertyHandler` used now.

Depends-on openthread/openthread#10272

### DIFF
--- a/src/ncp/CMakeLists.txt
+++ b/src/ncp/CMakeLists.txt
@@ -29,6 +29,8 @@
 add_library(otbr-ncp
     ncp_host.cpp
     ncp_host.hpp
+    ncp_spinel.cpp
+    ncp_spinel.hpp
     rcp_host.cpp
     rcp_host.hpp
     thread_host.cpp
--- a/src/ncp/ncp_host.cpp
+++ b/src/ncp/ncp_host.cpp
@@ -57,17 +57,18 @@ const char *NcpHost::GetCoprocessorVersion(void)
 void NcpHost::Init(void)
 {
     otSysInit(&mConfig);
+    mNcpSpinel.Init(&mSpinelDriver);
 }
 
 void NcpHost::Deinit(void)
 {
+    mNcpSpinel.Deinit();
     otSysDeinit();
 }
 
 void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
 {
-    // TODO: Implement the API with NCP Spinel
-    aHandler(OT_ERROR_NOT_IMPLEMENTED, OT_DEVICE_ROLE_DISABLED);
+    mNcpSpinel.GetDeviceRole(aHandler);
 }
 
 void NcpHost::Process(const MainloopContext &aMainloop)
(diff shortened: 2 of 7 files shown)


Irving-cl:
@bukepo @zhanglongxia @abtink @superwhd @wgtdkp 

I think the change is a non-trivial one. The PR is still under development. There are many things to do: adding test, refactoring. But I hope that I can get some feedback on the framework and the direction in early stage. PTAL when you have a moment.

Irving-cl:
This PR is ready for review now. It seems to be difficult to break
the PR down into smaller PRs. Here is an abstraction of what the PR
does:
1. Defined an interface (`IControllerOpenThread`) where we place the
unified thread control APIs (for both NCP and RCP cases). In this PR,
only one API (`GetDeviceRole`) is added for testing.
2. The original core class `ControllerOpenThread` is renamed as `ControllerOpenThreadRcp`
and refactor'ed as an implementation class of `IControllerOpenThread`.
Most of its original methods still exist.
3. Another class `ControllerOpenThreadNcp` is added to implement `IControllerOpenThread`
and is used under NCP case.
4. The dependencies on `ControllerOpenThread` of the current modules
are not changed at this staged. For example, `BackboneAgent`, `NdProxyManager`,
`BorderAgent` all depend on an instance of `ControllerOpenThread`. 
Now they still depend on `ControllerOpenThreadRcp`, instead of `IControllerOpenThread`.
Currently these modules won't be initialized and started under NCP
mode. We may change this when later working on enabling these modules
under NCP mode. 
5. Refactor'ed the initialization process. Currently the modules hold
a reference to `ControllerOpenThread`. And the reference is passed
in the class constructors. In this PR, th reference is changed to
pointer and pointer is passed in `Init` method because the instance
is created later based on the co-processor type.
6. Renamed the variable `Ncp`. Currently the variable name for an
instance of `ControllerOpenThread` is `mNcp` or `aNcp`. This is
confusing from my perspective. The PR changes all these names to
`Ctrlr` (Controller).
7. Added a test case to test the dbus API under NCP mode.

There is still a pending PR in openthread. Without it the CI here
fails. Please take a loook when you have a chance.
@zhanglongxia @abtink @superwhd

Irving-cl:
Per discussion, the PR is still too large for review. I'll try to breakdown it.

The first breakdown PR is: #2294

Please review the simple one first. Thanks!

Irving-cl:
The second breakdown PR is: #2299

Irving-cl:
This PR has been breakdown into a few PRs. Close this one.

PR #2284, opened 2024-05-11
[telemetry] add external route related telemetry

zesonzhang:
This commit added new entry `ExternalRoutes` inside message `wpan_border_router` to indicate the information about the external route added in leader network data

The `ExternalRoutes` has three bool fields:
 - `has_default_route_added`: if leader network data has `route ::/0`
 - `has_ula_route_added`: if leader network data has route `fc00::/7`
 - `has_others_route_added`: if leader network has others route

These information are collectly only when both TELEMETRY_DATA_API and BORDER_ROUTING are enabled.

This commit also enhanced the class `otbr::Ip6Address` and class `otbr::Ip6Prefix` by adding some more constructors, so that the ipv6 address/prefix can be easily constructed by a string. The `operator==` of `Ip6Prefix` is also added to compare the prefix length and first N-bit of the prefix, where N is length of the prefix.

### DIFF
--- a/src/common/types.cpp
+++ b/src/common/types.cpp
@@ -104,6 +104,30 @@ Ip6Address Ip6Address::FromString(const char *aStr)
     return addr;
 }
 
+bool Ip6Prefix::operator==(const Ip6Prefix &aOther) const
+{
+    bool    isEqual = false;
+    uint8_t lengthFullBytes;
+    uint8_t lengthRemainingBits;
+
+    VerifyOrExit(mLength == aOther.mLength);
+
+    lengthFullBytes     = mLength / 8;
+    lengthRemainingBits = mLength % 8;
+    VerifyOrExit(memcmp(mPrefix.m8, aOther.mPrefix.m8, lengthFullBytes) == 0);
+
+    if (lengthRemainingBits > 0)
+    {
+        uint8_t mask = 0xff << (8 - lengthRemainingBits);
+        VerifyOrExit((mPrefix.m8[lengthFullBytes] & mask) == (aOther.mPrefix.m8[lengthFullBytes] & mask));
+    }
+
+    isEqual = true;
+
+exit:
+    return isEqual;
+}
+
 void Ip6Prefix::Set(const otIp6Prefix &aPrefix)
 {
     memcpy(reinterpret_cast<void *>(this), &aPrefix, sizeof(*this));
--- a/src/common/types.cpp
+++ b/src/common/types.cpp
@@ -107,8 +107,8 @@ Ip6Address Ip6Address::FromString(const char *aStr)
 bool Ip6Prefix::operator==(const Ip6Prefix &aOther) const
 {
     bool    isEqual = false;
-    uint8_t lengthFullBytes;
-    uint8_t lengthRemainingBits;
+    uint8_t lengthFullBytes;     // the number of complete bytes in the prefix length
+    uint8_t lengthRemainingBits; // the number of remaining bits in the prefix length that do not form a complete byte
 
     VerifyOrExit(mLength == aOther.mLength);
 
--- a/src/common/types.hpp
+++ b/src/common/types.hpp
@@ -387,6 +387,14 @@ public:
      */
     bool IsDefaultRoutePrefix(void) const { return (*this == Ip6Prefix("::", 0)); }
 
+    /**
+     * This method checks if the object is the ULA prefix ("fc00::/7")
+     *
+     * @returns true if the object is the ULA prefix, false otherwise.
+     *
+     */
+    bool IsUlaPrefix(void) const { return (*this == Ip6Prefix("fc00::", 7)); }
+
     Ip6Address mPrefix; ///< The IPv6 prefix.
     uint8_t    mLength; ///< The IPv6 prefix length (in bits).
 };
--- a/src/common/types.hpp
+++ b/src/common/types.hpp
@@ -142,6 +142,14 @@ public:
     Ip6Address(const uint8_t (&aAddress)[16]);
 
     /**
+     * Constructor with a string.
+     *
+     * @param[in] aString The string representing the IPv6 address.
+     *
+     */
+    Ip6Address(const char *aString) { FromString(aString, *this); }
+
+    /**
      * This method overloads `<` operator and compares if the Ip6 address is smaller than the other address.
      *
      * @param[in] aOther  The other Ip6 address to compare with.
@@ -315,6 +323,33 @@ public:
     Ip6Prefix(void) { Clear(); }
 
     /**
+     * Constructor with an Ip6 address string and prefix length.
+     *
+     * @param[in] aIp6AddrStr The IPv6 address string.
+     * @param[in] aLength     The prefix length.
+     *
+     */
+    Ip6Prefix(const char *aIp6AddrStr, uint8_t aLength)
+        : mPrefix(aIp6AddrStr)
+        , mLength(aLength)
+    {
+    }
+
+    /**
+     * This method overloads `==` operator for comparing two Ip6Prefix objects by comparing their prefix and length.
+     *
+     * Two IpPrefix objects are considered equal if:
+     *  - their lengths are equal, and
+     *  - their first n-bits of the addresses are the same, where n is the length of the prefix.
+     *
+     * @param[in] aOther The Ip6Prefix object to compare with.
+     *
+     * @returns True if the two objects are equal, false otherwise.
+     *
+     */
+    bool operator==(const Ip6Prefix &aOther) const;
+
+    /**
      * This method sets the Ip6 prefix to an `otIp6Prefix` value.
      *
      * @param[in] aPrefix  The `otIp6Prefix` value to set the Ip6 prefix.
@@ -343,6 +378,14 @@ public:
      *
      */
     bool IsValid(void) const { return mLength > 0 && mLength <= 128; }
+
+    /**
+     * This method checks if the object is the default route prefix ("::/0")
+     *
+     * @returns true if the object is the default route prefix, false otherwise.
+     *
+     */
+    bool IsDefaultRoutePrefix(void) const { return (*this == Ip6Prefix("::", 0)); }
 
     Ip6Address mPrefix; ///< The IPv6 prefix.
     uint8_t    mLength; ///< The IPv6 prefix length (in bits).
--- a/src/proto/thread_telemetry.proto
+++ b/src/proto/thread_telemetry.proto
@@ -496,14 +496,17 @@ message TelemetryData {
     optional uint32 global_unicast_address_count = 7;
   }
 
-  // infomation on divergent BRs resulted from multi-AIL
-  // TODO: add more fields to this message to provide more information
-  message NodeDivergenceInfo {
-    // Indicate whether the BR's external route is published
-    optional bool route_ail = 1;
+  // Message to indicate the information of external routes in network data.
+  message ExternalRoutes {
+    // Indicates whether the a zero-length prefix (::/0) added from this BR
+    optional bool has_default_route_added = 1;
 
-    // Indicate whether the BR's external route is published as default route
-    optional bool route_default = 2;
+    // Indicates whether the a ULA prefix (fc00::/7) added from this BR
+    optional bool has_ula_route_added = 2;
+
+    // Indicates whether the other prefixes (other than "::/0" or "fc00::/7") added
+    // from this BR. (BR is a managed infrastructure router).
+    optional bool has_others_route_added = 3;
   }
 
   message WpanBorderRouter {
@@ -543,8 +546,8 @@ message TelemetryData {
     // Information about the infra link
     optional InfraLinkInfo infra_link_info = 12;
 
-    // Information about the node divergence (multi-AIL)
-    optional NodeDivergenceInfo node_divergence_info = 13;
+    // Information about the external routes in network data.
+    optional ExternalRoutes external_route_info = 13;
   }
 
   message RcpStabilityStatistics {
--- a/src/proto/thread_telemetry.proto
+++ b/src/proto/thread_telemetry.proto
@@ -496,6 +496,16 @@ message TelemetryData {
     optional uint32 global_unicast_address_count = 7;
   }
 
+  // infomation on divergent BRs resulted from multi-AIL
+  // TODO: add more fields to this message to provide more information
+  message NodeDivergenceInfo {
+    // Indicate whether the BR's external route is published
+    optional bool route_ail = 1;
+
+    // Indicate whether the BR's external route is published as default route
+    optional bool route_default = 2;
+  }
+
   message WpanBorderRouter {
     // Border routing counters
     optional BorderRoutingCounters border_routing_counters = 1;
@@ -532,6 +542,9 @@ message TelemetryData {
 
     // Information about the infra link
     optional InfraLinkInfo infra_link_info = 12;
+
+    // Information about the node divergence (multi-AIL)
+    optional NodeDivergenceInfo node_divergence_info = 13;
   }
 
   message RcpStabilityStatistics {
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -920,6 +920,39 @@ void ThreadHelper::DetachGracefullyCallback(void)
 }
 
 #if OTBR_ENABLE_TELEMETRY_DATA_API
+#if OTBR_ENABLE_BORDER_ROUTING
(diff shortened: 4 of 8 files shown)


src/utils/thread_helper.cpp
```diff
@@ -920,6 +920,42 @@ void ThreadHelper::DetachGracefullyCallback(void)
...
+{
+    bool isExternalRoutePublished = false;
+    bool isDefaultRoutePublished  = false;
+    Ip6Prefix prefix;
+    uint16_t rloc16               = otThreadGetRloc16(mInstance);
+
+    otNetworkDataIterator iterator = OT_NETWORK_DATA_ITERATOR_INIT;
+    otExternalRouteConfig config;
+
+    // iterate over leader network data
```
morningboata, line 934: iterate => Iterate. Correct the following comment's first word as well.

src/utils/thread_helper.cpp
```diff
@@ -920,6 +920,42 @@ void ThreadHelper::DetachGracefullyCallback(void)
...
+        }
+
+        // network data has the BR's external route
+        isExternalRoutePublished = true;
+
+        // network data has the default route of the BR if the prefix is matched
+        prefix.Set(config.mPrefix);
+        if (prefix.IsDefaultRoutePrefix())
+        {
+            isDefaultRoutePublished  = true;
```
morningboata, line 950: Do you want to break out the loop when this value is set to true?

src/utils/thread_helper.cpp
```diff
@@ -1281,6 +1317,9 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
             infraLinkInfo->set_global_unicast_address_count(addressCounters.mGlobalUnicastAddresses);
         }
         // End of InfraLinkInfo section.
+
+        // NodeDivergenceInfo section
+        GetNodeDivergenceInfo(wpanBorderRouter->mutable_node_divergence_info());
```
morningboata, line 1322: Please use same MACRO condition when invoking this function, i.e., add the OTBR_ENABLE_BORDER_ROUTING condition.

zesonzhang: Thanks!  The marco is already in Line #1302

src/utils/thread_helper.cpp
zesonzhang: I would like to iterate over in case there are multiple entries

src/utils/thread_helper.cpp
zesonzhang: updated! Thanks!

src/common/types.hpp
```diff
@@ -314,6 +322,30 @@ class Ip6Prefix
...
+    /**
+     *
+     * This method overloads `==` operator for comparing two Ip6Prefix objects by comparing their prefix and length.
+     *
+     * @param[in] aOther The Ip6Prefix object to compare with.
+     *
+     * @returns True if the prefix and length of the two objects are equal, false otherwise.
+     *
+     */
+    bool operator==(const Ip6Prefix &aOther) const { return mPrefix == aOther.mPrefix && mLength == aOther.mLength; }
```
morningboata, line 347: For mPrefix == aOther.mPrefix, do you only compare the pointer, or you want to compare the content of the char?
For mLength == aOther.mLength, this may be moved before mPrefix compare since int compare could be faster than string content comparison.

src/utils/thread_helper.cpp
morningboata: When isDefaultRoutePublished = true, it seems that this is not to be assigned back to false again, do you need any more iterations?

src/common/types.hpp
zesonzhang: 1. I expect it calls the `operator==` of class `Ip6Prefix` in line# 347
2. good point! updated

src/utils/thread_helper.cpp
zesonzhang: good point, update! will break after isDefaultRoutePublished set to true.

src/common/types.hpp
```diff
@@ -141,6 +141,14 @@ class Ip6Address
      */
     Ip6Address(const uint8_t (&aAddress)[16]);
 
+    /**
+     * Constructor with a string
```
superwhd, line 145:
```suggestion
     * Constructor with a string.
```

src/common/types.hpp
```diff
@@ -344,6 +376,14 @@ class Ip6Prefix
      */
     bool IsValid(void) const { return mLength > 0 && mLength <= 128; }
 
+    /**
+     * This method checks if the object is the default route prefix ("::/0")
+     *
+     * @returns true if the object is the default route prefix, false otherwise.
+     *
+     */
+    bool IsDefaultRoutePrefix(void) const { return (*this == Ip6Prefix("::", 0)); }
```
superwhd, line 385: What about we just check the prefix length is zero?

src/common/types.hpp
```diff
@@ -344,6 +376,14 @@ class Ip6Prefix
      */
     bool IsValid(void) const { return mLength > 0 && mLength <= 128; }
```
superwhd, line 377: Do we need to update this? We're now having prefixes with zero length.

src/proto/thread_telemetry.proto
```diff
@@ -496,6 +496,14 @@ message TelemetryData {
     optional uint32 global_unicast_address_count = 7;
   }
 
+  message NodeDivergenceInfo {
+    // Indicate whether the BR's external route is published
```
superwhd, line 500: We may need some more details about what it means.

src/utils/thread_helper.cpp
```diff
@@ -920,6 +920,43 @@ void ThreadHelper::DetachGracefullyCallback(void)
 }
 
 #if OTBR_ENABLE_TELEMETRY_DATA_API
+#if OTBR_ENABLE_BORDER_ROUTING
+void ThreadHelper::GetNodeDivergenceInfo(threadnetwork::TelemetryData_NodeDivergenceInfo *aNodeDivergenceInfo)
```
superwhd, line 924: I think the netdata and rloc16 are already included in telemetry data. Do we want to process them here again?

src/common/types.hpp
zesonzhang: both way should work good. but I prefer doing this as this is more common, so that similar method can be created using same way, e.g.
```
bool IsUlaPrefix(void) const { return (*this == Ip6Prefix("fc00::", 7)); }
```

src/common/types.hpp
zesonzhang: sounds reasonable! but this is not in the scope of this change. I can create another commit to fix this.

src/utils/thread_helper.cpp
```diff
@@ -920,6 +920,43 @@ void ThreadHelper::DetachGracefullyCallback(void)
...
+{
+    bool      isExternalRoutePublished = false;
```
abtink, line 926: There is a difference between "publish" and "add" terms used within OT regarding Network Data. I mentioed more details here https://github.com/openthread/openthread/pull/10199#discussion_r1591346486

Suggest renaming to `isExternalRouteAdded` as this is what is being checked below.

src/utils/thread_helper.cpp
```diff
@@ -920,6 +920,43 @@ void ThreadHelper::DetachGracefullyCallback(void)
...
+    Ip6Prefix prefix;
+    uint16_t  rloc16 = otThreadGetRloc16(mInstance);
+
+    otNetworkDataIterator iterator = OT_NETWORK_DATA_ITERATOR_INIT;
+    otExternalRouteConfig config;
+
+    // Iterate leader network data
+    while (otNetDataGetNextRoute(mInstance, &iterator, &config) == OT_ERROR_NONE)
+    {
+        // filter out the config that is not stable or unmatched rloc16
```
abtink, line 937: Suggest removing this comment here and ones below as they simply repeat what the code does and the logic is clear from the code.

This link has some useful advice on this: [code-health-to-comment-or-not-to-comment](https://testing.googleblog.com/2017/07/code-health-to-comment-or-not-to-comment.html) (or Google internal link [go/tott/563](https://goto.google.com/tott/563))

abtink (commented):
Thansk @zesonzhang. Overall looks good.
Some suggestions below.

src/utils/thread_helper.cpp
zesonzhang: In some system, e.g. Android, the network data is not able to be uploaded, we might need to explicitly get this metrics here.

src/proto/thread_telemetry.proto
zesonzhang: added some more description

src/common/types.hpp
```diff
@@ -314,6 +322,30 @@ class Ip6Prefix
...
+    /**
+     *
+     * This method overloads `==` operator for comparing two Ip6Prefix objects by comparing their prefix and length.
+     *
+     * @param[in] aOther The Ip6Prefix object to compare with.
+     *
+     * @returns True if the prefix and length of the two objects are equal, false otherwise.
+     *
+     */
+    bool operator==(const Ip6Prefix &aOther) const { return mLength == aOther.mLength && mPrefix == aOther.mPrefix; }
```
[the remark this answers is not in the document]
zesonzhang: `Ip6Address::operator==()` checking all 128 bits sounds reasonable `Ip6Address` does not contain any length related information. 

I just updated the `Ip6Prefix::operator==()` in types.cpp, this function just check the first N bits, where N is the length of the prefix.

Thanks!

src/utils/thread_helper.cpp
zesonzhang: Thanks! Updated!

src/utils/thread_helper.cpp
zesonzhang: Thanks! The link is helpful! And I removed the extra comment line.

abtink (approved):
Looks great. Thanks @zesonzhang

src/proto/thread_telemetry.proto
```diff
@@ -496,6 +496,16 @@ message TelemetryData {
     optional uint32 global_unicast_address_count = 7;
   }
 
+  // infomation on divergent BRs resulted from multi-AIL
```
bukepo, line 499:
```suggestion
  // information on divergent BRs resulted from multi-AIL
```

src/proto/thread_telemetry.proto
bukepo: Perhaps we should have 3 bool metrics: whether the route is publishing route for `::/0`(any network), `fc00::/7`(any ULA network), and other ULA prefix.

src/proto/thread_telemetry.proto
zesonzhang: Thanks!

src/proto/thread_telemetry.proto
zesonzhang: updated, added a new dedicated message `ExternalRoutes` to show the information about external route, from which we can get the information about `::/0`, `fc00::/7` or others

zesonzhang:
> Looks great. Thanks @zesonzhang

I just updated a new version per comment, can you please take a look? The diff commit is [here](https://github.com/openthread/ot-br-posix/pull/2284/commits/4518d77cf63e2b1a457a3c3550c887dd50f29823). The main change is to collect more information about external route added. And then replace `NodeDivergenceInfo` with `ExternalRoutes` in proto to make the message dedicated for external route info. `NodeDivergenceInfo` will be added in the future.

Thanks!

src/proto/thread_telemetry.proto
```diff
@@ -496,6 +496,19 @@ message TelemetryData {
...
+  message ExternalRoutes {
+    // Indicates whether the a zero-length prefix (::/0) added from this BR
+    optional bool has_default_route_added = 1;
+
+    // Indicates whether the a ULA prefix (fc00::/7) added from this BR
+    optional bool has_ula_route_added = 2;
+
+    // Indicates whether the other prefixes (other than "::/0" or "fc00::/7") added
+    // from this BR. (BR is a managed infrastructure router).
+    optional bool has_others_route_added = 3;
```
superwhd, line 509:
```suggestion
    optional bool has_other_route_added = 3;
```

src/utils/thread_helper.cpp
```diff
@@ -920,6 +920,46 @@ void ThreadHelper::DetachGracefullyCallback(void)
 }
 
 #if OTBR_ENABLE_TELEMETRY_DATA_API
+#if OTBR_ENABLE_BORDER_ROUTING
+void ThreadHelper::GetExternalRouteInfo(threadnetwork::TelemetryData_ExternalRoutes *aExternalRouteInfo)
+{
+    bool      isDefaultRouteAdded = false;
+    bool      isUlaRouteAdded     = false;
+    bool      isOthersRouteAdded  = false;
```
superwhd, line 928:
```suggestion
    bool      isOtherRouteAdded  = false;
```

src/utils/thread_helper.cpp
```diff
@@ -1281,6 +1321,10 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
             infraLinkInfo->set_global_unicast_address_count(addressCounters.mGlobalUnicastAddresses);
         }
         // End of InfraLinkInfo section.
+
+        // ExternalRoutes section
+        GetExternalRouteInfo(wpanBorderRouter->mutable_external_route_info());
+
```
superwhd, line 1327: I'd suggest to add a test for this.

abtink (approved):
LGTM.

src/utils/thread_helper.cpp
superwhd: ack

src/proto/thread_telemetry.proto
superwhd: ack

zesonzhang: Thanks for raising this! My thinkings are:
1. "others" acts as a pronoun here, referring to all routes other than "default" and "ULA"
2. "has_XXX_route_added" maintains grammatical consistency with the other fields

src/utils/thread_helper.cpp
zesonzhang: added in the latest push

src/utils/thread_helper.hpp
```diff
@@ -302,6 +302,10 @@ class ThreadHelper
 
     void ActiveDatasetChangedCallback(void);
 
+#if OTBR_ENABLE_TELEMETRY_DATA_API && OTBR_ENABLE_BORDER_ROUTING
+    void GetExternalRouteInfo(threadnetwork::TelemetryData_ExternalRoutes *aExternalRouteInfo);
```
superwhd, line 306: nit: After a second look I think the `Get` method usually has a return value but this one seems more like filling in the field. Maybe call it `Fill` or `Retrieve`.

superwhd (approved):
LGTM 👍  with a nit

src/utils/thread_helper.hpp
morningboata: +1 Populate/Fill/Retrieve seems to be a better name.

src/common/types.cpp
```diff
@@ -104,6 +104,30 @@ Ip6Address Ip6Address::FromString(const char *aStr)
...
+    uint8_t lengthFullBytes;
+    uint8_t lengthRemainingBits;
+
+    VerifyOrExit(mLength == aOther.mLength);
+
+    lengthFullBytes     = mLength / 8;
+    lengthRemainingBits = mLength % 8;
+    VerifyOrExit(memcmp(mPrefix.m8, aOther.mPrefix.m8, lengthFullBytes) == 0);
+
+    if (lengthRemainingBits > 0)
```
morningboata, line 119: Can you add a comment what this lengthRemainingBits compares?

src/utils/thread_helper.hpp
zesonzhang: Thanks for the suggestion @superwhd and @morningboata , I just renamed to `RetrieveExternalRouteInfo` in the latest push

src/common/types.cpp
zesonzhang: added in the latest push

bukepo (approved):

zesonzhang:
@jwhui can we get this PR merged?

jwhui (approved):

jwhui:
> @jwhui can we get this PR merged?

It looks like you have a number of commits in this PR. Did you want to keep them separate or are you OK to squash them?

zesonzhang:
> > @jwhui can we get this PR merged?
> 
> It looks like you have a number of commits in this PR. Did you want to keep them separate or are you OK to squash them?

OK to squash! Thanks!

PR #2286, opened 2024-05-14
[thread-helper] update `otDatasetSetActive()` use now being `void`

abtink:
Depends-On: openthread/openthread#10237


abtink:
Closing this PR as now in https://github.com/openthread/openthread/pull/10237 we retain the public API definition for backward compatibility (still returning `otError` but documenting that `OT_ERROR_NONE` is always returned).

PR #2293, opened 2024-05-21
[mbedtls] stop checking configuration explicitly

LuDuda:
This commit removes configuration sanity check for MbedTLS versions higher or equal to 3.0.0.

This is a follow-up after https://github.com/openthread/openthread/pull/10263 and needed to migrate to Mbed TLS 3.6.0.


jwhui (approved):
Thanks! 👍🏼

PR #2299, opened 2024-05-27
[controller] replace reference of controller with pointer

Irving-cl:
This PR changes the reference of `Ncp::HostRcp` inside all modules
to a pointer to `Ncp::HostRcp`. Instead of passing the reference of
`Ncp::HostRcp` to all modules in their contructors, this PR changes
it as passing a pointer in the `Init` of these modules.

This PR is also a subset of #2283. This change is required because:
Currently there is an instance of `Ncp::HostRcp` as a member of the
main class `Application`. Most modules inside `Application` have a
reference to the instance and the reference is passed to these modules
in their constructor (in `Application`'s initialization list).
However, in the NCP case, `Ncp::HostRcp` shouldn't be contructed at
all. So updating the reference to pointer allows `Application` to not
initialize these modules in NCP case and initialize a different `Host`.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -49,8 +49,6 @@ const struct timeval Application::kPollTimeout = {10, 0};
 
 Application::Application(const std::string               &aInterfaceName,
                          const std::vector<const char *> &aBackboneInterfaceNames,
-                         const std::vector<const char *> &aRadioUrls,
-                         bool                             aEnableAutoAttach,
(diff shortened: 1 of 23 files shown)


superwhd (commented):
I think the two-step initialization (constructor + `Init()`) is a bit more error-prone than the one-step initialization.

What about refactor it in a different way: Define interfaces for each component and let the RCP-based implementation store the reference to the `RcpHost` while the NCP one doesn't.

PR #2302, opened 2024-05-28
[application] refactor constructor to allow flexible initialization

Irving-cl:
This PR refactors the constructor of the `Application` class so that
the initialization could be flexible.

This PR is a replacement of #2299 and a child PR of #2283. With this
change, we can choose to **not** initialize other submodules when
the co-processor is an NCP.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -60,45 +60,48 @@ Application::Application(const std::string               &aInterfaceName,
 #else
     , mBackboneInterfaceName(aBackboneInterfaceNames.empty() ? "" : aBackboneInterfaceNames.front())
 #endif
-    , mHost(mInterfaceName.c_str(), aRadioUrls, mBackboneInterfaceName, /* aDryRun */ false, aEnableAutoAttach)
 #if OTBR_ENABLE_MDNS
     , mPublisher(Mdns::Publisher::Create([this](Mdns::Publisher::State aState) { this->HandleMdnsState(aState); }))
-#endif
-#if OTBR_ENABLE_BORDER_AGENT
-    , mBorderAgent(mHost, *mPublisher)
-#endif
-#if OTBR_ENABLE_BACKBONE_ROUTER
-    , mBackboneAgent(mHost, aInterfaceName, mBackboneInterfaceName)
-#endif
-#if OTBR_ENABLE_SRP_ADVERTISING_PROXY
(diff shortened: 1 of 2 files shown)


src/agent/application.cpp
```diff
@@ -60,45 +60,50 @@ Application::Application(const std::string               &aInterfaceName,
...
 #endif
-    , mHost(mInterfaceName.c_str(), aRadioUrls, mBackboneInterfaceName, /* aDryRun */ false, aEnableAutoAttach)
 #if OTBR_ENABLE_MDNS
     , mPublisher(Mdns::Publisher::Create([this](Mdns::Publisher::State aState) { this->HandleMdnsState(aState); }))
 #endif
+#if OTBR_ENABLE_VENDOR_SERVER
+    , mVendorServer(vendor::VendorServer::newInstance(*this))
+#endif
+{
+    mHost = std::unique_ptr<Ncp::RcpHost>(new Ncp::RcpHost(mInterfaceName.c_str(), aRadioUrls, mBackboneInterfaceName,
```
superwhd, line 70: I think you can use https://github.com/openthread/ot-br-posix/blob/5a7972eb22434f3eea2d1651bfd6a3e60f0821ba/src/common/code_utils.hpp#L167

src/agent/application.cpp
```diff
@@ -60,45 +60,50 @@ Application::Application(const std::string               &aInterfaceName,
...
+                                                           /* aDryRun */ false, aEnableAutoAttach));
+
 #if OTBR_ENABLE_BORDER_AGENT
-    , mBorderAgent(mHost, *mPublisher)
+    mBorderAgent = std::unique_ptr<BorderAgent>(new BorderAgent(*mHost, *mPublisher));
```
superwhd, line 74: I'm still inclined to reserving the initialization list for the constructor.

superwhd (approved):
LGTM :+1:

src/agent/application.cpp
Irving-cl: I agree initialization list is better. As discussed offline, let's keep this as a “transition". Because currently all these modules don't work under NCP mode, let's not initialize them and pass the `host` to it at this moment. When later we add NCP support to these modules, we can put them back to initialization list.

zhanglongxia (approved):
LGTM

abtink (approved):
👍

jwhui (approved):

PR #2306, opened 2024-05-30
[meshcop] add ThreadRole in state bitmap txt

librasungirl:


bukepo:
could you add a test to verify this change?

librasungirl:
`73 62 3d 00 00 0f b1` locally validated state bitmap for BR leader with BBR functionality.

Here want to keep the the change simple as it's quite straightforward, and prefer to add a meshcop test to validate key fields in meshcop in separate PR.

superwhd:
> Here want to keep the the change simple as it's quite straightforward, and prefer to add a meshcop test to validate key fields in meshcop in separate PR.

FYI you may add a test step in https://github.com/openthread/openthread/blob/main/tests/scripts/thread-cert/border_router/test_publish_meshcop_service.py.

abtink (approved):
👍

bukepo (approved):

lanyuwen (approved):
LGTM, thanks!

jwhui (approved):

PR #2308, opened 2024-06-02
[meshcop] support non-standard TXT entries at runtime

wgtdkp:
See also https://android-review.git.corp.google.com/c/platform/external/ot-br-posix/+/3111060
for how this change is used and tested.

### DIFF
--- a/src/border_agent/border_agent.cpp
+++ b/src/border_agent/border_agent.cpp
@@ -149,20 +149,30 @@ BorderAgent::BorderAgent(otbr::Ncp::RcpHost &aHost, Mdns::Publisher &aPublisher)
     mHost.AddThreadStateChangedCallback([this](otChangedFlags aFlags) { HandleThreadStateChanged(aFlags); });
 }
 
-otbrError BorderAgent::SetMeshCopServiceValues(const std::string          &aServiceInstanceName,
-                                               const std::string          &aProductName,
-                                               const std::string          &aVendorName,
-                                               const std::vector<uint8_t> &aVendorOui)
+otbrError BorderAgent::SetMeshCopServiceValues(const std::string              &aServiceInstanceName,
+                                               const std::string              &aProductName,
+                                               const std::string              &aVendorName,
+                                               const std::vector<uint8_t>     &aVendorOui,
+                                               const Mdns::Publisher::TxtList &aNonStandardTxtEntries)
 {
     otbrError error = OTBR_ERROR_NONE;
 
     VerifyOrExit(aProductName.size() <= kMaxProductNameLength, error = OTBR_ERROR_INVALID_ARGS);
     VerifyOrExit(aVendorName.size() <= kMaxVendorNameLength, error = OTBR_ERROR_INVALID_ARGS);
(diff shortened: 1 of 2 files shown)


src/border_agent/border_agent.hpp
```diff
@@ -161,6 +163,10 @@ class BorderAgent : private NonCopyable
     std::map<std::string, std::vector<uint8_t>> mMeshCopTxtUpdate;
```
superwhd, line 163: Instead of introducing `mNonStandardTxtEntries`, I'd suggest to reuse `mMeshCopTxtUpdate`

wgtdkp: done!

src/border_agent/border_agent.cpp
```diff
@@ -149,20 +149,29 @@ BorderAgent::BorderAgent(otbr::Ncp::RcpHost &aHost, Mdns::Publisher &aPublisher)
...
+    {
+        VerifyOrExit(!txtEntry.mKey.empty() && txtEntry.mKey.front() == 'v', error = OTBR_ERROR_INVALID_ARGS);
+    }
 
     mProductName = aProductName;
     mVendorName  = aVendorName;
     mVendorOui   = aVendorOui;
+    for (const auto &txtEntry : aNonStandardTxtEntries)
+    {
+        mMeshCopTxtUpdate[txtEntry.mKey] = txtEntry.mValue;
```
superwhd, line 173: Shall we clear the `mMeshCopTxtUpdate` before adding the items?

superwhd (approved):
LGTM overall 👍 

In the future we may want to unify the two ways (D-BUS, Android overlay) for customizing meshcop TXT for better readability.

bukepo (commented):
can we use  the existing HandleUpdateVendorMeshCoPTxtEntries?

wgtdkp:
> can we use the existing HandleUpdateVendorMeshCoPTxtEntries?

That one doesn't support overriding the vendor or product name

bukepo:
> > can we use the existing HandleUpdateVendorMeshCoPTxtEntries?
> 
> 
> 
> That one doesn't support overriding the vendor or product name

I think we should allow overriding these fields by dbus api as well.

wgtdkp:
> > > can we use the existing HandleUpdateVendorMeshCoPTxtEntries?
> > 
> > 
> > That one doesn't support overriding the vendor or product name
> 
> I think we should allow overriding these fields by dbus api as well.

I am not sure that's a good idea. The `SetMeshCopServiceValues` method for Android is documented clearly that this must be called before BorderAgent is enabled, so that it won't cause a vendor or model name change which will cause unexpected behavior - mobile devices are depending on the names to take specific actions. Generally, it doesn't sound reasonable for a device to be Google Border Router at T1 and then changes to be Apple Border Router at T2.

It would require the dbus client side change if you want to ensure that advertised vendor name never changes for dbus-controlled device. But this is out of scope I think.

bukepo (approved):

jwhui (approved):

PR #2309, opened 2024-06-03
[controller] allows polymorphism of the controller

Irving-cl:
This PR is also a breakdown of #2283 

This PR adds a factory method `CreateInstance` to provide polymorphism
for the `ThreadController`. This PR moves the construction and initialization
of the components to RCP specific methods and only contructs these
modules when the Co-processor is an RCP.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -60,6 +60,11 @@ Application::Application(const std::string               &aInterfaceName,
 #else
     , mBackboneInterfaceName(aBackboneInterfaceNames.empty() ? "" : aBackboneInterfaceNames.front())
 #endif
+    , mHost(Ncp::ThreadController::Create(mInterfaceName.c_str(),
+                                          aRadioUrls,
+                                          mBackboneInterfaceName,
+                                          /* aDryRun */ false,
+                                          aEnableAutoAttach))
 #if OTBR_ENABLE_MDNS
     , mPublisher(Mdns::Publisher::Create([this](Mdns::Publisher::State aState) { this->HandleMdnsState(aState); }))
 #endif
@@ -67,91 +72,28 @@ Application::Application(const std::string               &aInterfaceName,
     , mVendorServer(vendor::VendorServer::newInstance(*this))
 #endif
 {
-    mHost = MakeUnique<Ncp::RcpHost>(mInterfaceName.c_str(), aRadioUrls, mBackboneInterfaceName,
-                                     /* aDryRun */ false, aEnableAutoAttach);
-
-#if OTBR_ENABLE_BORDER_AGENT
-    mBorderAgent = MakeUnique<BorderAgent>(*mHost, *mPublisher);
-#endif
-#if OTBR_ENABLE_BACKBONE_ROUTER
-    mBackboneAgent = MakeUnique<BackboneRouter::BackboneAgent>(*mHost, aInterfaceName, mBackboneInterfaceName);
-#endif
-#if OTBR_ENABLE_SRP_ADVERTISING_PROXY
-    mAdvertisingProxy = MakeUnique<AdvertisingProxy>(*mHost, *mPublisher);
-#endif
-#if OTBR_ENABLE_DNSSD_DISCOVERY_PROXY
-    mDiscoveryProxy = MakeUnique<Dnssd::DiscoveryProxy>(*mHost, *mPublisher);
-#endif
-#if OTBR_ENABLE_TREL
-    mTrelDnssd = MakeUnique<TrelDnssd::TrelDnssd>(*mHost, *mPublisher);
-#endif
-#if OTBR_ENABLE_OPENWRT
-    mUbusAgent = MakeUnique<ubus::UBusAgent>(*mHost);
-#endif
-#if OTBR_ENABLE_REST_SERVER
-    mRestWebServer = MakeUnique<rest::RestWebServer>(*mHost, aRestListenAddress, aRestListenPort);
-#endif
-#if OTBR_ENABLE_DBUS_SERVER && OTBR_ENABLE_BORDER_AGENT
-    mDBusAgent = MakeUnique<DBus::DBusAgent>(*mHost, *mPublisher);
-#endif
-
-    OT_UNUSED_VARIABLE(aRestListenAddress);
-    OT_UNUSED_VARIABLE(aRestListenPort);
+    if (mHost->GetCoprocessorType() == OT_COPROCESSOR_RCP)
+    {
+        CreateRcpMode(aRestListenAddress, aRestListenPort);
+    }
 }
 
 void Application::Init(void)
 {
     mHost->Init();
 
-#if OTBR_ENABLE_MDNS
-    mPublisher->Start();
-#endif
-#if OTBR_ENABLE_BORDER_AGENT
-// This is for delaying publishing the MeshCoP service until the correct
-// vendor name and OUI etc. are correctly set by BorderAgent::SetMeshCopServiceValues()
-#if OTBR_STOP_BORDER_AGENT_ON_INIT
-    mBorderAgent->SetEnabled(false);
-#else
-    mBorderAgent->SetEnabled(true);
-#endif
-#endif
-#if OTBR_ENABLE_BACKBONE_ROUTER
-    mBackboneAgent->Init();
-#endif
-#if OTBR_ENABLE_SRP_ADVERTISING_PROXY
-    mAdvertisingProxy->SetEnabled(true);
-#endif
-#if OTBR_ENABLE_DNSSD_DISCOVERY_PROXY
-    mDiscoveryProxy->SetEnabled(true);
-#endif
-#if OTBR_ENABLE_OPENWRT
-    mUbusAgent->Init();
-#endif
-#if OTBR_ENABLE_REST_SERVER
-    mRestWebServer->Init();
-#endif
-#if OTBR_ENABLE_DBUS_SERVER
-    mDBusAgent->Init();
-#endif
-#if OTBR_ENABLE_VENDOR_SERVER
-    mVendorServer->Init();
-#endif
+    if (mHost->GetCoprocessorType() == OT_COPROCESSOR_RCP)
+    {
+        InitRcpMode();
+    }
 }
 
 void Application::Deinit(void)
 {
-#if OTBR_ENABLE_SRP_ADVERTISING_PROXY
-    mAdvertisingProxy->SetEnabled(false);
-#endif
-#if OTBR_ENABLE_DNSSD_DISCOVERY_PROXY
-    mDiscoveryProxy->SetEnabled(false);
-#endif
-#if OTBR_ENABLE_BORDER_AGENT
-    mBorderAgent->SetEnabled(false);
-#endif
-#if OTBR_ENABLE_MDNS
-    mPublisher->Stop();
-#endif
+    if (mHost->GetCoprocessorType() == OT_COPROCESSOR_RCP)
+    {
+        DeinitRcpMode();
+    }
 
     mHost->Deinit();
 }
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -145,6 +145,8 @@ jobs:
         cmake --version | grep 3.10.3
     - name: Build
       run: script/test package
+    - name: Setup tmate session
+      uses: mxschmitt/action-tmate@v3
```
superwhd, line 149: I don't think we should add this step as part of CI.

Irving-cl: Was doing some debugging. Have removed it.

src/trel_dnssd/trel_dnssd.hpp
```diff
@@ -80,6 +79,14 @@ class TrelDnssd
      */
     void Initialize(std::string aTrelNetif);
 
+    /**
+     * This method sets the host instance.
+     *
+     * @param[in] aHost  A reference to the Thread Controller instance.
```
superwhd, line 85:
```suggestion
     * @param[in] aHost  A pointer to the Thread Controller instance.
```

src/trel_dnssd/trel_dnssd.cpp
```diff
@@ -103,6 +103,16 @@ void TrelDnssd::Initialize(std::string aTrelNetif)
     }
 }
 
+void TrelDnssd::SetHost(Ncp::RcpHost *aHost)
```
superwhd, line 106: Do we always expect a non-null value? If so please consider using a reference.

src/ncp/thread_controller.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include <openthread/openthread-system.h>
+
+#include "lib/spinel/coprocessor_type.h"
+
+#include "rcp_host.hpp"
+
+namespace otbr {
+namespace Ncp {
+
+ThreadController *ThreadController::CreateInstance(const char                      *aInterfaceName,
```
superwhd, line 43: What about making it return a `unique_ptr`?

src/ncp/rcp_host.hpp
```diff
@@ -194,9 +188,19 @@ class RcpHost : public MainloopProcessor, public ThreadController
 
     ~RcpHost(void) override;
 
-    // Thread Control APIs
+    // Thread Control virtual methods
     void GetDeviceRole(const DeviceRoleHandler aHandler) override;
```
superwhd, line 192: Maybe in a future PR: make this parameter either a non-const value or a const reference.

src/agent/application.cpp
```diff
@@ -255,4 +199,89 @@ void Application::HandleSignal(int aSignal)
     signal(aSignal, SIG_DFL);
 }
 
+void Application::ConstructRcpMode(const std::string &aRestListenAddress, int aRestListenPort)
+{
+    otbr::Ncp::RcpHost &rcpHost = static_cast<otbr::Ncp::RcpHost &>(*mHost);
```
superwhd, line 204: Ideally we should avoid down-casting. Is there a plan to components accepting ThreadController instead of RcpHost?

src/agent/application.cpp
```diff
@@ -255,4 +199,89 @@ void Application::HandleSignal(int aSignal)
     signal(aSignal, SIG_DFL);
 }
 
+void Application::ConstructRcpMode(const std::string &aRestListenAddress, int aRestListenPort)
```
superwhd, line 202: I'm concerned that there may be uninitialized members after some time, since the compiler doesn't throw an error if somebody forgets to initialize a member.

src/ncp/rcp_host.cpp
```diff
@@ -213,7 +167,15 @@ void RcpHost::Init(void)
...
-    mInstance = otSysInit(&mConfig);
+    // There are two cases:
+    // 1. The `Init` is called after `CreateInstance`. In this case, `otSysInit` has been called
+    // and an otInstance has been created. So there is no need to create an instance again.
+    // 2. `Reset` is called. `Deinit` is called first and then `Init` is called again. In this
+    // case need to call `otSysInit` to reset the co-processor and creat an otInstance again.
+    if (mInstance == nullptr)
+    {
+        mInstance = otSysInit(&mConfig);
+    }
```
superwhd, line 178: It looks error-prone that we have to create the otInstance at different places. Probably we should look for a better approach.

src/ncp/thread_controller.hpp
```diff
@@ -58,13 +58,66 @@ class ThreadController
...
+     * This method deinitialize the OpenThread controller.
+     *
+     */
+    virtual void Deinit(void) = 0;
+
+    /**
+     * The decontructor.
+     *
+     */
+    virtual ~ThreadController(void) {}
```
superwhd, line 117:
```suggestion
    virtual ~ThreadController(void) = default;
```

src/ncp/thread_controller.hpp
```diff
@@ -58,13 +58,66 @@ class ThreadController
...
     virtual void GetDeviceRole(DeviceRoleHandler aHandler) = 0;
+
+    /**
+     * Returns the co-processor type.
+     *
+     */
+    virtual CoprocessorType GetCoprocessorType(void) = 0;
+
+    /**
+     * Returns the Co-processor version string.
```
superwhd, line 96: Unify the `Co-processor` to upper/lower case initial letter.

src/ncp/thread_controller.hpp
```diff
@@ -58,13 +58,66 @@ class ThreadController
...
+     *
+     */
+    virtual const char *GetCoprocessorVersion(void) = 0;
+
+    /**
+     * This method initialize the OpenThread controller.
```
superwhd, line 102: nit: Thread controller

src/ncp/rcp_host.hpp
```diff
@@ -72,30 +72,24 @@ class RcpHost : public MainloopProcessor, public ThreadController
...
      */
-    RcpHost(const char                      *aInterfaceName,
-            const std::vector<const char *> &aRadioUrls,
-            const char                      *aBackboneInterfaceName,
-            bool                             aDryRun,
-            bool                             aEnableAutoAttach);
+    RcpHost(otInstance *aInstance, const otPlatformConfig &aConfig, bool aEnableAutoAttach);
 
     /**
      * This method initialize the NCP controller.
```
superwhd, line 83: Shall we update the comment?

src/ncp/thread_controller.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    instance = otSysInit(&config);
+    if (config.mCoprocessorType == OT_COPROCESSOR_RCP)
+    {
+        host = new RcpHost(instance, config, aEnableAutoAttach);
+    }
+    else
+    {
+        // TODO: add NCP type
+        VerifyOrDie(false, "Unknown coprocessor type!");
```
superwhd, line 77: use `DieNow`

src/ncp/thread_controller.hpp
```diff
@@ -58,13 +58,66 @@ class ThreadController
...
+     *
+     */
+    virtual void Init(void) = 0;
+
+    /**
+     * This method deinitialize the OpenThread controller.
```
superwhd, line 108: nit: Thread controller.

src/ncp/thread_controller.hpp
```diff
@@ -58,13 +58,66 @@ class ThreadController
...
+     *
+     */
+    virtual void Deinit(void) = 0;
+
+    /**
+     * The decontructor.
```
superwhd, line 114:
```suggestion
     * The destructor.
```

src/agent/application.cpp
Irving-cl: Yes, ideally we hope all the components work with the interface `ThreadController` instead of the implementation class. But at this stage, there are too many dependencies on `RcpHost` so we still let them depend on `RcpHost`. Later we will gradually change the dependence of `RcpHost` to `ThreadController`.

src/trel_dnssd/trel_dnssd.hpp
Irving-cl: Has updated to use reference.

src/ncp/rcp_host.hpp
Irving-cl: SG. Let's do it in a seperate PR.

src/ncp/thread_controller.cpp
Irving-cl: SG! updated.

src/ncp/rcp_host.cpp
Irving-cl: I agree. 

The difficulty here is that `otSysInit` actually contains 3 steps:
1. `platformInit`
2. create otInstance 
3. `platformSetUp`

And now we do need `platformInit` to run at the very beginning so that we know the co-processor type. However we do not have access to `platformInit`. We can only call `otSysInit` which includes creating the otInstance.

When we do something like `Reset`. We also need to call `otSysInit` again. This caused the problem we are facing now.

I think we can refactor `otSysInit` and extract `platformSpinelManagerInit` out of it as a new posix otSys Api. And at the beginning we only call `platformSpinelManagerInit` to know the co-processor type. Then `otSysInit` will still be called later. It will also try to call `platformSpinelManagerInit` but it finds that it has been initialized and won't initialize it again. Thoughts? @superwhd

Irving-cl: I have submitted a change in openthread: https://github.com/openthread/openthread/pull/10344

With this change, we don't need to call `otSysInit` and create otInstance when trying to determine the co-processor type. Please take a look at that PR first.

src/agent/application.cpp
Irving-cl: As discussed offline, we still try to make these members be initialized in the initliazer list as possible in the future when other components can work under RCP mode.

superwhd (approved):
LGTM overall

abtink (approved):
LGTM.

src/ncp/thread_controller.hpp
```diff
@@ -58,13 +59,66 @@ class ThreadController
...
+     *
+     * This is a factory method that will decide which implementation class will be created.
+     *
+     * @param[in]   aInterfaceName          A string of the Thread interface name.
+     * @param[in]   aRadioUrls              The radio URLs (can be IEEE802.15.4 or TREL radio).
+     * @param[in]   aBackboneInterfaceName  The Backbone network interface name.
+     * @param[in]   aDryRun                 TRUE to indicate dry-run mode. FALSE otherwise.
+     * @param[in]   aEnableAutoAttach       Whether or not to automatically attach to the saved network.
+     *
+     * @retval Non-null OpenThread Controller instance.
```
jwhui, line 73:
```suggestion
     * @returns Non-null OpenThread Controller instance.
```

src/agent/application.cpp
```diff
@@ -60,98 +60,40 @@ Application::Application(const std::string               &aInterfaceName,
 #else
     , mBackboneInterfaceName(aBackboneInterfaceNames.empty() ? "" : aBackboneInterfaceNames.front())
 #endif
+    , mHost(Ncp::ThreadController::CreateInstance(mInterfaceName.c_str(),
```
jwhui, line 63: nit:
```suggestion
    , mHost(Ncp::ThreadController::Create(mInterfaceName.c_str(),
```

src/agent/application.cpp
```diff
@@ -60,98 +60,40 @@ Application::Application(const std::string               &aInterfaceName,
...
-#endif
-#if OTBR_ENABLE_DBUS_SERVER && OTBR_ENABLE_BORDER_AGENT
-    mDBusAgent = MakeUnique<DBus::DBusAgent>(*mHost, *mPublisher);
-#endif
-
-    OT_UNUSED_VARIABLE(aRestListenAddress);
-    OT_UNUSED_VARIABLE(aRestListenPort);
+    if (mHost->GetCoprocessorType() == OT_COPROCESSOR_RCP)
+    {
+        ConstructRcpMode(aRestListenAddress, aRestListenPort);
```
jwhui, line 77: nit: to be consistent?
```suggestion
        CreateRcpMode(aRestListenAddress, aRestListenPort);
```

jwhui (approved):

PR #2331, opened 2024-06-15
[docker] add ability to set debug-level

### LINKED ISSUE #2323: Cannot disable Spinel debug logging
I have built using 
```OTBR_OPTIONS=-DOT_LOG_LEVEL=INFO ...```

However my resultant output still contains a ton of 
```
otbr                 | Jun  7 13:18:18 c66b5ad37c98 otbr-agent[158]: 00:00:44.796 [D] P-RadioSpinel-: Sent spinel frame, flg:0x2, iid:0, tid:13, cmd:PROP_VALUE_SET, key:PHY_CHAN, channel:11
otbr                 | Jun  7 13:18:18 c66b5ad37c98 otbr-agent[158]: 00:00:44.796 [D] P-RadioSpinel-: Wait response: tid=13 key=33

```
Is there a bug in the cmake stuff? I don't know how to fix it but it seems it does not pass the OT_LOG_LEVEL into the third_party openthread build which is where the radio_spinel.cpp code is.

### PR DESCRIPTION
**Pull request by krbvroc1**
Passing OTBR_OPTIONS=-DOT_LOG_LEVEL=INFO does not actually change the debug level as expected. Instead a -d command line option is required.

The current docker_entrypoint.sh file hard codes -d7 (Debug level) on the command line.

This PR changes the hard coded value to a ```--debug-level``` command line argument like the existing ```--radio_url```, ```--interface``` options.

This allows setting the debug level in the docker-compose file.

Closes #2323

### DIFF
--- a/etc/docker/docker_entrypoint.sh
+++ b/etc/docker/docker_entrypoint.sh
@@ -58,6 +58,11 @@ function parse_args()
                 shift
                 shift
                 ;;
+            --debug-level)
+                DEBUG_LEVEL=$2
+                shift
+                shift
+                ;;
             *)
                 shift
                 ;;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: etc/docker/docker_entrypoint.sh [hunk 1]
```diff
@@ -58,6 +58,11 @@ function parse_args()
                 shift
                 shift
                 ;;
+            --debug-level)
```
superwhd, line 61: Shall we name it 'log-level' instead?

krbvroc1: I don't think so. It maps directly to the 'debug-level' or '-d' command line option of otbr-agent.

https://github.com/openthread/ot-br-posix/blob/32462a163c609f7ed06af852d0998d21b96c5bf9/src/agent/main.cpp#L90

jwhui (approved):
Thanks! 👍🏼

PR #2350, opened 2024-06-28
[ncp] add ncp spinel and implement GetDeviceRole

Irving-cl:
This PR implements NcpSpinel for the host to interact with the host.

In this PR, the first method `GetDeviceRole` is implemented and used
in NcpHost to get the device role.

### DIFF
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -1,0 +1,299 @@
+/*
+ *  Copyright (c) 2024, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#define OTBR_LOG_TAG "NcpSpinel"
+
+#include "ncp_spinel.hpp"
+
+#include <stdarg.h>
+
+#include <openthread/thread.h>
+
+#include "common/code_utils.hpp"
+#include "common/logging.hpp"
+#include "lib/spinel/spinel.h"
+#include "lib/spinel/spinel_driver.hpp"
+
+namespace otbr {
+namespace Ncp {
+
+static constexpr char kSpinelDataUnpackFormat[] = "CiiD";
+
+NcpSpinel::NcpSpinel(void)
+    : mSpinelDriver(nullptr)
+    , mCmdTidsInUse(0)
+    , mCmdNextTid(1)
+    , mDeviceRole(OT_DEVICE_ROLE_DISABLED)
+    , mGetDeviceRoleHandler(nullptr)
+{
+    memset(mWaitingKeyTable, 0xff, sizeof(mWaitingKeyTable));
+}
+
+void NcpSpinel::Init(ot::Spinel::SpinelDriver &aSpinelDriver)
+{
+    mSpinelDriver = &aSpinelDriver;
+    mSpinelDriver->SetFrameHandler(&HandleReceivedFrame, &HandleSavedFrame, this);
+}
+
+void NcpSpinel::Deinit(void)
+{
+    mSpinelDriver = nullptr;
+}
+
+void NcpSpinel::GetDeviceRole(GetDeviceRoleHandler aHandler)
+{
+    otError      error = OT_ERROR_NONE;
+    spinel_tid_t tid   = GetNextTid();
+    va_list      args;
+
+    error = mSpinelDriver->SendCommand(SPINEL_CMD_PROP_VALUE_GET, SPINEL_PROP_NET_ROLE, tid, nullptr, args);
+    if (error != OT_ERROR_NONE)
+    {
+        FreeTid(tid);
+        mTaskRunner.Post([aHandler, error](void) { aHandler(error, OT_DEVICE_ROLE_DISABLED); });
+        ExitNow();
+    }
+
+    mWaitingKeyTable[tid] = SPINEL_PROP_NET_ROLE;
+    mGetDeviceRoleHandler = aHandler;
+
+exit:
+    return;
+}
+
+otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDataLen, const char *aPackFormat, ...)
+{
+    otbrError      error = OTBR_ERROR_NONE;
+    spinel_ssize_t unpacked;
+    va_list        args;
+
+    va_start(args, aPackFormat);
+    unpacked = spinel_datatype_vunpack(aDataIn, aDataLen, aPackFormat, args);
+    va_end(args);
+
+    VerifyOrExit(unpacked > 0, error = OTBR_ERROR_PARSE);
+
+exit:
+    return error;
+}
+
+void NcpSpinel::HandleReceivedFrame(const uint8_t *aFrame,
+                                    uint16_t       aLength,
+                                    uint8_t        aHeader,
+                                    bool          &aSave,
+                                    void          *aContext)
+{
+    static_cast<NcpSpinel *>(aContext)->HandleReceivedFrame(aFrame, aLength, aHeader, aSave);
+}
+
+void NcpSpinel::HandleReceivedFrame(const uint8_t *aFrame, uint16_t aLength, uint8_t aHeader, bool &aShouldSaveFrame)
+{
+    spinel_tid_t tid = SPINEL_HEADER_GET_TID(aHeader);
+
+    if (tid == 0)
+    {
+        HandleNotification(aFrame, aLength);
+    }
+    else if (tid < kMaxTids)
+    {
+        HandleResponse(tid, aFrame, aLength);
+    }
+    else
+    {
+        otbrLogCrit("Received unexpected tid: %u", tid);
+    }
+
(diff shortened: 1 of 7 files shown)


src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,283 @@
...
+    case SPINEL_PROP_NET_ROLE:
+    {
+        spinel_net_role_t role;
+
+        unpacked = spinel_datatype_unpack(aBuffer, aLength, "C", &role);
+        VerifyOrExit(unpacked > 0, error = OTBR_ERROR_PARSE);
+
+        mDeviceRole = SpinelRoleToDeviceRole(role);
+
+        otbrLogInfo("Device role changed to %s", otThreadDeviceRoleToString(mDeviceRole));
```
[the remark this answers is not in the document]
Irving-cl: `LogSpinelFrame` is already called in `SpinelDriver` (https://github.com/openthread/openthread/blob/main/src/lib/spinel/spinel_driver.cpp#L279), which means the these frames will be logged (no matter in NCP mode or RCP mode).

But LogSpinelFrame will only show when the log level is debug. Probably we can keep the log here or simply remove it. Thoughts?

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,283 @@
...
+    VerifyOrExit(unpacked > 0, error = OT_ERROR_PARSE);
+
+    VerifyOrExit(cmd == SPINEL_CMD_PROP_VALUE_IS && key == mWaitingKeyTable[aTid], error = OT_ERROR_INVALID_STATE);
+
+    switch (key)
+    {
+    case SPINEL_PROP_NET_ROLE:
+    {
+        spinel_net_role_t spinelRole;
+        unpacked = spinel_datatype_unpack(data, len, "i", &spinelRole);
```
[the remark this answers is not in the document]
Irving-cl: Updated.

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,283 @@
...
+        break;
+    }
+
+exit:
+    FreeTid(aTid);
+    if (error == OT_ERROR_INVALID_STATE)
+    {
+        otbrLogCrit("Received unexpected response with cmd:%u, key:%u, waiting key:%u for tid:%u", cmd, key,
+                    mWaitingKeyTable[aTid], aTid);
+    }
```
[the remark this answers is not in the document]
Irving-cl: Updated.

And in the cases where parsing fails (`VerifyOrExit` jumps to the `exit`), we still needs to call the handler with error code. So I introduced `FailureHandler`. In the handling of different keys, we set the `FailureHandler` first. And if the parsing isn't successful, the original handler will still be called in the end.

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,283 @@
...
+    unpacked = spinel_datatype_unpack(aFrame, aLength, "CiiD", &header, &cmd, &key, &data, &len);
+    VerifyOrExit(unpacked > 0, error = OT_ERROR_PARSE);
+
+    VerifyOrExit(cmd == SPINEL_CMD_PROP_VALUE_IS && key == mWaitingKeyTable[aTid], error = OT_ERROR_INVALID_STATE);
+
+    switch (key)
+    {
+    case SPINEL_PROP_NET_ROLE:
+    {
+        spinel_net_role_t spinelRole;
```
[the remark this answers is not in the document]
Irving-cl: Updated.

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,283 @@
...
+{
+    otError      error = OT_ERROR_NONE;
+    spinel_tid_t tid   = GetNextTid();
+    va_list      args;
+
+    error = mSpinelDriver->SendCommand(SPINEL_CMD_PROP_VALUE_GET, SPINEL_PROP_NET_ROLE, tid, nullptr, args);
+    if (error != OT_ERROR_NONE)
+    {
+        FreeTid(tid);
+        aHandler(error, OT_DEVICE_ROLE_DISABLED);
```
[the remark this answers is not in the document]
Irving-cl: Thanks! Updated to use post to call the handler.

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,283 @@
...
+    case SPINEL_PROP_NET_ROLE:
+    {
+        spinel_net_role_t spinelRole;
+        unpacked = spinel_datatype_unpack(data, len, "i", &spinelRole);
+
+        mDeviceRole = SpinelRoleToDeviceRole(spinelRole);
+
+        if (mGetDeviceRoleHandler)
+        {
+            mGetDeviceRoleHandler(error, mDeviceRole);
```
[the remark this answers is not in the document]
Irving-cl: Updated.

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,291 @@
...
+    : mSpinelDriver(nullptr)
+    , mCmdTidsInUse(0)
+    , mCmdNextTid(1)
+    , mDeviceRole(OT_DEVICE_ROLE_DISABLED)
+    , mGetDeviceRoleHandler(nullptr)
+{
+    memset(mWaitingKeyTable, 0xff, sizeof(mWaitingKeyTable));
+}
+
+NcpSpinel::~NcpSpinel(void) = default;
```
abtink, line 55: Better not including destructor definition if using `=default`. 
Can lead to poor code optimization by some tool chains (e.g. see [link](https://andreasfertig.blog/2023/11/why-you-shouldnt-provide-an-empty-destructor/))

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,291 @@
...
+void NcpSpinel::HandleResponse(spinel_tid_t aTid, const uint8_t *aFrame, uint16_t aLength)
+{
+    spinel_prop_key_t key;
+    spinel_size_t     len = 0;
+    spinel_ssize_t    unpacked;
+    uint8_t          *data = nullptr;
+    uint32_t          cmd;
+    uint8_t           header;
+    otError           error = OT_ERROR_NONE;
+    FailureHandler    failureHandler;
```
abtink, line 158: Does `FailureHandler` have a default constructor to set as `null` or empty?

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+    if (error == OT_ERROR_INVALID_STATE)
+    {
+        otbrLogCrit("Received unexpected response with cmd:%u, key:%u, waiting key:%u for tid:%u", cmd, key,
+                    mWaitingKeyTable[aTid], aTid);
+    }
+    else if (error == OT_ERROR_PARSE)
+    {
+        otbrLogCrit("Error parsing response with tid:%u", aTid);
+        failureHandler(error);
```
abtink, line 194: Should we ehck if `faiureHandler` is not `null`/empty? Can we call `std::function` it if it is empty?
Not sure if we can... https://en.cppreference.com/w/cpp/utility/functional/function/operator()

src/ncp/ncp_spinel.cpp
abtink: Check`std::function<>`. Seems like it does init as empty.

abtink (approved):
LGTM. Thanks. 
Couple of questions/suggestions below

src/ncp/ncp_spinel.cpp
Irving-cl: Thanks! Though it inits as empty, I think we'd better to explicitly init it for better readability. Have added an explicit initialization.

src/ncp/ncp_spinel.cpp
Irving-cl: Yeah we should add a check. I've added a check here.

src/ncp/ncp_spinel.cpp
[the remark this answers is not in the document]
zhanglongxia: Thanks for the explanation. I think we can keep the log here.

zhanglongxia (approved):
LGTM

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,292 @@
...
+    error = mSpinelDriver->SendCommand(SPINEL_CMD_PROP_VALUE_GET, SPINEL_PROP_NET_ROLE, tid, nullptr, args);
+    if (error != OT_ERROR_NONE)
+    {
+        FreeTid(tid);
+        mTaskRunner.Post([aHandler, error](void) { aHandler(error, OT_DEVICE_ROLE_DISABLED); });
+        ExitNow();
+    }
+
+    mWaitingKeyTable[tid] = SPINEL_PROP_NET_ROLE;
+    mGetDeviceRoleHandler = aHandler;
```
superwhd, line 81: What's the expected behavior if there's already a `mGetDeviceRoleHandler`?

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,292 @@
...
+{
+    spinel_prop_key_t key;
+    spinel_size_t     len = 0;
+    spinel_ssize_t    unpacked;
+    uint8_t          *data = nullptr;
+    uint32_t          cmd;
+    uint8_t           header;
+    otbrError         error = OTBR_ERROR_NONE;
+
+    unpacked = spinel_datatype_unpack(aFrame, aLength, "CiiD", &header, &cmd, &key, &data, &len);
```
superwhd, line 134: Better to make constants for the string literals

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+        otbrLogInfo("NCP last status: %s", spinel_status_to_cstr(status));
+        break;
+    }
+    case SPINEL_PROP_NET_ROLE:
+    {
+        spinel_net_role_t role;
+
+        unpacked = spinel_datatype_unpack(aBuffer, aLength, "C", &role);
+        VerifyOrExit(unpacked > 0, error = OTBR_ERROR_PARSE);
```
superwhd, line 221: nit: may wrap the function `spinel_datatype_unpack` and let it return an OT error.

src/ncp/ncp_spinel.hpp
```diff
@@ -0,0 +1,140 @@
...
+     *
+     */
+    void GetDeviceRole(GetDeviceRoleHandler aHandler);
+
+private:
+    using FailureHandler = std::function<void(otError)>;
+
+    static constexpr uint8_t kMaxTids = 16;
+
+    template <typename Function, typename... Args> void CallAndClear(Function &aFunc, Args &&...aArgs)
```
superwhd, line 100: It could be static

src/ncp/ncp_spinel.cpp
Irving-cl: Updated.

src/ncp/ncp_spinel.cpp
Irving-cl: Considering the user of `NcpSpinel` is `NcpHost` which is also internal code, I prefer letting the new one override the existing one and there will be only one call to the handler. (The previous one will be discarded). I added some description in the API doc.

src/ncp/ncp_spinel.cpp
```diff
@@ -0,0 +1,299 @@
...
+    VerifyOrExit(cmd == SPINEL_CMD_PROP_VALUE_IS && key == mWaitingKeyTable[aTid], error = OTBR_ERROR_INVALID_STATE);
+
+    switch (key)
+    {
+    case SPINEL_PROP_NET_ROLE:
+    {
+        spinel_net_role_t spinelRole;
+        failureHandler = [this](otError aError) { CallAndClear(mGetDeviceRoleHandler, aError, mDeviceRole); };
+
+        SuccessOrExit(error = SpinelDataUnpack(data, len, "i", &spinelRole));
```
superwhd, line 181: nit: May also make `"i"` a constant.

src/ncp/ncp_spinel.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * @param[in]  aHandler   A handler to return the role.
+     *
+     */
+    void GetDeviceRole(GetDeviceRoleHandler aHandler);
+
+private:
+    using FailureHandler = std::function<void(otError)>;
+
+    static constexpr uint8_t kMaxTids = 16;
+    static const char       *kSpinelDataUnpackFormat;
```
superwhd, line 102: Do you need this constant to be part of the class?

superwhd (approved):
LGTM overall 👍

src/ncp/ncp_spinel.hpp
```diff
@@ -0,0 +1,145 @@
...
+     * Returns the Co-processor version string.
+     *
+     */
+    const char *GetCoprocessorVersion(void) { return mSpinelDriver->GetVersion(); }
+
+    /**
+     * This method gets the device role and return the role through the handler.
+     *
+     * If this method is called again before the handler is called, the previous handler will be
+     * overriden and there will be only one call to the latest handler.
```
jwhui, line 91:
```suggestion
     * overridden and there will be only one call to the latest handler.
```

jwhui (approved):

PR #2353, opened 2024-07-03
[tests] enhance expect test framework for ncp_mode

Irving-cl:
This PR enhances the expect test framework for testing NCP mode:
1. Add util methods to simplify the tests
2. Allow the bash script to run specific test for easier local testing
3. Build ot-cli node and allow create CLI node in the tests

PR #2355, opened 2024-07-03
[controller] add join and leave api

Irving-cl:
This PR adds 'Join' and 'Leave' APIs to `ThreadHost` and implements them under NCP mode.

The PR also adds an expect test to test the two APIs under NCP mode.

### DIFF
--- a/src/dbus/server/dbus_thread_object_ncp.cpp
+++ b/src/dbus/server/dbus_thread_object_ncp.cpp
@@ -57,6 +57,12 @@ otbrError DBusThreadObjectNcp::Init(void)
     RegisterAsyncGetPropertyHandler(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_PROPERTY_DEVICE_ROLE,
                                     std::bind(&DBusThreadObjectNcp::AsyncGetDeviceRoleHandler, this, _1));
 
+    RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_JOIN_METHOD,
+                   std::bind(&DBusThreadObjectNcp::JoinHandler, this, _1));
+    RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_LEAVE_NETWORK_METHOD,
+                   std::bind(&DBusThreadObjectNcp::LeaveHandler, this, _1));
+
+    SuccessOrExit(error = Signal(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_SIGNAL_READY, std::make_tuple()));
 exit:
     return error;
 }
@@ -86,6 +92,40 @@ exit:
     {
         aRequest.ReplyOtResult(error);
     }
+}
+
+void DBusThreadObjectNcp::JoinHandler(DBusRequest &aRequest)
+{
+    std::vector<uint8_t>     dataset;
+    otOperationalDatasetTlvs activeOpDatasetTlvs;
+    otError                  error = OT_ERROR_NONE;
+
+    auto args = std::tie(dataset);
+
+    SuccessOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args), error = OT_ERROR_INVALID_ARGS);
+
+    VerifyOrExit(dataset.size() <= sizeof(activeOpDatasetTlvs.mTlvs), error = OT_ERROR_INVALID_ARGS);
+    std::copy(dataset.begin(), dataset.end(), activeOpDatasetTlvs.mTlvs);
+    activeOpDatasetTlvs.mLength = dataset.size();
+
+    mHost.Join(activeOpDatasetTlvs, [aRequest](otError aError, const std::string &aErrorInfo) mutable {
+        OT_UNUSED_VARIABLE(aErrorInfo);
+        aRequest.ReplyOtResult(aError);
+    });
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        aRequest.ReplyOtResult(error);
+    }
+}
+
+void DBusThreadObjectNcp::LeaveHandler(DBusRequest &aRequest)
+{
+    mHost.Leave([aRequest](otError aError, const std::string &aErrorInfo) mutable {
+        OT_UNUSED_VARIABLE(aErrorInfo);
+        aRequest.ReplyOtResult(aError);
+    });
 }
 
 } // namespace DBus
--- a/src/ncp/async_task.cpp
+++ b/src/ncp/async_task.cpp
@@ -1,0 +1,99 @@
+/*
+ *  Copyright (c) 2024, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "async_task.hpp"
+
+#include <assert.h>
+#include <memory>
+
+#include "common/code_utils.hpp"
+
+namespace otbr {
+namespace Ncp {
+
+AsyncTask::AsyncTask(const ResultHandler &aResultHandler)
+    : mResultHandler(aResultHandler)
+{
+}
+
+AsyncTask::~AsyncTask()
+{
+    if (!mNext)
+    {
+        if (mResultHandler)
+        {
+            mResultHandler(OT_ERROR_FAILED, "AsyncTask ends without setting any result.");
+        }
+    }
+}
+
+void AsyncTask::Run(void)
+{
+    SetResult(OT_ERROR_NONE, "");
+}
+
+void AsyncTask::SetResult(otError aError, const std::string &aErrorInfo)
+{
+    if (mNext)
+    {
+        if (aError == OT_ERROR_NONE)
+        {
+            mThen(std::move(mNext));
+        }
+        else
+        {
+            mNext->SetResult(aError, aErrorInfo);
+        }
+        mThen = nullptr;
+    }
+    else
+    {
+        mResultHandler(aError, aErrorInfo);
+        mResultHandler = nullptr;
+    }
+}
+
+AsyncTaskPtr &AsyncTask::First(const ThenHandler &aFirst)
+{
+    assert(mNext == nullptr);
+
+    return Then(aFirst);
+}
+
+AsyncTaskPtr &AsyncTask::Then(const ThenHandler &aThen)
+{
+    assert(mNext == nullptr);
+
+    mNext = std::make_shared<AsyncTask>(mResultHandler);
+    mThen = aThen;
+
+    return mNext;
+}
+
+} // namespace Ncp
+} // namespace otbr
--- a/src/ncp/async_task.hpp
+++ b/src/ncp/async_task.hpp
@@ -1,0 +1,120 @@
+/*
+ *  Copyright (c) 2024, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *  2. Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *  3. Neither the name of the copyright holder nor the
+ *     names of its contributors may be used to endorse or promote products
+ *     derived from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *  ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes definitions for chained async task.
+ *   The utility class is used to support the usage of a `Then`-style chained async operations.
+ */
+
+#ifndef OTBR_AGENT_ASYNC_TASK_HPP_
+#define OTBR_AGENT_ASYNC_TASK_HPP_
+
+#include <functional>
+#include <memory>
+
+#include <openthread/error.h>
+
+namespace otbr {
+namespace Ncp {
+
+class AsyncTask;
+using AsyncTaskPtr = std::shared_ptr<AsyncTask>;
+
+class AsyncTask
+{
+public:
+    using ThenHandler   = std::function<void(AsyncTaskPtr)>;
+    using ResultHandler = std::function<void(otError, const std::string &)>;
+
+    /**
+     * Constructor.
+     *
+     * @param[in]  The error handler called when the result is not OT_ERROR_NONE;
+     *
+     */
+    AsyncTask(const ResultHandler &aResultHandler);
+
+    /**
+     * Destructor.
+     *
+     */
+    ~AsyncTask(void);
+
+    /**
+     * Trigger the initial action of the chained async operations.
+     *
+     * This method should be called to trigger the chained async operations.
+     *
+     */
+    void Run(void);
+
+    /**
+     * Set the result of the previous async operation.
+     *
+     * This method should be called when the result of the previous async operation is ready.
+     * This method will pass the result to next operation.
+     *
+     * @param[in] aError  The result for the previous async operation.
+     *
+     */
+    void SetResult(otError aError, const std::string &aErrorInfo);
+
+    /**
+     * Set the initial operation of the chained async operations.
+     *
+     * @param[in] aFirst  A reference to a function object for the initial action.
+     *
+     * @returns  A shared pointer to a AsyncTask object created in this method.
+     *
+     */
+    AsyncTaskPtr &First(const ThenHandler &aFirst);
+
+    /**
+     * Set the next operation of the chained async operations.
+     *
+     * @param[in] aThen  A reference to a function object for the next action.
+     *
+     * @returns A shared pointer to a AsyncTask object created in this method.
+     *
+     */
+    AsyncTaskPtr &Then(const ThenHandler &aThen);
+
+private:
+    union
+    {
+        ThenHandler   mThen;          // Only valid when `mNext` is not nullptr
+        ResultHandler mResultHandler; // Only valid when `mNext` is nullptr
+    };
+    AsyncTaskPtr mNext;
+};
+
+} // namespace Ncp
+} // namespace otbr
+
+#endif // OTBR_AGENT_ASYNC_TASK_HPP_
--- a/src/ncp/ncp_host.cpp
+++ b/src/ncp/ncp_host.cpp
@@ -30,12 +30,16 @@
 
 #include "ncp_host.hpp"
 
+#include <memory>
+
 #include <openthread/error.h>
 #include <openthread/thread.h>
 
 #include <openthread/openthread-system.h>
 
 #include "lib/spinel/spinel_driver.hpp"
+
+#include "ncp/async_task.hpp"
 
 namespace otbr {
 namespace Ncp {
@@ -83,6 +87,31 @@ void NcpHost::Deinit(void)
 {
     mNcpSpinel.Deinit();
     otSysDeinit();
+}
+
+void NcpHost::Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const AsyncResultReceiver &aReceiver)
+{
+    AsyncTaskPtr task;
+    auto errorHandler = [aReceiver](otError aError, const std::string &aErrorInfo) { aReceiver(aError, aErrorInfo); };
+
+    task = std::make_shared<AsyncTask>(errorHandler);
+    task->First([this, aActiveOpDatasetTlvs](AsyncTaskPtr aNext) {
+            mNcpSpinel.DatasetSetActiveTlvs(aActiveOpDatasetTlvs, std::move(aNext));
+        })
+        ->Then([this](AsyncTaskPtr aNext) { mNcpSpinel.Ip6SetEnabled(true, std::move(aNext)); })
+        ->Then([this](AsyncTaskPtr aNext) { mNcpSpinel.ThreadSetEnabled(true, std::move(aNext)); });
+    task->Run();
+}
+
+void NcpHost::Leave(const AsyncResultReceiver &aReceiver)
+{
+    AsyncTaskPtr task;
+    auto errorHandler = [aReceiver](otError aError, const std::string &aErrorInfo) { aReceiver(aError, aErrorInfo); };
+
+    task = std::make_shared<AsyncTask>(errorHandler);
+    task->First([this](AsyncTaskPtr aNext) { mNcpSpinel.ThreadDetachGracefully(std::move(aNext)); })
+        ->Then([this](AsyncTaskPtr aNext) { mNcpSpinel.ThreadErasePersistentInfo(std::move(aNext)); });
+    task->Run();
 }
 
 void NcpHost::Process(const MainloopContext &aMainloop)
--- a/src/ncp/ncp_host.hpp
+++ b/src/ncp/ncp_host.hpp
@@ -86,6 +86,8 @@ public:
     ~NcpHost(void) override = default;
 
     // ThreadHost methods
+    void Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const AsyncResultReceiver &aReceiver) override;
+    void Leave(const AsyncResultReceiver &aReceiver) override;
     CoprocessorType GetCoprocessorType(void) override { return OT_COPROCESSOR_NCP; }
     const char     *GetCoprocessorVersion(void) override;
     const char     *GetInterfaceName(void) const override { return mConfig.mInterfaceName; }
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -32,12 +32,17 @@
 
 #include <stdarg.h>
 
+#include <algorithm>
+
+#include <openthread/dataset.h>
 #include <openthread/thread.h>
 
 #include "common/code_utils.hpp"
 #include "common/logging.hpp"
 #include "lib/spinel/spinel.h"
+#include "lib/spinel/spinel_decoder.hpp"
 #include "lib/spinel/spinel_driver.hpp"
+#include "lib/spinel/spinel_helper.hpp"
 
 namespace otbr {
 namespace Ncp {
@@ -48,15 +53,20 @@ NcpSpinel::NcpSpinel(void)
     : mSpinelDriver(nullptr)
     , mCmdTidsInUse(0)
     , mCmdNextTid(1)
+    , mNcpBuffer(mTxBuffer, kTxBufferSize)
+    , mEncoder(mNcpBuffer)
+    , mIid(SPINEL_HEADER_INVALID_IID)
     , mPropsObserver(nullptr)
 {
-    memset(mWaitingKeyTable, 0xff, sizeof(mWaitingKeyTable));
+    std::fill_n(mWaitingKeyTable, SPINEL_PROP_LAST_STATUS, sizeof(mWaitingKeyTable));
+    memset(mCmdTable, 0, sizeof(mCmdTable));
 }
 
 void NcpSpinel::Init(ot::Spinel::SpinelDriver &aSpinelDriver, PropsObserver &aObserver)
 {
     mSpinelDriver  = &aSpinelDriver;
     mPropsObserver = &aObserver;
+    mIid           = mSpinelDriver->GetIid();
     mSpinelDriver->SetFrameHandler(&HandleReceivedFrame, &HandleSavedFrame, this);
 }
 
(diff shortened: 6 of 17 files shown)


Irving-cl:
This PR depends on the test enhancement in #2353 .

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +71,71 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
+    auto handle_1 = [this, handle_2](otError aError) {
+        if (aError != OT_ERROR_NONE)
+        {
+            CallAsyncReceivers(mJoinReceivers, aError, "Failed to set active dataset!");
+        }
+        else
+        {
+            mNcpSpinel.Ip6SetEnabled(true, handle_2);
+        }
+    };
```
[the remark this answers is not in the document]
superwhd: I think you can declare a class `AsyncResultReceiver` and let it store what the next receiver is. When a receiver is notified, you add the next action's receiver into the receiver list. In this way you can make chained actions.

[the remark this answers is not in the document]
wgtdkp: This is obviously not manageable. Can the NcpSpinel APIs be blocking?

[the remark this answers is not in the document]
Irving-cl: @wgtdkp 

> This is obviously not manageable. Can the NcpSpinel APIs be blocking?

Per our initial discussion, we decided to make all spinel operations for NCP as non-blocking. Because the blocking operations block the whole program, which we don't want to see.

We have an alternative: Add new spinel commands which bundle these operations together. For example, a command for 'join' is sent to NCP. And NCP will set active dataset, enable IP6, start thread and then respond to the host. The downside is that we need to add new spinel command. And if we don't update NCP, the behavior is hard to change. But I think this is still an option as we planned to make NCP updateable.

What do you mean by 'not manageable'? I prefer that we first find some better way to interpret the async process into code.

src/ncp/ncp_spinel.cpp
```diff
@@ -102,6 +111,130 @@ otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDat
...
+                                    SPINEL_DATATYPE_STRUCT_S(SPINEL_DATATYPE_UINT_PACKED_S SPINEL_DATATYPE_DATA_S)
+                                        SPINEL_DATATYPE_STRUCT_S(SPINEL_DATATYPE_UINT_PACKED_S SPINEL_DATATYPE_UINT16_S
+                                                                     SPINEL_DATATYPE_UINT8_S SPINEL_DATATYPE_UINT8_S),
+        SPINEL_PROP_DATASET_ACTIVE_TIMESTAMP, dataset.mActiveTimestamp.mSeconds, SPINEL_PROP_NET_NETWORK_KEY,
+        dataset.mNetworkKey.m8, sizeof(dataset.mNetworkKey.m8), SPINEL_PROP_NET_NETWORK_NAME, dataset.mNetworkName.m8,
+        SPINEL_PROP_NET_XPANID, dataset.mExtendedPanId.m8, sizeof(dataset.mExtendedPanId.m8),
+        SPINEL_PROP_IPV6_ML_PREFIX, &addrMlPrefix, OT_IP6_PREFIX_BITSIZE, SPINEL_PROP_MAC_15_4_PANID, dataset.mPanId,
+        SPINEL_PROP_PHY_CHAN, dataset.mChannel, SPINEL_PROP_NET_PSKC, dataset.mPskc.m8, sizeof(dataset.mPskc.m8),
+        SPINEL_PROP_DATASET_SECURITY_POLICY, dataset.mSecurityPolicy.mRotationTime, flagsSecurityPolicy[0],
+        flagsSecurityPolicy[1]);
```
superwhd, line 148: - It's kind of counter-intuitive that the items at the same level have different indentions. Not sure if this is a defect of the clang-format when dealing with macros.
- There might be too many repetitions of `SPINEL_DATATYPE` and `SPINEL_PROP` which makes it harder to notice what the actual data is.

Irving-cl: Yeah this is affected by clang-format... I don't find any good ways to handle this.

abtink: In the OpenThread NCP code, we use the `Spinel::Decoder` class, which offers convenient helper methods for decoding and reading different data types (`Read{Type}()`) from the received Spinel frames.

- This approach simplifies the code and enhances readability.
- It also helps prevent potential errors where the input variables passed might not align with the expected `SPINEL_DATATYPE` specified in the format string (e.g., passing a `uint32_t` with `SPINEL_DATA_TYPE_UINT16`).

We could consider adopting the same approach here as well. We also have the `Spinel::Encoder` class for constructing spinel frames.

Irving-cl: I have updated to to `Spinel::Encoder`. I think it's much nicer now. Please take a look at the usage of `Spinel::Encoder`.

src/ncp/ncp_host.cpp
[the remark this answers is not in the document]
Irving-cl: I've updated to use a 'Then' style to implement the `Join` method. I introduced a utilitiy class `ChainedAsyncResult` which is basically as @superwhd mentioned.

The code is like:
```
    ChainedAsyncResultPtr result;
    bool                  isAlready = !mJoinReceivers.empty();

    VerifyOrExit(!isAlready);

    result = std::make_shared<ChainedAsyncResult>();
    result
        ->Then([this](otError aError, ChainedAsyncResultPtr aNext) {
            if (aError != OT_ERROR_NONE)
            {
                CallAsyncReceivers(mJoinReceivers, aError, "Failed to set active dataset!");
            }
            else
            {
                mNcpSpinel.Ip6SetEnabled(true, aNext);
            }
        })
        ->Then([this](otError aError, ChainedAsyncResultPtr aNext) {
            if (aError != OT_ERROR_NONE)
            {
                CallAsyncReceivers(mJoinReceivers, aError, "Failed to enable the network interface!");
            }
            else
            {
                mNcpSpinel.ThreadSetEnabled(true, aNext);
            }
        })
        ->Then([this](otError aError, ChainedAsyncResultPtr aNext) {
            OTBR_UNUSED_VARIABLE(aNext);
            CallAsyncReceivers(mJoinReceivers, aError,
                               aError == OT_ERROR_NONE ? "Success" : "Failed to enable the Thread network!");
        });

exit:
    mJoinReceivers.push_back(aReceiver);
    if (!isAlready)
    {
        mNcpSpinel.DatasetSetActiveTlvs(aActiveOpDatasetTlvs, result);
    }
```
What do you think?

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +75,76 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
+            else
+            {
+                mNcpSpinel.ThreadSetEnabled(true, aNext);
+            }
+        })
+        ->Then([this](otError aError, ChainedAsyncResultPtr aNext) {
+            OTBR_UNUSED_VARIABLE(aNext);
+            CallAsyncReceivers(mJoinReceivers, aError,
+                               aError == OT_ERROR_NONE ? "Success" : "Failed to enable the Thread network!");
+        });
```
superwhd, line 111: The order of `Then` is reversed from the actual execution order. Is it possible to use the execution order?

superwhd: I misread this. Never mind.

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +75,76 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
+
+exit:
+    mJoinReceivers.push_back(aReceiver);
+    if (!isAlready)
+    {
+        mNcpSpinel.DatasetSetActiveTlvs(aActiveOpDatasetTlvs, result);
```
superwhd, line 117: Could this be put at the beginning of the chain of `Then`?

Irving-cl: Thanks! This is great. I've added a method `First` and `Finally` to write the first and last action. In this way we can get intuitive processes.

src/ncp/ncp_host.cpp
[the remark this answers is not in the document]
wgtdkp: > Per our initial discussion, we decided to make all spinel operations for NCP as non-blocking. Because the blocking operations block the whole program, which we don't want to see.

Current OpenThread code is already implemented with blocking radio RPCs, isn't it?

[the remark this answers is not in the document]
Irving-cl: > Current OpenThread code is already implemented with blocking radio RPCs, isn't it?
Yes, but it's actually not a desired model. (It uses two `select`) The intial detailed discussion is [here](https://github.com/openthread/openthread/pull/9898#issuecomment-1984699024).

For RCP the blocking model can work because the radio operations are all small and quick to finish. When it comes to NCP, the operations can have much larger delay. For example, `DetachGracefully`. And a blocking model means the whole program will block on that operation. Then all the components in otbr-agent, including the platform things like netif, infrastructure if will all be blocked. So our idea is that, since we start NCP host from scratch, we use non-blocking model.

And the ThreadHost APIs are also async. It seems that the major concern is that the implementation code is not very readable. I think that with introduction of the utility class and the `Then` style operations, it's acceptable. Thoughts?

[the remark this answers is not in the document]
wgtdkp: The new result->Then()->Then() style is not readable neither, can we avoid it if possible?

> For RCP the blocking model can work because the radio operations are all small and quick to finish. When it comes to NCP, the operations can have much larger delay. For example, DetachGracefully. 

Can we make "setThreadEnabled()" blocking and make "DetachGracefully()" async? DetachGracefully() sends traffic to a peer device and waits for response, so it makes sense for it to be async.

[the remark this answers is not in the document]
wgtdkp: My concerns are around readability and maintenance. If there is a way to shorten this code to around three lines (because the blocking calls need only three lines), it will be great!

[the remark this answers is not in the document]
Irving-cl: @abtink @jwhui 

Do you have any preference on the implementation of the `Join` here? There are 3 options now:
1. Use the 'then' style chained async operations. (The current implementation). The major concern is the code readibility.
2. Adds a block mechanism for `NcpSpinel` and make the three operations (`SetDatasetActive`, `Ip6SetEnabled`, `ThreadStart`) as blocking methods. Will use the same approach as `RadioSpinel`.
3. Add a new spinel command which combines the 3 operations and let the NCP handle the new command.

Or do you have any other suggestions?

src/ncp/ncp_spinel.cpp
abtink: Thank you! Looks great 😃

src/ncp/ncp_host.cpp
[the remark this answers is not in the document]
abtink: Thanks guys.

I don't have any strong preferences or suggestions. Each model has its advantages and disadvantages. The `then` model seems like a good option.

Just to add another idea to the mix, we could also consider using co-routines (or classes that mimic co-routine behavior). `wpantund` used a co-routine-style mechanism called proto-thread, but I wouldn't necessarily recommend it here due to its potential complexity.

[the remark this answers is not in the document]
Irving-cl: Hostly coroutine is something new to me so I learned some basic things of it. And I think coroutine should be the ultimate solution for our case. 

As I understand, if we use coroutine, it will work like: we make the async ThreadHost APIs all as coroutines. And when it executes to spinel action, the coroutine is suspended and the flow returns to the caller of the Thread Host API. And then the code can continue to execute. And when a spinel response is obtained, the coroutine is resumed and execute again. This seems to be best solution that can give consideration to both performance and code readabilty maintainence.

The problem is how to use coroutine in `ot-br-posix`. The native c++ coroutine is available in c++20 and so cannot be used in `ot-br-posix` now. I noticed `boost` also has [coroutine2](https://www.boost.org/doc/libs/1_85_0/libs/coroutine2/doc/html/coroutine2/intro.html) and can be used with c++11. And boost is actually already used in `ot-br-posix` (used by `otbr-web`). So I think this is the most possbile choice. `proto-thread` may also be considered.

But anyway I hope we can first move forward with the current implementation to make otbr-agent can run with NCP. And since these are implementations (and no changes on NCP), we update and change them anytime. I'll follow up to seek a coroutine solution. Thoughts? @wgtdkp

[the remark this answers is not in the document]
wgtdkp: I am not okay with the fact that you need to write around 30 lines to just implement what can be implemented with three lines:

```
SuccessOrExit(error = setDataset());
SuccessOrExit(error = setThreadEnabled());
SuccessOrExit(error = setInterfaceUp());
```

Can you try boost coroutine and see how it works?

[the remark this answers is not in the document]
Irving-cl: Thanks! @wgtdkp 

I've further shorten the implementation. Now it's less than 10 lines:
```
    result = std::make_shared<ChainedAsyncResult>(errorHandler);
    result
        ->First([this, &aActiveOpDatasetTlvs](ChainedAsyncResultPtr aNext) {
            mNcpSpinel.DatasetSetActiveTlvs(aActiveOpDatasetTlvs, aNext);
        })
        ->Then([this](ChainedAsyncResultPtr aNext) { mNcpSpinel.Ip6SetEnabled(true, aNext); })
        ->Then([this](ChainedAsyncResultPtr aNext) { mNcpSpinel.ThreadSetEnabled(true, aNext); });
    result->Run();
```

---

I also tried to use `boost coroutine2`. It can be easily added to the project and build. However there are some difficulties to embedded it with our main loop.

If we use coroutine, the flow will be like:
1. Code executes a spinel operation, this starts a new 'context' and the action is suspend.
2. Then switch to the **main context** to process various events and tasks.

However in our current model, the call to spinel operation itself is from the main context. And there is no way to suspend the current main context while continuing to execute the main context.

In the introduction of boost.coroutine, it says
> Boost.Asio provides with its new asynchronous result feature a new framework combining event-driven model and coroutines, hiding the complexity of event-driven programming and permitting the style of classic sequential code. 

And gives an example:
```
void session(boost::asio::io_service& io_service){
    // construct TCP-socket from io_service
    boost::asio::ip::tcp::socket socket(io_service);

    try{
        for(;;){
            // local data-buffer
            char data[max_length];

            boost::system::error_code ec;

            // read asynchronous data from socket
            // execution context will be suspended until
            // some bytes are read from socket
            std::size_t length=socket.async_read_some(
                    boost::asio::buffer(data),
                    boost::asio::yield[ec]);
            if (ec==boost::asio::error::eof)
                break; //connection closed cleanly by peer
            else if(ec)
                throw boost::system::system_error(ec); //some other error

            // write some bytes asynchronously
            boost::asio::async_write(
                    socket,
                    boost::asio::buffer(data,length),
                    boost::asio::yield[ec]);
            if (ec==boost::asio::error::eof)
                break; //connection closed cleanly by peer
            else if(ec)
                throw boost::system::system_error(ec); //some other error
        }
    } catch(std::exception const& e){
        std::cerr<<"Exception: "<<e.what()<<"\n";
    }
}
```
I think in this example, with asio::yield, it transfer the flow to the main context which runs `io_service`. So if we want to use in that way we need to replace the current mainloop with asio. So I don't think it's worthwhile to do it now.

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -95,5 +101,39 @@ void DBusThreadObjectNcp::ReplyAsyncGetProperty(DBusRequest &aRequest, const std
...
 
+void DBusThreadObjectNcp::JoinHandler(DBusRequest &aRequest)
+{
+    std::vector<uint8_t>     dataset;
+    otOperationalDatasetTlvs activeOpDatasetTlvs;
+    otError                  error = OT_ERROR_NONE;
+
+    auto args = std::tie(dataset);
+
+    VerifyOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args) == OTBR_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
```
superwhd, line 112:
```suggestion
    SuccessOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args), error = OT_ERROR_INVALID_ARGS);
```

src/ncp/chained_async_result.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+ChainedAsyncResultPtr &ChainedAsyncResult::First(const ThenHandler &aFirst)
+{
+    return Then(aFirst);
+}
+
+ChainedAsyncResultPtr &ChainedAsyncResult::Then(const ThenHandler &aThen)
+{
+    mThen = aThen;
```
superwhd, line 79: I think there should be an assertion that the `mThen` is not yet set.

src/ncp/ncp_host.hpp
```diff
@@ -63,7 +63,9 @@ class NcpHost : public MainloopProcessor, public ThreadHost
     ~NcpHost(void) override = default;
 
     // ThreadHost methods
-    void            GetDeviceRole(const DeviceRoleHandler aHandler) override;
+    void GetDeviceRole(const DeviceRoleHandler aHandler) override;
+    void Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const AsyncResultReceiver aReceiver) override;
+    void Leave(const AsyncResultReceiver aReceiver) override;
```
superwhd, line 68: For the receiver parameters, I'd suggest to either pass by const reference or pass by value.

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +75,50 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
+    mJoinReceivers.push_back(aReceiver);
+    VerifyOrExit(!isAlready);
+
+    result = std::make_shared<ChainedAsyncResult>(errorHandler);
+    result
+        ->First([this, &aActiveOpDatasetTlvs](ChainedAsyncResultPtr aNext) {
+            mNcpSpinel.DatasetSetActiveTlvs(aActiveOpDatasetTlvs, aNext);
+        })
+        ->Then([this](ChainedAsyncResultPtr aNext) { mNcpSpinel.Ip6SetEnabled(true, aNext); })
+        ->Then([this](ChainedAsyncResultPtr aNext) { mNcpSpinel.ThreadSetEnabled(true, aNext); });
```
superwhd, line 95: nit: `std::move(aNext)`

src/ncp/chained_async_result.cpp
```diff
@@ -0,0 +1,85 @@
...
+        if (mErrorHandler)
+        {
+            mErrorHandler(aError, aErrorInfo);
+        }
+    }
+}
+
+ChainedAsyncResultPtr &ChainedAsyncResult::First(const ThenHandler &aFirst)
+{
+    return Then(aFirst);
```
superwhd, line 74: I think there should be an assertion that the mThen is not yet set.

src/ncp/chained_async_result.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <openthread/error.h>
+
+namespace otbr {
+namespace Ncp {
+
+class ChainedAsyncResult;
+using ChainedAsyncResultPtr = std::shared_ptr<ChainedAsyncResult>;
+
+class ChainedAsyncResult
```
superwhd, line 49: Probably the name can be shorter. You may consider removing `Chained`.

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +75,50 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
+    auto                  errorHandler = [this](otError aError, const std::string &aErrorInfo) {
+        CallAsyncReceivers(mJoinReceivers, aError, aErrorInfo);
+    };
+
+    mJoinReceivers.push_back(aReceiver);
+    VerifyOrExit(!isAlready);
+
+    result = std::make_shared<ChainedAsyncResult>(errorHandler);
+    result
+        ->First([this, &aActiveOpDatasetTlvs](ChainedAsyncResultPtr aNext) {
```
superwhd, line 91: It seems unsafe to capture the `aActiveOpDatasetTlvs` by reference

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +75,50 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
 
+void NcpHost::Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const AsyncResultReceiver aReceiver)
+{
+    ChainedAsyncResultPtr result;
+    bool                  isAlready    = !mJoinReceivers.empty();
+    auto                  errorHandler = [this](otError aError, const std::string &aErrorInfo) {
+        CallAsyncReceivers(mJoinReceivers, aError, aErrorInfo);
+    };
+
+    mJoinReceivers.push_back(aReceiver);
```
superwhd, line 86: What if this call is using a different operational dataset from another ongoing call?

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +75,50 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
+    VerifyOrExit(!isAlready);
+
+    result = std::make_shared<ChainedAsyncResult>(errorHandler);
+    result
+        ->First([this, &aActiveOpDatasetTlvs](ChainedAsyncResultPtr aNext) {
+            mNcpSpinel.DatasetSetActiveTlvs(aActiveOpDatasetTlvs, aNext);
+        })
+        ->Then([this](ChainedAsyncResultPtr aNext) { mNcpSpinel.Ip6SetEnabled(true, aNext); })
+        ->Then([this](ChainedAsyncResultPtr aNext) { mNcpSpinel.ThreadSetEnabled(true, aNext); });
+    result->Run();
```
superwhd, line 96: How do you ensure that `result` is still alive when it goes out of this method's scope?

src/ncp/ncp_spinel.cpp
```diff
@@ -102,6 +112,111 @@ otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDat
...
 
+void NcpSpinel::DatasetSetActiveTlvs(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs,
+                                     ChainedAsyncResultPtr           aAsyncResult)
+{
+    otError      error        = OT_ERROR_NONE;
+    EncodingFunc encodingFunc = [this, &aActiveOpDatasetTlvs] {
+        return EncodeDatasetSetActiveTlvs(aActiveOpDatasetTlvs);
+    };
+
+    otbrLogWarning("DatasetSetActive");
```
superwhd, line 123: The log level seems inappropriate.

src/ncp/ncp_host.cpp
Irving-cl: Good point! What would be the expected behavior on Android? 

Under current implementation, the different operational dataset won't take effect. But it will still get callbacked when previous operation completes.

I think we may probably change to let each call of 'Join' just go through the process. Sometimes if the spinel has already started an operation, then returns `BUSY`. Thoughts?

src/ncp/ncp_host.cpp
Irving-cl: So long as we called any NcpSpinel methods that takes this shared pointer. Because in the NcpSpinel methods, the shared pointer will be stored until there is a response. When there is a response, the shared pointer will be reset and the object will be released.

src/ncp/ncp_host.cpp
Irving-cl: Not sure if this is needed here `aNext` is a shared_pointer. If we use `std::move`, as I understand, it just destructs the `aNext` within the scope of this lambda function right? I think maybe we should move it when calling `mThen(mNext)` in `SetResult` => `mThen(std::move(mNext))`.  Thoughts?

src/ncp/ncp_host.cpp
Irving-cl: Have changed the implementation to allow a consequent call (even previous one is not allowed). Both calls will be get callbacked anyway.

src/ncp/async_result.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <openthread/error.h>
+
+namespace otbr {
+namespace Ncp {
+
+class AsyncResult;
+using AsyncResultPtr = std::shared_ptr<AsyncResult>;
+
+class AsyncResult
```
abtink, line 49: This class allows us to link a sequence of actions (`ThenHandler()` functions) that are called sequentially, one after another, asynchronously. This happens when the result of the currently active action is determined, i.e., `SetResult()` is called on it.

The name `AsyncResult` didn't immediately convey this to me. I initially assumed it referred to the "result of an asynchronous" operation, not the action/operation itself.

How about we call it `AsyncAction` (or `AsyncTask`)?

abtink (approved):
Looks great. Thanks. 
I have one suggestion below.

src/ncp/async_result.hpp
Irving-cl: Fully agree! Changed to `AsyncTask` which corresponds to the method `run`. This also avoids the confusion with `AsyncResultReceiver` for ThreadHost APIs.

src/ncp/async_task.cpp
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+    else
+    {
+        if (mErrorHandler)
+        {
+            mErrorHandler(aError, aErrorInfo);
+        }
+    }
+}
```
abtink, line 71: One thought came to me related to this class:

`ThenHandler` receives an `aAsyncTaskPtr` (which is a `shared_ptr`). We use `std::move(mNext)` when calling the `ThenHandler`, so we relinquish our ownership of this `shared_ptr` and pass it to the `ThenHandler`.

We assume the `ThenHandler` will keep the `shared_ptr` alive, and at some point, `SetResult()` will be called on the `aAsyncTaskPtr` instance for it to then proceed with the next chained task and/or report the outcome by calling `ErrorHandler`.

However, what if the `ThenHandler` lets the passed-in `aAsyncTaskPtr` go out of scope (get deleted and freed)? In that case, if I'm not mistaken, we would never call `ErrorHandler`. Is this the desired behavior? I believe we should still call `ErrorHandler` in this scenario to report the outcome (perhaps as failed?).

We could consider handling such a case in the destructor of `AsyncTask`. One thing to consider is that when a chain of linked `AsyncTask` objects is being destroyed, the destructor will be called on each one (`mNext` will be destroyed before `AsyncTask` itself is destroyed), and we need to ensure `ErrorHandler` is called only once (across all chained `AsyncTask`s).

src/ncp/async_task.hpp
```diff
@@ -0,0 +1,111 @@
...
+class AsyncTask
+{
+public:
+    using ThenHandler  = std::function<void(AsyncTaskPtr)>;
+    using ErrorHandler = std::function<void(otError, const std::string &)>;
+
+    /**
+     * Constructor.
+     *
+     * @param[in]  The error handler called when the result is not OT_ERROR_NONE;
```
abtink, line 58: I think if all the chained tasks are finished successfully we will call `mErrorHandler` with `OT_ERROR_NONE` to indicate this.
```c++
void AsyncTask::SetResult(otError aError, const std::string &aErrorInfo)
{
    if (aError == OT_ERROR_NONE)
    {
        if (mThen)
        {
            mThen(std::move(mNext));
        }
        else if (mErrorHandler)
        {
            // The Final case
            mErrorHandler(aError, aErrorInfo);
        }
    }
```

How about renaming it to `ResultHandler` (it is not just used for reporting errors but successful completion of executing all the chained tasks?)

src/ncp/async_task.cpp
Irving-cl: Good catch! I think the essence is that we need handle the case where an AsyncTask has ended its lifetime but `SetResult` hasn't been called. 

I added a step to clear the `ResultHandler` when calling `SetResult` because it moves forward anyway and, in the destructor, if `ResultHandler` is not `nullptr`, it's called with a failure. (`SetResult` is never called for this object).

src/ncp/async_task.hpp
Irving-cl: Have renamed it.

src/ncp/async_task.hpp
```diff
@@ -0,0 +1,117 @@
...
+     *
+     */
+    AsyncTask(const ResultHandler &aResultHandler);
+
+    /**
+     * Destructor.
+     *
+     */
+    ~AsyncTask();
```
superwhd, line 67:
```suggestion
    ~AsyncTask(void);
```
nit

src/ncp/ncp_host.cpp
```diff
@@ -71,6 +75,31 @@ void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
...
+{
+    AsyncTaskPtr task;
+    auto errorHandler = [aReceiver](otError aError, const std::string &aErrorInfo) { aReceiver(aError, aErrorInfo); };
+
+    task = std::make_shared<AsyncTask>(errorHandler);
+    task->First([this, aActiveOpDatasetTlvs](AsyncTaskPtr aNext) {
+            mNcpSpinel.DatasetSetActiveTlvs(aActiveOpDatasetTlvs, aNext);
+        })
+        ->Then([this](AsyncTaskPtr aNext) { mNcpSpinel.Ip6SetEnabled(true, aNext); })
+        ->Then([this](AsyncTaskPtr aNext) { mNcpSpinel.ThreadSetEnabled(true, aNext); });
```
superwhd, line 88: You may consider applying `std::move()` to every `aNext` to convert the copy construction into move construction.

src/ncp/ncp_spinel.cpp
```diff
@@ -102,6 +112,106 @@ otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDat
...
+    VerifyOrExit(mThreadDetachGracefullyTask == nullptr);
+
+    error                       = SetProperty(SPINEL_PROP_NET_LEAVE_GRACEFULLY, encodingFunc);
+    mThreadDetachGracefullyTask = aAsyncTask;
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        mTaskRunner.Post(
+            [aAsyncTask, error](void) { aAsyncTask->SetResult(error, "Failed to enable the network interface!"); });
```
superwhd, line 185:
```suggestion
            [aAsyncTask, error](void) { aAsyncTask->SetResult(error, "Failed to detach gracefully!"); });
```

src/ncp/ncp_spinel.cpp
```diff
@@ -102,6 +112,106 @@ otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDat
...
+    }
+    return;
+}
+
+void NcpSpinel::ThreadDetachGracefully(AsyncTaskPtr aAsyncTask)
+{
+    otError      error        = OT_ERROR_NONE;
+    EncodingFunc encodingFunc = [] { return OT_ERROR_NONE; };
+
+    VerifyOrExit(mThreadDetachGracefullyTask == nullptr);
```
superwhd, line 176: When this check fails, it returns an `OT_ERROR_NONE` without setting the result. Is this expected?

src/ncp/ncp_spinel.cpp
```diff
@@ -102,6 +112,106 @@ otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDat
...
+    va_list      args;
+
+    VerifyOrExit(mThreadErasePersistentInfoTask == nullptr);
+
+    error = mSpinelDriver->SendCommand(SPINEL_CMD_NET_CLEAR, SPINEL_PROP_LAST_STATUS, tid, nullptr, args);
+    if (error != OT_ERROR_NONE)
+    {
+        FreeTid(tid);
+        mTaskRunner.Post(
+            [aAsyncTask, error](void) { aAsyncTask->SetResult(error, "Failed to enable the Thread network!"); });
```
superwhd, line 203: Seems the error message is not relevant.

src/ncp/ncp_spinel.cpp
```diff
@@ -102,6 +112,106 @@ otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDat
...
+
+void NcpSpinel::ThreadErasePersistentInfo(AsyncTaskPtr aAsyncTask)
+{
+    otError      error = OT_ERROR_NONE;
+    spinel_tid_t tid   = GetNextTid();
+    va_list      args;
+
+    VerifyOrExit(mThreadErasePersistentInfoTask == nullptr);
+
+    error = mSpinelDriver->SendCommand(SPINEL_CMD_NET_CLEAR, SPINEL_PROP_LAST_STATUS, tid, nullptr, args);
```
superwhd, line 198: What about using `SuccessOrExit` and put the error handling in `exit` section?

src/ncp/ncp_spinel.cpp
```diff
@@ -266,6 +425,180 @@ spinel_tid_t NcpSpinel::GetNextTid(void)
...
+{
+    static constexpr uint8_t kObtainNetworkKeyMask        = 1 << 7;
+    static constexpr uint8_t kNativeCommissioningMask     = 1 << 6;
+    static constexpr uint8_t kRoutersMask                 = 1 << 5;
+    static constexpr uint8_t kExternalCommissioningMask   = 1 << 4;
+    static constexpr uint8_t kCommercialCommissioningMask = 1 << 2;
+    static constexpr uint8_t kAutonomousEnrollmentMask    = 1 << 1;
+    static constexpr uint8_t kNetworkKeyProvisioningMask  = 1 << 0;
+    static constexpr uint8_t kTobleLinkMask               = 1 << 7;
+    static constexpr uint8_t kNonCcmRoutersMask           = 1 << 6;
```
superwhd, line 477: Probably document the byte index of the masks because there are same mask values in different bytes representing different meanings.

src/ncp/ncp_spinel.cpp
```diff
@@ -266,6 +425,180 @@ spinel_tid_t NcpSpinel::GetNextTid(void)
...
+{
+    otError              error = OT_ERROR_NONE;
+    otOperationalDataset dataset;
+    otIp6Address         addrMlPrefix;
+    uint8_t              flagsSecurityPolicy[2];
+
+    SuccessOrExit(error = otDatasetParseTlvs(&aActiveOpDatasetTlvs, &dataset));
+    GetFlagsFromSecurityPolicy(&dataset.mSecurityPolicy, flagsSecurityPolicy, sizeof(flagsSecurityPolicy));
+    memcpy(addrMlPrefix.mFields.m8, dataset.mMeshLocalPrefix.m8, 8);
+    memset(addrMlPrefix.mFields.m8 + 8, 0, 8);
```
superwhd, line 548: better to avoid magic numbers

src/ncp/ncp_spinel.cpp
```diff
@@ -233,7 +369,14 @@ void NcpSpinel::HandleValueIs(spinel_prop_key_t aKey, const uint8_t *aBuffer, ui
...
     }
 
+    case SPINEL_PROP_NET_LEAVE_GRACEFULLY:
+    {
+        CallAndClear(mThreadDetachGracefullyTask, OT_ERROR_NONE);
+        break;
+    }
+
     default:
+        otbrLogWarning("Received uncognized key:%u", aKey);
```
superwhd, line 379:
```suggestion
        otbrLogWarning("Received uncognized key: %u", aKey);
```
nit

src/ncp/ncp_spinel.hpp
```diff
@@ -124,19 +201,41 @@ class NcpSpinel
...
+
+    void GetFlagsFromSecurityPolicy(const otSecurityPolicy *aSecurityPolicy, uint8_t *aFlags, uint8_t aFlagsLength);
+
+    otError EncodeDatasetSetActiveTlvs(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs);
+
     ot::Spinel::SpinelDriver *mSpinelDriver;
     uint16_t                  mCmdTidsInUse;              ///< Used transaction ids.
     spinel_tid_t              mCmdNextTid;                ///< Next available transaction id.
     spinel_prop_key_t         mWaitingKeyTable[kMaxTids]; ///< The property keys of ongoing transactions.
+    spinel_command_t mCmdTable[kMaxTids]; ///< The mapping of spinel command and tids when the response is LAST_STATUS.
```
superwhd, line 221: I wonder why this line is not aligned with others? Doesn't the formatter apply here?

src/ncp/ncp_spinel.cpp
Irving-cl: Set to `OT_ERROR_BUSY` now.

src/ncp/ncp_spinel.hpp
Irving-cl: Yeah that's interesting and annoying. I had a few tries and it seems the pretty tries to make it < 120 word by compacting the line instead of starting a new line. So I manually start a new line for the comment.

src/ncp/async_task.hpp
abtink: Thanks.

src/ncp/async_task.cpp
abtink: Thanks.

However, I believe now we may encounter the issue of the `ResultHandler` potentially being called multiple times. We clear `mResultHandler` when we call it on an `AsyncTask`, but other `AsyncTask`s (in the same chain) will still retain their own copy of the handler in their own `mResultHandler`. When the destructor is called on them, they will also call the same `ResultHandler` again.

abtink: Here's my suggestion on this:

We modify `AsyncTask` to track either a `ThenHandler` or a `ResultHandler` (not both - we can use a `union`). An `AsyncTask` instance that tracks a `ResultHandler` will always be the end of a chain of tasks. So, if `mNext` is `nullptr`, this is the last entry in the linked list, and therefore it's tracking a `ResultHandler`; otherwise, it's a `ThenHandler`.

```c++
private:
    union
    {
        ThenHandler   mThen;
        ResultHandler mResultHandler;
    };

    AsyncTaskPtr mNext;
```

We'll need to update `First()` and `Then()` to accommodate this new data model.

`SetResult()` would then check if it's not the last entry (has `mNext`) and call `SetResult()` on `mNext`. Otherwise, if it's the last entry, it invokes `ResultHandler` and clears it.

The `~AsyncTask()` destructor can simply call `SetResult()` with the error (in all cases). 

With this model, `ResultHandler` is only invoked from one place.

Thoughts?

src/ncp/ncp_host.cpp
superwhd: The idea here is to improve the efficiency of argument passing by avoid copy constructions.

There are 2 factors here determining what constructor is triggered when passing arguments:
- The type of the argument to be passed in: Here `aNext` is a non-const value.
- The type of the parameter in the function/method declaration: `NcpSpinel::Ip6SetEnabled` takes an `AsyncTaskPtr` by value.

Hence if you pass `aNext` as is, it will be regarded as an lvalue reference and result in a copy construction of `AsyncTaskPtr` when it's passed to `NcpSpinel::Ip6SetEnabled`. 

However, if you pass in `std::move(aNext)`, it will be passed in as an rvalue reference and result in a move construction, which is a bit cheaper.

src/ncp/async_task.cpp
Irving-cl: Thanks a lot! You're correct about this.

I've changed to the union implementation. This actually simplies the logic and saves memory.

The very simple class needs to be taken great care. I also added a unit test `test_async_tasks.cpp` for it.

src/ncp/ncp_host.cpp
Irving-cl: Oh I see, thanks! Have updated to use `std::move`.

superwhd (approved):
LGTM 👍

src/ncp/async_task.cpp
abtink: Great. Thank you.

Irving-cl:
@wgtdkp Could you please take another look? Thanks!

wgtdkp (approved):

src/ncp/async_task.cpp
```diff
@@ -0,0 +1,100 @@
+
```
jwhui, line 1: Remove empty line.

jwhui (approved):

src/ncp/ncp_spinel.cpp
```diff
@@ -236,6 +394,181 @@ spinel_tid_t NcpSpinel::GetNextTid(void)
...
+    otOperationalDataset dataset;
+    otIp6Address         addrMlPrefix;
+    uint8_t              flagsSecurityPolicy[2];
+
+    SuccessOrExit(error = otDatasetParseTlvs(&aActiveOpDatasetTlvs, &dataset));
+    GetFlagsFromSecurityPolicy(&dataset.mSecurityPolicy, flagsSecurityPolicy, sizeof(flagsSecurityPolicy));
+    memcpy(addrMlPrefix.mFields.m8, dataset.mMeshLocalPrefix.m8, sizeof(otMeshLocalPrefix));
+    memset(addrMlPrefix.mFields.m8 + 8, 0, sizeof(otMeshLocalPrefix));
+
+    SuccessOrExit(error = mEncoder.OpenStruct());
```
jwhui, line 520: Ideally, we should always transport Operational Datasets as TLVs. In particular, it should be possible to configure an Operational Dataset where the TLVs are not defined in the current Thread Specification.

Irving-cl: Make sense! I've checked the existing spinel properties and currently there is no TLV version of `SPINEL_PROP_THREAD_ACTIVE_DATASET` or `SPINEL_PROP_THREAD_PENDING_DATASET`. So we first need to add new spinel properties  `SPINEL_PROP_THREAD_ACTIVE_DATASET_TLV` and the handling on NCP side. Shall we use another PR to do the change after I add the new properties? I'll create an issue to track this.

PR #2361, opened 2024-07-07
[telemetry] add `peer_br_count` in InfraLinkInfo

zesonzhang:
- Add new field `peer_br_count` which is the total number of peer border routers on the same infra link
- Also move all InfraLinkInfo section into a new function `RetrieveInfraLinkInfo()`

### DIFF
--- a/src/proto/thread_telemetry.proto
+++ b/src/proto/thread_telemetry.proto
@@ -494,6 +494,8 @@ message TelemetryData {
     optional uint32 link_local_address_count = 5;
     optional uint32 unique_local_address_count = 6;
     optional uint32 global_unicast_address_count = 7;
+    // Indicates how many peer BRs (connected to the same Thread mesh network) are on the infra link.
+    optional uint32 peer_br_count = 8;
   }
 
   // Message to indicate the information of external routes in network data.
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -1429,12 +1429,8 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
 #endif // OTBR_ENABLE_TREL
 
 #if OTBR_ENABLE_BORDER_ROUTING
-        // InfraLinkInfo section.
         RetrieveInfraLinkInfo(*wpanBorderRouter->mutable_infra_link_info());
-
-        // ExternalRoutes section
         RetrieveExternalRouteInfo(*wpanBorderRouter->mutable_external_route_info());
-
 #endif
 
 #if OTBR_ENABLE_SRP_ADVERTISING_PROXY
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -944,7 +944,44 @@ void ThreadHelper::DetachGracefullyCallback(void)
 
 #if OTBR_ENABLE_TELEMETRY_DATA_API
 #if OTBR_ENABLE_BORDER_ROUTING
-void ThreadHelper::RetrieveExternalRouteInfo(threadnetwork::TelemetryData::ExternalRoutes *aExternalRouteInfo)
+void ThreadHelper::RetrieveInfraLinkInfo(threadnetwork::TelemetryData::InfraLinkInfo &aInfraLinkInfo)
+{
+    {
+        otSysInfraNetIfAddressCounters addressCounters;
+        uint32_t                       ifrFlags = otSysGetInfraNetifFlags();
+
+        otSysCountInfraNetifAddresses(&addressCounters);
+
+        aInfraLinkInfo.set_name(otSysGetInfraNetifName());
+        aInfraLinkInfo.set_is_up((ifrFlags & IFF_UP) != 0);
+        aInfraLinkInfo.set_is_running((ifrFlags & IFF_RUNNING) != 0);
+        aInfraLinkInfo.set_is_multicast((ifrFlags & IFF_MULTICAST) != 0);
+        aInfraLinkInfo.set_link_local_address_count(addressCounters.mLinkLocalAddresses);
+        aInfraLinkInfo.set_unique_local_address_count(addressCounters.mUniqueLocalAddresses);
+        aInfraLinkInfo.set_global_unicast_address_count(addressCounters.mGlobalUnicastAddresses);
+    }
+
+    //---- peer_br_count
+    {
+        uint32_t                           count = 0;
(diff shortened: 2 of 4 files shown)


src/utils/thread_helper.cpp
```diff
@@ -944,6 +944,44 @@ void ThreadHelper::DetachGracefullyCallback(void)
...
+void ThreadHelper::RetrieveInfraLinkInfo(threadnetwork::TelemetryData::InfraLinkInfo *aInfraLinkInfo)
+{
+    {
+        otSysInfraNetIfAddressCounters addressCounters;
+        uint32_t                       ifrFlags = otSysGetInfraNetifFlags();
+
+        otSysCountInfraNetifAddresses(&addressCounters);
+
+        aInfraLinkInfo->set_name(otSysGetInfraNetifName());
+        aInfraLinkInfo->set_is_up((ifrFlags & IFF_UP) != 0);
```
morningboata, line 956: Since you are uploading the ifrFlags contents, how about directly upload ifrFlags itself, if there is no sensitive information inside? This will save the upload payload size.

zesonzhang: thanks @morningboata , these block is just the refactor, I think it's better to keep as it original is?  cc. @superwhd

superwhd: I don't think ifr_flags' integer values are the same across platforms. Hence we have to extract every flag using the platform-specific enums before post-processing.

src/utils/thread_helper.cpp
```diff
@@ -944,6 +944,44 @@ void ThreadHelper::DetachGracefullyCallback(void)
...
+
+    //---- infra_br_count
+    {
+        uint8_t                            count = 0;
+        otBorderRoutingPrefixTableIterator iterator;
+        otBorderRoutingRouterEntry         entry;
+
+        otBorderRoutingPrefixTableInitIterator(mInstance, &iterator);
+
+        while (otBorderRoutingGetNextRouterEntry(mInstance, &iterator, &entry) == OT_ERROR_NONE)
```
superwhd, line 972: I don't think `otBorderRoutingGetNextRouterEntry` only returns BRs. It also returns ordinary WiFi/Ethernet routers.

zesonzhang: Thanks for pointing this out! I will update a new version

zesonzhang: updated

src/proto/thread_telemetry.proto
```diff
@@ -494,6 +494,8 @@ message TelemetryData {
     optional uint32 link_local_address_count = 5;
     optional uint32 unique_local_address_count = 6;
     optional uint32 global_unicast_address_count = 7;
+    // Indicates how many peer BRs (connected to the same Thread mesh network) on the infra link.
```
superwhd, line 497:
```suggestion
    // Indicates how many peer BRs (connected to the same Thread mesh network) are on the infra link.
```
nit

superwhd (approved):
LGTM 👍

bukepo (approved):

morningboata (approved):

src/utils/thread_helper.cpp
```diff
@@ -944,6 +944,43 @@ void ThreadHelper::DetachGracefullyCallback(void)
 
 #if OTBR_ENABLE_TELEMETRY_DATA_API
 #if OTBR_ENABLE_BORDER_ROUTING
+void ThreadHelper::RetrieveInfraLinkInfo(threadnetwork::TelemetryData::InfraLinkInfo *aInfraLinkInfo)
```
abtink, line 947: How about using a reference instead of a pointer here? (Since we don't anticipate or handle the case where the passed-in input pointer is `nullptr`.)

src/utils/thread_helper.cpp
```diff
@@ -944,6 +944,43 @@ void ThreadHelper::DetachGracefullyCallback(void)
...
+        aInfraLinkInfo->set_is_running((ifrFlags & IFF_RUNNING) != 0);
+        aInfraLinkInfo->set_is_multicast((ifrFlags & IFF_MULTICAST) != 0);
+        aInfraLinkInfo->set_link_local_address_count(addressCounters.mLinkLocalAddresses);
+        aInfraLinkInfo->set_unique_local_address_count(addressCounters.mUniqueLocalAddresses);
+        aInfraLinkInfo->set_global_unicast_address_count(addressCounters.mGlobalUnicastAddresses);
+    }
+
+    //---- peer_br_count
+    {
+        uint8_t                            count = 0;
```
abtink, line 966: How about using `uint16_t` or `uint32_t` for this to be safe (it's unlikely to have more than 255 BRs, but it is technically possible).

What is the type for this in the `.proto` definition? I believe it's defined as `uint32_t` (so maybe using the same type here for consistency?)

abtink (commented):
Looks good. Thanks.
Couple of suggestions below:

src/utils/thread_helper.cpp
zesonzhang: sg! Thanks for the suggestions!

src/utils/thread_helper.cpp
zesonzhang: Yes, it's uint32 in `.proto`. It sounds reasonable to make it consistent!

abtink (approved):
👍

src/utils/thread_helper.cpp
```diff
@@ -1392,26 +1429,11 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
...
-            infraLinkInfo->set_unique_local_address_count(addressCounters.mUniqueLocalAddresses);
-            infraLinkInfo->set_global_unicast_address_count(addressCounters.mGlobalUnicastAddresses);
-        }
-        // End of InfraLinkInfo section.
+        // InfraLinkInfo section.
+        RetrieveInfraLinkInfo(*wpanBorderRouter->mutable_infra_link_info());
 
         // ExternalRoutes section
-        RetrieveExternalRouteInfo(wpanBorderRouter->mutable_external_route_info());
+        RetrieveExternalRouteInfo(*wpanBorderRouter->mutable_external_route_info());
```
superwhd, line 1436: nit: I think the method names are clear enough so maybe you don't have to document them in comments.

superwhd (approved):

src/utils/thread_helper.cpp
zesonzhang: updated, thanks

jwhui (approved):

PR #2362, opened 2024-07-08
[tests] use more accurate EXPECT_* in tests

superwhd:
This CL replaces `EXPECT_TRUE` and `EXPECT_FALSE` with more accurate `EXPECT_*`s to make the test failures more informative.

This PR also implements `!=` for common types if the `==` operator is overriden.


bukepo (approved):

tests/gtest/test_logging.cpp
```diff
@@ -48,7 +48,7 @@ TEST(Logging, TestLoggingHigherLevel)
 
     char cmd[128];
     snprintf(cmd, sizeof(cmd), "grep '%s.*cool-higher' /var/log/syslog", ident);
-    EXPECT_TRUE(0 != system(cmd));
+    EXPECT_NE(0, system(cmd));
```
wgtdkp, line 51: Put the value being tested on the left side of comparisons.

Having a consistent order throughout this repo makes it easier for code readers to identify which part of a comparison is the computed result and which part is the expectation. This ordering was chosen because it reads more naturally in English, but you can also think of it as an arbitrary choice.

There are more such changes, please change all.
```suggestion
    EXPECT_NE(system(cmd), 0);
```

superwhd: Good idea. Updated.

jwhui (approved):

PR #2368, opened 2024-07-10
[ePSKc] runtime enable/disable EphemeralKey feature

mia1yang:
Set Enabled Dbus API for unit test


src/agent/application.cpp
```diff
@@ -68,9 +68,6 @@ Application::Application(const std::string               &aInterfaceName,
 #if OTBR_ENABLE_MDNS
     , mPublisher(Mdns::Publisher::Create([this](Mdns::Publisher::State aState) { this->HandleMdnsState(aState); }))
 #endif
-#if OTBR_ENABLE_DBUS_SERVER && OTBR_ENABLE_BORDER_AGENT
```
Irving-cl, line 71: Let's still put the construction of `DBusAgent` in initializer list. And let's first add ePSK feature only for RCP. We can change in this way:
1. Do not change `mDBusAgent` here.
2. Add a parameter to `DBusAgent::Init`, which is a pointer to `BorderAgent`.
3. In Application::InitRcpMode, pass `mBorderAgent to` `mDBusAgent.Init()`.
4. In `DBusAgent::Init`, in the case `OT_COPROCESSOR_RCP`, de-reference the pointer and pass it to the constructor of `DBusThreadObjectRcp`.

Then `DBusThreadObjectRcp` can have a reference to `BorderAgent`.

mia1yang: Thanks! Modified according to the comments, could check again.

src/border_agent/border_agent.cpp
```diff
@@ -257,6 +267,8 @@ void BorderAgent::PublishEpskcService()
     otbrLogInfo("Publish meshcop-e service %s.%s.local. port %d", mServiceInstanceName.c_str(),
                 kBorderAgentEpskcServiceType, port);
 
+    VerifyOrExit(IsEpskcEnabled() == true);
```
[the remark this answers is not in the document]
superwhd: BTW I'd suggest to guard the below logic by an `if` block instead of a `VerifyOrExit`.

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1957,6 +1960,19 @@ otError DBusThreadObjectRcp::SetNat64Cidr(DBusMessageIter &aIter)
...
 #endif // OTBR_ENABLE_NAT64
 
+void DBusThreadObjectRcp::SetAdminPasscodeEnabled(DBusRequest &aRequest)
+{
+    otError error = OT_ERROR_NONE;
+    bool    enable;
+    auto    args = std::tie(enable);
+
+    VerifyOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args) == OTBR_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
+    mBorderAgent.SetEpskcEnabled(enable);
```
superwhd, line 1970: I'd suggest to do following instead:

- In ot-br-posix's `BorderAgent`, use [otBorderAgentSetEphemeralKeyCallback](https://github.com/openthread/openthread/blob/20c9bcbc4142fd74e490bd50faa7def9e1b3a823/include/openthread/border_agent.h#L275) to listen to the Epskc state changes and updates the meshcop service accordingly.
- In `DBusThreadObjectRcp::SetAdminPasscodeEnabled`, just enable/disable the Epskc mode by `otBorderAgentSetEphemeralKey`.

In this way you have a single source of truth from OT Core and don't have to maintain the state at ot-br-posix's `BorderAgent`. Also you don't have to let `DBusThreadObjectRcp` control `BorderAgent`.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -371,6 +371,18 @@ class ThreadApiDBus
...
+     * This method sets the Thread Admin Passcode switch.
+     *
+     * @param[in] aEnable  A boolean to enable/disable the TAP.
+     *
+     * @retval ERROR_NONE  Successfully performed the dbus function call
+     * @retval ERROR_DBUS  dbus encode/decode error
+     * @retval ...         OpenThread defined error value otherwise
+     *
+     */
+    ClientError SetAdminPasscodeEnabled(bool aEnabled);
```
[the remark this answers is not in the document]
librasungirl: +1. may be consistent with the ot API signature as https://github.com/openthread/openthread/blob/20c9bcbc4142fd74e490bd50faa7def9e1b3a823/include/openthread/border_agent.h#L208 
 
```suggestion
    ClientError SetEphemeralKeyEnabled(bool aEnabled);
```

src/border_agent/border_agent.cpp
```diff
@@ -130,7 +130,7 @@ struct StateBitmap
         , mBbrIsActive(0)
         , mBbrIsPrimary(0)
         , mThreadRole(kThreadRoleDisabledOrDetached)
-        , mEpskcSupported(OTBR_ENABLE_EPSKC)
+        , mEpskcSupported(0)
```
librasungirl, line 133: How about enable by default if OPENTHREAD_CONFIG_THREAD_VERSION is OT_THREAD_VERSION_1_4?

src/dbus/server/dbus_thread_object_rcp.cpp
librasungirl: This PR is to update the `epskc capability` bit in state bitmap of meshcop service according to feature flags, other than publish/unpublish meshcop-e service according to epskc setting.

src/dbus/server/dbus_agent.cpp
```diff
@@ -71,7 +71,7 @@ void DBusAgent::Init(void)
     {
     case OT_COPROCESSOR_RCP:
         mThreadObject = MakeUnique<DBusThreadObjectRcp>(*mConnection, mInterfaceName,
```
librasungirl, line 73: if RCP object, will it cause feature divergency between RCP and NCP. feature parity between these two modes should be expected? @Irving-cl

src/border_agent/border_agent.hpp
```diff
@@ -126,6 +126,14 @@ class BorderAgent : private NonCopyable
      */
     void SetEnabled(bool aIsEnabled);
 
+    /**
+     * This method enables/disables the Border Agent ePSKc feature.
+     *
+     * @param[in] aIsEnabled  Whether to enable the BA ePSKc feature.
+     *
+     */
+    void SetEpskcEnabled(bool aIsEnabled);
```
librasungirl, line 135: nit: may be consistent with the ot API signature as https://github.com/openthread/openthread/blob/20c9bcbc4142fd74e490bd50faa7def9e1b3a823/include/openthread/border_agent.h#L208

```suggestion
    void SetEphemeralKeyEnabled(bool aIsEnabled);
```

src/dbus/server/introspect.xml
```diff
@@ -231,6 +231,10 @@
       <arg name="enable" type="b" direction="in"/>
     </method>
 
+    <method name="SetAdminPasscodeEnabled">
```
librasungirl, line 234: How about defining a property `EphemeralKeyEnabled` with `readwrite` access instead?

src/dbus/server/dbus_agent.cpp
Irving-cl: NCP currently only has very limited feature and it's not ready to support ePSKc in it now. Let's first implement it in RCP first. I'll work with @mia1yang on this later when it's ready.

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -150,7 +152,8 @@ otbrError DBusThreadObjectRcp::Init(void)
                    std::bind(&DBusThreadObjectRcp::LeaveNetworkHandler, this, _1));
     RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_SET_NAT64_ENABLED_METHOD,
                    std::bind(&DBusThreadObjectRcp::SetNat64Enabled, this, _1));
-
+    RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_SET_ADMIN_PASSCODE_ENABLED_METHOD,
```
Irving-cl, line 155: I think this looks like a `PropertySet` method. So you may consider registering it as a `SetPropertyHandler`. For example, https://github.com/openthread/ot-br-posix/blob/main/src/dbus/server/dbus_thread_object_rcp.cpp#L157

And the implementation of `SetPropertyHandler` will be even simpler. Current approach is fine. It's up to you if to update it.

Irving-cl (approved):
LGTM

src/dbus/server/dbus_thread_object_rcp.cpp
superwhd: Thanks for the context! I misinterpreted this as the API to active/deactivate the Epskc mode. 

Since that's for the capability feature flag, I'd suggest to directly add a new field in [feature flag proto instead of adding a new D-BUS method. New D-BUS method introduces more fragmentations of the feature flag framework.

https://github.com/openthread/ot-br-posix/blob/b7ebdc361c4c051448b1c611e9dde3d20c1bc5c7/src/ncp/rcp_host.cpp#L231C18-L231C30

src/dbus/client/thread_api_dbus.hpp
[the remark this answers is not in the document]
mia1yang: OK. Will use ePSKc mode and obsolete "Admin Passcode mode"

src/dbus/server/introspect.xml
mia1yang: Accept

src/border_agent/border_agent.hpp
mia1yang: Accept. Thanks~

src/border_agent/border_agent.cpp
```diff
@@ -207,6 +207,16 @@ void BorderAgent::SetEnabled(bool aIsEnabled)
     return;
 }
 
+void BorderAgent::SetEpskcEnabled(bool aIsEnabled)
+{
+    VerifyOrExit(IsEpskcEnabled() != aIsEnabled);
+    mIsEpskcEnabled = aIsEnabled;
+    UpdateMeshCopService();
```
abtink, line 214: Do we intend to allow `SetEpskcEnabled()` to be called at any point, especially while e-PSK is actively in use?

Consider this scenario:

* The e-PSK feature is enabled and actively used (e-PSKs are set and the related service is being published).
* While this is happening, `SetEpskcEnabled(false)` is called.

Will the system correctly stop using e-PSK and unpublish the associated service?

While I see that this PR adds code to `PublishEpskcService()` to check `IsEpskcEnabled()` before publishing, I'm not sure how we would handle unpublishing the service in the scenario described above.

abtink: My suggestion is that whenever `mIsEpskcEnabled` transitions from `true` to `false`, we call `otBorderAgentClearEphemeralKey()`. 

* If there's no active e-PSK in use, this call will have no effect.
* If there is an active e-PSK, it will be canceled, and later triggers the `otBorderAgentEphemeralKeyCallback`, which will then lead to unpublishing the service. 

Thoughts?

mia1yang: Thanks for the comments! Will follow the suggestion to modify SetEpskcEnabled code and do verification accordingly.

src/border_agent/border_agent.cpp
```diff
@@ -496,8 +510,9 @@ void BorderAgent::PublishMeshCopService(void)
     // "xa" stands for Extended MAC Address (64-bit) of the Thread Interface of the Border Agent.
     txtList.emplace_back("xa", extAddr->m8, sizeof(extAddr->m8));
 
-    state       = GetStateBitmap(*instance);
-    stateUint32 = htobe32(state.ToUint32());
+    state                 = GetStateBitmap(*instance);
+    state.mEpskcSupported = (IsEpskcEnabled() ? 1 : 0);
```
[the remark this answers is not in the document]
mia1yang: Accept. Thanks!

src/border_agent/border_agent.cpp
librasungirl: To achieve this, you may consider initialise this variable according to the value returned from [otThreadGetVersion](https://github.com/openthread/openthread/blob/aba7aede1476339abe3032a250349d67a4ec1605/include/openthread/thread.h#L252)

src/border_agent/border_agent.cpp
```diff
@@ -207,6 +207,27 @@ void BorderAgent::SetEnabled(bool aIsEnabled)
     return;
 }
 
+void BorderAgent::SetEpskcEnabled(bool aIsEnabled)
+{
+    VerifyOrExit(IsEpskcEnabled() != aIsEnabled);
+    mIsEpskcEnabled = aIsEnabled;
+
+    if (mIsEpskcEnabled)
```
abtink, line 215: Should it be `if (!mIsEpskcEnabled)`?

abtink (approved):
Thanks. Looks good overall.
One questions/suggestion below:

src/border_agent/border_agent.cpp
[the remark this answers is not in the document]
mia1yang: Have removed this ```VerifyOrExit```  after discussion. Thanks!

src/border_agent/border_agent.cpp
mia1yang: Thanks!

src/border_agent/border_agent.cpp
mia1yang: Accept，add in the initiation, let me know there is more concern. thanks~

src/dbus/server/dbus_thread_object_rcp.cpp
mia1yang: Modify according to the suggestion, thanks!

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -371,6 +371,30 @@ class ThreadApiDBus
      */
     ClientError SetNat64Enabled(bool aEnabled);
 
+    /**
+     * This method sets the Ephemeral Key switch.
+     *
+     * @param[in] aEnable  A boolean to enable/disable the TAP.
```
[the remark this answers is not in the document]
mia1yang: Accept thanks!

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1957,6 +1962,29 @@ otError DBusThreadObjectRcp::SetNat64Cidr(DBusMessageIter &aIter)
 }
 #endif // OTBR_ENABLE_NAT64
 
+otError DBusThreadObjectRcp::GetEphemeralKeyEnabled(DBusMessageIter &aIter)
+{
+    otError error = OT_ERROR_NONE;
+
+    VerifyOrExit(DBusMessageEncodeToVariant(&aIter, mBorderAgent.GetEpskcEnabled()) == OTBR_ERROR_NONE,
```
superwhd, line 1969: `SuccessOrExit` should be more convenient in this use case.

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1957,6 +1962,29 @@ otError DBusThreadObjectRcp::SetNat64Cidr(DBusMessageIter &aIter)
...
+exit:
+    return error;
+}
+
+otError DBusThreadObjectRcp::SetEphemeralKeyEnabled(DBusMessageIter &aIter)
+{
+    otError error = OT_ERROR_NONE;
+    bool    enable;
+
+    VerifyOrExit(DBusMessageExtractFromVariant(&aIter, enable) == OTBR_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
```
superwhd, line 1981: `SuccessOrExit` should be more convenient in this use case.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -371,6 +371,30 @@ class ThreadApiDBus
...
+     * This method sets the Ephemeral Key switch.
+     *
+     * @param[in] aEnable  A boolean to enable/disable the Ephemeral Key.
+     *
+     * @retval ERROR_NONE  Successfully performed the dbus function call
+     * @retval ERROR_DBUS  dbus encode/decode error
+     * @retval ...         OpenThread defined error value otherwise
+     *
+     */
+    ClientError SetEphemeralKeyEnabled(bool &aEnabled);
```
superwhd, line 384:
```suggestion
    ClientError SetEphemeralKeyEnabled(bool aEnabled);
```

src/agent/application.cpp
```diff
@@ -303,7 +303,7 @@ void Application::DeinitRcpMode(void)
 void Application::InitNcpMode(void)
 {
 #if OTBR_ENABLE_DBUS_SERVER
-    mDBusAgent->Init();
+    mDBusAgent->Init(*mBorderAgent.get());
```
superwhd, line 306:
```suggestion
    mDBusAgent->Init(*mBorderAgent);
```

src/agent/application.cpp
```diff
@@ -277,7 +277,7 @@ void Application::InitRcpMode(void)
     mRestWebServer->Init();
 #endif
 #if OTBR_ENABLE_DBUS_SERVER
-    mDBusAgent->Init();
+    mDBusAgent->Init(*mBorderAgent.get());
```
superwhd, line 280:
```suggestion
    mDBusAgent->Init(*mBorderAgent);
```

src/dbus/client/thread_api_dbus.hpp
superwhd, line 1: What's the use case of the new methods added to this file? I think generally the methods here are used for writing unit tests.

src/agent/application.cpp
mia1yang: Thanks! Accepted.

src/dbus/client/thread_api_dbus.hpp
mia1yang: accept, thanks~

src/dbus/server/dbus_thread_object_rcp.cpp
mia1yang: Accept. Thanks~

src/dbus/client/thread_api_dbus.hpp
mia1yang: unit tests are added in this PR. Thanks!

src/border_agent/border_agent.cpp
```diff
@@ -207,6 +209,30 @@ void BorderAgent::SetEnabled(bool aIsEnabled)
...
+        otBorderAgentClearEphemeralKey(mHost.GetInstance());
+    }
+
+    UpdateMeshCopService();
+
+exit:
+    return;
+}
+
+bool BorderAgent::GetEpskcEnabled()
```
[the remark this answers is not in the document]
mia1yang: for unit test cases to get the init state.

src/border_agent/border_agent.cpp
```diff
@@ -158,6 +158,8 @@ BorderAgent::BorderAgent(otbr::Ncp::RcpHost &aHost, Mdns::Publisher &aPublisher)
     , mBaseServiceInstanceName(OTBR_MESHCOP_SERVICE_INSTANCE_NAME)
 {
     mHost.AddThreadStateChangedCallback([this](otChangedFlags aFlags) { HandleThreadStateChanged(aFlags); });
+    mIsEpskcEnabled = (otThreadGetVersion() >= 5);
```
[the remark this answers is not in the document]
mia1yang: Will add ON_INIT to decide the initiate value per versions in late CL, add issue to track it.

src/dbus/server/dbus_thread_object_rcp.cpp
mia1yang: done, thanks!

src/border_agent/border_agent.cpp
```diff
@@ -158,6 +158,9 @@ BorderAgent::BorderAgent(otbr::Ncp::RcpHost &aHost, Mdns::Publisher &aPublisher)
     , mBaseServiceInstanceName(OTBR_MESHCOP_SERVICE_INSTANCE_NAME)
 {
     mHost.AddThreadStateChangedCallback([this](otChangedFlags aFlags) { HandleThreadStateChanged(aFlags); });
+    mIsEphemeralKeyEnabled = (otThreadGetVersion() >= 5);
```
superwhd, line 161: As discussed offline, could you use a macro to configure whether it's enabled by default?

src/border_agent/border_agent.cpp
```diff
@@ -207,6 +210,25 @@ void BorderAgent::SetEnabled(bool aIsEnabled)
...
 }
 
+void BorderAgent::SetEphemeralKeyEnabled(bool aIsEnabled)
+{
+    VerifyOrExit(GetEphemeralKeyEnabled() != aIsEnabled);
+    mIsEphemeralKeyEnabled = aIsEnabled;
+
+    if (!mIsEphemeralKeyEnabled)
+    {
+        // If the ePSK feature is enabled, we trigger the clear function
```
superwhd, line 220:
```suggestion
        // If the ePSKc feature is enabled, we trigger the clear function
```

src/border_agent/border_agent.cpp
```diff
@@ -496,8 +518,9 @@ void BorderAgent::PublishMeshCopService(void)
     // "xa" stands for Extended MAC Address (64-bit) of the Thread Interface of the Border Agent.
     txtList.emplace_back("xa", extAddr->m8, sizeof(extAddr->m8));
 
-    state       = GetStateBitmap(*instance);
-    stateUint32 = htobe32(state.ToUint32());
+    state                 = GetStateBitmap(*instance);
+    state.mEpskcSupported = GetEphemeralKeyEnabled();
```
superwhd, line 522: why not put this processing in `GetStateBitmap`?

src/border_agent/border_agent.hpp
```diff
@@ -126,6 +126,20 @@ class BorderAgent : private NonCopyable
...
+    /**
+     * This method enables/disables the Border Agent Ephemeral Key feature.
+     *
+     * @param[in] aIsEnabled  Whether to enable the BA Ephemeral Key feature.
+     *
+     */
+    void SetEphemeralKeyEnabled(bool aIsEnabled);
+
+    /**
+     * This method return the Border Agent Ephemeral Key feature state.
```
superwhd, line 138:
```suggestion
     * This method returns the Border Agent Ephemeral Key feature state.
```

src/dbus/server/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ target_include_directories(otbr-dbus-server PRIVATE
 add_dependencies(otbr-dbus-server otbr-dbus-introspect-header)
 
 target_link_libraries(otbr-dbus-server PUBLIC
+    $<$<BOOL:${OTBR_MDNS}>:otbr-mdns>
```
superwhd, line 53: I wonder why this is necessary. It seems the dependency is on BorderAgent.

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1957,6 +1963,28 @@ otError DBusThreadObjectRcp::SetNat64Cidr(DBusMessageIter &aIter)
 }
 #endif // OTBR_ENABLE_NAT64
 
+otError DBusThreadObjectRcp::GetEphemeralKeyEnabled(DBusMessageIter &aIter)
+{
+    otError error = OT_ERROR_NONE;
+
+    SuccessOrExit(DBusMessageEncodeToVariant(&aIter, mBorderAgent.GetEphemeralKeyEnabled()));
```
superwhd, line 1970:
```suggestion
    SuccessOrExit(error = DBusMessageEncodeToVariant(&aIter, mBorderAgent.GetEphemeralKeyEnabled()));
```

tests/dbus/test_dbus_client.cpp
```diff
@@ -261,6 +261,19 @@ void CheckNat64(ThreadApiDBus *aApi)
 #endif
 }
 
+void CheckEphemeralKey(ThreadApiDBus *aApi)
+{
+    OTBR_UNUSED_VARIABLE(aApi);
```
superwhd, line 266: Do we need this?

tests/dbus/test_dbus_client.cpp
```diff
@@ -261,6 +261,19 @@ void CheckNat64(ThreadApiDBus *aApi)
...
+    bool enabled;
+    bool original;
+
+    TEST_ASSERT(aApi->GetEphemeralKeyEnabled(original) == OTBR_ERROR_NONE);
+    TEST_ASSERT(aApi->SetEphemeralKeyEnabled(!original) == OTBR_ERROR_NONE);
+    TEST_ASSERT(aApi->GetEphemeralKeyEnabled(enabled) == OTBR_ERROR_NONE);
+    TEST_ASSERT(enabled == !original);
+    TEST_ASSERT(aApi->GetEphemeralKeyEnabled(original) == OTBR_ERROR_NONE);
```
superwhd, line 274: As mentioned earlier in a different PR, please use a more deterministic routine. For example, set it to false at the beginning and set it to true later.

src/border_agent/border_agent.cpp
mia1yang: accept, thanks!

src/border_agent/border_agent.hpp
mia1yang: done, thanks!

src/dbus/server/dbus_thread_object_rcp.cpp
mia1yang: done, thanks~

tests/dbus/test_dbus_client.cpp
mia1yang: done, thanks~

tests/dbus/test_dbus_client.cpp
mia1yang: As there are no dependencies on the ePSKc original state, it should be OK to use deterministic way. Thanks~

src/border_agent/border_agent.cpp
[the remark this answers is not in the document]
mia1yang: Current modification will set init to false; later to add  "OT_THREAD_VERSION_1_4" in check, as it's in the core now.

src/dbus/server/CMakeLists.txt
mia1yang: This is due to the build errors introducing by border_agent.hpp which newly added in the dbus; move this check to border_agent.cpp to avoid the build errors and also unnecessary lib in dbus.

src/border_agent/border_agent.cpp
mia1yang: Current modification will set init to false; later to add "OT_THREAD_VERSION_1_4" in check, as it's in the core now.

src/border_agent/border_agent.cpp
```diff
@@ -495,9 +519,10 @@ void BorderAgent::PublishMeshCopService(void)
 
     // "xa" stands for Extended MAC Address (64-bit) of the Thread Interface of the Border Agent.
     txtList.emplace_back("xa", extAddr->m8, sizeof(extAddr->m8));
-
-    state       = GetStateBitmap(*instance);
-    stateUint32 = htobe32(state.ToUint32());
+    state = GetStateBitmap(*instance);
+    // Todo: refact GetStateBitmap to support more parameters to process epskc
```
superwhd, line 523:
```suggestion
    // TODO: refactor GetStateBitmap to support more parameters to process epskc
```
nit

superwhd (approved):
LGTM :+1:

src/border_agent/border_agent.cpp
mia1yang: thanks～

librasungirl (approved):

src/border_agent/border_agent.cpp
```diff
@@ -207,6 +212,25 @@ void BorderAgent::SetEnabled(bool aIsEnabled)
...
+void BorderAgent::SetEphemeralKeyEnabled(bool aIsEnabled)
+{
+    VerifyOrExit(GetEphemeralKeyEnabled() != aIsEnabled);
+    mIsEphemeralKeyEnabled = aIsEnabled;
+
+    if (!mIsEphemeralKeyEnabled)
+    {
+        // If the ePSKc feature is enabled, we trigger the clear function
+        // which will wait session disconnected if it is actively used before
+        // removing ephemeral key and then lead to unpublishing the service.
```
jwhui, line 224:
```suggestion
        // If the ePSKc feature is enabled, we call the clear function
        // which will wait for the session to close if it is in active use before
        // removing ephemeral key and unpublishing the service.
```

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -371,6 +371,30 @@ class ThreadApiDBus
...
+     * @retval ERROR_DBUS  dbus encode/decode error
+     * @retval ...         OpenThread defined error value otherwise
+     *
+     */
+    ClientError SetEphemeralKeyEnabled(bool aEnabled);
+
+    /**
+     * This method gets the Ephemeral Key switch.
+     *
+     * @param[out] aEnable  A boolean of enable/disable for Ephemeral Key state.
```
jwhui, line 389:
```suggestion
     * @param[out] aEnabled  A boolean of enable/disable for Ephemeral Key state.
```

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -371,6 +371,30 @@ class ThreadApiDBus
      */
     ClientError SetNat64Enabled(bool aEnabled);
 
+    /**
+     * This method sets the Ephemeral Key switch.
+     *
+     * @param[in] aEnable  A boolean to enable/disable the Ephemeral Key.
```
jwhui, line 377:
```suggestion
     * @param[in] aEnabled  A boolean to enable/disable the Ephemeral Key.
```

mia1yang: Accept. Thanks！

src/dbus/client/thread_api_dbus.hpp
mia1yang: Accept. Thanks!

src/border_agent/border_agent.cpp
mia1yang: Accept. Thanks!

jwhui (approved):

PR #2372, opened 2024-07-11
[dbus] Add leader takeover and local leader weight API

nand-nor:
Adds DBus APIs to takeover leader role and to set the local leader weight


nand-nor:
This depends on if/when this OT PR merges https://github.com/openthread/openthread/pull/9186

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1754,6 +1771,17 @@ void DBusThreadObjectRcp::LeaveNetworkHandler(DBusRequest &aRequest)
     });
 }
 
+void DBusThreadObjectRcp::TakeoverLeaderHandler(DBusRequest &aRequest)
+{
+    auto    threadHelper = mHost.GetThreadHelper();
+    otError error        = OT_ERROR_NONE;
+
+    SuccessOrExit(error = otThreadBecomeLeader(threadHelper->GetInstance()));
```
bukepo, line 1933: From the API doc, this should only be called when the device is attached. Could you add a check before issuing this call?

nand-nor: Added a check for this in [95dfb9b](https://github.com/openthread/ot-br-posix/commit/95dfb9b6e19e8c8ffb7d67632d21e3b8bb5bccc0)

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1906,6 +1923,19 @@ void DBusThreadObjectRcp::LeaveNetworkHandler(DBusRequest &aRequest)
...
+    auto         threadHelper = mHost.GetThreadHelper();
+    otError      error        = OT_ERROR_NONE;
+    otDeviceRole role         = otThreadGetDeviceRole(threadHelper->GetInstance());
+
+    VerifyOrExit(role != OT_DEVICE_ROLE_DISABLED && role != OT_DEVICE_ROLE_DETACHED, error = OT_ERROR_INVALID_STATE);
+    SuccessOrExit(error = otThreadBecomeLeader(threadHelper->GetInstance()));
+
+exit:
+    aRequest.ReplyOtResult(error);
+}
```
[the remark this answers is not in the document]
nand-nor: @bukepo are you Ok with leaving this as-is per your feedback [here](https://github.com/openthread/ot-br-posix/pull/2372#discussion_r2621830456)? I can remove also

PR #2373, opened 2024-07-12
[docker] add environment variable for HTTP server

bukepo:
This commit adds an environment variable to specify the otbr-web port in docker, so that it can avoid port conflict with other services in host network mode.


superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2378, opened 2024-07-14
[telemetry] move WpanTopoFull section to `RetrieveWpanTopoFull()`

zesonzhang:
- Add `RetrieveWpanTopoFull()` to populate WpanTopoFull section;
- Add `TelemetryGetNeighborTable()`, the neighbor table is used by WpanTopoFull, TopoEntry and LinkMetricsEntry;
- Add `TelemetryGetChildTable()`, the child table is used by both WpanTopoFull and TopoEntry

Also move the including of `<openthread/thread_ftd.h>` from thread_helper.cpp to thread_helper.hpp for `otChildInfo` declaration

### DIFF
--- a/src/utils/thread_helper.cpp
+++ b/src/utils/thread_helper.cpp
@@ -57,7 +57,6 @@
 #if OTBR_ENABLE_SRP_ADVERTISING_PROXY
 #include <openthread/srp_server.h>
 #endif
-#include <openthread/thread_ftd.h>
 #if OTBR_ENABLE_TREL
 #include <openthread/trel.h>
 #endif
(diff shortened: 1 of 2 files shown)


morningboata:
Check why RetrieveWpanTopoFull() is created? Basically we need to apply the changes on Android (with a similar set of methods). We can either keep changes smaller by not introduce this new method, or we shall sync the new methods so we can do the diff comparison easier. Either is OK but effort needs to be kept in mind.

zesonzhang:
> Check why RetrieveWpanTopoFull() is created? Basically we need to apply the changes on Android (with a similar set of methods). We can either keep changes smaller by not introduce this new method, or we shall sync the new methods so we can do the diff comparison easier. Either is OK but effort needs to be kept in mind.

just want to it more easy to read and modify. If this change would make a big effort in Android side, maybe it's not a good time to do this refactor. Any suggestions?

morningboata:
OK to have this refactor, but remember the Android side requires refactor as well.

morningboata (approved):

zesonzhang:
not urgent needed, close this temporarily,  will re-open later

PR #2387, opened 2024-07-18
[host] add network properties class

Irving-cl:
This PR adds a new interface `NetworkProperties` to provide a unified, synchronized way to access network properties for both NCP and RCP.

* For RCP, the property getters are implemented by OT APIs with an OT instance.
* For NCP, the property getters are implemented by the replica of the data that comes from NCP.

This is important for adding NCP support because in some cases the network properties are required by some functionality and we don't want to make an asynchronous call simply to get the property. As one example, in AdvertisingProxy, we need to the MeshLocal Prefix to check if an address is mesh local address. Currently this is done by OT API, which is not available under NCP mode. However, the information is contained in the active dataset and NCP will always notify the host about the change. With `NetworkProperties`, we can always get these properties in a synchronized way.

This PR also replaces a few usage of `otThreadGetDeviceRole` by `NetworkProperties`. In this way we will gradually remove the direct usage of OT APIs and mInstance in ot-br-posix modules. In this way we can develop for both NCP&RCP more easily.

### DIFF
--- a/src/ncp/ncp_host.cpp
+++ b/src/ncp/ncp_host.cpp
@@ -40,6 +40,25 @@
 namespace otbr {
 namespace Ncp {
 
+// =============================== NcpNetworkProperties ===============================
+
+NcpNetworkProperties::NcpNetworkProperties(void)
+    : mDeviceRole(OT_DEVICE_ROLE_DISABLED)
+{
+}
+
+otDeviceRole NcpNetworkProperties::GetDeviceRole(void) const
+{
+    return mDeviceRole;
+}
+
+void NcpNetworkProperties::SetDeviceRole(otDeviceRole aRole)
+{
+    mDeviceRole = aRole;
+}
+
+// ===================================== NcpHost ======================================
+
 NcpHost::NcpHost(const char *aInterfaceName, bool aDryRun)
     : mSpinelDriver(*static_cast<ot::Spinel::SpinelDriver *>(otSysGetSpinelDriver()))
 {
@@ -57,18 +76,13 @@ const char *NcpHost::GetCoprocessorVersion(void)
 void NcpHost::Init(void)
 {
     otSysInit(&mConfig);
-    mNcpSpinel.Init(mSpinelDriver);
+    mNcpSpinel.Init(mSpinelDriver, *this);
 }
 
 void NcpHost::Deinit(void)
 {
     mNcpSpinel.Deinit();
     otSysDeinit();
-}
-
-void NcpHost::GetDeviceRole(DeviceRoleHandler aHandler)
-{
-    mNcpSpinel.GetDeviceRole(aHandler);
 }
 
 void NcpHost::Process(const MainloopContext &aMainloop)
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -48,42 +48,21 @@ NcpSpinel::NcpSpinel(void)
     : mSpinelDriver(nullptr)
     , mCmdTidsInUse(0)
     , mCmdNextTid(1)
-    , mDeviceRole(OT_DEVICE_ROLE_DISABLED)
-    , mGetDeviceRoleHandler(nullptr)
+    , mPropsObserver(nullptr)
 {
     memset(mWaitingKeyTable, 0xff, sizeof(mWaitingKeyTable));
 }
 
-void NcpSpinel::Init(ot::Spinel::SpinelDriver &aSpinelDriver)
+void NcpSpinel::Init(ot::Spinel::SpinelDriver &aSpinelDriver, PropsObserver &aObserver)
 {
-    mSpinelDriver = &aSpinelDriver;
+    mSpinelDriver  = &aSpinelDriver;
+    mPropsObserver = &aObserver;
     mSpinelDriver->SetFrameHandler(&HandleReceivedFrame, &HandleSavedFrame, this);
 }
 
 void NcpSpinel::Deinit(void)
 {
     mSpinelDriver = nullptr;
-}
-
-void NcpSpinel::GetDeviceRole(GetDeviceRoleHandler aHandler)
-{
-    otError      error = OT_ERROR_NONE;
-    spinel_tid_t tid   = GetNextTid();
-    va_list      args  = {};
-
-    error = mSpinelDriver->SendCommand(SPINEL_CMD_PROP_VALUE_GET, SPINEL_PROP_NET_ROLE, tid, nullptr, args);
-    if (error != OT_ERROR_NONE)
-    {
-        FreeTid(tid);
-        mTaskRunner.Post([aHandler, error](void) { aHandler(error, OT_DEVICE_ROLE_DISABLED); });
-        ExitNow();
-    }
-
-    mWaitingKeyTable[tid] = SPINEL_PROP_NET_ROLE;
-    mGetDeviceRoleHandler = aHandler;
-
-exit:
-    return;
 }
 
 otbrError NcpSpinel::SpinelDataUnpack(const uint8_t *aDataIn, spinel_size_t aDataLen, const char *aPackFormat, ...)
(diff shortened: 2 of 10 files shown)


superwhd:
> With NetworkProperties, we can always get these properties in a synchronized way.

The idea is to maintain a cache of the properties at the host (by receiving updates from the callbacks). Right? 

I think 'synchronized' may not be accurate here.

src/ncp/ncp_host.cpp
```diff
@@ -40,6 +40,25 @@
...
+    : mDeviceRole(OT_DEVICE_ROLE_DISABLED)
+{
+}
+
+otDeviceRole NcpNetworkProperties::GetDeviceRole(void) const
+{
+    return mDeviceRole;
+}
+
+void NcpNetworkProperties::UpdateDeviceRole(otDeviceRole aRole)
```
superwhd, line 55:
```suggestion
void NcpNetworkProperties::SetDeviceRole(otDeviceRole aRole)
```

src/ncp/rcp_host.cpp
```diff
@@ -221,10 +235,11 @@ void RcpHost::Init(void)
 #endif
 #endif // OTBR_ENABLE_FEATURE_FLAGS
 
-    mThreadHelper = std::unique_ptr<otbr::agent::ThreadHelper>(new otbr::agent::ThreadHelper(mInstance, this));
+    mThreadHelper      = std::unique_ptr<otbr::agent::ThreadHelper>(new otbr::agent::ThreadHelper(mInstance, this));
```
superwhd, line 238: better to use `MakeUnique`

src/ncp/rcp_host.cpp
```diff
@@ -372,6 +387,12 @@ const char *RcpHost::GetThreadVersion(void)
     return version;
 }
 
+const NetworkProperties &RcpHost::GetNetworkProperties(void)
+{
+    assert(mNetworkProperties != nullptr);
+    return *mNetworkProperties.get();
```
superwhd, line 393:
```suggestion
    return *mNetworkProperties;
```

src/ncp/ncp_spinel.cpp
```diff
@@ -44,12 +44,19 @@ namespace Ncp {
 
 static constexpr char kSpinelDataUnpackFormat[] = "CiiD";
 
+#define SAFE_OBS(updateProp)        \
+    if (mPropsObserver)             \
```
superwhd, line 48: Is the observer optional? If not, we should make it an assertion.

Irving-cl: At first I tried to make it flexible and even support multiple observers. But let's start with simple implementations. I have put the initialization of propsObserver into `Init` method using a reference. In this way we don't need to check nullptr.

src/ncp/ncp_spinel.hpp
```diff
@@ -49,6 +49,28 @@
 namespace otbr {
 namespace Ncp {
 
+/**
+ * This interface is an observer to subscribe the network properties from NCP.
+ *
+ */
+class PropsObserver
```
superwhd, line 56: Is it necessary to define an interface class for this?

Irving-cl: This has a key advantage: The user of `NetworkProperties` can only see and use the Get methods and cannot set the values. Otherwise we need to introduce 'friend', which is not a preferred way.

superwhd (approved):
LGTM 👍

src/ncp/thread_host.hpp
```diff
@@ -79,6 +102,14 @@ class ThreadHost
                                               bool                             aDryRun,
                                               bool                             aEnableAutoAttach);
 
+    /**
+     * Returns the Network properties.
+     *
+     * @returns the Network properties.
+     *
+     */
+    virtual const NetworkProperties &GetNetworkProperties(void) = 0;
```
zhanglongxia, line 111: Could we make the `NetworkProperties` as the base class of the `ThreadHost`? So that we can directly call the function like `mHost->GetDeviceRole();` rather than `mHost->GetNetworkProperties().GetDeviceRole();`.

Irving-cl: Thanks! Have updated.

src/ncp/ncp_host.cpp
```diff
@@ -40,6 +40,25 @@
...
+{
+    return mDeviceRole;
+}
+
+void NcpNetworkProperties::SetDeviceRole(otDeviceRole aRole)
+{
+    mDeviceRole = aRole;
+}
+
+// =============================== NcpHost ===============================
```
zhanglongxia, line 60: nit: make the line 43 and 60 have the same length of comment

zhanglongxia (approved):
LGTM

src/ncp/ncp_spinel.cpp
```diff
@@ -181,6 +183,7 @@ void NcpSpinel::HandleResponse(spinel_tid_t aTid, const uint8_t *aFrame, uint16_
         SuccessOrExit(error = SpinelDataUnpack(data, len, SPINEL_DATATYPE_UINT_PACKED_S, &spinelRole));
 
         mDeviceRole = SpinelRoleToDeviceRole(spinelRole);
+        mPropsObserver->SetDeviceRole(mDeviceRole);
```
wgtdkp, line 186: How do you guarantee that the device role is always propagated back to the host before the callback of `registerStateCallback()` is invoked?

src/ncp/ncp_spinel.cpp
```diff
@@ -181,6 +183,7 @@ void NcpSpinel::HandleResponse(spinel_tid_t aTid, const uint8_t *aFrame, uint16_
...
         CallAndClear(mGetDeviceRoleHandler, OtbrErrorToOtError(error), mDeviceRole);
```
wgtdkp, line 187: What happens if there are multiple simultaneous call of `GetDeviceRole()`? It looks like a single `mGetDeviceRoleHandler` is recorded and invoked here. We don't support multiple calls to `GetDeviceRole()`?

src/ncp/ncp_spinel.cpp
Irving-cl: Because the state callback is triggered by the update of the properties.

For example, the device role is updated on NCP -> NCP sends a notification to the host about the change -> the state callback is invoked on host.

src/ncp/ncp_spinel.cpp
Irving-cl: As discussed, let's remove the async version of `GetDeviceRole` and only use the sync version. The sync version returns immediately so can support multiple calls.

Irving-cl: Have removed the async version of `GetDeviceRole` and changed to use the `GetDeviceRole` provided by network properties.

src/ncp/rcp_host.hpp
```diff
@@ -60,11 +60,34 @@ class FeatureFlagList;
...
+
+    // NetworkProperties methods
+    otDeviceRole GetDeviceRole(void) const override;
+
+    // Set the otInstance
+    void SetInstance(otInstance *aInstance);
+
+private:
+    otInstance *mInstance;
+};
```
abtink, line 84: This is applicable to RCP but seems to be even more general-purpose. 

I think it can be used whenever the OpenThread stack itself is directly accessible (i.e., we can call OpenThread APIs directly). Should we perhaps rename/define it to indicate such a model?

This model might be helpful in other (future) architectures we could support, e.g., an architecture where the entire OpenThread stack, along with the radio driver and platform, is part of the same firmware (running on the same MCU).

abtink (approved):
Looks great. Thanks @Irving-cl. 

One (optional) suggestion below:

src/ncp/rcp_host.hpp
Irving-cl: SG! I think of a few namings. It should reflect that "OpenThread APIs is directly accessible". What about using `OtNetworkProperties`?

abtink: Yes. Sounds good to me. 👍

Irving-cl: Updated.

wgtdkp (approved):

jwhui (approved):

PR #2395, opened 2024-07-22
[ncp] fix a compilation issue in `ncp_spinel.cpp`

suveshpratapa:
Fix an issue in `ncp_spinel.cpp` related to a dummy va_list variable being uninitialized. It causes problems with certain architectures and compiler flags, and initializing it using va_start doesn't work since it's not in a variadic function.

### DIFF
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -69,7 +69,7 @@ void NcpSpinel::GetDeviceRole(GetDeviceRoleHandler aHandler)
 {
     otError      error = OT_ERROR_NONE;
     spinel_tid_t tid   = GetNextTid();
-    va_list      args;
+    va_list      args  = {};
 
     error = mSpinelDriver->SendCommand(SPINEL_CMD_PROP_VALUE_GET, SPINEL_PROP_NET_ROLE, tid, nullptr, args);
     if (error != OT_ERROR_NONE)


third_party/openthread/CMakeLists.txt
```diff
@@ -85,6 +85,10 @@ if (NOT OT_THREAD_VERSION STREQUAL "1.1")
     set(OT_LINK_METRICS_SUBJECT ON CACHE STRING "enable link metrics subject" FORCE)
 endif()
 
+if ((OT_BORDER_AGENT_EPSKC) AND (NOT OT_THREAD_VERSION STREQUAL "1.4"))
+    message(FATAL_ERROR "ePSKc requires Thread protocol version >= 1.4")
+endif()
```
superwhd, line 90: I wonder why it's necessary to enforce version 1.4 if we want to enable ePSKc feature.

superwhd: You can enable `OT_BORDER_AGENT_EPSKC` to enable ePSKc feature. 

The value here is just a default value which can be overriden:
https://github.com/openthread/openthread/blob/d0fbfb8c768eae85d6bf5099bd8129db1c0a94b1/src/core/config/border_agent.h#L84

librasungirl: There was some discussion in https://github.com/openthread/ot-br-posix/pull/2259#discussion_r1591872385 regarding deprecating OTBR_ESPKC/OTBR_ENABLE_EPSKC, and https://github.com/openthread/openthread/pull/10503#issuecomment-2242492237 about the thread version + OT ephemeral key flag.

Thread 1.4 should have OPENTHREAD_CONFIG_BORDER_AGENT_EPHEMERAL_KEY_ENABLE, meanwhile it's hard to forbid enabling it for version less than 1.4.

For reference support, it seems all right to add Thread 1.4 build with `-DOT_THREAD_VERSION=1.4` specified Thoughts?

src/border_agent/border_agent.cpp
```diff
@@ -69,9 +69,11 @@
 
 namespace otbr {
 
-static const char    kBorderAgentServiceType[]      = "_meshcop._udp";   ///< Border agent service type of mDNS
-static const char    kBorderAgentEpskcServiceType[] = "_meshcop-e._udp"; ///< Border agent ePSKc service
-static constexpr int kBorderAgentServiceDummyPort   = 49152;
+static const char kBorderAgentServiceType[] = "_meshcop._udp"; ///< Border agent service type of mDNS
+#if OTBR_ENABLE_EPSKC
```
librasungirl, line 73: remove  `OTBR_ENABLE_EPSKC` condition here and below?

Irving-cl (approved):
LGTM 👍  Thanks!

librasungirl (approved):

superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2401, opened 2024-07-25
[ncp] add posix netif for NCP

Irving-cl:
This PR adds a posix platform module for NCP mode. Under NCP mode, the platform operation won't depend OT posix platform implementation and will use the posix platform implementation added in this PR instead.

This PR only adds the netif module and only support udpate IP6 addresses on the Thread network interface. The netif module hasn't been integrated with the NcpHost. In this PR, the posix platform is added as a library and has its own unit test (also implemented by gtest).

The netif implementations are different on different platforms: linux, apple, bsd... This PR only implements it on linux and adds an empty implementation for other posix platforms.

### DIFF
--- a/include/openthread-br/otbr-posix-config.h
+++ b/include/openthread-br/otbr-posix-config.h
@@ -1,0 +1,58 @@
+/*
+ *    Copyright (c) 2024, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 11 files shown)


Irving-cl:
I'm adding more unit tests for this PR. But I want to get some early feedback on the design and the unit tests. Please take a look at your convenience. @bukepo

Irving-cl:
Will breakdown the PR into smaller ones.

Irving-cl:
Obselete.

PR #2416, opened 2024-08-06
[ncp] update the frame handling of ncp spinel

Irving-cl:
This PR updates the general logic of the handling of incoming spinel frames in NcpSpinel.

Currently the dispatching depends on the incoming key (what we actually received), which brings some implementation troubles. For example, the response of a get command and a set command could sometimes be the same thing: the property.

In this PR, we use `mCmdTable` together with `mWaitingKeyTable` to record each sent command and do the handling based on `mCmdTable[tid]` and `mWaitingKeyTable[tid]` (what we expect to receive). This makes it much easier to further add more handlings later.

### DIFF
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -184,7 +184,7 @@ void NcpSpinel::ThreadErasePersistentInfo(AsyncTaskPtr aAsyncTask)
 exit:
     if (error != OT_ERROR_NONE)
     {
-        FreeTid(tid);
+        FreeTidTableItem(tid);
         mTaskRunner.Post(
             [aAsyncTask, error](void) { aAsyncTask->SetResult(error, "Failed to erase persistent info!"); });
     }
@@ -257,33 +257,21 @@ void NcpSpinel::HandleResponse(spinel_tid_t aTid, const uint8_t *aFrame, uint16_
 
     SuccessOrExit(error = SpinelDataUnpack(aFrame, aLength, kSpinelDataUnpackFormat, &header, &cmd, &key, &data, &len));
 
-    VerifyOrExit(cmd == SPINEL_CMD_PROP_VALUE_IS && key == mWaitingKeyTable[aTid], error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(cmd == SPINEL_CMD_PROP_VALUE_IS, error = OTBR_ERROR_INVALID_STATE);
 
-    switch (key)
+    switch (mCmdTable[aTid])
     {
-    case SPINEL_PROP_LAST_STATUS:
+    case SPINEL_CMD_PROP_VALUE_SET:
+    {
+        error = HandleResponseForPropSet(aTid, key, data, len);
+        break;
+    }
+    case SPINEL_CMD_NET_CLEAR:
     {
         spinel_status_t status = SPINEL_STATUS_OK;
 
-        failureHandler = [this, aTid](otError aError) { HandleResponseForCommand(aTid, aError); };
         SuccessOrExit(error = SpinelDataUnpack(data, len, SPINEL_DATATYPE_UINT_PACKED_S, &status));
-
-        HandleResponseForCommand(aTid, ot::Spinel::SpinelStatusToOtError(status));
-        break;
-    }
-    case SPINEL_PROP_THREAD_ACTIVE_DATASET:
-    {
-        CallAndClear(mDatasetSetActiveTask, OT_ERROR_NONE);
-        break;
-    }
-    case SPINEL_PROP_NET_IF_UP:
-    {
-        CallAndClear(mIp6SetEnabledTask, OT_ERROR_NONE);
-        break;
-    }
-    case SPINEL_PROP_NET_STACK_UP:
-    {
-        CallAndClear(mThreadSetEnabledTask, OT_ERROR_NONE);
+        CallAndClear(mThreadErasePersistentInfoTask, ot::Spinel::SpinelStatusToOtError(status));
         break;
     }
     default:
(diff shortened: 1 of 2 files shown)


src/ncp/ncp_spinel.cpp
```diff
@@ -354,20 +338,42 @@ void NcpSpinel::HandleValueIs(spinel_prop_key_t aKey, const uint8_t *aBuffer, ui
...
+                                              spinel_prop_key_t aKey,
+                                              const uint8_t    *aData,
+                                              uint16_t          aLength)
 {
-    switch (mCmdTable[aTid])
+    OTBR_UNUSED_VARIABLE(aData);
+    OTBR_UNUSED_VARIABLE(aLength);
+
+    otbrError error = OTBR_ERROR_NONE;
+
```
zhanglongxia, line 350: I think we can check the key before the `switch` to avoid check the key in each case:

```C++
     VerifyOrExit(aKey == mWaitingKeyTable[aTid], error = OTBR_ERROR_INVALID_STATE);
```

Irving-cl: This is not always the case. For some SetProperty, the key in spinel response is the same property. For example, host sends Set `SPINEL_PROP_NET_IF_UP` and the response is VALUE_IS `SPINEL_PROP_NET_IF_UP`. However, for some properties, the response is VALUE_IS `SPINEL_PROP_LAST_STATUS` to indicate if the set was successful. So we have to handle them differently. In existing code, the responses are all the same prop key.

zhanglongxia (approved):
LGTM

src/ncp/ncp_spinel.cpp
```diff
@@ -257,55 +257,39 @@ void NcpSpinel::HandleResponse(spinel_tid_t aTid, const uint8_t *aFrame, uint16_
...
+        otbrLogCrit("Received unexpected response with (cmd:%u, key:%u), waiting (cmd:%u, key:%u) for tid:%u", cmd, key,
+                    mCmdTable[aTid], mWaitingKeyTable[aTid], aTid);
     }
     else if (error == OTBR_ERROR_PARSE)
     {
         otbrLogCrit("Error parsing response with tid:%u", aTid);
-        if (failureHandler)
-        {
-            failureHandler(OtbrErrorToOtError(error));
-        }
```
abtink, line 307: This block calling `failureHandler` seems to be removed? Is this intentional?

src/ncp/ncp_spinel.cpp
```diff
@@ -354,20 +338,42 @@ void NcpSpinel::HandleValueIs(spinel_prop_key_t aKey, const uint8_t *aBuffer, ui
...
+
+    switch (mWaitingKeyTable[aTid])
     {
-    case SPINEL_CMD_NET_CLEAR:
+    case SPINEL_PROP_THREAD_ACTIVE_DATASET:
+    {
+        VerifyOrExit(aKey == SPINEL_PROP_THREAD_ACTIVE_DATASET, error = OTBR_ERROR_INVALID_STATE);
+        CallAndClear(mDatasetSetActiveTask, OT_ERROR_NONE);
+        break;
+    }
```
abtink, line 358: Can consider removing the `{ }`  for this and the other `case` blocks below?
```suggestion
        VerifyOrExit(aKey == SPINEL_PROP_THREAD_ACTIVE_DATASET, error = OTBR_ERROR_INVALID_STATE);
        CallAndClear(mDatasetSetActiveTask, OT_ERROR_NONE);
        break;
```

src/ncp/ncp_spinel.cpp
```diff
@@ -354,20 +338,42 @@ void NcpSpinel::HandleValueIs(spinel_prop_key_t aKey, const uint8_t *aBuffer, ui
...
+    case SPINEL_PROP_NET_STACK_UP:
     {
-        CallAndClear(mThreadErasePersistentInfoTask, aError);
+        VerifyOrExit(aKey == SPINEL_PROP_NET_STACK_UP, error = OTBR_ERROR_INVALID_STATE);
+        CallAndClear(mThreadSetEnabledTask, OT_ERROR_NONE);
         break;
     }
     default:
         break;
```
abtink, line 372: Does it make sense in `default` case to do `VerifyOrExit(aKey == mWaitingKeyTable[aTid], error = ...)`?

(since we remove a similar check in `HandleResponse()` in `VerifyOrExit(cmd == SPINEL_CMD_PROP_VALUE_IS && key == mWaitingKeyTable[aTid], error = OTBR_ERROR_INVALID_STATE);` ?

src/ncp/ncp_spinel.hpp
```diff
@@ -203,13 +203,17 @@ class NcpSpinel
...
+    void      HandleResponse(spinel_tid_t aTid, const uint8_t *aFrame, uint16_t aLength);
+    void      HandleValueIs(spinel_prop_key_t aKey, const uint8_t *aBuffer, uint16_t aLength);
+    otbrError HandleResponseForPropSet(spinel_tid_t      aTid,
+                                       spinel_prop_key_t aKey,
+                                       const uint8_t    *aData,
+                                       uint16_t          aLength);
 
     spinel_tid_t GetNextTid(void);
     void         FreeTid(spinel_tid_t tid) { mCmdTidsInUse &= ~(1 << tid); }
+    void         FreeTidTableItem(spinel_tid_t aTid);
```
abtink, line 216: An idea/suggestion: Can we consider removing `FreeTid()` and always use the `FreeTidTableItem()`? (basically having one way to track pending TIDs)? 

Maybe something to consider in its own PR for future?

abtink (commented):
Thanks @Irving-cl. Looks good overall. 

Some smaller suggestions/questions below:

src/ncp/ncp_spinel.cpp
Irving-cl: Yes, it's intentional. My idea is that both `INVALID_STATE` and `PARSE` errors are unexpected and users (the caller of NcpSpinel async APIs) cannot handle such errors. When theses errors happened, that means the stack has run into some problems. It's unlike the case where user calls a NcpSpinel API, and the API returns an error such as `INVALID_STATE` through the callback because that's the result of the operation. So I think we probably don't return errors through callback to the caller in these cases.

src/ncp/ncp_spinel.cpp
Irving-cl: Make sense! Updated.

src/ncp/ncp_spinel.hpp
Irving-cl: Updated! This is a small change.

abtink (approved):
LGTM. Thanks. 👍

jwhui (approved):

PR #2423, opened 2024-08-07
[epskc] feature flag to enable/disble ePSKc

mia1yang:


src/border_agent/border_agent.cpp
```diff
@@ -251,6 +254,13 @@ void BorderAgent::Start(void)
     UpdateMeshCopService();
 
     otBorderAgentSetEphemeralKeyCallback(mHost.GetInstance(), BorderAgent::HandleEpskcStateChanged, this);
+
+#if OTBR_ENABLE_FEATURE_FLAGS
+    FeatureFlagList featureFlagList;
```
abtink, line 259: I noticed that we create a local variable of `FeatureFlagList` type here.  Does the `FeatureFlagList` constructor properly initialize it? How does it get set?

This might be unrelated to the changes in this PR, and more about how `FeatureFlagList` is intended to work. Is this the intended usage, with different modules creating their own instances of `FeatureFlagList`? Should we consider making it a singleton instance instead?

src/border_agent/border_agent.cpp
```diff
@@ -251,6 +254,13 @@ void BorderAgent::Start(void)
...
+
+    SetEphemeralKeyEnabled(featureFlagList.enable_ephemeralkey());
+    otbrLogInfo("BA Epskc Feature has been %sabled in init", (featureFlagList.enable_ephemeralkey() ? "en" : "dis"));
```
abtink, line 262: These seems to do the same as `ApplyFeatureFlagList()` method below? Should we call it instead?
```suggestion
    ApplyFeatureFlagList(fearetuFlagList);
```

src/border_agent/border_agent.cpp
superwhd: +1 on Abtin's concern. 

I think we should keep using a centralized place to initialize/apply the feature flags instead of creating a new feature flag instance here.

Currently the feature flag's input is from dbus:
https://github.com/openthread/ot-br-posix/blob/f2fcf8b5b8564bd42c36c249b2d3f250e00802c2/src/dbus/server/dbus_thread_object_rcp.cpp#L1239

Here's how it's takes effect: https://github.com/openthread/ot-br-posix/blob/f2fcf8b5b8564bd42c36c249b2d3f250e00802c2/src/ncp/rcp_host.cpp#L252

mia1yang: Thanks for input! We have a discussion on how to handle featureflaglist in OTBR ePSKC feature; summary: would not use featureflag(default) in init, ePSKc has it's own init vale; keep current method for applyFeature list, and refactor to callback function(tracking [issue]([b/358293040](http://b/358293040))) later. [Note](https://docs.google.com/document/d/1m_-2xhL43CaMNfa10nWDdxrd1X_YUbdLQUD9UUxfIJo/edit?usp=sharing&resourcekey=0-40QZF81WWOKyQtWoWfM8tA)

src/border_agent/border_agent.cpp
mia1yang: Thanks, init part has been removed.

src/proto/feature_flag.proto
```diff
@@ -66,4 +66,6 @@ message FeatureFlagList {
   optional bool enable_dhcp6_pd = 6 [default = false];
   // Whether to enable link metrics manager.
   optional bool enable_link_metrics_manager = 7 [default = false];
+  // Whether to enable the ePSKc feature.
+  optional bool enable_ephemeralkey = 8 [default = false];
```
Irving-cl, line 70: Is it better to name it as `enable_epskc` to keep consistency?

src/border_agent/border_agent.hpp
```diff
@@ -144,10 +144,24 @@ class BorderAgent : private NonCopyable
      */
     void HandleMdnsState(Mdns::Publisher::State aState);
 
+#if OTBR_ENABLE_FEATURE_FLAGS
+    /**
+     * Apply the feature flag values to Border Agent features.
+     *
+     * @param[in] aFeatureFlagList  The feature flag list to be applied to OT/OTBR.
```
Irving-cl, line 151:
```suggestion
     * @param[in] aFeatureFlagList  A const reference to the feature flag list.
```

Irving-cl (approved):
LGTM 👍

src/border_agent/border_agent.cpp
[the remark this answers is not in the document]
mia1yang: Thanks！ Have removed the init part of feature flag as It's better to have ePSKc init itself.

src/border_agent/border_agent.hpp
mia1yang: remove this part and note with TODO:refactor

src/proto/feature_flag.proto
mia1yang: Modify the log to keep consistency. Thanks!

librasungirl (approved):

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1236,6 +1236,10 @@ otError DBusThreadObjectRcp::SetFeatureFlagListDataHandler(DBusMessageIter &aIte
 
     VerifyOrExit(DBusMessageExtractFromVariant(&aIter, data) == OTBR_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(featureFlagList.ParseFromString(std::string(data.begin(), data.end())), error = OT_ERROR_INVALID_ARGS);
+    // apply FeatureFlagList to BorderAgent. Todo: refactor to support various components
+    mBorderAgent.SetEphemeralKeyEnabled(featureFlagList.enable_ephemeralkey());
+    otbrLogInfo("BA Ephemeral Key Feature has been %sabled by feature flag list",
+                (featureFlagList.enable_ephemeralkey() ? "en" : "dis"));
```
superwhd, line 1242:
```suggestion
    otbrLogInfo("Border Agent Ephemeral Key Feature has been %s by feature flag",
                (featureFlagList.enable_ephemeralkey() ? "enabled" : "disabled"));
```

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1236,6 +1236,10 @@ otError DBusThreadObjectRcp::SetFeatureFlagListDataHandler(DBusMessageIter &aIte
 
     VerifyOrExit(DBusMessageExtractFromVariant(&aIter, data) == OTBR_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
     VerifyOrExit(featureFlagList.ParseFromString(std::string(data.begin(), data.end())), error = OT_ERROR_INVALID_ARGS);
+    // apply FeatureFlagList to BorderAgent. Todo: refactor to support various components
```
superwhd, line 1239:
```suggestion
// TODO: implement the feature flag handler at every component
```

abtink (approved):
👍

src/dbus/server/dbus_thread_object_rcp.cpp
mia1yang: Accept. Thanks!

src/dbus/server/dbus_thread_object_rcp.cpp
mia1yang: Accept. Thanks!

jwhui (approved):

PR #2427, opened 2024-08-09
[dbus] add dbus ScheduleMigration API for NCP

Irving-cl:
This PR adds the 'Migrate' Dbus API under interface.

This PR only adds the dbus interface with a pseudo implementation. (not calling `ThreadHost` API to do the actual migration) This will be added later.

The 'Migrate' API shares some common implementation as `ThreadHelper::AttachAllNodesTo` method. So the PR does a minimum refactor to expose the process to validate/append PENDING_TIMESTAMP and DELAY tlvs from `ThreadHelper` so that `DBusThreadObjectNcp` can reuse the code.

This PR adds a small expect test to check the handling of input dataset.


src/dbus/common/constants.hpp
```diff
@@ -63,6 +63,7 @@
 #define OTBR_DBUS_GET_PROPERTIES_METHOD "GetProperties"
 #define OTBR_DBUS_LEAVE_NETWORK_METHOD "LeaveNetwork"
 #define OTBR_DBUS_SET_NAT64_ENABLED_METHOD "SetNat64Enabled"
+#define OTBR_DBUS_MIGRATE_METHOD "Migrate"
```
superwhd, line 66: What about calling it `ScheduleMigration` to be consistent with [Android API](https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/Connectivity/thread/framework/java/android/net/thread/ThreadNetworkController.java;l=549;drc=5fc74b9fd8a31d6703ba760fcafe9bcfb4b99b1e)?

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -128,5 +131,32 @@ void DBusThreadObjectNcp::LeaveHandler(DBusRequest &aRequest)
...
+void DBusThreadObjectNcp::MigrateHandler(DBusRequest &aRequest)
+{
+    std::vector<uint8_t>     dataset;
+    uint32_t                 delayInMilli;
+    otOperationalDatasetTlvs pendingOpDatasetTlvs;
+    otError                  error = OT_ERROR_NONE;
+
+    auto args = std::tie(dataset, delayInMilli);
+
+    VerifyOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args) == OTBR_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
```
superwhd, line 143:
```suggestion
    SuccessOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args), error = OT_ERROR_INVALID_ARGS);
```

src/utils/thread_helper.cpp
```diff
@@ -1694,5 +1668,42 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
...
+otError ThreadHelper::ProcessDatasetForMigration(otOperationalDatasetTlvs &aDatasetTlvs, uint32_t aDelayMilli)
+{
+    otError  error = OT_ERROR_NONE;
+    Tlv     *tlv;
+    timespec currentTime;
+    uint64_t pendingTimestamp = 0;
+
+    VerifyOrExit(FindTlv(OT_MESHCOP_TLV_PENDINGTIMESTAMP, aDatasetTlvs.mTlvs, aDatasetTlvs.mLength) == nullptr &&
+                     FindTlv(OT_MESHCOP_TLV_DELAYTIMER, aDatasetTlvs.mTlvs, aDatasetTlvs.mLength) == nullptr,
+                 error = OT_ERROR_INVALID_ARGS);
```
superwhd, line 1681: I'd suggest to split this into 2 `VerifyOrExit`s

superwhd (approved):
LGTM  👍

zhanglongxia (approved):
LGTM

src/utils/thread_helper.cpp
```diff
@@ -1694,5 +1668,43 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
...
+                         + (sizeof(uint8_t) + sizeof(uint8_t) + sizeof(uint32_t))) // Delay Timer TLV (6 bytes)
+            <= int{sizeof(aDatasetTlvs.mTlvs)},
+        error = OT_ERROR_INVALID_ARGS);
+
+    tlv = reinterpret_cast<Tlv *>(aDatasetTlvs.mTlvs + aDatasetTlvs.mLength);
+    tlv->SetType(OT_MESHCOP_TLV_PENDINGTIMESTAMP);
+    clock_gettime(CLOCK_REALTIME, &currentTime);
+    pendingTimestamp |= (static_cast<uint64_t>(currentTime.tv_sec) << 16);
+    pendingTimestamp |= (((static_cast<uint64_t>(currentTime.tv_nsec) * 32768 / 1000000000) & 0x7fff) << 1);
+    tlv->SetValue(pendingTimestamp);
```
abtink, line 1697: Should we rely on the caller to provide the pending timestamp already in the dataset?

abtink, line 1697: Here we are setting value as `uint64` value? Is the endian order handled properly?

Is this the correct format for timestamp TLV?
Here is how it is defined in OT `meshcop/timestamp.h`

```c++
    uint16_t mSeconds16; // bits 32-47
    uint32_t mSeconds32; // bits 0-31
    uint16_t mTicksAndAuthFlag;
```

I see we calculate it as 
```c++
    pendingTimestamp |= (static_cast<uint64_t>(currentTime.tv_sec) << 16);
    pendingTimestamp |= (((static_cast<uint64_t>(currentTime.tv_nsec) * 32768 / 1000000000) & 0x7fff) << 1);
```
but not clear to me that it mapps correctly to the above format? (if it does, maybe adding a comment can help?)

src/utils/thread_helper.cpp
```diff
@@ -1694,5 +1668,43 @@ otError ThreadHelper::RetrieveTelemetryData(Mdns::Publisher *aPublisher, threadn
...
+
+    tlv = tlv->GetNext();
+    tlv->SetType(OT_MESHCOP_TLV_DELAYTIMER);
+    tlv->SetValue(aDelayMilli);
```
abtink, line 1701: Is the endian-ness correctly handled when setting a `uint32_t` value on TLV?

Irving-cl: I've checked and it's handled (written in big endian). These [code](https://github.com/openthread/ot-br-posix/blob/main/src/common/tlv.hpp#L155) were in introduced since this PR: https://github.com/openthread/ot-br-posix/pull/1051

I think the code doesn't reflect that the endianness is handled correctly because in ot-br-posix the tlv handling code isn't that complete as that in OT. Let's further improve this in future PRs.

src/utils/thread_helper.cpp
Irving-cl: I've checked and added a comment to explain it. The implementation takes `mSeconds16` and `mSeconds32` as a whole and fill it with the unix timestamp. The ticks is fractional Unix Time value in 32.768 kHz
resolution. The remaining bit is unset. And the `SetValue` uses big-endian. So it will follow the same order as timestamp defined in OT.

Irving-cl: I think for some historical reasons the dbus API `AttachAllNodesTo` is changed like this: the API first verifies that the input dataset doesn't contain a pending timestamp and then generate the pending timestamp) https://github.com/openthread/ot-br-posix/pull/1051  

So I prefer not changing this in this PR to avoid changing the behavior on products. And considering this is a dbus API, it's not easy for the caller to generate a timestamp tlv. But the ThreadHost API for migration will takes the entire dataset as the input including pending timestamp and delay TLVs. Thoughts?

src/utils/thread_helper.cpp
abtink: Thanks. SGTM.

src/utils/thread_helper.cpp
abtink: Make sense. Thanks.

abtink (approved):
👍

jwhui (approved):

PR #2431, opened 2024-08-13
[posix] add unicast address update in netif module

Irving-cl:
This PR implements `UpdateIp6UnicastAddresses` method for `Posix::Netif` module to add/remove unicast address on the wpan interface. A unit test is added together.

The method `UpdateIp6UnicastAddresses` takes a vector of Ip6 addresses as input (because the address update from the NCP is the entire address table) and update the entire address list to the interface. However, it doesn't do it by removing all and then adding all. Instead, it compares the existing unicast addresses against the new addresses to decide which addresses to add/remove and then issue netlink requests to do the changes.

### DIFF
--- a/src/common/types.hpp
+++ b/src/common/types.hpp
@@ -43,6 +43,7 @@
 #include <vector>
 
 #include <openthread/error.h>
+#include <openthread/ip6.h>
 
 #include "common/byteswap.hpp"
 
@@ -142,6 +143,14 @@ public:
      *
      */
     Ip6Address(const uint8_t (&aAddress)[16]);
+
+    /**
+     * Constructor with an otIp6Address.
+     *
+     * @param[in] aAddress  A const reference to an otIp6Address.
+     *
+     */
+    explicit Ip6Address(const otIp6Address &aAddress);
 
     /**
      * Constructor with a string.
--- a/src/ncp/posix/netif.cpp
+++ b/src/ncp/posix/netif.cpp
@@ -41,6 +41,8 @@
 #include <sys/socket.h>
 #include <unistd.h>
 
+#include <algorithm>
+
 #include "common/code_utils.hpp"
 #include "common/logging.hpp"
 #include "common/types.hpp"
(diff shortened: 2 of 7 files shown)


src/common/types.hpp
```diff
@@ -143,6 +144,12 @@ class Ip6Address
      */
     Ip6Address(const uint8_t (&aAddress)[16]);
 
+    /**
+     * Constructor with an otIp6Address.
+     *
+     */
+    Ip6Address(const otIp6Address &aAddress);
```
bukepo, line 151:
```suggestion
    explicit Ip6Address(const otIp6Address &aAddress);
```

src/ncp/posix/netif.cpp
```diff
@@ -85,6 +87,38 @@ void Netif::Deinit(void)
...
+void Netif::UpdateIp6UnicastAddresses(const std::vector<otIp6AddressInfo> &aAddresses)
+{
+    mIp6UnicastAddressesSwap.clear();
+    for (const otIp6AddressInfo &address : aAddresses)
+    {
+        mIp6UnicastAddressesSwap.emplace_back(address);
+    }
+
+    // Remove stale addresses
+    for (const Ip6AddressInfo &address : mIp6UnicastAddresses)
```
bukepo, line 99: avoid swap array with iterator?

src/common/types.hpp
Irving-cl: Added 'explicit' but keep the comment for consistency.

src/ncp/posix/netif.cpp
Irving-cl: Thanks! Updated the implementation and removed the swap.

src/ncp/posix/netif.cpp
```diff
@@ -85,6 +87,38 @@ void Netif::Deinit(void)
...
+        {
+            otbrLogInfo("Add address: %s", Ip6Address(addressInfo.mAddress).ToString().c_str());
+            ProcessUnicastAddressChange(addressInfo, true);
+        }
+    }
+
+    mIp6UnicastAddresses.clear();
+    for (const otIp6AddressInfo &address : aAddresses)
+    {
+        mIp6UnicastAddresses.emplace_back(address);
```
bukepo, line 118:
```suggestion
        mIp6UnicastAddresses.emplace_back(Ip6AddressInfo(address));
```

src/ncp/posix/netif.cpp
```diff
@@ -85,6 +87,38 @@ void Netif::Deinit(void)
...
+void Netif::UpdateIp6UnicastAddresses(const std::vector<otIp6AddressInfo> &aAddresses)
+{
+    // Remove stale addresses
+    for (const Ip6AddressInfo &address : mIp6UnicastAddresses)
+    {
+        auto comparator = [&address](const otIp6AddressInfo &aInfo) { return address == aInfo; };
+        if (std::find_if(aAddresses.begin(), aAddresses.end(), comparator) == aAddresses.end())
+        {
+            otbrLogInfo("Remove address: %s", Ip6Address(address.mAddress).ToString().c_str());
+            ProcessUnicastAddressChange(address, false);
```
bukepo, line 99: TODO: verify success of the deletion or addition.

bukepo (approved):

src/ncp/posix/netif_linux.cpp
```diff
@@ -195,6 +195,54 @@ void Netif::SetAddrGenModeToNone(void)
...
+
+    req.ifa.ifa_family    = AF_INET6;
+    req.ifa.ifa_prefixlen = aAddressInfo.mPrefixLength;
+    req.ifa.ifa_flags     = IFA_F_NODAD;
+    req.ifa.ifa_scope     = aAddressInfo.mScope;
+    req.ifa.ifa_index     = mNetifIndex;
+
+    AddRtAttr(&req.nh, sizeof(req), IFA_LOCAL, &aAddressInfo.mAddress, sizeof(aAddressInfo.mAddress));
+
+    if (!aAddressInfo.mPreferred || aAddressInfo.mMeshLocal)
```
abtink, line 224: Should we still add non-preferred addresses? 

For example we deprecate SLAAC addresses (e.g., an OMR address) before fully removing it and while it is deprecating we should still accept and receive msgs destined to that address?

abtink (approved):
Looks good.

One question/suggestion below:

src/ncp/posix/netif_linux.cpp
abtink: Please ignore this. I assume were were skipping adding the address (but that's not the case).

jwhui (approved):

PR #2433, opened 2024-08-13
[spinel] fix compilation error while building for openwrt

sarveshkumarv3:
Avoid passing unused variable argument to SendSpinel command which leads to compilation error. This PR depends on another [OT PR](https://github.com/openthread/openthread/pull/10602) which allows to invoke SendCommand API without the variable arguments.


abtink (approved):
👍

Irving-cl (approved):
LGTM 👍 Thanks

PR #2440, opened 2024-08-16
[logging] use the same default log level for Android and other platforms

superwhd:
There's no need currently to reduce the logs on Android user images.


superwhd:
Tentatively closing this PR until we have a consensus.

PR #2442, opened 2024-08-16
[logging] support disable syslog for otbr-web as well

agners:
Commit c88a5fce6659 ("[logging] add command line arg to disable syslog") allows to disable logging to syslog via command line argument. Support the same -s command line argument for otbr-web as well.


jwhui (approved):
Thanks! 👍🏼

PR #2443, opened 2024-08-17
[ncp] use `SendCommand` with no `va_args` requirement to avoid compilation issues

suveshpratapa:
Fix an issue in `ncp_spinel.cpp` related to a dummy va_list variable being uninitialized.

### DIFF
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -354,6 +354,16 @@ void NcpSpinel::HandleValueIs(spinel_prop_key_t aKey, const uint8_t *aBuffer, ui
         break;
     }
 
+    case SPINEL_PROP_IPV6_ADDRESS_TABLE:
+    {
+        std::vector<Ip6AddressInfo> addressInfoTable;
+
+        VerifyOrExit(ParseIp6AddressTable(aBuffer, aLength, addressInfoTable) == OT_ERROR_NONE,
+                     error = OTBR_ERROR_PARSE);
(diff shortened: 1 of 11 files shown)


src/ncp/ncp_spinel.cpp
```diff
@@ -190,7 +190,7 @@ void NcpSpinel::ThreadErasePersistentInfo(AsyncTaskPtr aAsyncTask)
 {
     otError      error = OT_ERROR_NONE;
     spinel_tid_t tid   = GetNextTid();
-    va_list      args;
+    va_list      args  = {};
```
abtink, line 193: Suggest using the newly added version of `SendCommand` with no extra args instead. This way there is no need to define an empty `args`.

```c++
    /*
     * Sends a spinel command without arguments to the co-processor.
     *
     * @param[in] aCommand    The spinel command.
     * @param[in] aKey        The spinel property key.
     * @param[in] aTid        The spinel transaction id.
     *
     * @retval  OT_ERROR_NONE           Successfully sent the command through spinel interface.
     * @retval  OT_ERROR_INVALID_STATE  The spinel interface is in an invalid state.
     * @retval  OT_ERROR_NO_BUFS        The spinel interface doesn't have enough buffer.
     *
     */
    otError SendCommand(uint32_t aCommand, spinel_prop_key_t aKey, spinel_tid_t aTid);

```

abtink (approved):
👍 Thanks.

jwhui (approved):

PR #2452, opened 2024-08-23
[posix] add netif TUN Ip6 Sending

Irving-cl:
This PR implements netif TUN Ip6 forwarding.

In specific, when a packet from the system is sent to the wpan interface, the Netif module will do the Ip6 sending. In this PR the actual sending (let NCP send the IP6 packet) hasn't been implemented and an empty implementation is used.

The PR also adds a unit test to verify IP packet sent through wpan interface can be successfully handled in netif.

### DIFF
--- a/src/ncp/posix/netif.cpp
+++ b/src/ncp/posix/netif.cpp
@@ -59,9 +59,12 @@ Netif::Netif(void)
 {
 }
 
-otbrError Netif::Init(const std::string &aInterfaceName)
+otbrError Netif::Init(const std::string &aInterfaceName, const Ip6SendFunc &aIp6SendFunc)
 {
     otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(aIp6SendFunc, error = OTBR_ERROR_INVALID_ARGS);
+    mIp6SendFunc = aIp6SendFunc;
 
     mIpFd = SocketWithCloseExec(AF_INET6, SOCK_DGRAM, IPPROTO_IP, kSocketNonBlock);
     VerifyOrExit(mIpFd >= 0, error = OTBR_ERROR_ERRNO);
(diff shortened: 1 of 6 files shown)


src/ncp/posix/netif.cpp
```diff
@@ -87,6 +90,35 @@ void Netif::Deinit(void)
...
+    assert(mTunFd >= 0);
+    assert(mIpFd >= 0);
+
+    FD_SET(mTunFd, &aContext->mReadFdSet);
+    FD_SET(mTunFd, &aContext->mErrorFdSet);
+
+    if (mTunFd > aContext->mMaxFd)
+    {
+        aContext->mMaxFd = mTunFd;
+    }
```
abtink, line 119: Not for this PR but a possible enhancement idea (for future PR)

I assume this pattern of adding certain `fd` to `MainloopContext` (updating `mReadFdSet` and `mErrorFdSet` and then also updating `mMaxFd`) might be done in several places in the code.

It could be helpful to add a method on `MainloopContext` for this. This would simplify the code and ensure we always update `mMaxFd` as well.

src/ncp/posix/netif.cpp
```diff
@@ -202,6 +234,25 @@ void Netif::SetNetifState(bool aState)
...
+    ssize_t   rval;
+    uint8_t   packet[kIp6Mtu];
+    otbrError error = OTBR_ERROR_NONE;
+
+    rval = read(mTunFd, packet, sizeof(packet));
+    VerifyOrExit(rval > 0, error = OTBR_ERROR_ERRNO);
+
+    otbrLogInfo("[netif] Send packet (%hu bytes)", static_cast<uint16_t>(rval));
+
+    error = mIp6SendFunc(packet, rval);
```
abtink, line 248: Does it make sense to check if `mIp6SendFunc` may be null (not set)?

abtink (approved):
LGTM. Thanks.

src/ncp/posix/netif.cpp
Irving-cl: This is great! I'll do that in a seperate PR.

zhanglongxia (approved):
LGTM

jwhui (approved):

PR #2455, opened 2024-08-23
[posix] add netif TUN Ip6Receive

Irving-cl:
This PR implements netif TUN Ip6 receiving.

When netif module receives an Ip6 packet, it will pass it to the system network stack through wpan interface. The PR also adds a unit test to verify an application can receive UDP packets from wpan through netif.

This PR hasn't integrated Ip6 receiving with NCP(forward the packet from NCP to the system network stack). This will be done in future PRs.

### DIFF
--- a/src/ncp/posix/netif.cpp
+++ b/src/ncp/posix/netif.cpp
@@ -202,6 +202,23 @@ exit:
     }
 }
 
+void Netif::Ip6Receive(const uint8_t *aBuf, uint16_t aLen)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(aLen <= kIp6Mtu, error = OTBR_ERROR_DROPPED);
+    VerifyOrExit(mTunFd > 0, error = OTBR_ERROR_INVALID_STATE);
+
+    otbrLogInfo("Packet from NCP (%u bytes)", aLen);
+    VerifyOrExit(write(mTunFd, aBuf, aLen) == aLen, error = OTBR_ERROR_ERRNO);
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        otbrLogWarning("Failed to receive, error:%s", otbrErrorString(error));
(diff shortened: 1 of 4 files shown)


src/ncp/posix/netif.cpp
```diff
@@ -202,6 +202,23 @@ void Netif::SetNetifState(bool aState)
...
 }
 
+void Netif::Ip6Receive(const uint8_t *aBuf, uint16_t aLen)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(aLen <= kIp6Mtu, error = OTBR_ERROR_DROPPED);
+    VerifyOrExit(mTunFd > 0, error = OTBR_ERROR_INVALID_ARGS);
+
+    otbrLogInfo("Packet from NCP (%u bytes)", static_cast<uint16_t>(aLen));
```
abtink, line 212: Can we remove `static_cast<uint16_t>()` as it is already `uint16_t aLen`?
 ```suggestion
    otbrLogInfo("Packet from NCP (%u bytes)", aLen);
```

File: src/ncp/posix/netif.cpp [hunk 2]
```diff
@@ -202,6 +202,23 @@ void Netif::SetNetifState(bool aState)
     }
 }
 
+void Netif::Ip6Receive(const uint8_t *aBuf, uint16_t aLen)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(aLen <= kIp6Mtu, error = OTBR_ERROR_DROPPED);
+    VerifyOrExit(mTunFd > 0, error = OTBR_ERROR_INVALID_ARGS);
```
abtink, line 210: How about `INVALID_STATE` instead of `INVALID_ARGS`? since `mTunFd` is not an arg?

abtink (approved):
LGTM. Thanks.

jwhui (approved):

PR #2459, opened 2024-08-27
[ncp] integrate netif isUp state update

Irving-cl:
This PR integrates the `Netif::SetNetifState` with `NcpSpinel` so that when the netif state on NCP changes, the wpan state on the host updates accordingly.

### DIFF
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -72,7 +72,9 @@ void NcpSpinel::Init(ot::Spinel::SpinelDriver &aSpinelDriver, PropsObserver &aOb
 
 void NcpSpinel::Deinit(void)
 {
-    mSpinelDriver = nullptr;
+    mSpinelDriver              = nullptr;
+    mIp6AddressTableCallback   = nullptr;
+    mNetifStateChangedCallback = nullptr;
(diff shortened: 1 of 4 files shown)


src/ncp/ncp_spinel.cpp
```diff
@@ -72,7 +72,9 @@ void NcpSpinel::Init(ot::Spinel::SpinelDriver &aSpinelDriver, PropsObserver &aOb
 
 void NcpSpinel::Deinit(void)
 {
-    mSpinelDriver = nullptr;
+    mSpinelDriver              = nullptr;
+    mIp6AddressTableCallback   = nullptr;
+    mNetifStateChangedCallback = nullptr;
```
zhanglongxia, line 77: Do we need to initialize these variables to `nullptr` in the constructor?

Irving-cl: The `std::function` has default constructor to let the member be `nullptr`. So I think it's not necessary to initialize them explicitly. Thoughts?

abtink (approved):
LGTM. 👍

zhanglongxia (approved):
LGTM

jwhui (approved):

PR #2463, opened 2024-08-28
[telemetry] remove uploading epskc_state and border_agent_state

sunytt:
The epskc_state and border_agent_state has no clear metrics, remove from uploading.


bukepo (approved):

morningboata (approved):
Normally the removed proto fields should be deprecated instead of delete. It seems that there are no offspring apps using these fields yet, LGTM to remove them.

librasungirl (approved):

PR #2478, opened 2024-09-05
[netif] add wpan multicast group join for NCP

Irving-cl:
This PR adds wpan multicast group join functionality for NCP.

This PR allows the applications to join/leave multicast groups in the Thread network. The functionality is implemented by listening to MLD reports sent by kernel. And then the Netif module will send the join/leave request to the NCP and let NCP join/leave the multicast group. Previous work on this in OT: https://github.com/openthread/openthread/pull/4687

In this PR, `NcpSpinel::Ip6MulAddrUpdateSubscription` is implemented as empty. This will be implemented in a following PR. This PR adds a unit test case to verify that the MLD reports can be successfully received and handled.

### DIFF
--- a/src/ncp/posix/netif.cpp
+++ b/src/ncp/posix/netif.cpp
@@ -30,8 +30,10 @@
 
 #include "netif.hpp"
 
+#include <arpa/inet.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <ifaddrs.h>
 #include <net/if.h>
 #include <net/if_arp.h>
 #include <netinet/in.h>
@@ -54,13 +56,51 @@ otbrError Netif::Dependencies::Ip6Send(const uint8_t *aData, uint16_t aLength)
 {
     OTBR_UNUSED_VARIABLE(aData);
     OTBR_UNUSED_VARIABLE(aLength);
+
     return OTBR_ERROR_NONE;
 }
+
+otbrError Netif::Dependencies::Ip6MulAddrUpdateSubscription(const otIp6Address &aAddress, bool aIsAdd)
+{
+    OTBR_UNUSED_VARIABLE(aAddress);
+    OTBR_UNUSED_VARIABLE(aIsAdd);
+
+    return OTBR_ERROR_NONE;
+}
+
+OT_TOOL_PACKED_BEGIN
+struct Mldv2Header
+{
+    uint8_t  mType;
+    uint8_t  _rsv0;
+    uint16_t mChecksum;
+    uint16_t _rsv1;
+    uint16_t mNumRecords;
+} OT_TOOL_PACKED_END;
+
+OT_TOOL_PACKED_BEGIN
+struct Mldv2Record
+{
+    uint8_t         mRecordType;
+    uint8_t         mAuxDataLen;
+    uint16_t        mNumSources;
+    struct in6_addr mMulticastAddress;
+} OT_TOOL_PACKED_END;
+
+enum
+{
+    kIcmpv6Mldv2Type                      = 143,
+    kIcmpv6Mldv2ModeIsIncludeType         = 1,
+    kIcmpv6Mldv2ModeIsExcludeType         = 2,
+    kIcmpv6Mldv2RecordChangeToIncludeType = 3,
+    kIcmpv6Mldv2RecordChangeToExcludeType = 4,
+};
 
 Netif::Netif(Dependencies &aDependencies)
     : mTunFd(-1)
     , mIpFd(-1)
     , mNetlinkFd(-1)
+    , mMldFd(-1)
     , mNetlinkSequence(0)
     , mNetifIndex(0)
     , mDeps(aDependencies)
@@ -79,6 +119,8 @@ otbrError Netif::Init(const std::string &aInterfaceName)
 
     mNetifIndex = if_nametoindex(mNetifName.c_str());
     VerifyOrExit(mNetifIndex > 0, error = OTBR_ERROR_INVALID_STATE);
+
+    SuccessOrExit(error = InitMldListener());
 
     PlatformSpecificInit();
 
@@ -103,9 +145,20 @@ void Netif::Process(const MainloopContext *aContext)
         DieNow("Error on Tun Fd!");
     }
 
+    if (FD_ISSET(mMldFd, &aContext->mErrorFdSet))
+    {
+        close(mMldFd);
+        DieNow("Error on MLD Fd!");
+    }
+
     if (FD_ISSET(mTunFd, &aContext->mReadFdSet))
     {
         ProcessIp6Send();
+    }
+
+    if (FD_ISSET(mMldFd, &aContext->mReadFdSet))
+    {
+        ProcessMldEvent();
     }
 }
 
@@ -114,8 +167,10 @@ void Netif::UpdateFdSet(MainloopContext *aContext)
     assert(aContext != nullptr);
     assert(mTunFd >= 0);
     assert(mIpFd >= 0);
+    assert(mMldFd >= 0);
 
     aContext->AddFdToSet(mTunFd, MainloopContext::kErrorFdSet | MainloopContext::kReadFdSet);
+    aContext->AddFdToSet(mMldFd, MainloopContext::kErrorFdSet | MainloopContext::kReadFdSet);
 }
 
 void Netif::UpdateIp6UnicastAddresses(const std::vector<Ip6AddressInfo> &aAddrInfos)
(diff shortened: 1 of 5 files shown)


src/ncp/posix/netif.cpp
```diff
@@ -57,10 +59,48 @@ otbrError Netif::Dependencies::Ip6Send(const uint8_t *aData, uint16_t aLength)
...
+otbrError Netif::Dependencies::Ip6MulAddrUpdateSubscription(const otIp6Address &aAddress, bool aIsAdd)
+{
+    OTBR_UNUSED_VARIABLE(aAddress);
+    OTBR_UNUSED_VARIABLE(aIsAdd);
+    return OTBR_ERROR_NONE;
+}
+
+#if NETIF_USE_MLD_MONITOR
+OT_TOOL_PACKED_BEGIN
+struct MLDv2Header
```
superwhd, line 71: Maybe it's a good chance to make it conforming to the OT's naming style. i.e. Mldv2Header

I'm fine with keeping it as-is, though.

src/ncp/posix/netif.cpp
```diff
@@ -57,10 +59,48 @@ otbrError Netif::Dependencies::Ip6Send(const uint8_t *aData, uint16_t aLength)
     return OTBR_ERROR_NONE;
 }
 
+otbrError Netif::Dependencies::Ip6MulAddrUpdateSubscription(const otIp6Address &aAddress, bool aIsAdd)
+{
+    OTBR_UNUSED_VARIABLE(aAddress);
+    OTBR_UNUSED_VARIABLE(aIsAdd);
+    return OTBR_ERROR_NONE;
```
superwhd, line 66:
```suggestion

    return OTBR_ERROR_NONE;
```
nit

src/ncp/posix/netif.cpp
```diff
@@ -57,10 +59,48 @@ otbrError Netif::Dependencies::Ip6Send(const uint8_t *aData, uint16_t aLength)
...
+{
+    uint8_t         mRecordType;
+    uint8_t         mAuxDataLen;
+    uint16_t        mNumSources;
+    struct in6_addr mMulticastAddress;
+} OT_TOOL_PACKED_END;
+
+enum
+{
+    kICMPv6MLDv2Type                      = 143,
```
superwhd, line 91: ditto about the naming

src/ncp/posix/netif.hpp
```diff
@@ -46,6 +46,12 @@
 
 namespace otbr {
 
+#if defined(RTM_NEWMADDR) || defined(__NetBSD__)
+#define NETIF_USE_MLD_MONITOR 0
```
superwhd, line 50: This is not following the naming style of OTBR

Irving-cl: Updated, thanks!

superwhd (approved):

src/ncp/posix/netif.cpp
```diff
@@ -289,9 +354,137 @@ void Netif::Clear(void)
...
+    {
+        if (ifAddr->ifa_addr != nullptr && ifAddr->ifa_addr->sa_family == AF_INET6 &&
+            strncmp(mNetifName.c_str(), ifAddr->ifa_name, IFNAMSIZ) == 0)
+        {
+            struct sockaddr_in6 *addr6 = reinterpret_cast<struct sockaddr_in6 *>(ifAddr->ifa_addr);
+
+            if (memcmp(&addr6->sin6_addr, &srcAddr.sin6_addr, sizeof(in6_addr)) == 0)
+            {
+                fromSelf = true;
+                break;
```
sunytt, line 437: nit: maybe fromSelf is not needed if using `ExitNow();` here.

src/ncp/posix/netif.cpp
```diff
@@ -289,9 +354,137 @@ void Netif::Clear(void)
...
+            else if (record->mRecordType == kIcmpv6Mldv2RecordChangeToExcludeType)
+            {
+                error = mDeps.Ip6MulAddrUpdateSubscription(address, /* isAdd */ false);
+            }
+
+            offset += sizeof(Mldv2Record) + sizeof(in6_addr) * ntohs(record->mNumSources);
+
+            if (error != OTBR_ERROR_NONE)
+            {
+                otbrLogWarning("Failed to Update multicast subscription: %s", otbrErrorString(error));
```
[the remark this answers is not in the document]
Irving-cl: Updated!

src/ncp/posix/netif.cpp
Irving-cl: It seems we cannot remove it because the condition is to verify that at least one address matches and only exit when all addresses don't match.

src/ncp/posix/netif.cpp
```diff
@@ -289,9 +356,143 @@ void Netif::Clear(void)
...
+    }
+#endif
+
     mNetifIndex = 0;
     mIp6UnicastAddresses.clear();
     mIp6MulticastAddresses.clear();
 }
 
+#if OTBR_NETIF_USE_MLD_MONITOR
+static const otIp6Address kMLDv2MulticastAddress = {
```
abtink, line 373:
```suggestion
static const otIp6Address kMldv2MulticastAddress = {
```

src/ncp/posix/netif.cpp
```diff
@@ -289,9 +356,143 @@ void Netif::Clear(void)
...
+    {{0xff, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x16}}};
+static const otIp6Address kAllRouterLocalMulticastAddress = {
+    {{0xff, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02}}};
+
+static bool IsMulAddrFiltered(const otIp6Address &aAddr)
+{
+    return memcmp(&aAddr, &kMLDv2MulticastAddress, sizeof(otIp6Address)) == 0 ||
```
abtink, line 380: This is fine, but we can also consider using `otIp6IsAddressEqual()`?

abtink (approved):
LGTM. 👍 

src/ncp/posix/netif.cpp
```diff
@@ -54,13 +56,55 @@ otbrError Netif::Dependencies::Ip6Send(const uint8_t *aData, uint16_t aLength)
...
 
+otbrError Netif::Dependencies::Ip6MulAddrUpdateSubscription(const otIp6Address &aAddress, bool aIsAdd)
+{
+    OTBR_UNUSED_VARIABLE(aAddress);
+    OTBR_UNUSED_VARIABLE(aIsAdd);
+
+    return OTBR_ERROR_NONE;
+}
+
+#if OTBR_NETIF_USE_MLD_MONITOR
```
bukepo, line 71: Is it better to use MLD only?

Irving-cl: This isn't a configuration but an evaluation if we should use MLD monitor. In OT posix netif.cpp:
```
#if defined(RTM_NEWMADDR) || defined(__NetBSD__)                                                                                                                                                                                                       
// on some BSDs (mac OS, FreeBSD), we get RTM_NEWMADDR/RTM_DELMADDR messages, so we don't need to monitor using MLD                                                                                                                                    
// on NetBSD, MLD monitoring simply doesn't work                                                                                                                                                                                                       
#define OPENTHREAD_POSIX_USE_MLD_MONITOR 0                                                                                                                                                                                                             
#else                                                                                                                                                                                                                                                  
// on some platforms (Linux, but others might be made to work), we do not get information about multicast                                                                                                                                              
// group joining via AF_NETLINK or AF_ROUTE sockets.  on those platform, we must listen for IPv6 ICMP                                                                                                                                                  
// MLDv2 messages to know when multicast memberships change                                                                                                                                                                                            
//              https://stackoverflow.com/questions/37346289/using-netlink-is-it-possible-to-listen-whenever-multicast-group-membership-is-ch                                                                                                          
#define OPENTHREAD_POSIX_USE_MLD_MONITOR 1                                                                                                                                                                                                             
#endif // defined(RTM_NEWMADDR) || defined(__NetBSD__)       
```
It said when `defined(RTM_NEWMADDR) || defined(__NetBSD__)` MLD doesn't work. So I think we can use macro to turn this off on these platforms. Thoughts?

Irving-cl: As discussed offline, we will always use MLD_MONITOR. Will do special handling on some platforms where MLD_MONITOR doesn't work.

Irving-cl: Updated!

bukepo (approved):

jwhui (approved):

PR #2479, opened 2024-09-06
[netif] refactor Netif and NcpSpinel

Irving-cl:
This PR does a small refactoring for `Netif` and `NcpSpinel`.

`Netif::Ip6Send` depends on the implementation of `NcpSpinel`. And there are more methods to come. (For example, `SubcribeMulticastAddress`). Currently we pass the Ip6Send implementation to `Netif` in `Netif::Init` through a std::function object. But then, each time we add a new method, we need to update the signature of `Netif::Init`(making it very long) and the calling of `Netif::Init`, which is not preferred. This PR adds a dependencies class `NetifDependencies` to encapsulate all the implementation functions for Netif and let `NcpSpinel` implement this interface.

### DIFF
--- a/src/ncp/ncp_host.cpp
+++ b/src/ncp/ncp_host.cpp
@@ -65,7 +65,7 @@ void NcpNetworkProperties::SetDeviceRole(otDeviceRole aRole)
 
 NcpHost::NcpHost(const char *aInterfaceName, bool aDryRun)
     : mSpinelDriver(*static_cast<ot::Spinel::SpinelDriver *>(otSysGetSpinelDriver()))
-    , mNetif()
+    , mNetif(mNcpSpinel)
 {
     memset(&mConfig, 0, sizeof(mConfig));
     mConfig.mInterfaceName = aInterfaceName;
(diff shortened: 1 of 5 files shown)


src/ncp/ncp_host.cpp
```diff
@@ -80,10 +80,12 @@ const char *NcpHost::GetCoprocessorVersion(void)
 
 void NcpHost::Init(void)
 {
+    otbr::NetifIp6Implementor implementor = {
+        [this](const uint8_t *aData, uint16_t aLength) { return mNcpSpinel.Ip6Send(aData, aLength); }};
```
superwhd, line 84: In AOSP, there's a pattern of making its dependencies a class which can help reduce the number of parameters and make the class more testable. I wonder if we can borrow the idea here.

https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/Connectivity/service/src/com/android/server/connectivity/MulticastRoutingCoordinatorService.java;l=761;drc=5005c919c7cd6d27edad9954785d8f59ae6f78fb

Irving-cl: Thanks! I think this is a good model. I have updated to use virtual interface as the dependencies class.

src/ncp/posix/netif.hpp
```diff
@@ -46,14 +46,18 @@
 
 namespace otbr {
 
-class Netif
+class NetifDependencies
```
superwhd, line 49: nit: maybe making it a nested class can reduce the repetitions of the name 'netif'.

I don't have a strong preference on this though.

zhanglongxia (approved):
LGTM

abtink (approved):
👍

superwhd (approved):
LGTM :+1:

jwhui (approved):

PR #2514, opened 2024-09-27
[rest] Add REST JSON:API for commissioner and diagnostics

martinzi:
In this PR, we propose an extended `REST` API functionality providing capabilities for commissioning and on-mesh diagnostics to generic off-mesh http clients. The implementation is guided by the [JSON:API specification](https://jsonapi.org/format/) and provides following parts

- `api/node` - pointing to api/devices/_thisdevice_
- `api/devices` - a collection of discovered devices
- `api/actions` - indirect processing of tasks, in particular for those that may need more time
    - `addThreadDeviceTask` - starts the on-mesh commissioner and adds the joiner candidate into the joiner table
   - `getNetworkDiagnosticTask` - sends diagnostic requests and diagnostic queries to the destination
   - `resetNetworkDiagCounterTask` - resets the network diagnostic mle and/or mac counter
   - `getEnergyScanTask` - starts the commissioner and sends energy scan requests to the destination
- `api/diagnostics`
   -  networkDiagnostic - all the Diagnostic TLV details
   -  energyScanReport - results of energy scans

The commit also provides integration tests, see tests/restjsonapi.
[The Readme provides more details](src/rest/extensions) in src/rest/extensions folder.

Issue #2492, opened 2024-09-16
REST: implement server-sent events


OmegaRelay:
# Server Sent Events in REST API server

**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem is.

At my company we currently use the http REST API to control the otbr instance with our own application.
The issue is we currently need to poll the /node/state endpoint to be updated of the connection status.

**Describe the solution you'd like** A clear and concise description of what you want to happen.

Server-sent event implementation defined in [RFC-8895](https://datatracker.ietf.org/doc/html/rfc8895) for http clients to recieve push updates by starting a session with a GET request on a stream endpoint e.g /node/state/stream.

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features you've considered.

We've considered using the D-BUS api but opted for the REST api for the following reasons:
 - We do not want our application running with the escalated privileges that is required to connect to the system daemon. 
 - We prefer a http && json data implementation due to our current knowledge base
 - Application portablility is important to us
 

**Additional context** Add any other context or screenshots about the feature request here.

inthesquarehole:
> We do not want our application running with the escalated privileges that is required to connect to the system daemon.

Which escalated privileges are that and why would those be such a big deal? I mean, *I* know what they are, but do you?

PR #2515, opened 2024-09-30
[rest] add commissioner api

OmegaRelay:
Added endpoints to the REST server to expose commissioner functionality over the REST API

These changes intend to implement commissioner functionality in the same style as the current rest API offering.
 - The `/node/commissioner/state` endpoint allows the user to get/set the commissioner state using GET and PUT requests respectively.
 - The `/node/commissioner/joiner` allows the user to manage joiners accepted by the commissioner, supporting POST, GET and DELETE request methods to add, get and remove a joiner respectively. 

All added endpoints and their usage have been documented in the openapi.yaml

Example usage:
image

Please give me feedback and let me know of any required changes 😄

### DIFF
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -374,7 +374,11 @@ paths:
           application/json:
             schema:
                 type: string
-                description: Eui64 address or discerner of joiner to remove
+                description: |-
+                  Joiner ID to remove, can be either:
+                   - An EUI64 in the form of a 16 character hex string
+                   - A discerner in the form of the discerner hex value 
+                     (optionally with leading 0x) and bit length separated by a '/'
                 example: "0xabc/12"
       responses:
         "200":
@@ -575,7 +579,10 @@ components:
           example: "0123456789abcdef"
         Discerner:
           type: string
-          description: A string of the joiner discerner, mutually exclusive with JoinerId and Eui64
+          description: |- 
+            A discerner in the form of the discerner hex value (optionally with leading 0x) 
+            and bit length separated by a '/'.
+            Field is mutually exclusive with JoinerId and Eui64.
           example: "0xabc/12"
         Timeout:
           type: integer
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -363,6 +363,8 @@ paths:
           description: Invalid request body.
         "409":
           description: Adding joiner rejected because commissioner is not active.
+        "507":
+          description: Number of joiners the commissioner supports is full and the new one cannot be added.
     delete:
       tags: 
         - node
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -345,10 +345,10 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/JoinerData"
-    put:
+    post:
       tags:
         - node
-      summary: Creates a new joiner
+      summary: Adds a new joiner
       requestBody:
         content:
           application/json:
@@ -370,11 +370,13 @@ paths:
           application/json:
             schema:
                 type: string
-                description: Eui64 address of joiner to remove
-                example: "0123456789abcdef"
+                description: Eui64 address or discerner of joiner to remove
+                example: "0xabc/12"
       responses:
         "200":
           description: Successfully removed joiner.
+        "204":
+          description: Joiner not found.
         "400":
           description: Invalid request body.
         "409":
@@ -558,12 +560,20 @@ components:
           type: string
           description: Joining device's pre-shared key
           example: J01N
+        JoinerId:
+          type: string
+          description: A string of the EUI-64, Discerner, or "*", mutually exclusive with Eui64 and Discerner
+          example: "0xabc/12"
+          default: "*"
         Eui64:
           type: string
-          description: A string of the EUI-64 of the device to join or "*" for any
+          description: A string of the EUI-64, mutually exclusive with JoinerId and Discerner
           example: "0123456789abcdef"
-          default: "*"
+        Discerner:
+          type: string
+          description: A string of the joiner discerner, mutually exclusive with JoinerId and Eui64
+          example: "0xabc/12"
         Timeout:
-          type: uint32
+          type: integer
           description: Joiner expiration time in milliseconds on response and seconds on request 
           default: 60 
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -321,6 +321,8 @@ paths:
       responses:
         "200":
           description: Successful operation.
+        "204":
+          description: Already in state.
         "409":
           description: Cannot set commissioner state because border router state is not active
       requestBody:
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -317,7 +317,7 @@ paths:
         - commissioner
       summary: Set current Commissioner state.
       description: |-
-        Enable and disable the Commissioner.
+        Enable or disable the Commissioner.
       responses:
         "200":
           description: Successful operation.
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -291,6 +291,83 @@ paths:
           description: Successfully created the pending operational dataset.
         "400":
           description: Invalid request body.
+  /node/commissioner/state:
+    get:
+      tags: 
+        - node
+        - commissioner
+      summary: Get current Commissioner state.
+      description: |-
+        State describing the current Commissioner role of this Thread node.
+        - disabled
+        - petitioning
+        - active
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                type: string
+                description: Current state
+                example: "active"
+    put:
+      tags:
+        - node
+        - commissioner
+      summary: Set current Commissioner state.
+      description: |-
+        Enable and disable the Commissioner.
+      responses:
+        "200":
+          description: Successful operation.
+        "304":
+          description: Commissioner is already in requested state
+      requestBody:
+        description: New Commissioner state
+        content:
+          application/json:
+            schema:
+              type: string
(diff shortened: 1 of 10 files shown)


marijnacht:
Will this PR be picked up as we kind of need this functionality. I also see another PR #2514 but that is really large change.

src/rest/openapi.yaml
```diff
@@ -291,6 +291,101 @@ paths:
...
+      tags: 
+        - node
+      summary: Removes a joiner from the node
+      requestBody:
+        content:
+          application/json:
+            schema:
+                type: string
+                description: Eui64 address or discerner of joiner to remove
+                example: "0xabc/12"
```
wgtdkp, line 378: Could you add document explains the format 0xabc/12? What does "/12" mean?

src/rest/resource.cpp
```diff
@@ -800,6 +809,252 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+{
+    otbrError   error = OTBR_ERROR_NONE;
+    std::string errorCode;
+    std::string body;
+
+    VerifyOrExit(Json::JsonString2String(aRequest.GetBody(), body), error = OTBR_ERROR_INVALID_ARGS);
+    if (body == "enable")
+    {
+        VerifyOrExit(otCommissionerGetState(mInstance) == OT_COMMISSIONER_STATE_DISABLED,
+                     error = OTBR_ERROR_DUPLICATED);
```
wgtdkp, line 835: What about accepting it in this case so that the REST API is idempotent?

src/rest/resource.cpp
```diff
@@ -800,6 +809,252 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+        VerifyOrExit(otCommissionerStart(mInstance, NULL, NULL, NULL) == OT_ERROR_NONE,
+                     error = OTBR_ERROR_INVALID_STATE);
+    }
+    else if (body == "disable")
+    {
+        VerifyOrExit(otCommissionerGetState(mInstance) != OT_COMMISSIONER_STATE_DISABLED,
+                     error = OTBR_ERROR_DUPLICATED);
```
wgtdkp, line 842: Ditto, it's better to make the REST API idempotent.

src/rest/resource.cpp
```diff
@@ -800,6 +809,252 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+        ErrorHandler(aResponse, HttpStatusCode::kStatusConflict);
+    }
+    else if (error == OTBR_ERROR_INVALID_ARGS)
+    {
+        ErrorHandler(aResponse, HttpStatusCode::kStatusBadRequest);
+    }
+    else if (error != OTBR_ERROR_NONE)
+    {
+        ErrorHandler(aResponse, HttpStatusCode::kStatusInternalServerError);
+    }
```
wgtdkp, line 869: It's probably safer to have an `else {}` to capture any error which may be missed from the `if ()` statements.

src/rest/resource.cpp
```diff
@@ -800,6 +809,252 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+exit:
+    if (error == OTBR_ERROR_INVALID_STATE)
+    {
+        ErrorHandler(aResponse, HttpStatusCode::kStatusConflict);
+    }
+    else if (error == OTBR_ERROR_INVALID_ARGS)
+    {
+        ErrorHandler(aResponse, HttpStatusCode::kStatusBadRequest);
+    }
+    else if (error == OTBR_ERROR_OPENTHREAD)
```
wgtdkp, line 958: Instead of using OTBR_ERROR_OPENTHREAD for InsufficientStorage, it's probably better to just use OT error directly, instead of casting OT error to OTBR error.

src/rest/resource.cpp
```diff
@@ -800,6 +809,252 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+        }
+        else if (error != OTBR_ERROR_NONE)
+        {
+            ExitNow(error = OTBR_ERROR_INVALID_ARGS);
+        }
+    }
+
+    if (discerner.mLength == 0)
+    {
+        VerifyOrExit(otCommissionerRemoveJoiner(mInstance, addrPtr) == OT_ERROR_NONE, error = OTBR_ERROR_NOT_FOUND);
```
wgtdkp, line 1001: Can we consider it success if the joiner to be removed has already been removed? This will make the API idempotent.

src/rest/resource.cpp
```diff
@@ -800,6 +809,252 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+    }
+    else
+    {
+        VerifyOrExit(otCommissionerRemoveJoinerWithDiscerner(mInstance, &discerner) == OT_ERROR_NONE,
```
wgtdkp, line 1005: Ditto

src/utils/hex.cpp
```diff
@@ -92,6 +92,24 @@ int Hex2Bytes(const char *aHex, uint8_t *aBytes, uint16_t aBytesLength)
     return static_cast<int>(cur - aBytes);
 }
 
+size_t Uint2Hex(const uint64_t aUint, const uint8_t aUintLength, char *aHex)
```
wgtdkp, line 95: Can't `Long2Hex()` in this file fulfill your requirements?

wgtdkp (commented):
Thanks for the contribution! Overall loooks good to me, a few suggestions below:

src/utils/hex.cpp
OmegaRelay: The issue with long2Hex is that the string in in little endian. I could write a string reversal if that would be preferable

OmegaRelay: using `snprint()` for direct conversion gives compiler warnings that `%016lx` must be used for unsigned long but this then breaks 32bit architecture compatibility where it needs to be `%016llx`

OmegaRelay:
Hey @wgtdkp, I've made the requested changes, do you want resolve the comments?

src/rest/resource.cpp
wgtdkp: @OmegaRelay What about this comment?

src/rest/resource.cpp
wgtdkp: @OmegaRelay What about this comment?

wgtdkp:
> Hey @wgtdkp, I've made the requested changes, do you want me to resolve the comments?

@OmegaRelay It's typically the PR author to resolve the comments if the suggests in the comment is followed. If you have different ideas on the topic, reply to the comment and leave it open, the reviewers will take a look again and close it when a consensus is reached.

I did take a glance and it seems my comments are not resolved. Could you check the comments again? thank you!

src/rest/resource.cpp
OmegaRelay: I updated the error handling so that `NOT_FOUND` also sets a status ok response

wgtdkp: It would be clearer to return `OT_ERROR_NONE` in this case.

src/rest/resource.cpp
```diff
@@ -800,6 +809,256 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+        ErrorHandler(aResponse, HttpStatusCode::kStatusInternalServerError);
+    }
+}
+
+void Resource::RemoveJoiner(const Request &aRequest, Response &aResponse) const
+{
+    otbrError         error = OTBR_ERROR_NONE;
+    std::string       errorCode;
+    otExtAddress      eui64;
+    otExtAddress     *addrPtr   = nullptr;
```
wgtdkp, line 980: I think this `addrPtr` is not necessary.

src/rest/resource.cpp
```diff
@@ -800,6 +809,256 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+        }
+        else if (error != OTBR_ERROR_NONE)
+        {
+            ExitNow(error = OTBR_ERROR_INVALID_ARGS);
+        }
+    }
+
+    if (discerner.mLength == 0)
+    {
+        otCommissionerRemoveJoiner(mInstance, addrPtr);
```
wgtdkp, line 1008: The CI test is failing with this line, see https://github.com/openthread/ot-br-posix/actions/runs/11502608180/job/32019937960?pr=2515. This is because OpenThread API requires checking the returned `otError` value. If you are sure that this method can't fail, you can ignore it by adding the **(void)** prefix.

```suggestion
        (void) otCommissionerRemoveJoiner(mInstance, addrPtr);
```

There are more errors like this, you would want to fix all of them to pass the CI.

wgtdkp (approved):
Overall LGTM, thank you! 👍

src/rest/resource.cpp
OmegaRelay: it's there to provide a NULL argument to remove all if the wildcard `*` was given or to point to the eui64 if the wildcard was not given

wgtdkp: Oh, I missed that. The current code looks good, thanks!

martinzi:
@OmegaRelay A Joiner drops from the joiner table after timeout, but also after successfully joining the network. Could your API differentiate between these two scenarios?

OmegaRelay:
> @OmegaRelay A Joiner drops from the joiner table after timeout, but also after successfully joining the network. Could your API differentiate between these two scenarios?

Hi @martinzi, this PR is intended to be a simple extension of the current stateless API and as such implementing this joiner status feature feels outside the scope of this PR. 

On a side note, I am currently looking at implementing server-sent events which can also cover this use case and keep the API itself stateless, I do have an open [feature request](https://github.com/openthread/ot-br-posix/issues/2492) to discuss this but it hasn't had much interaction.

jwhui (approved):
Thanks! 👍🏼

PR #2526, opened 2024-10-08
[thread host] add support for vendor thread host

Irving-cl:
This PR adds support for vendors to create customized `ThreadHost`.

This allows vendors to provide customized implementation for `ThreadHost` APIs.

### DIFF
--- a/src/ncp/thread_host.hpp
+++ b/src/ncp/thread_host.hpp
@@ -232,6 +232,9 @@ public:
      * The destructor.
      */
     virtual ~ThreadHost(void) = default;
+
+protected:
+    static CoprocessorType InitCoprocessor(const std::vector<const char *> &aRadioUrls);
 };
 
 } // namespace Ncp
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -158,6 +158,11 @@ if (OTBR_VENDOR_SERVER)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_VENDOR_SERVER=1)
 endif()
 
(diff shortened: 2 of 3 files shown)


Irving-cl:
@wgtdkp @superwhd 

As discussed earlier, we hope that we can avoid changing the implementation in public ot-br-posix when we try to change the behavior of Thread Host Api on a specific platform. This PR provides a way to customize the `ThreadHost` child class on vendor platforms. In this way, vendors can override the implementation of `RcpHost` or `NcpHost`. What do you think of the change?

src/ncp/thread_host.hpp
```diff
@@ -170,6 +170,14 @@ class ThreadHost : virtual public NetworkProperties
 
 protected:
     static otLogLevel ConvertToOtLogLevel(otbrLogLevel aLevel);
+
+private:
+    static std::unique_ptr<ThreadHost> CreateThreadHost(CoprocessorType                  aType,
```
superwhd, line 175: Where should a vendor implement this method?

I feel it kind of inconspicuous that a vendor needs to implement this private static method.

Irving-cl: I've refactor it. Let's just use the static `Create` method for vendor to provide a different implementation.

In terms of Android, I plan to put the vendor implementation at `src/ncp/android`, which is a new directory. See this draft [CL](https://android-review.git.corp.google.com/c/platform/external/ot-br-posix/+/3305636).

Irving-cl:
As discussed offline, we don't need this change for now because we can quickly sync the change here to aosp.

Irving-cl:
Per latest discussion, we will have this change to allow vendor platforms to extend ThreadHost API while using the implementation of ThreadHost API here.

PR #2532, opened 2024-10-10
[docker] fix access to DBus when restarting container / rebooting host

bouvrais:
This fixes an issue where the `otbr-agent` would not be properly stopped when the container receives SIGTERM (e.g. `docker stop otbr`).
In such scenarios, the agent would not be able to use dbus upon restart and the service would be in a broken state. This happens on container restart, docker service restart, host reboot, etc.

### DIFF
--- a/script/_initrc
+++ b/script/_initrc
@@ -107,9 +107,9 @@ start_service()
 stop_service()
 {
     local service_name=$1
-    if [[ $HAVE_SYSTEMCTL == 1 ]]; then
+    if [[ ${HAVE_SYSTEMCTL} == 1 ]]; then
         systemctl is-active "$service_name" && sudo systemctl stop "$service_name" || echo "Failed to stop $service_name!"
-    elif [[ $HAVE_SERVICE == 1 ]]; then
+    elif [[ ${HAVE_SERVICE} == 1 ]]; then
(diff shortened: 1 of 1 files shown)


script/_initrc
```diff
@@ -107,9 +107,9 @@ start_service()
 stop_service()
 {
     local service_name=$1
-    if $HAVE_SYSTEMCTL; then
+    if [[ $HAVE_SYSTEMCTL == 1 ]]; then
```
superwhd, line 110: nit: be consistent with the style in `start_service()`
```suggestion
    if [[ ${HAVE_SYSTEMCTL} == 1 ]]; then
```

script/_initrc
```diff
@@ -107,9 +107,9 @@ start_service()
...
         systemctl is-active "$service_name" && sudo systemctl stop "$service_name" || echo "Failed to stop $service_name!"
-    elif $HAVE_SERVICE; then
+    elif [[ $HAVE_SERVICE == 1 ]]; then
```
superwhd, line 112:
```suggestion
    elif [[ ${HAVE_SERVICE} == 1 ]]; then
```

superwhd (approved):
LGTM overall with nits

jwhui (approved):
Thanks! 👍🏼

PR #2534, opened 2024-10-11
[infra if] implement infra if send icmp6 nd

Irving-cl:
This PR implements `SendIcmp6Nd` in the `InfraIf` module.

The method is used to send an ICMPv6 ND message (on host) when NCP calls `otPlatInfraIfSendIcmp6Nd`. In this PR, the method hasn't been called. It will be later integrated with `NcpSpinel`. The implementation of this method is the same as OT posix [implementation](https://github.com/openthread/openthread/blob/main/src/posix/platform/infra_if.cpp#L214).

### DIFF
--- a/src/ncp/posix/infra_if.cpp
+++ b/src/ncp/posix/infra_if.cpp
@@ -28,6 +28,10 @@
 
 #define OTBR_LOG_TAG "INFRAIF"
 
+#ifdef __APPLE__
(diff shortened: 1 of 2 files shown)


superwhd (approved):
LGTM 👍

abtink (approved):
👍

jwhui (approved):

PR #2538, opened 2024-10-12
[host] add detach gracefully method

Irving-cl:
This PR adds a method `DetachGracefully` in the `ThreadHost` APIs.

This method instructs the device to detach from the network gracefully without erasing the persistent network info. The method is required in `SetThreadEnabled` on Android platform where detachGracefully is called first and then disable the ifconfig and Thread stack.

This PR also implements the `Leave` method for `RcpHost` for it to be called in Android source code.

### DIFF
--- a/src/ncp/rcp_host.cpp
+++ b/src/ncp/rcp_host.cpp
@@ -295,6 +295,9 @@ void RcpHost::Deinit(void)
     OtNetworkProperties::SetInstance(nullptr);
     mThreadStateChangedCallbacks.clear();
     mResetHandlers.clear();
+
+    mDetachGracefullyResultReceiver = nullptr;
+    mLeaveResultReceiver            = nullptr;
 }
 
 void RcpHost::HandleStateChanged(otChangedFlags aFlags)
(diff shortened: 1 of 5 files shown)


src/ncp/rcp_host.cpp
```diff
@@ -402,10 +405,65 @@ void RcpHost::Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const A
...
 {
-    // TODO: Implement Leave under RCP mode.
-    mTaskRunner.Post([aReceiver](void) { aReceiver(OT_ERROR_NOT_IMPLEMENTED, "Not implemented!"); });
+    if (mInstance == nullptr)
+    {
+        mTaskRunner.Post([&aReceiver](void) { aReceiver(OT_ERROR_INVALID_STATE, ""); });
+        ExitNow();
+    }
+
+    mLeaveResultReceiver = aReceiver;
```
sunytt, line 463: I think it's better if `Leave()` follows the same style as `DetachGracefully()` to return `OT_ERROR_BUSY` if there is already ongoing request? Otherwise it's easy for callers to mess up with the receiver.

Another approach is that both of the functions keep a list of receivers and return success for all concurrent requests. Thoughts?

src/ncp/thread_host.hpp
```diff
@@ -112,7 +112,19 @@ class ThreadHost : virtual public NetworkProperties
...
+     * @param[in] aReceiver  A receiver to get the async result of this operation.
+     */
+    virtual void DetachGracefully(const AsyncResultReceiver &aRecevier) = 0;
+
+    /**
+     * This method instructs the device to detach from the current network gracefully and then erase the persistent
+     * network credentials.
      *
      * 1. If there is already an ongoing 'Leave' operation, no action will be taken and @p aReceiver
      *    will be called after the previous request is completed. The previous @p aReceiver will also
```
sunytt, line 130: This doesn't seem true with the current code, only latest Receiver will be called it seems?

Irving-cl:
This method isn't needed anymore.

PR #2540, opened 2024-10-14
[host] add active dataset as a network property

Irving-cl:
This PR adds active dataset tlvs to `NetworkProperties` so that for both NCP and RCP active dataset can be accessed directly.

The active dataset is used in many places in the platform code. For example, in android `SetThreadEnabled`, it will only enable Thread when the active dataset is a valid one.

I verified the NCP case locally by log (dumping the active set and see if it's correctly set). If writing a test for it, we need add one more dbus API, which seems unworthy now.


zhanglongxia (approved):
LGTM

abtink (approved):
Looks great. 👍

jwhui (approved):

PR #2542, opened 2024-10-15
[host] add SetThreadEnabled to ThreadHost APIs

Irving-cl:
This PR adds a new method `SetThreadEnabled` as a ThreadHost API.

The method is added so that the Android binder API can be easily implemented without the need to handle NCP/RCP case. In this PR, `RcpHost::SetThreadEnabled` is impelmented and `NcpHost::SetThreadEnabled` will be implemented later.

### DIFF
--- a/src/ncp/rcp_host.cpp
+++ b/src/ncp/rcp_host.cpp
@@ -295,6 +295,8 @@ void RcpHost::Deinit(void)
     OtNetworkProperties::SetInstance(nullptr);
     mThreadStateChangedCallbacks.clear();
     mResetHandlers.clear();
+
+    mSetThreadEnabledReceiver = nullptr;
 }
 
 void RcpHost::HandleStateChanged(otChangedFlags aFlags)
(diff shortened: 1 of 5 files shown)


src/ncp/rcp_host.cpp
```diff
@@ -417,6 +419,63 @@ void RcpHost::ScheduleMigration(const otOperationalDatasetTlvs &aPendingOpDatase
...
+    VerifyOrExit(mInstance != nullptr, error = OT_ERROR_INVALID_STATE);
+    VerifyOrExit(mSetThreadEnabledReceiver == nullptr, error = OT_ERROR_BUSY);
+    mSetThreadEnabledReceiver = aReceiver;
+
+    if (aEnabled)
+    {
+        otOperationalDatasetTlvs datasetTlvs;
+
+        if (otDatasetGetActiveTlvs(mInstance, &datasetTlvs) != OT_ERROR_NOT_FOUND && datasetTlvs.mLength > 0 &&
+            !IsAttached())
```
abtink, line 436: We can have situation where `otThreadSetEnabled()` is already done but device is in `DETACHED` state?  We would go through and call `otThreadSetEnabled() again which I think will return an error.

How about we check if we have dataset and device is `DISABLED` role here?

abtink (approved):
LGTM. One smaller suggestion below:

sunytt (approved):
LGTM, thanks!

jwhui (approved):

PR #2543, opened 2024-10-16
[test] increase reply timeout for AttachToAllNodes

bukepo:
This commit increases the reply timeout for AttachToAllNodes due to the increased leader restore timeout.


abtink (approved):
👍

jwhui (approved):

PR #2547, opened 2024-10-18
[infra if] add infra if icmp6 nd receiving

Irving-cl:
This PR implements Icmp6Nd receiving in the InfraIf module.

The method is used to receive ICMPv6 ND messages (on host) and forward them to the NCP. The implementation of this method is the same as OT posix [implementation](https://github.com/openthread/openthread/blob/main/src/posix/platform/infra_if.cpp#L593). In this PR, the forwarding to NCP hasn't been implemented. It will be later integrated with NcpSpinel. 

This PR also adds a test case to verify that `HandleIcmp6Nd` is called correctly when the infrastructure interface receives an ICMP6 ND message.

### DIFF
--- a/src/ncp/posix/infra_if.cpp
+++ b/src/ncp/posix/infra_if.cpp
@@ -60,6 +60,11 @@ otbrError InfraIf::Dependencies::SetInfraIf(unsigned int                   aInfr
     return OTBR_ERROR_NONE;
 }
 
+otbrError InfraIf::Dependencies::HandleIcmp6Nd(uint32_t, const Ip6Address &, const uint8_t *, uint16_t)
+{
+    return OTBR_ERROR_NONE;
+}
+
 InfraIf::InfraIf(Dependencies &aDependencies)
     : mDeps(aDependencies)
     , mInfraIfIndex(0)
@@ -118,31 +123,41 @@ void InfraIf::Deinit(void)
 
 void InfraIf::Process(const MainloopContext &aContext)
 {
-    OT_UNUSED_VARIABLE(aContext);
+    VerifyOrExit(mInfraIfIcmp6Socket != -1);
 #ifdef __linux__
     VerifyOrExit(mNetlinkSocket != -1);
+#endif
 
+    if (FD_ISSET(mInfraIfIcmp6Socket, &aContext.mReadFdSet))
+    {
+        ReceiveIcmp6Message();
+    }
+#ifdef __linux__
     if (FD_ISSET(mNetlinkSocket, &aContext.mReadFdSet))
     {
         ReceiveNetlinkMessage();
     }
+#endif
 
 exit:
-#endif
     return;
 }
 
 void InfraIf::UpdateFdSet(MainloopContext &aContext)
 {
-    OT_UNUSED_VARIABLE(aContext);
+    VerifyOrExit(mInfraIfIcmp6Socket != -1);
 #ifdef __linux__
     VerifyOrExit(mNetlinkSocket != -1);
+#endif
 
+    FD_SET(mInfraIfIcmp6Socket, &aContext.mReadFdSet);
+    aContext.mMaxFd = std::max(aContext.mMaxFd, mInfraIfIcmp6Socket);
+#ifdef __linux__
     FD_SET(mNetlinkSocket, &aContext.mReadFdSet);
     aContext.mMaxFd = std::max(aContext.mMaxFd, mNetlinkSocket);
+#endif
 
 exit:
-#endif
     return;
 }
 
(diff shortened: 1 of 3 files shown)


src/ncp/posix/infra_if.cpp
```diff
@@ -376,6 +399,75 @@ bool InfraIf::HasLinkLocalAddress(const std::vector<Ip6Address> &aAddrs)
...
+    bufferLength = static_cast<uint16_t>(rval);
+
+    for (cmh = CMSG_FIRSTHDR(&msg); cmh; cmh = CMSG_NXTHDR(&msg, cmh))
+    {
+        if (cmh->cmsg_level == IPPROTO_IPV6 && cmh->cmsg_type == IPV6_PKTINFO &&
+            cmh->cmsg_len == CMSG_LEN(sizeof(struct in6_pktinfo)))
+        {
+            struct in6_pktinfo pktinfo;
+
+            memcpy(&pktinfo, CMSG_DATA(cmh), sizeof pktinfo);
```
bukepo, line 447:
```suggestion
            const struct in6_pktinfo *pktinfo = reinterpret_cast<struct in6_pktinfo *>(CMSG_DATA(cmh));
```

src/ncp/posix/infra_if.cpp
```diff
@@ -376,6 +399,75 @@ bool InfraIf::HasLinkLocalAddress(const std::vector<Ip6Address> &aAddrs)
...
+        {
+            hopLimit = *(int *)CMSG_DATA(cmh);
+        }
+    }
+
+    VerifyOrExit(ifIndex == mInfraIfIndex, error = OTBR_ERROR_DROPPED);
+
+    // We currently accept only RA & RS messages for the Border Router and it requires that
+    // the hoplimit must be 255 and the source address must be a link-local address.
+    VerifyOrExit(hopLimit == 255 && IN6_IS_ADDR_LINKLOCAL(&srcAddr.sin6_addr), error = OTBR_ERROR_DROPPED);
```
bukepo, line 462:
```suggestion
            VerifyOrExit(pktinfo.ipi6_ifindex == mInfraIfIndex, error = OTBR_ERROR_DROPPED);
        }
        else if (cmh->cmsg_level == IPPROTO_IPV6 && cmh->cmsg_type == IPV6_HOPLIMIT &&
                 cmh->cmsg_len == CMSG_LEN(sizeof(int)))
        {
            int hopLimit = *(int *)CMSG_DATA(cmh);
            VerifyOrExit(hopLimit == 255, error = OTBR_ERROR_DROPPED);
        }
    }
```

src/ncp/posix/infra_if.cpp
```diff
@@ -376,6 +399,75 @@ bool InfraIf::HasLinkLocalAddress(const std::vector<Ip6Address> &aAddrs)
...
+    msg.msg_control    = cmsgbuf;
+    msg.msg_controllen = sizeof(cmsgbuf);
+
+    rval = recvmsg(mInfraIfIcmp6Socket, &msg, 0);
+    if (rval < 0)
+    {
+        otbrLogWarning("Failed to receive ICMPv6 message: %s", strerror(errno));
+        ExitNow(error = OTBR_ERROR_DROPPED);
+    }
+
```
bukepo, line 437:
```suggestion

    VerifyOrExit(ifIndex == mInfraIfIndex, error = OTBR_ERROR_DROPPED);

    // We currently accept only RA & RS messages for the Border Router and it requires that
    // the hoplimit must be 255 and the source address must be a link-local address.
    VerifyOrExit(IN6_IS_ADDR_LINKLOCAL(&srcAddr.sin6_addr), error = OTBR_ERROR_DROPPED);
```

src/ncp/posix/infra_if.cpp
```diff
@@ -60,6 +60,19 @@ otbrError InfraIf::Dependencies::SetInfraIf(unsigned int                   aInfr
...
+                                               const uint8_t    *aData,
+                                               uint16_t          aDataLen)
+{
+    OTBR_UNUSED_VARIABLE(aInfraIfIndex);
+    OTBR_UNUSED_VARIABLE(aSrcAddress);
+    OTBR_UNUSED_VARIABLE(aData);
+    OTBR_UNUSED_VARIABLE(aDataLen);
+
+    return OTBR_ERROR_NONE;
+}
```
bukepo, line 74: Perhaps we can omit variable names for dummy methods.

@abtink thoughts?

```suggestion
otbrError InfraIf::Dependencies::HandleIcmp6Nd(uint32_t          ,
                                               const Ip6Address &,
                                               const uint8_t    *,
                                               uint16_t          )
{
    return OTBR_ERROR_NONE;
}
```

src/ncp/posix/infra_if.cpp
Irving-cl: `ifIndex` is obtained later in the parsing so I don't think it can be moved here. Thoughts?

src/ncp/posix/infra_if.cpp
Irving-cl: I think making `hopLimit` as a function-scope variable and verify it outside the loop is better. Because in this way if the message doesn't contain the hoplimit, it will be dropped. Otherwise it will be processed. Thoughts?

bukepo (approved):

src/ncp/posix/infra_if.cpp
```diff
@@ -376,6 +391,73 @@ bool InfraIf::HasLinkLocalAddress(const std::vector<Ip6Address> &aAddrs)
     return hasLla;
 }
 
+void InfraIf::ReceiveIcmp6Message(void)
+{
+    otbrError error = OTBR_ERROR_NONE;
+    uint8_t   buffer[1500];
```
abtink, line 397: Should we define a constant for `1500`?

abtink (approved):
LGTM. 👍

src/ncp/posix/infra_if.cpp
Irving-cl: Updated!

jwhui (approved):

PR #2549, opened 2024-10-19
[test] add unit test framework for ThreadHost APIs (RCP)

Irving-cl:
This PR adds a unit test framework for testing ThreadHost APIs (RCP).

It's implemeneted by building `rcp_host.cpp` with `FakePlatform` in OT. A few small changes are done to make this work:
* implement openthread posix system APIs in  `fake_posix_platform.cpp`. (like `otSysInit`, `otSysMainloopProcess `) `ot::FakePlatform` doesn't include these things and I think it's better to put it here.
* Remove the fixed build dependency of `otbr-common` on `openthread-posix` so that the unit test executable can link to the fake system API implementation instead those in `openthread-posix`.
* Move `ThreadHost::ConvertToOtLogLevel` to `logging.hpp`. Because otherwise we have to build `thread_host.cpp` into the test executable and then have more dependencies on NcpHost.

This PR also implements the first unit test case to test the `RcpHost::SetThreadEnabled` method.

### DIFF
--- a/tests/gtest/test_rcp_host_api.cpp
+++ b/tests/gtest/test_rcp_host_api.cpp
@@ -1,0 +1,131 @@
+/*
+ *    Copyright (c) 2024, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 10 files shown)


bukepo (approved):

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -0,0 +1,129 @@
...
+        resultReceived = true;
+        error          = aError;
+    };
+    otbr::Ncp::RcpHost host("wpan0", std::vector<const char *>(), /* aBackboneInterfaceName */ "", /* aDryRun */ false,
+                            /* aEnableAutoAttach */ false);
+
+    host.Init();
+
+    // 1. Active dataset hasn't been set, should succeed with device role still being disabled.
+    host.SetThreadEnabled(true, [&resultReceived, &error](otError aError, const std::string &aErrorMsg) {
```
sunytt, line 58: should it use `receiver` here?

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -0,0 +1,129 @@
...
+        OT_UNUSED_VARIABLE(aErrorMsg);
+        resultReceived = true;
+        error          = aError;
+    });
+    while (!resultReceived)
```
sunytt, line 63: this may get stuck forever if the `SetThreadEnabled` api is not implemented correctly? May give it a guard for max time or max times checked. Same for the other while loops below, may also give each one a corresponding error message to indicate which received failed to be called?

Irving-cl: Good point! I added a util function `MainloopProcessUtil` which allows to check timeout as well.

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -0,0 +1,134 @@
...
+                                 uint32_t                  aTimeoutSec,
+                                 std::function<bool(void)> aCondition)
+{
+    timeval startTime;
+    timeval now;
+    gettimeofday(&startTime, nullptr);
+
+    while (!aCondition())
+    {
+        if (aTimeoutSec > 0)
```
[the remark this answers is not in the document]
Irving-cl: Make sense. Updated!

sunytt (approved):
LGTM 👍

PR #2557, opened 2024-10-22
[rest] fix: rename of Dataset to ActiveDataset in otbr open-api.yaml

harterSe:
There was an link to an not existing component which resulted in an error when generating an open api client.

The error was in the Route

/node/dataset/active(get) 

The missing component was:
$ref: "#/components/schemas/Dataset"

I changed the route to use the existing component:
$ref: "#/components/schemas/ActiveDataset"


wgtdkp (approved):
Thanks for the fix!

jwhui (approved):
Thanks! 👍🏼

PR #2559, opened 2024-10-24
[thread host] implement schedule migration for rcp host

Irving-cl:
This PR implements the `ScheduleMigration` method in `RcpHost`.


src/ncp/rcp_host.hpp
```diff
@@ -269,7 +273,7 @@ class RcpHost : public MainloopProcessor, public ThreadHost, public OtNetworkPro
     bool                                       mEnableAutoAttach = false;
 
     AsyncResultReceiver mSetThreadEnabledReceiver;
-
+    AsyncResultReceiver mScheduleMigrationReceiver;
```
sunytt, line 276: nit: new line after this?

sunytt (approved):
LGTM 👍

src/ncp/rcp_host.cpp
```diff
@@ -424,10 +425,43 @@ void RcpHost::Leave(const AsyncResultReceiver &aReceiver)
...
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        mTaskRunner.Post([aReceiver, error, errorMsg](void) { aReceiver(error, errorMsg); });
+    }
+    else
+    {
+        // otDatasetSendMgmtPendingSet() returns OT_ERROR_BUSY if it has already been called before but the
+        // callback hasn't been invoked. So we can guarantee that mMigrationReceiver is always nullptr here
+        assert(mScheduleMigrationReceiver == nullptr);
```
jwhui, line 452: Should we use `OT_ASSERT` here?

Irving-cl: I think we should just use `assert` because `OT_ASSERT` is within OT core (`openthread/src/core/common/debug.hpp`). We try to avoid including OT core headers in ot-br-posix because that may cause conflicts. In ot-br-posix `assert` is already used in many places.

jwhui (approved):

PR #2560, opened 2024-10-24
[mdns] refactor mDNS State subscription

Irving-cl:
This PR refactors the mDNS State subscription using the Subject-Observer pattern.

The PR makes the state subscription more flexible. Currently the mDNS state is published in `Application::HandleMdnsState`. Now we have both NCP and RCP case. This function will be difficult to implement. With this new pattern, we can register different observers during NCP/RCP initialization.

### DIFF
--- a/src/border_agent/border_agent.hpp
+++ b/src/border_agent/border_agent.hpp
@@ -75,7 +75,7 @@ namespace otbr {
 /**
  * This class implements Thread border agent functionality.
  */
-class BorderAgent : private NonCopyable
+class BorderAgent : public Mdns::StateObserver, private NonCopyable
 {
 public:
     /** The callback for receiving ephemeral key changes. */
@@ -139,7 +139,7 @@ public:
      *
      * @param[in] aState  The state of mDNS publisher.
      */
-    void HandleMdnsState(Mdns::Publisher::State aState);
+    void HandleMdnsState(Mdns::Publisher::State aState) override;
 
     /**
      * This method creates ephemeral key in the Border Agent.
--- a/src/mdns/mdns.hpp
+++ b/src/mdns/mdns.hpp
@@ -666,6 +666,64 @@ protected:
 };
 
 /**
+ * This interface is a mDNS State Observer.
+ */
+class StateObserver
+{
+public:
+    /**
+     * This method notifies the mDNS state to the observer.
+     *
+     * @param[in] aState  The mDNS State.
+     */
+    virtual void HandleMdnsState(Publisher::State aState) = 0;
+
+    /**
+     * The destructor.
+     */
+    virtual ~StateObserver(void) = default;
+};
+
+/**
+ * This class defines a mDNS State Subject.
+ */
+class StateSubject
+{
+public:
+    /**
+     * Constructor.
+     */
+    StateSubject(void) = default;
+
+    /**
+     * Destructor.
+     */
+    ~StateSubject(void) = default;
+
+    /**
+     * This method adds an mDNS State Observer to this subject.
+     *
+     * @param[in] aObserver  A reference to the observer. If it's nullptr, it won't be added.
+     */
+    void AddObserver(StateObserver &aObserver);
+
+    /**
+     * This method updates the mDNS State.
+     *
+     * @param[in] aState  The mDNS State.
+     */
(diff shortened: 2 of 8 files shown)


src/border_agent/border_agent.hpp
```diff
@@ -75,7 +75,7 @@ namespace otbr {
 /**
  * This class implements Thread border agent functionality.
  */
-class BorderAgent : private NonCopyable
+class BorderAgent : private NonCopyable, public Mdns::StateObserver
```
superwhd, line 78: Put the public inheritance in front of private inheritance

src/mdns/mdns.hpp
```diff
@@ -665,6 +665,59 @@ class Publisher : private NonCopyable
...
+
+    /**
+     * This method updates the mDNS State.
+     *
+     * @param[in] aState  The mDNS State.
+     */
+    void UpdateState(Publisher::State aState);
+
+private:
+    std::vector<StateObserver *> mObservers;
```
superwhd, line 718: I feel it kind of dangerous to store all the observers using raw pointers. Is it guaranteed that the observers and the subjects all have the same lifetime?

src/mdns/mdns.hpp
```diff
@@ -665,6 +665,59 @@ class Publisher : private NonCopyable
...
+public:
+    /**
+     * Constructor.
+     */
+    StateSubject(void) = default;
+
+    /**
+     * Destructor.
+     */
+    ~StateSubject(void) = default;
```
superwhd, line 701: virtual

src/sdp_proxy/advertising_proxy.hpp
```diff
@@ -52,7 +52,7 @@ namespace otbr {
 /**
  * This class implements the Advertising Proxy.
  */
-class AdvertisingProxy : private NonCopyable
+class AdvertisingProxy : private NonCopyable, public Mdns::StateObserver
```
superwhd, line 55: ditto

src/sdp_proxy/discovery_proxy.hpp
```diff
@@ -56,7 +56,7 @@ namespace Dnssd {
 /**
  * This class implements the DNS-SD Discovery Proxy.
  */
-class DiscoveryProxy : private NonCopyable
+class DiscoveryProxy : private NonCopyable, public Mdns::StateObserver
```
superwhd, line 59: ditto

src/mdns/mdns.hpp
superwhd: Never mind. I overlooked this.

src/mdns/mdns.hpp
Irving-cl: Good point! It's guaranteed because all the observers are components of `Application` and they will be released at the end of the program (released by unique_ptr in Application's destructor). Before that the mDNSPublisher has stopped working. I also added `StateSubject::Deinit` to remove all observers in de-initialization which is ahead of the destructors. And in otbr-agent, we won't have any cases of run-time obserser allocation and release.

src/mdns/mdns.hpp
```diff
@@ -665,6 +665,64 @@ class Publisher : private NonCopyable
...
+
+    /**
+     * This method adds an mDNS State Observer to this subject.
+     *
+     * @param[in] aObserver  A pointer to the observer. If it's nullptr, it won't be added.
+     */
+    void AddObserver(StateObserver *aObserver);
```
superwhd, line 708: nit: What about declaring the parameter as a reference so it enforces the non-nullity?

src/mdns/mdns.hpp
```diff
@@ -665,6 +665,64 @@ class Publisher : private NonCopyable
...
+     * This method updates the mDNS State.
+     *
+     * @param[in] aState  The mDNS State.
+     */
+    void UpdateState(Publisher::State aState);
+
+    /**
+     * This method removes all the observers.
+     */
+    void Deinit(void);
```
superwhd, line 720: It may be more clear and consistent to name it as `Clear`, for there's no `Init` method.

superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2564, opened 2024-10-25
[tests] add unit test for RcpHost SetCountryCode

Irving-cl:
This PR adds a unit test for `RcpHost::SetCountryCode`.

The PR also fixes an issue of current usage of `FakePlatform`. In each gtest case, we should get a new OT Instance. Otherwise the OT instance used in last gtest case will have some remaining tasks which are invalid to process.

### DIFF
--- a/tests/gtest/fake_posix_platform.cpp
+++ b/tests/gtest/fake_posix_platform.cpp
@@ -28,10 +28,12 @@
 
 #include "fake_platform.hpp"
 
+#include <assert.h>
+
 #include "openthread/openthread-system.h"
 
-otPlatResetReason       gPlatResetReason = OT_PLAT_RESET_REASON_POWER_ON;
-static ot::FakePlatform sFakePlatform;
+otPlatResetReason        gPlatResetReason = OT_PLAT_RESET_REASON_POWER_ON;
+static ot::FakePlatform *sFakePlatform;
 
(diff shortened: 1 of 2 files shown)


src/ncp/rcp_host.cpp
```diff
@@ -322,16 +322,23 @@ void RcpHost::HandleStateChanged(otChangedFlags aFlags)
 
 void RcpHost::Update(MainloopContext &aMainloop)
 {
+    VerifyOrExit(mInstance != nullptr);
```
zhanglongxia, line 325: Why we can skip the code when the `mInstance == nullptr`? Shouldn't we add `assert(mInstance != nullptr)`?

Irving-cl: This is for test purpose. Because the fuzzer tests in Android will call APIs when the host hasn't been initialized. And the unit tests here will also do the same thing. However we need to run the loop to process the tasks and get the async results. And `otTaskletProcess` will be called when `mInstance` is `nullptr`. So we skip here and avoid `otTaskletProcess` from being called in unit tests.

Irving-cl: I've just come up with a better solution. I removed the `VerifyOrExit(mInstance != nullptr);`.

zhanglongxia (approved):
LGTM

jwhui (approved):

PR #2574, opened 2024-10-31
[application] move ThreadHost::Create out of the Application class

Irving-cl:
This PR refactors the `Application` class.

In specific:
1. Decouple `Application` with `InfraLinkSelector` by moving `InfraLinkSelector` out of the `Application` class. The current dependency causes that we cannot move the creation of `ThreadHost` instance out of the constructor of the `Application` class.
2. Move `Ncp::ThreadHost::Create` out of the `Application` constructor.

This change solves the fuzzer test issue in Android: When we create an `Application` object, we have to initlialize the radio which will cause error in the fuzz environment. With this change, we can pass a FuzzThreadHost object to the `Application`.


wgtdkp (approved):

zhanglongxia (approved):
LGTM

PR #2579, opened 2024-11-04
[trel] updates for switching trel interface

sunytt:
As it's possible to switch network interface for TREL in https://github.com/openthread/openthread/pull/10872, updated trel_dnssd.cpp in 2 places:
1). Initialize() can be called multiple times now, so it should clear the mTrelNetifIndex at the beginning to refresh the index according to the given interface name.
2). HandleMdnsState() may be invoked when TrelDnssd is not initialized, we should still update the mMdnsPublisherReady status, so it's ready to publish a service when TrelDnssd becomes initialized.

### DIFF
--- a/src/trel_dnssd/trel_dnssd.cpp
+++ b/src/trel_dnssd/trel_dnssd.cpp
@@ -91,6 +91,9 @@ TrelDnssd::TrelDnssd(Ncp::RcpHost &aHost, Mdns::Publisher &aPublisher)
 void TrelDnssd::Initialize(std::string aTrelNetif)
 {
     mTrelNetif = std::move(aTrelNetif);
+    // Reset mTrelNetifIndex to 0 so that when this function is called with a different aTrelNetif
+    // than the current mTrelNetif, CheckTrelNetifReady() will update mTrelNetifIndex accordingly.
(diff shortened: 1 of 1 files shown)


src/trel_dnssd/trel_dnssd.cpp
```diff
@@ -91,6 +91,9 @@ TrelDnssd::TrelDnssd(Ncp::RcpHost &aHost, Mdns::Publisher &aPublisher)
 void TrelDnssd::Initialize(std::string aTrelNetif)
 {
     mTrelNetif = std::move(aTrelNetif);
+    // Reset mTrelNetifIndex to 0 so that When this function is called with a different aTrelNetif
+    // than the current mTrelNetif, CheckTrelNetifReady() will update mTrelNetifIndex accordingly.
```
superwhd, line 95:
```suggestion
    // Reset mTrelNetifIndex to 0 so that when this function is called with a different aTrelNetif
    // than the current mTrelNetif, CheckTrelNetifReady() will update mTrelNetifIndex accordingly.
```

superwhd (approved):
LGTM 👍

abtink (approved):

jwhui (approved):

PR #2590, opened 2024-11-08
[options] add an option to control power calibration

Irving-cl:
This PR adds an option to enable/disable the power calibration functionality in otbr.

The recently introduced ThreadHost API `SetChannelMaxPowers` depends on the radio platform API `otPlatRadioSetChannelTargetPower` which may not be enabled on some platforms/products. In those cases, we will have undefined symbols error. This PR adds a build option to disable the power calibration function in otbr. On those platforms/products who don't enable `otPlatRadioSetChannelTargetPower`, they should turn off this option as well.


mia1yang (commented):
LGTM

mia1yang (approved):

jwhui (approved):

PR #2595, opened 2024-11-13
[pd] pd daemon service: remove redundant subprocess 'sudo' in python script

yangsong-cnyn:
Removes unnecessary sudo commands from dhcp6_pd_daemon.py.

The script was already running as root due to the `User=root` setting in the `systemd` service file. The internal `sudo` calls were redundant and caused issues with dbus context switching, preventing the script from receiving dbus signals correctly. This change removes the `sudo` calls, ensuring the script operates within a consistent dbus context and resolves the signal handling problems.

This commit addresses the issue where the dhcp6_pd_daemon was not responding to changes in the Dhcp6PdState dbus property.

### DIFF
--- a/script/reference-device/dhcp6_pd_daemon.py
+++ b/script/reference-device/dhcp6_pd_daemon.py
@@ -54,10 +54,9 @@ def restart_dhcpcd_service(config_path):
         logging.error(f"{config_path} not found. Cannot apply configuration.")
         return
     try:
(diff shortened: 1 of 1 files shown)


superwhd (approved):
LGTM 👍

erjiaqing (approved):

jwhui (approved):

PR #2596, opened 2024-11-13
[github-actions] fix MacOS build error: unlink pkg-config before reinstalling

yangsong-cnyn:
Fix macOS build issue caused by pkg-config conflicts.

Unlinking pkg-config before reinstalling it in the MacOS build workflow. This resolves issues arising from stale or conflicting installations of pkg-config, which can lead to build failures.


superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2598, opened 2024-11-13
[thread-host] add common network properties

Irving-cl:
This PR adds a few common network properties to `ThreadHost`.

The direct usage of these properties is in Android platform when refreshing the state. With this PR, that process can work for both NCP & RCP. This PR only implements the Get methods for RCP. The NCP version will be implemented later.


zhanglongxia (approved):
LGTM

jwhui (approved):

PR #2600, opened 2024-11-14
[PD] pd daemon service: ensure dhcp6_pd_daemon starts after otbr-agent

yangsong-cnyn:
The `dhcp6_pd_daemon` occasionally fails to receive dbus messages on initial boot due to a race condition with `otbr-agent`.  This commit fixes the issue by explicitly defining a service dependency on `otbr-agent`.


superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2602, opened 2024-11-14
[rest] add coprocessor version string API

puddly:
We'd like to access the current NCP/RCP coprocessor firmware version string from the REST API. I've added a new API endpoint under `/node/coprocessor/version` that fetches it with `GetCoprocessorVersion()`.

### DIFF
--- a/src/rest/resource.cpp
+++ b/src/rest/resource.cpp
@@ -49,6 +49,8 @@
 #define OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING "/node/dataset/pending"
 #define OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_STATE "/node/commissioner/state"
 #define OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_JOINER "/node/commissioner/joiner"
+#define OT_REST_RESOURCE_PATH_NODE_COPROCESSOR "/node/coprocessor"
+#define OT_REST_RESOURCE_PATH_NODE_COPROCESSOR_VERSION "/node/coprocessor/version"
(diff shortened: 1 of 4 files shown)


src/rest/resource.cpp
```diff
@@ -800,6 +803,33 @@ void Resource::DatasetPending(const Request &aRequest, Response &aResponse) cons
...
+{
+    std::string errorCode;
+
+    if (aRequest.GetMethod() == HttpMethod::kGet)
+    {
+        GetCoprocessorVersion(aResponse);
+    }
+    else
+    {
+        ErrorHandler(aResponse, HttpStatusCode::kStatusMethodNotAllowed);
```
bukepo, line 829: Perhaps we can have a refactor to these resource handlers to return whether the request is successfully processed so that this error can be responded in one place.

bukepo (approved):

src/rest/resource.cpp
puddly: Do you mean in this specific PR or in general?

puddly:
Is there anything I can do to progress this PR forward?

jwhui:
> Is there anything I can do to progress this PR forward?

@puddly , can you help rebase and resolve conflicts?

puddly:
Done! Let me know if there's anything that you would like changed.

puddly:
It looks like CI failed due to a network issue:

```
              error: downloading 'https://cmake.org/files/v3.10/cmake-3.10.3.tar.gz' failed
              status_code: 28
              status_string: "Timeout was reached"
```

jwhui (approved):

PR #2604, opened 2024-11-15
C++ version update

Deepan-Shiva:
In Yocto 5.0 protobuf version as been updated and requires to link the absl which requires minimum C++14.

`/include/absl/base/policy_checks.h:79:2: error: #error "C++ versions less than C++14 are not supported."
|    79 | #error "C++ versions less than C++14 are not supported."
|       |  ^~~~~
| compilation terminated due to -Wfatal-errors.`

and absl uses strings/string_view which is supported from c++17

`include/absl/strings/string_view.h:53:26: error: 'string_view' in namespace 'std' does not name a type
|    53 | using string_view = std::string_view;
|       |                          ^~~~~~~~~~~
| compilation terminated due to -Wfatal-errors.`


CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ if(NOT CMAKE_C_STANDARD)
 endif()
 
 if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD 17)
```
superwhd, line 55: My thoughts:
- Could this be limited to yocto? I think we're targeting to make OTBR work on C++11.
- Could we specify an older version of protobuf version so it doesn't have to depend on `std::string_view`? cc @morningboata

morningboata: Currently the protobuf is install by the system's libprotobuf-dev and protobuf-compiler library. It's hard to config the custom proto generator & library, and change the CMAKER's default proto generator & library MACROs. I don't know what is the proper way yet.

superwhd: @Deepan-Shiva Could you set the `CMAKE_CXX_STANDARD` to 17 in yocto-specific build configuration?

jwhui:
Closing stale PR.

PR #2605, opened 2024-11-15
[pd] daemon service: reconnect to dbus on dbus reset

yangsong-cnyn:
This commit addresses a potential issue where the pd daemon loses its dbus connection when original dbus is reset. The daemon now monitors the io.openthread.BorderRouter.wpan0 dbus name and reconnects if the name owner changes, indicating a likely otbr-agent reload. This ensures the daemon continues to function correctly after dbus is reset by otbr-agent.


script/reference-device/dhcp6_pd_daemon.py
```diff
@@ -107,8 +107,17 @@ def connect_to_signal():
 
 
 def check_and_reconnect(dbus_obj):
+    global bus
     if dbus_obj is None:
         connect_to_signal()
+    bus.watch_name_owner(bus_name='io.openthread.BorderRouter.wpan0',
```
superwhd, line 113: For this [watch_name_owner API](https://dbus.freedesktop.org/doc/dbus-python/dbus.bus.html?highlight=watch_name_owner#dbus.bus.BusConnection.watch_name_owner), what's the behavior of calling it multiple times?

superwhd, line 113: For this [watch_name_owner API](https://dbus.freedesktop.org/doc/dbus-python/dbus.bus.html?highlight=watch_name_owner#dbus.bus.BusConnection.watch_name_owner), what's the behavior of calling it multiple times? I failed to find detailed description regarding it.

script/reference-device/dhcp6_pd_daemon.py
```diff
@@ -107,8 +107,17 @@ def connect_to_signal():
...
+                         callback=handle_name_owner_changed)
+
+
+def handle_name_owner_changed(new_owner):
+    if not new_owner:
+        logging.warning("Otbr-agent likely reloaded, reconnecting to D-Bus")
```
superwhd, line 119: By `reload`, do you mean restart? 'Reloading' may be confused with reloading the config in this context.

yangsong-cnyn: I agree the message is confusing because it makes reader think about "systemctl reload otbr-agent" -- in fact "Requires=otbr-agent.service" already guarantees pd daemon restarts on otbr-agent.

To clarify, it is triggered by otbr-agent's internal reset mechanism (e.g. during `factoryreset`), which also re-establishes the dbus connection.

I'll update the message to use 'reset' instead of 'reload' to avoid confusion.

script/reference-device/dhcp6_pd_daemon.py
yangsong-cnyn: The [dbus watch implementation](https://dbus.freedesktop.org/doc/api/html/structDBusWatch.html) shows dbus lib keeps internal lists of watch entries. When a new watch is registered for the same name and the same callback, it simply updates the existing entry in the table.

yangsong-cnyn: The [source code](https://github.com/TelepathyIM/telepathy-glib/blob/72412c944b771f3214ddc40fa9dea82cea3a5651/telepathy-glib/dbus-daemon.c#L519C7-L519C26) shows more details

superwhd (approved):
LGTM 👍

script/reference-device/dhcp6_pd_daemon.py
```diff
@@ -90,6 +90,7 @@ def properties_changed_handler(interface_name, changed_properties,
 
 
 def connect_to_signal():
+    global bus
```
superwhd, line 93: nit: Is this needed?

jwhui (approved):

PR #2612, opened 2024-11-20
[thread-host] add ThreadEnabled state callback

Irving-cl:
This PR adds `AddThreadEnabledStateChangedCallback` to `ThreadHost` and implements it for RcpHost.

Since the `ThreadEnabledState` is subscribed by some platform services, (for example, on Android) and this state is also required to implement `Join`, `Leave`, `ScheduleMigration` and `SetThreadEnabled`, this PR adds the `ThreadEnabledState` to `RcpHost` and provide an API to allow other components to subscribe to this state.

This PR makes the behavior of `ScheduleMigration` and `SetThreadEnabled` same as the Android binder API. And the unit test is updated accordingly.

### DIFF
--- a/tests/gtest/test_rcp_host_api.cpp
+++ b/tests/gtest/test_rcp_host_api.cpp
@@ -64,8 +64,9 @@ static void MainloopProcessUntil(otbr::MainloopContext    &aMainloop,
 
 TEST(RcpHostApi, DeviceRoleChangesCorrectlyAfterSetThreadEnabled)
 {
-    otError                                    error          = OT_ERROR_FAILED;
-    bool                                       resultReceived = false;
+    otError                                    error              = OT_ERROR_FAILED;
+    bool                                       resultReceived     = false;
+    otbr::Ncp::ThreadEnabledState              threadEnabledState = otbr::Ncp::ThreadEnabledState::kStateInvalid;
     otbr::MainloopContext                      mainloop;
     otbr::Ncp::ThreadHost::AsyncResultReceiver receiver = [&resultReceived, &error](otError            aError,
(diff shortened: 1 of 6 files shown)


tests/gtest/test_rcp_host_api.cpp
```diff
@@ -110,6 +111,7 @@ TEST(RcpHostApi, DeviceRoleChangesCorrectlyAfterSetThreadEnabled)
     MainloopProcessUntil(mainloop, /* aTimeoutSec */ 1, [&resultReceived]() { return resultReceived; });
```
sunytt, line 111: may verify the state is disabling before the delay?

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -64,40 +64,41 @@ static void MainloopProcessUntil(otbr::MainloopContext    &aMainloop,
 
 TEST(RcpHostApi, DeviceRoleChangesCorrectlyAfterSetThreadEnabled)
```
sunytt, line 65: may also add cases that enable after enable Thread remains enabled; disable after disable Thread remains disabled?

tests/gtest/test_rcp_host_api.cpp
Irving-cl: Added the check.

tests/gtest/test_rcp_host_api.cpp
Irving-cl: Added "enable after enable" in step 3 and "disable after disable" in step 5.

jwhui (approved):

PR #2618, opened 2024-11-25
[thread-host] update Leave method

Irving-cl:
This PR updates the `ThreadHost::Leave` method.

This PR adds a parameter `aEraseDataset` to the method. This PR also implements this method for `RcpHost`. The implementation is consistent with the android implementation. The PR also adds a unit test for `RcpHost::Leave`.

### DIFF
--- a/src/ncp/rcp_host.cpp
+++ b/src/ncp/rcp_host.cpp
@@ -333,6 +333,7 @@ void RcpHost::Deinit(void)
 
     mSetThreadEnabledReceiver  = nullptr;
     mScheduleMigrationReceiver = nullptr;
+    mDetachGracefullyCallbacks.clear();
 }
 
 void RcpHost::HandleStateChanged(otChangedFlags aFlags)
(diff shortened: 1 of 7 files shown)


src/ncp/rcp_host.cpp
```diff
@@ -595,9 +621,36 @@ void RcpHost::SetChannelMaxPowers(const std::vector<ChannelMaxPower> &aChannelMa
...
+    SafeInvokeAndClear(mScheduleMigrationReceiver, OT_ERROR_ABORT, "Aborted by leave/disable operation");
+
+    for (auto &callback : mDetachGracefullyCallbacks)
+    {
+        callback();
+    }
+    mDetachGracefullyCallbacks.clear();
+}
+
+void RcpHost::ErasePersistentInfo(bool aErase)
```
superwhd, line 648: Seems when `aErase` is false it's a no-op, which kind of conflicts with the method's name.

Irving-cl: What about naming it as `ConditionalErasePersistentInfo`? This is basically an inline function to write shorter code.

superwhd: SGTM name it as `ConditionalErasePersistentInfo`. In a longer term we may want to introduce a util for this pattern.

superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2624, opened 2024-11-27
[ncp] integrate SRP Server and Advertising Proxy for NCP

Irving-cl:
This PR integrates the mDNS Publisher with NcpHost to make advertising proxy work under NCP mode.

The PR adds an expect test to verify the advertising proxy function under NCP mode. To do the test, `avahi-browser` is used. So it's added into the bootstrap of the CI for ncp_mode.

As an abstraction: 
1. The PR adds `NcpSpinel::SrpServerSetEnabled` and `NcpSpinel::SrpServerSetAutoEnableMode` to control the SRP server on the NCP.
2. The PR adds the handling of `SPINEL_PROP_DNSSD_HOST`, `SPINEL_PROP_DNSSD_SERVICE` and `SPINEL_PROP_DNSSD_KEY_RECORD` to help NCP to publish the dnssd entries (by using the mDNSPublisher on the host).

### DIFF
--- a/src/ncp/ncp_spinel.cpp
+++ b/src/ncp/ncp_spinel.cpp
@@ -36,6 +36,7 @@
 
 #include <openthread/dataset.h>
 #include <openthread/thread.h>
+#include <openthread/platform/dnssd.h>
 
 #include "common/code_utils.hpp"
 #include "common/logging.hpp"
@@ -44,6 +45,7 @@
 #include "lib/spinel/spinel_driver.hpp"
 #include "lib/spinel/spinel_encoder.hpp"
 #include "lib/spinel/spinel_helper.hpp"
+#include "lib/spinel/spinel_prop_codec.hpp"
 
 namespace otbr {
 namespace Ncp {
@@ -58,6 +60,9 @@ NcpSpinel::NcpSpinel(void)
     , mEncoder(mNcpBuffer)
     , mIid(SPINEL_HEADER_INVALID_IID)
     , mPropsObserver(nullptr)
+#if OTBR_ENABLE_SRP_ADVERTISING_PROXY
+    , mPublisher(nullptr)
+#endif
 {
     std::fill_n(mWaitingKeyTable, SPINEL_PROP_LAST_STATUS, sizeof(mWaitingKeyTable));
     memset(mCmdTable, 0, sizeof(mCmdTable));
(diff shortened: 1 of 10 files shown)


src/ncp/ncp_spinel.cpp
```diff
@@ -439,6 +497,183 @@ void NcpSpinel::HandleValueIs(spinel_prop_key_t aKey, const uint8_t *aBuffer, ui
...
+static std::string KeyNameFor(const otPlatDnssdKey &aKey)
+{
+    std::string name(aKey.mName);
+
+    if (aKey.mServiceType != nullptr)
+    {
+        name += ".";
+        name += aKey.mServiceType;
+    }
+    return name;
```
abtink, line 510: We should be careful here, as the first label in a "service instance name" may itself contain a dot (`.`) character. So, if we construct a full name by appending the labels, it may be parsed incorrectly later.

abtink (approved):
Looks good overll. 
Couple of smaller suggestions. and one question/suggestion about key name.

src/ncp/ncp_spinel.cpp
Irving-cl: Added a check to only add the dot when the name doesn't contain one in the end.

abtink: The `.` can be in the middle and part of the instance name label itself (not a separator). The dot `.`  is generally not used (or allowed) in DNS labels, but service instance labels allow this. For example, a service instance label can be "device.1234" as a single label (`.` being part of the label itself).

If we want to handle this, I can think of two options:

1.  Keep the first label separate and pass it as a separate "string" to all APIs (mDNS-related APIs).
2.  Add a new escaping mechanism so we can construct a full name as a single string, using an escape character to indicate when `.` is part of the label and not a separator.

I prefer option 1 (that's why we use this way to represent the names from the OT core).

I know this is rather tricky, so it may be fine to leave it for the future (accepting that the current code would not work with instance labels that include a `.`).

Irving-cl: SGTM! I added a comment in the code.

zhanglongxia (approved):
LGTM

src/ncp/ncp_spinel.hpp
```diff
@@ -245,10 +249,48 @@ class NcpSpinel : public Netif::Dependencies, public InfraIf::Dependencies
...
+     *
+     */
+    void SrpServerSetAutoEnableMode(bool aEnabled);
+
+    /**
+     * This method sets the dnssd state on NCP.
+     *
+     * @param[in] aState  The dnssd state.
+     *
+     */
```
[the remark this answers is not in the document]
Irving-cl: Thanks! Updated.

jwhui (approved):

PR #2631, opened 2024-12-04
[thread-host] implement join for rcp host

Irving-cl:
This PR implements the RcpHost::Join method.

The implementation is exactly the same as Android binder API's implementation. This PR also adds a unit test to check the Join method.

### DIFF
--- a/src/ncp/rcp_host.cpp
+++ b/src/ncp/rcp_host.cpp
@@ -331,6 +331,7 @@ void RcpHost::Deinit(void)
     mThreadEnabledStateChangedCallbacks.clear();
     mResetHandlers.clear();
 
+    mJoinReceiver              = nullptr;
     mSetThreadEnabledReceiver  = nullptr;
     mScheduleMigrationReceiver = nullptr;
     mDetachGracefullyCallbacks.clear();
@@ -344,6 +345,12 @@ void RcpHost::HandleStateChanged(otChangedFlags aFlags)
     }
 
     mThreadHelper->StateChangedCallback(aFlags);
+
+    if ((aFlags & OT_CHANGED_THREAD_ROLE) && IsAttached() && mJoinReceiver != nullptr)
+    {
+        otbrLogInfo("Join succeeded");
+        SafeInvokeAndClear(mJoinReceiver, OT_ERROR_NONE, "Join succeeded");
+    }
 }
 
 void RcpHost::Update(MainloopContext &aMainloop)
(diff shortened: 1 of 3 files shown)


src/ncp/rcp_host.cpp
```diff
@@ -331,19 +331,28 @@ void RcpHost::Deinit(void)
...
+    mJoinReceiver              = nullptr;
     mSetThreadEnabledReceiver  = nullptr;
     mScheduleMigrationReceiver = nullptr;
     mDetachGracefullyCallbacks.clear();
 }
 
 void RcpHost::HandleStateChanged(otChangedFlags aFlags)
 {
+    otOperationalDatasetTlvs datasetTlvs;
+    GetDatasetActiveTlvs(datasetTlvs);
```
sunytt, line 343: I wonder if we need to do this every time there is a state change. Maybe only when there is `OT_CHANGED_ACTIVE_DATASET` change?

Irving-cl: Good idea! I move the checking of dataset into the if loop. So it will only be fetched when other conditions are all met.

src/ncp/rcp_host.cpp
```diff
@@ -438,12 +447,83 @@ const char *RcpHost::GetThreadVersion(void)
...
+                  errorMsg = "Failed to set Active Operational Dataset");
+
+    // TODO(b/273160198): check how we can implement join as a child
+    SuccessOrExit(error = otIp6SetEnabled(mInstance, true), errorMsg = "Failed to bring up Thread interface");
+    SuccessOrExit(error = otThreadSetEnabled(mInstance, true), errorMsg = "Failed to bring up Thread stack");
+
+    // Abort an ongoing join()
+    if (mJoinReceiver != nullptr)
+    {
+        SafeInvoke(mJoinReceiver, OT_ERROR_ABORT, "Join() is aborted");
```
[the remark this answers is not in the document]
Irving-cl: Added handling in `ThreadDetachGracefullyCallback`.

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -317,3 +317,79 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterScheduleMigration)
...
+    EXPECT_EQ(host.GetDeviceRole(), OT_DEVICE_ROLE_LEADER);
+
+    // 4. Call Join with the same dataset.
+    error          = OT_ERROR_NONE;
+    resultReceived = false;
+    host.Join(datasetTlvs, receiver);
+    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 0, [&resultReceived]() { return resultReceived; });
+    EXPECT_EQ(error, OT_ERROR_NONE);
+    EXPECT_STREQ(errorMsg.c_str(), "Already Joined the target network");
+
```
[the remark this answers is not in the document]
Irving-cl: Added a test case where join is folloed by disable.

sunytt (approved):
LGTM 👍

src/ncp/rcp_host.cpp
```diff
@@ -438,12 +451,83 @@ const char *RcpHost::GetThreadVersion(void)
...
+{
+    bool result = false;
+
+    otOperationalDataset     lhsDataset;
+    otOperationalDataset     rhsDataset;
+    otOperationalDatasetTlvs lhsNormalizedTlvs;
+    otOperationalDatasetTlvs rhsNormalizedTlvs;
+
+    // Sort the TLVs in the TLV byte arrays by leveraging the deterministic nature of the two OT APIs
+    SuccessOrExit(otDatasetParseTlvs(&aLhs, &lhsDataset));
```
jwhui, line 464: This will drop an TLVs that are unknown to `otDatasetParseTlvs`. Will this cause any issues?

Is it sufficient to simply compare the Network Key and possibly Extended PAN ID?

src/ncp/rcp_host.cpp
```diff
@@ -438,12 +451,83 @@ const char *RcpHost::GetThreadVersion(void)
...
+    otError                  error = OT_ERROR_NONE;
+    std::string              errorMsg;
+    bool                     receiveResultHere = true;
+    otOperationalDatasetTlvs curDatasetTlvs;
+
+    VerifyOrExit(mInstance != nullptr, error = OT_ERROR_INVALID_STATE, errorMsg = "OT is not initialized");
+    VerifyOrExit(mThreadEnabledState != ThreadEnabledState::kStateDisabling, error = OT_ERROR_BUSY,
+                 errorMsg = "Thread is disabling");
+    VerifyOrExit(mThreadEnabledState == ThreadEnabledState::kStateEnabled, error = OT_ERROR_INVALID_STATE,
+                 errorMsg = "Thread is disabled");
```
jwhui, line 487:
```suggestion
                 errorMsg = "Thread is not enabled");
```

src/ncp/rcp_host.cpp
Irving-cl: As we discussed offline, this step here is to avoid unnecessary re-join. So it is sufficient to compare network key and extended PAN ID (and we also add channel which may effect connectivity if updated).

> This will drop an TLVs that are unknown to otDatasetParseTlvs. Will this cause any issues?
I added one step to update the new dataset in this if block. So even we decide to skip the re-join, the new dataset will still be set.

src/ncp/rcp_host.cpp
```diff
@@ -344,6 +345,18 @@ void RcpHost::HandleStateChanged(otChangedFlags aFlags)
     }
 
     mThreadHelper->StateChangedCallback(aFlags);
+
+    if ((aFlags & OT_CHANGED_THREAD_ROLE) && IsAttached() && mJoinReceiver != nullptr)
+    {
+        otOperationalDatasetTlvs datasetTlvs;
+
+        GetDatasetActiveTlvs(datasetTlvs);
+        if (datasetTlvs.mLength != 0)
```
abtink, line 354: What's the reason to do this check for non-empty dataset (in addition to `IsAttahced()` check)?

src/ncp/rcp_host.cpp
```diff
@@ -438,12 +451,82 @@ const char *RcpHost::GetThreadVersion(void)
...
+    otOperationalDataset rhsDataset;
+
+    SuccessOrExit(otDatasetParseTlvs(&aLhs, &lhsDataset));
+    SuccessOrExit(otDatasetParseTlvs(&aRhs, &rhsDataset));
+
+    result =
+        (lhsDataset.mChannel == rhsDataset.mChannel) &&
+        (memcmp(lhsDataset.mNetworkKey.m8, rhsDataset.mNetworkKey.m8, sizeof(lhsDataset.mNetworkKey)) == 0) &&
+        (memcmp(lhsDataset.mExtendedPanId.m8, rhsDataset.mExtendedPanId.m8, sizeof(lhsDataset.mExtendedPanId)) == 0);
+
```
abtink, line 468: The method name is `needRejoin()`. If the channel, key, and PAN ID match, we return `true`, which indicates that we need to rejoin. Should it be the other way around? Maybe I am missing something?

src/ncp/rcp_host.cpp
```diff
@@ -438,12 +451,82 @@ const char *RcpHost::GetThreadVersion(void)
...
+    otbrLogInfo("Start joining...");
+
+    error = otDatasetGetActiveTlvs(mInstance, &curDatasetTlvs);
+    if (error == OT_ERROR_NONE && needRejoin(aActiveOpDatasetTlvs, curDatasetTlvs) && IsAttached())
+    {
+        // Do not leave and re-join if this device has already joined the same network. This can help elimilate
+        // unnecessary connectivity and topology disruption and save the time for re-joining. It's more useful for use
+        // cases where Thread networks are dynamically brought up and torn down (e.g. Thread on mobile phones).
+        SuccessOrExit(error    = otDatasetSetActiveTlvs(mInstance, &aActiveOpDatasetTlvs),
+                      errorMsg = "Failed to set Active Operational Dataset");
```
abtink, line 495: If already joined to same network, do we need to set the active dataset again? 
This can trigger registration/resync of dataset with leader? Is this desired/interntional?

abtink (commented):
Looks good. Couple of questions below. Thanks

src/ncp/rcp_host.cpp
Irving-cl: Yes, this is out of consideration for unknown TLVs in the dataset. It's very difficult to check if two datasets are completely the same with the existence of unknown TLVs (because of the order). So we simply set the active dataset again. But we want to avoid rejoin as possible. Thoughts?

src/ncp/rcp_host.cpp
Irving-cl: This code is converted from the Android implementation. But thinking on this, this is not needed.

src/ncp/rcp_host.cpp
Irving-cl: Renamed the method to `noNeedRejoin`. That's what I had meant to do.

src/ncp/rcp_host.cpp
abtink: I think it sounds okay to do so. The leader should combine them. This may add some extra traffic/exchanges when the parameters fully match, but I expect it should be harmless otherwise.

abtink (approved):
👍

jwhui (approved):

PR #2633, opened 2024-12-05
[thread host] add ThreadHost Error code

Irving-cl:
This PR adds ThreadHost error code and use it as part of the ThreadHost API.

Following https://github.com/openthread/openthread/pull/10990, currently `otError`s are used in `ThreadHost` APIs. But some specific errors in `ThreadHost` APIs cannot be denoted by `otError` and these errors do need to be differentiated by the client code. For example, `kErrorThreadDisabled` means some action is taken when Thread is 'Disabled' while 'Thread Disabled' is a concept for `ThreadHost` APIs and beyond the scope of openthread.

This PR adds ThreadHost error code which is a superset of `otError`s and contains a few proprietary errors that could be used to represent some specific cases.

This PR also replaces the existing usage of otError in the client code of `ThreadHost` APIs.


abtink (approved):
Thanks. LGTM 👍

sunytt (approved):

src/host/thread_host.hpp
```diff
@@ -106,6 +106,56 @@ enum ThreadEnabledState
...
+constexpr Error kErrorReassemblyTimeout          = OT_ERROR_REASSEMBLY_TIMEOUT;
+constexpr Error kErrorNotTmf                     = OT_ERROR_NOT_TMF;
+constexpr Error kErrorNotLowpanDataFrame         = OT_ERROR_NOT_LOWPAN_DATA_FRAME;
+constexpr Error kErrorLinkMarginLow              = OT_ERROR_LINK_MARGIN_LOW;
+constexpr Error kErrorInvalidCommand             = OT_ERROR_INVALID_COMMAND;
+constexpr Error kErrorPending                    = OT_ERROR_PENDING;
+constexpr Error kErrorRejected                   = OT_ERROR_REJECTED;
+constexpr Error kErrorGeneric                    = OT_ERROR_GENERIC;
+
+// Proprietary errors: -32768 ~ -1
```
jwhui, line 154:
```suggestion
// Errors specific to ThreadHost: -32768 ~ -1
```

src/host/thread_host.hpp
```diff
@@ -106,6 +106,56 @@ enum ThreadEnabledState
...
+constexpr Error kErrorUnsupportedChannel = -1; // The channel provided is not supported.
+constexpr Error kErrorDisabled           = -2; // The action didn't succeed because Thread is in Disabled state.
+constexpr Error kErrorFailedPrecondition = -3; // The action didn't succeed because some precondition isn't satisfied.
```
jwhui, line 157: How is this different than `kInvalidState`?

src/host/thread_host.hpp
```diff
@@ -106,6 +106,56 @@ enum ThreadEnabledState
...
+constexpr Error kErrorNotTmf                     = OT_ERROR_NOT_TMF;
+constexpr Error kErrorNotLowpanDataFrame         = OT_ERROR_NOT_LOWPAN_DATA_FRAME;
+constexpr Error kErrorLinkMarginLow              = OT_ERROR_LINK_MARGIN_LOW;
+constexpr Error kErrorInvalidCommand             = OT_ERROR_INVALID_COMMAND;
+constexpr Error kErrorPending                    = OT_ERROR_PENDING;
+constexpr Error kErrorRejected                   = OT_ERROR_REJECTED;
+constexpr Error kErrorGeneric                    = OT_ERROR_GENERIC;
+
+// Proprietary errors: -32768 ~ -1
+constexpr Error kErrorUnsupportedChannel = -1; // The channel provided is not supported.
```
jwhui, line 155: Is it possible to use `kErrorInvalidArgs`?

src/host/thread_host.hpp
```diff
@@ -106,6 +106,56 @@ enum ThreadEnabledState
...
+constexpr Error kErrorDisabled           = -2; // The action didn't succeed because Thread is in Disabled state.
```
jwhui, line 156: Is it possible to use `kInvalidState`?

src/host/thread_host.hpp
Irving-cl: `ERROR_UNSUPPORTED_CHANNEL` indicates the operation was rejected because the specified channel is currently not supported by this device in this country. For example, trying to join or migrate to a network with channel which is not supported. The caller should should change the channel or return an error to the user if the channel cannot be changed.

src/host/thread_host.hpp
Irving-cl: `ERROR_THREAD_DISABLED` indicates that the operation failed because the Thread network feature is disabled. The caller should not retry before Thread lis enabled.

bukepo: Then, perhaps we should use `kErrorNotImplemented`?

Irving-cl: This error specifically indicates that the caller should call the `ThreadHost::SetThreadEnabled` first. In Android, it indicates the caller to call [SetThreadEnabled API](https://cs.android.com/android/platform/superproject/main/+/main:out/soong/.intermediates/external/ot-br-posix/src/android/aidl/ot-daemon-aidl-ndk-source/gen/include/aidl/com/android/server/thread/openthread/IOtDaemon.h;l=106;drc=a94b8b985b7bdf228dea2588a1302a29ccc2443a;bpv=0;bpt=1). So it doesn't mean that the code, the Thread feature is not statically enabled.

src/host/thread_host.hpp
wgtdkp: We basically don't expose the underlying OT error as Android system APIs as OT error are implementation details and can change with a GitHub -> Android sync. The current Android Thread error codes are defined in https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/Connectivity/thread/framework/java/android/net/thread/ThreadNetworkException.java.

The categories are basically from [go/canonical-codes](https://goto.google.com/canonical-codes), they are designed to give users / developers clear indications of what to do to handle the errors:
1. Some are implementation errors which should never be thrown at runtime, e.g. ERROR_INVALID_ARGS. The programmer needs to fix their code to avoid this error. The API caller should also avoid ERROR_FAILED_PRECONDITION (and ERROR_THREAD_DISABLED) but it's harder to achieve because the internal Thread state can change between the caller checks the Thread state and calls the API.
2. Some indicates the the device capability missing, e.g. ERROR_UNSUPPORTED_FEATURE. Caller received this error code should not retry, and probably should give a clear message to the end users so that they wouldn't keep clicking the same button again and again.
3. Some indicates temporary failure, e.g. ERROR_ABORTED and ERROR_UNAVAILABLE. Caller received this error  code may retry to recover

Specially, an OT_ERROR_INVALID_STATE error will be translated to Android ERROR_INTERNAL error in the Android system service ([here](https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/Connectivity/thread/service/java/com/android/server/thread/ThreadNetworkControllerService.java;bpv=1;bpt=1;l=1221?gsn=otErrorToAndroidError&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fandroid.googlesource.com%2Fplatform%2Fsuperproject%2Fmain%2F%2Fmain%3Flang%3Djava%3Fpath%3Dcom.android.server.thread.ThreadNetworkControllerService%2384a0480b944d997aaada8eddadf3b9d210c8a95587f934855479a4634c786af6)). This is because we are always checking the preconditions right before all OT API calls and if there is a OT_ERROR_INVALID_STATE, it means there is a bug in our code. Note that we can safely assert on this because all OT API calls are dispatched sequentially on a single thread, there is no way that the internal Thread state can change between the state check and an OT API call. This is different with the higher level Android API which the implementation and the caller are in different processes.

src/host/thread_host.hpp
bukepo: Can we have a mapping between go/canonical-codes and Thread's error codes? It looks like the FAILED_PRECONDITION should be mapped to kInvalidState from the description of both error codes.

Irving-cl:
Will have more discussion on this.

PR #2644, opened 2024-12-18
[PD] make PD upstream interface consistent with infra_if name

yangsong-cnyn:


script/_dhcpv6_pd_ref
```diff
@@ -30,8 +30,7 @@
 #       This script manipulates DHCPv6-PD-REF configuration.
 #
 
-# TODO: set the upstream interface according to the environment variables of `script/setup`.
-UPSTREAM_INTERFACE="eth0"
+UPSTREAM_INTERFACE=${INFRA_IF_NAME}
```
Irving-cl, line 33: As I checked, `INFRA_IF_NAME` is defined as a environment variable only in [dockerfile](https://github.com/openthread/openthread/blob/main/tools/harness-simulation/posix/etc/Dockerfile#L44) for harness simulation. That means in other cases this variable will be undefined. Is it correct? Or we need to include other bash script from here?

yangsong-cnyn: You are correct, Li. Thanks for figuring it out.

Issue #2658, opened 2025-01-06
Refactor Border Agent (MeshCoP Service Publisher)


Irving-cl:
This issue tracks the process of refactoring Border Agent.

The background is that we need to support Border Agent under NCP mode as well. However the current implementation of Border Agent MeshCoP Publisher is difficult to work under NCP mode: The MeshCoP Service needs various data from OT core (to compose the txt records), which requires more data sync between host and the NCP. And it also entails more spinel properties and methods. In addition, moving the MeshCoP Service Publisher into OT core has these vantages:
* The meshcop service publishing module is identical for both NCP & RCP mode. They both rely on the platform mDNS API to publish the meshcop service.
* The code is more cohesive. When assembling the meshcop service, we can easily get any Thread data we want without calling OT API as it’s in OT core.
* When the state of the meshcop service changes, OT core can directly update the meshcop service. With the current implemention, the code in ot-br-posix needs to use OT APIs to set a callback to get notified when any data changes and then update the meshcop service. For example, the network state.
* The code is more flexible. When we want to change the meshcop service, we simply update the module and don’t need to change other things.
* Under NCP mode, the host program will have less work. Because we hope that the host wakes up as little as possible under NCP mode to save more power.

Related PRs:
- openthread
    - https://github.com/openthread/openthread/pull/11119
    - https://github.com/openthread/openthread/pull/11175
    - https://github.com/openthread/openthread/pull/11271
- ot-br-posix
    - https://github.com/openthread/ot-br-posix/pull/2664 
    - https://github.com/openthread/ot-br-posix/pull/2727
    - https://github.com/openthread/ot-br-posix/pull/2717

Irving-cl:
Refactor completed.

PR #2664, opened 2025-01-09
[platform] implement platform dnssd apis

Irving-cl:
This PR adds `DnssdPlatform` module to implement OT platform DNSSD APIs.

This PR follows up the unmerged PR: https://github.com/openthread/ot-br-posix/pull/2042

This PR is part of the effort to move the BorderAgent MeshCoP Service Publisher into OT core: https://github.com/openthread/ot-br-posix/issues/2658

In this PR, `DnssdPlatform` is added but not used yet. It will be enabled and used later when the native Border Agent MeshCoP Service publishing is ready.

In this PR, these platform DNSSD APIs are implemented while the rest of them has an empty implementation and will be implemented when necessary:
* otPlatDnssdGetState
* otPlatDnssdRegisterService
* otPlatDnssdUnregisterService
* otPlatDnssdRegisterHost
* otPlatDnssdUnregisterHost
* otPlatDnssdRegisterKey
* otPlatDnssdUnregisterKey

### DIFF
--- a/src/host/posix/dnssd.cpp
+++ b/src/host/posix/dnssd.cpp
@@ -1,0 +1,325 @@
+/*
+ *    Copyright (c) 2025, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file includes implementation of OpenThread DNS-SD platform APIs on the posix platform.
+ */
+
+#define OTBR_LOG_TAG "DNSSD"
+
+#include "host/posix/dnssd.hpp"
+
+#include <string>
+
+#include <openthread/platform/dnssd.h>
+
+#include "common/code_utils.hpp"
+#include "common/logging.hpp"
+#include "common/types.hpp"
+
+static otbr::DnssdPlatform::RegisterCallback MakeRegisterCallback(otInstance                 *aInstance,
+                                                                  otPlatDnssdRegisterCallback aCallback)
+{
+    return [aInstance, aCallback](otPlatDnssdRequestId aRequestId, otError aError) {
+        aCallback(aInstance, aRequestId, aError);
+    };
+}
+
+extern "C" otPlatDnssdState otPlatDnssdGetState(otInstance *aInstance)
+{
+    OTBR_UNUSED_VARIABLE(aInstance);
+
+    return otbr::DnssdPlatform::Get().GetState();
+}
+
+extern "C" void otPlatDnssdRegisterService(otInstance                 *aInstance,
+                                           const otPlatDnssdService   *aService,
+                                           otPlatDnssdRequestId        aRequestId,
+                                           otPlatDnssdRegisterCallback aCallback)
+{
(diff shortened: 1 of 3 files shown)


src/ncp/posix/dnssd.hpp
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Stops the `DnssdPlatform` module
+     *
+     */
+    void Stop(void);
+
+    /**
+     * Gets the singleton `DnssdPlatform` instance.
+     *
+     * @retval A reference to the `DnssdPlatform` instance.
```
superwhd, line 82: Should it be `@returns`?

superwhd (approved):
LGTM overall

src/host/posix/dnssd.cpp
```diff
@@ -0,0 +1,324 @@
...
+void DnssdPlatform::UnregisterKey(const Key &aKey, RequestId aRequestId, RegisterCallback aCallback)
+{
+    mPublisher.UnpublishKey(KeyNameFor(aKey), MakePublisherCallback(aRequestId, aCallback));
+}
+
+void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
+{
+    if (mPublisherState != aState)
+    {
+        mPublisherState = aState;
```
abtink, line 319: Do we need to track the `Mdns::Publisher::State` in `mPublisherState`? Is there an API where we can ask the publisher for its current state?

If we have such an API, using that in `UpdateState()` may be better. We would know we are using the most up-to-date state (no matter how `UpdateState()` is called) and we would not need to wait until the `HandleMdnsState()` callback is invoked to use the new publisher state.

Here, in `HandleMdnsState()` we can then simply call `UpdateState()`.

What do you think?

abtink (approved):
Looks good. 👍  Couple of questions/suggestions below:

src/host/posix/dnssd.hpp
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Sets a Dnssd State changed callback.
+     *
+     * The main usage of this method is to call `otPlatDnssdStateHandleStateChange` to notify OT core about the change
+     * when the dnssd state changes. We shouldn't directly call `otPlatDnssdStateHandleStateChange` in this module'
+     * because it only fits the RCP case.
+     *
+     * @param[in] aCallback  The callback to be invoked when the dnssd state changes.
+     */
+    typedef std::function<void(otPlatDnssdState)> DnssdStateChangeCallback;
```
abtink, line 96: Move this definition before the Doxygen comment for the method?

src/host/posix/dnssd.cpp
Irving-cl: That makes sense to me. However there isn't such API for now. And this API will be a virtual method thus need implementation for each sub-class. I'm open to add such API in Mdns if we want to do that. Maybe let's do that in a seperate PR?

src/host/posix/dnssd.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+std::string DnssdPlatform::KeyNameFor(const Key &aKey)
+{
+    std::string name(aKey.mName);
+
+    if (aKey.mServiceType != nullptr)
+    {
+        name += ".";
+        name += aKey.mServiceType;
+    }
```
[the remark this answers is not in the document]
Irving-cl: I just remembered that we have discussed this before: https://github.com/openthread/ot-br-posix/pull/2624#discussion_r1868266829

I left a TODO comment here. And after this PR, I'll update the publish operation in NcpSpinel. (using mPubhlisher) to use DnssdPlatform's Publish method. And `KeyNameFor` in NcpSpinel won't be needed anymore.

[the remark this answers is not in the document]
Irving-cl: I just remebered that we have discussed this [before](https://github.com/openthread/ot-br-posix/pull/2624#discussion_r1868266829). I also added a comment to indicate the `.` case doesn't work for now. Let's solve this problem in another PR.

I'd like to think a little bit of solving this tech debt. If using option 1 mentioned in previous [discussion](https://github.com/openthread/ot-br-posix/pull/2624#discussion_r1868266829), then we should do these things:
1. Add `mFirstLabel` and `mOtherLabels` in `otPlatDnssdService` to replace `mServiceInstance`
2. Also add 2 string fields in `otPlatDnssdKey`. If the type is host, then only one field is used.
3. In `PublishKey` method, we add escape `\` to all dots in the first label and then compose the full string and publish.

What do you think?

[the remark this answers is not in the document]
abtink: Sounds good to me. 
`otPlatDnssdKey` should handle this already (have two fields, `mName` and `mServiceType`).

src/host/posix/dnssd.cpp
abtink: Sounds good. Thanks. 
I did not know that we don't have an API to ask `Publisher` about its current state. Make sense to add it later.

jwhui (approved):

src/host/posix/dnssd.hpp
```diff
@@ -0,0 +1,139 @@
...
+ */
+class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
+{
+public:
+    /**
+     * Initializes the `DnssdPlatform` instance
+     *
+     * @param[in] aPublisher   A reference to `Mdns::Publisher` to use.
+     *
+     */
```
jwhui, line 64:
```suggestion
     * @param[in] aPublisher   A reference to `Mdns::Publisher` to use.
     */
```

src/host/posix/dnssd.hpp
```diff
@@ -0,0 +1,139 @@
...
+    DnssdPlatform(Mdns::Publisher &aPublisher);
+
+    /**
+     * Starts the `DnssdPlatform` module.
+     *
+     */
```
jwhui, line 70:
```suggestion
     * Starts the `DnssdPlatform` module.
     */
```

src/host/posix/dnssd.hpp
```diff
@@ -0,0 +1,139 @@
...
+    void Start(void);
+
+    /**
+     * Stops the `DnssdPlatform` module
+     *
+     */
```
jwhui, line 76:
```suggestion
     * Stops the `DnssdPlatform` module
     */
```

src/host/posix/dnssd.hpp
```diff
@@ -0,0 +1,139 @@
...
+    void Stop(void);
+
+    /**
+     * Gets the singleton `DnssdPlatform` instance.
+     *
+     * @returns  A reference to the `DnssdPlatform` instance.
+     *
+     */
```
jwhui, line 84:
```suggestion
     * @returns  A reference to the `DnssdPlatform` instance.
     */
```

src/host/posix/dnssd.hpp
```diff
@@ -0,0 +1,139 @@
...
+    static DnssdPlatform &Get(void) { return *sDnssdPlatform; }
+
+    typedef std::function<void(otPlatDnssdState)> DnssdStateChangeCallback;
+    /**
```
jwhui, line 88:
```suggestion
    typedef std::function<void(otPlatDnssdState)> DnssdStateChangeCallback;

    /**
```

Irving-cl:
@jwhui Updated!

PR #2677, opened 2025-01-21
[dnssd-plat] integrate DnssdPlatform into Application

superwhd:


script/bootstrap
```diff
@@ -66,6 +66,7 @@ install_packages_apt()
     # Avahi should be included for reference device builds.
     if [[ ${OTBR_MDNS} == "avahi" || ${OT_BACKBONE_CI} == 1 || ${REFERENCE_DEVICE} == 1 ]]; then
         sudo apt-get install --no-install-recommends -y avahi-daemon
+        sudo sed -i 's/^#objects-per-client-max=[0-9]\+/objects-per-client-max=30000/' /etc/avahi/avahi-daemon.conf
```
Irving-cl, line 69: Maybe add a comment here to explain this?

Irving-cl (approved):
LGTM 👍

superwhd:
The CI failure should be solved when https://github.com/openthread/ot-br-posix/pull/2702 is meregd.

abtink (approved):
LGTM. Thanks. 👍

superwhd:
Met CI failures in `test_mdns_restart.py` which I failed to reproduce locally. Debugging...

src/agent/application.cpp
```diff
@@ -140,6 +144,9 @@ otbrError Application::Run(void)
     // allow quitting elegantly
     signal(SIGTERM, HandleSignal);
 
+    // avoid exiting on SIGPIPE
+    signal(SIGPIPE, SIG_IGN);
```
superwhd, line 148: I noticed `SIGPIPE` signal was triggred and it made the program exit when calling mDNSResponder API. Not sure if ignoring the signal is fine. 

@abtink

abtink (approved):

src/agent/application.cpp
abtink: It would be better to understand why/how this is being signaled. Ignoring it may cause more harm later on, especially if it is happening due to some module not being ready or set up.

It seems to originate from us calling the mDNSResponder APIs?
My guess is that maybe we call an API somehow when the underlying mDNSResponder is not yet ready (too early)?

superwhd: It got SIGPIPE at this line: https://github.com/openthread/ot-br-posix/blob/3f28d4c990c98d22a0cb7a8f8ff2942c1b119b46/src/mdns/mdns_mdnssd.cpp#L748 by calling `DNSServiceRemoveRecord`. At that time the underlying mDNSResponder wasn't ready.

IIUC mDNSResponder supports turning the SIGPIPE off on OSX: https://github.com/apple-oss-distributions/mDNSResponder/blob/71e6611203d57c78b26fd505d98cb57a33d00880/mDNSShared/dnssd_clientstub.c#L839. So I think it should be fine for us to ignore it? Though the scope is different (global vs per socket).

abtink: I still think it would be better to investigate and understand this better.

We have `State` in `Publisher` to track when the underlying mDNS is ready or not, and there are explicit checks before all method calls, like this:

```c++
    if (mState!= State::kReady)
    {
        error = OTBR_ERROR_INVALID_STATE;
        std::move(aCallback)(error);
        ExitNow();
    }
```

So why do we get to this part of the code where it may not be ready? Do we not detect that `State` has changed and it is not ready? Or are there some other missing state checks somewhere?

superwhd: The scenario is 
1. `mdnsd` gets restarted.
2. the `Publisher` noticed this via `kDNSServiceErr_ServiceNotRunning` error code, then it tries to restart: https://github.com/openthread/ot-br-posix/blob/e2f3c1fe8f12bcd5f51cfc220261cd19a73bd81e/src/mdns/mdns_mdnssd.cpp#L391
3. In `Stop()`, it wants to remove all registrations. https://github.com/openthread/ot-br-posix/blob/e2f3c1fe8f12bcd5f51cfc220261cd19a73bd81e/src/mdns/mdns_mdnssd.cpp#L260
4. In the destructor of key registration, it unregisters itself by calling `DNSServiceRemoveRecord`. https://github.com/openthread/ot-br-posix/blob/e2f3c1fe8f12bcd5f51cfc220261cd19a73bd81e/src/mdns/mdns_mdnssd.cpp#L748

Per your idea I think we should add/move such checks into every registration type's `Register` or `Unregister` method. I can send a PR later.

However, I think this cannot fully solve the issue. The restart of `mdnsd` could happen at any time. When it restarts right before the `DNSServiceRemoveRecord` call, it will cause the SIGPIPE before the state callback of `Publisher`.

abtink: Thanks for the details. I think we should also change order in `Stop()` and set `mState` first before the list/entry deallocation calls.

superwhd: Sent https://github.com/openthread/ot-br-posix/pull/2725 to apply the suggested order of actions in `Stop()`.

abtink (approved):

jwhui (approved):

PR #2690, opened 2025-01-31
[dbus] add Join and SetThreadEnabled method to Dbus RCP

Irving-cl:
This PR adds DBus methods 'Join' and 'SetThreadEnabled' under RCP mode.

The background is to support these basic methods for both NCP & RCP mode so that we can write some test scripts that work for both NCP & RCP mode. For the long term, we also try to let NCP & RCP support the same DBus interfaces.


abtink (approved):
👍

PR #2697, opened 2025-02-06
[task-runner] only perform read when FD is set

Irving-cl:
This PR adds a check in TaskRunner::Process so that 'read' is only executed when the eventFd is set.

Currently if we use strace to see the i/o of otbr-agent, we can see many things like this:
```
read(7, 0x7ffc7c0f80ff, 1)              = -1 EAGAIN (Resource temporarily unavailable)
read(9, 0x7ffc7c0f80ff, 1)              = -1 EAGAIN (Resource temporarily unavailable)
read(4, 0x7ffc7c0f80ff, 1)              = -1 EAGAIN (Resource temporarily unavailable)
read(7, 0x7ffc7c0f80ff, 1)              = -1 EAGAIN (Resource temporarily unavailable)
read(9, 0x7ffc7c0f80ff, 1)              = -1 EAGAIN (Resource temporarily unavailable)
read(4, 0x7ffc7c0f80ff, 1)              = -1 EAGAIN (Resource temporarily unavailable)
```
Though this is harmless, we should avoid it as possible for easier debugging and reduce overhead.

### DIFF
--- a/src/common/task_runner.cpp
+++ b/src/common/task_runner.cpp
@@ -110,9 +110,9 @@ void TaskRunner::Update(MainloopContext &aMainloop)
 
 void TaskRunner::Process(const MainloopContext &aMainloop)
 {
-    OTBR_UNUSED_VARIABLE(aMainloop);
-
     ssize_t rval;
+
+    VerifyOrExit(FD_ISSET(mEventFd[kRead], &aMainloop.mReadFdSet));
 
     // Read any data in the pipe.
     do
@@ -125,6 +125,7 @@ void TaskRunner::Process(const MainloopContext &aMainloop)
     // Critical error happens, simply die.
     VerifyOrDie(errno == EAGAIN || errno == EWOULDBLOCK, strerror(errno));
 
+exit:
     PopTasks();
 }
 


src/common/task_runner.cpp
```diff
@@ -125,7 +125,9 @@ void TaskRunner::Process(const MainloopContext &aMainloop)
     // Critical error happens, simply die.
     VerifyOrDie(errno == EAGAIN || errno == EWOULDBLOCK, strerror(errno));
 
+exit:
     PopTasks();
+    return;
```
superwhd, line 130: nit: Per syntax, this 'return' is unnecessary. I'm not sure if having a 'return' here is suggested by the style? @jwhui

superwhd (approved):
LGTM 👍 

Thanks for fixing this!

src/common/task_runner.cpp
Irving-cl: Removed it. As I remember, we cannot put `exit:` at the end of a function (limited by compiler or pretty tool or code style). But now that we have a statement after `exit:`, we don't need the `return`.

jwhui (approved):

PR #2699, opened 2025-02-07
[docker] correct radio URL in docker_entrypoint.sh

Irving-cl:
OT platform driver compares the radio URL with `spinel+hdlc`. `+uart` is not needed.

### DIFF
--- a/etc/docker/docker_entrypoint.sh
+++ b/etc/docker/docker_entrypoint.sh
@@ -81,7 +81,7 @@ trap shutdown TERM INT
 
 parse_args "$@"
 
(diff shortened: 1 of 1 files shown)


etc/docker/docker_entrypoint.sh
```diff
@@ -81,7 +81,7 @@ trap shutdown TERM INT
 
 parse_args "$@"
 
-[ -n "$RADIO_URL" ] || RADIO_URL="spinel+hdlc+uart:///dev/ttyUSB0"
```
bukepo, line 84: there's another combination: `spinel+hdlc+forkpty`, right?

Irving-cl: Sorry I missed it. We need to keep it.

PR #2704, opened 2025-02-11
[meshcop] do not allow OpenThread vendor name in production build

bukepo:
This commit forbids using "OpenThread" as the vendor name in production build. This commit also discourages "BorderRouter" as product name in production build.


script/cmake-build
```diff
@@ -75,7 +75,8 @@ main()
     (
         cd "${builddir}" || die "Failed to enter ${builddir}"
 
-        cmake -GNinja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON "${OTBR_TOP_SRCDIR}" "$@"
+        cmake -GNinja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+            -DCMAKE_BUILD_TYPE=Debug "${OTBR_TOP_SRCDIR}" "$@"
```
morningboata, line 79: 1. Can we add instruction above (around line 47), on how to compile a Debug version, and how to compile a Release version (with the vendor_name and model name as parameter)?

2. if user run things like "script/cmake-build -DCMAKE_BUILD_TYPE=Release", the command line parameter will override this -DCMAKE_BUILD_TYPE=Debug here right?

bukepo: > if user run command "script/cmake-build -DCMAKE_BUILD_TYPE=Release", the command line parameter "DCMAKE_BUILD_TYPE=Release" will override "-DCMAKE_BUILD_TYPE=Debug" here right?

Yes.

morningboata (approved):

PR #2709, opened 2025-02-12
[trel] require link-local for peer discovered by mDNS

jwhui:


abtink (approved):
LGTM. Thanks. 👍

abtink (approved):
👍

bukepo (approved):

bukepo:
we may have to add link-local destination support in https://github.com/openthread/openthread/blob/e1adb71cf039d683047aaa9fcc21016ae90de4f5/src/posix/platform/trel.cpp#L222

EskoDijk:
@jwhui There could be a TREL peer using a (legacy?) mDNS stack that doesn't include any link-local address in host records if a non-LL address is available, such as Avahi.

Wouldn't this change break TREL with all such "legacy" peers?

jwhui:
> Wouldn't this change break TREL with all such "legacy" peers?

@EskoDijk , given the reported issues with existing "legacy" peers, maybe it is good to avoid connecting to such "legacy" peers.

PR #2710, opened 2025-02-12
[mdns-publisher] move link-local address filtering to DiscoveryProxy

abtink:
This commit modifies the `Publisher` to stop filtering link-local addresses during service or host resolution. Instead, the `DiscoveryProxy` now handles link-local address filtering. This change allows the `Publisher` to properly resolve TREL peers, which require link-local addresses.


superwhd (approved):
LGTM :+1:

jwhui (approved):

PR #2712, opened 2025-02-12
[mdns-publisher] use `FlagsMoreComing` in `Mdns::Publisher`

abtink:
This commit enhances `Mdns::Publisher` using mDNSResponder to utilize the `kDNSServiceFlagsMoreComing` flag. When this flag is present in a callback, it indicates that additional results are forthcoming. In such cases, we delay reporting the result to collect additional addresses.

This improves host and service resolution by preventing early "resolved" callbacks after receiving only the first IPv6 address. This allows the publisher to gather the additional addresses (from cache or all the addresses included in a query responses).

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -1265,7 +1265,8 @@ void PublisherMDnsSd::ServiceInstanceResolution::HandleGetAddrInfoResult(DNSServ
     OTBR_UNUSED_VARIABLE(aInterfaceIndex);
 
     Ip6Address address;
-    bool       isAdd = (aFlags & kDNSServiceFlagsAdd) != 0;
+    bool       isAdd      = (aFlags & kDNSServiceFlagsAdd) != 0;
+    bool       moreComing = (aFlags & kDNSServiceFlagsMoreComing) != 0;
 
     otbrLog(aErrorCode == kDNSServiceErr_NoError ? OTBR_LOG_INFO : OTBR_LOG_WARNING, OTBR_LOG_TAG,
             "DNSServiceGetAddrInfo reply: flags=%" PRIu32 ", host=%s, sa_family=%u, error=%" PRId32, aFlags, aHostName,
@@ -1292,7 +1293,7 @@ void PublisherMDnsSd::ServiceInstanceResolution::HandleGetAddrInfoResult(DNSServ
     mInstanceInfo.mTtl = aTtl;
 
 exit:
-    if (!mInstanceInfo.mAddresses.empty() || aErrorCode != kDNSServiceErr_NoError)
+    if ((!mInstanceInfo.mAddresses.empty() && !moreComing) || aErrorCode != kDNSServiceErr_NoError)
     {
         FinishResolution();
     }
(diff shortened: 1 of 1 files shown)


abtink:
It seems that, unfortunately, some existing tests actually assumed that the "resolve" outcome would include only one address, and they are force-checking this.

So, ironically, fixing/improving the `Publisher` implementation to return all addresses is now causing test failures.

As far as I can tell, `test_dnssd_server_multi_border_routers` and `test_dnssd_server` are among those tests.

The challenge here is that these tests are added in the main openthread repo, which makes it harder to update them (along with this change).

```
FAIL: test (__main__.TestDnssdServerOnMultiBr)
  File "/home/runner/work/ot-br-posix/ot-br-posix/third_party/openthread/repo/tests/scripts/thread-cert/border_router/test_dnssd_server_multi_border_routers.py", line 278, in test
    self._verify_discovery_proxy_meshcop(br2_addr, br2.get_network_name(), host)
  File "/home/runner/work/ot-br-posix/ot-br-posix/third_party/openthread/repo/tests/scripts/thread-cert/border_router/test_dnssd_server_multi_border_routers.py", line 347, in _verify_discovery_proxy_meshcop
    self._assert_dig_result_matches(dig_result, {
  File "/home/runner/work/ot-br-posix/ot-br-posix/third_party/openthread/repo/tests/scripts/thread-cert/border_router/test_dnssd_server_multi_border_routers.py", line 410, in _assert_dig_result_matches
    self.assertEqual(len(dig_result['ANSWER']), len(expected_result['ANSWER']), dig_result)
AssertionError: 2 != 1 : {'QUESTION': [('0ffa89ff8a1b.default.service.arpa.', 'IN', 'AAAA')], 'ANSWER': [('0ffa89ff8a1b.default.service.arpa.', 10, 'IN', 'AAAA', 'fdde:ad00:beef:cafe:42:c0ff:fea8:9003'), ('0ffa89ff8a1b.default.service.arpa.', 10, 'IN', 'AAAA', '9100::3')], 'ADDITIONAL': [], 'opcode': 'QUERY', 'status': 'NOERROR', 'id': '44835'}
```

abtink:
Submitted  the following PR in OT:
- https://github.com/openthread/openthread/pull/11254

The change in this PR along with above OT PR are validated together in the following customized github run https://github.com/openthread/ot-br-posix/actions/runs/13319321737.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1292,7 +1293,7 @@ void PublisherMDnsSd::ServiceInstanceResolution::HandleGetAddrInfoResult(DNSServ
     mInstanceInfo.mTtl = aTtl;
 
 exit:
-    if (!mInstanceInfo.mAddresses.empty() || aErrorCode != kDNSServiceErr_NoError)
+    if ((!mInstanceInfo.mAddresses.empty() && !moreComing) || aErrorCode != kDNSServiceErr_NoError)
```
superwhd, line 1296: I'm thinking if we should remove this `!mInstanceInfo.mAddresses.empty()` check and just check the `!moreComing` flag? There are addresses dropped by the filter at line 1279 so maybe there aren't any addresses remaining while the `moreComing` flag is false.

abtink: This PR's focus is on using `moreComing` flag (under mDNSResponder).

Removing `!mInstanceInfo.mAddresses.empty()` would be a bigger change. Basically saying if we think there are no (valid) addresses, we want to report it earlier (with an empty list), compared to current behavior where we would wait to see maybe new addresses are reported (wait for timeout).

Since this is a bigger change, I think may be good to be done in its own PR later.  I dont think it is good to add it in this PR. Always better to keep PRs smaller in scope and break changes into smaller PRs.

Also when adding this we should make sure all callbacks can properly handle when the `mAddresses` list is empty (exitsing code may have assumed that it would not be empty).

superwhd (approved):
LGTM 👍

PR #2725, opened 2025-02-18
[mdns] unify the behavior of `Stop()` between avahi and mDNSResponder

superwhd:
In the `Stop()` method, the `Publisher` will:
1. Early exit if it's already stopped. 
2. Mark the state as stopped.
3. Clear registrations and subscriptions. The destructors will free the mDNS library's client-side objects.
4. Trigger state callbacks.

### DIFF
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -519,6 +519,10 @@ bool PublisherAvahi::IsStarted(void) const
 
 void PublisherAvahi::Stop(void)
 {
+    VerifyOrExit(mState == State::kReady);
+
+    mState = State::kIdle;
+
     mServiceRegistrations.clear();
     mHostRegistrations.clear();
 
@@ -531,7 +535,10 @@ void PublisherAvahi::Stop(void)
         mClient = nullptr;
     }
 
-    mState = Mdns::Publisher::State::kIdle;
+    mStateCallback(mState);
+
+exit:
+    return;
 }
 
 void PublisherAvahi::HandleClientState(AvahiClient *aClient, AvahiClientState aState, void *aContext)
@@ -664,8 +671,6 @@ void PublisherAvahi::HandleClientState(AvahiClient *aClient, AvahiClientState aS
 
     case AVAHI_CLIENT_FAILURE:
         otbrLogErr("Avahi client failed to start: %s", avahi_strerror(avahi_client_errno(aClient)));
-        mState = State::kIdle;
-        mStateCallback(mState);
         Stop();
         Start();
         break;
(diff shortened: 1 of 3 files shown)


src/mdns/mdns_avahi.cpp
```diff
@@ -519,6 +519,11 @@ bool PublisherAvahi::IsStarted(void) const
 
 void PublisherAvahi::Stop(void)
 {
+    VerifyOrExit(mState == State::kReady);
+
+    mState = State::kIdle;
+    mStateCallback(mState);
```
bukepo, line 525: should this callback be called after everything is cleared?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -241,6 +241,9 @@ void PublisherMDnsSd::Stop(StopMode aStopMode)
 {
     VerifyOrExit(mState == State::kReady);
 
+    mState = State::kIdle;
+    mStateCallback(mState);
```
bukepo, line 245: Is it better to clear states and then update the state and trigger callback?

src/mdns/mdns_avahi.cpp
superwhd: See discussion here: https://github.com/openthread/ot-br-posix/pull/2677#discussion_r1958727572

bukepo (approved):

src/mdns/mdns_mdnssd.cpp
```diff
@@ -241,6 +241,8 @@ void PublisherMDnsSd::Stop(StopMode aStopMode)
 {
```
abtink, line 241: A general question about how the `Stop()` method is intended to be used:

1. `Stop()` can be called when the underlying mDNS provider (e.g., mDNSResponder) is itself stopped.
2. It may be called when the underlying mDNS provider is still active and running, but for some other reason, we are stopping all mDNS operations.

I noticed we have `StopMode` input with `Stop(kNormalStop)` or `kStopOnServiceNotRunningError`, which makes it seem that both cases are used and should be supported.

If we need to support both cases (which I think we need to), then I think we should handle them differently.

* In the first case, it is okay/desired to set `mState` early to ensure the `Clear()` methods (destructors and `Unregister()` calls) won't call mDNSResponder APIs anymore, as it is no longer running.
* In the second case, we do want to call mDNSResponder APIs to clear any previously registered entries (otherwise, it continues to publish the service/host info). But then setting `mState` early in the method will cause these to be skipped (which would be incorrect).

My suggestion is that we track whether the mDNSResponder is active in a separate variable other than `mState`, which we then check (along with `mState`) to decide if we can call any of the mDNSResponder (or Avahi) APIs. Thoughts?

superwhd:
Will file a new PR according to Abtin's suggestion in the future.

PR #2727, opened 2025-02-19
[border-agent] refactor to use MeshCoP Service state changed callback

Irving-cl:
This PR refactors the BorderAgent module to use the new MeshCoP service state changed callback.

Following https://github.com/openthread/openthread/pull/11215, this PR uses the new MeshCoP service state changed callback to obtain the border agent state, udp port and MeshCoP Txt values. With this PR, the BorderAgent module doesn't need to assemble Txt values from OT itself nor need to subscribe to various Thread state change.

This PR is also intended to decouple the BorderAgent module with OT instance (the RCP architecture) and make it can also work for the NCP architecture.

A new ThreadHost API `BorderAgentSetMeshCoPServiceChangedCallback` is added. In the RCP implementation, it simply uses `otBorderAgentSetMeshCoPServiceChangedCallback`. The NCP implementation will be added later. This API is invoked to connect the BorderAgent module with the MeshCoP service state change.

This PR also updates the DBus method 'Get/SetEphemeralKey' implementation to use OT API because with the new design the bit indicating IsEphemeralKeyEnabled is managed by OT core.


Irving-cl:
Pending on the this PR in openthread: https://github.com/openthread/openthread/pull/11276

Irving-cl:
Pending on https://github.com/openthread/openthread/pull/11310

Irving-cl:
Found another issue in openthread. Pending on https://github.com/openthread/openthread/pull/11320

Irving-cl:
The update of openthread/repo is to include the latest fix to pass the CI. Please ignore that. It will disappear after today's bump.

src/agent/application.cpp
```diff
@@ -255,7 +255,11 @@ void Application::InitRcpMode(void)
     mPublisher->Start();
 #endif
 #if OTBR_ENABLE_BORDER_AGENT
-    mBorderAgent->Init();
+    mHost.BorderAgentSetMeshCoPServiceChangedCallback(
+        [this](bool aIsActive, uint16_t aPort, const uint8_t *aTxtData, uint16_t aLength) {
+            mBorderAgent->HandleBorderAgentStateChange(aIsActive, aPort);
+            mBorderAgent->HandleOtMeshCoPTxtValueChange(std::vector<uint8_t>(aTxtData, aTxtData + aLength));
```
superwhd, line 261: Seems like there's a chance that it'll publish mesh cop services twice in a row? Can we merge the updates?

src/border_agent/border_agent.cpp
```diff
@@ -403,33 +377,44 @@ void BorderAgent::HandleMdnsState(Mdns::Publisher::State aState)
...
-        omrData.push_back(omrPrefix.mLength);
-        std::copy(omrPrefix.mPrefix.mFields.m8, omrPrefix.mPrefix.mFields.m8 + (omrPrefix.mLength + 7) / 8,
-                  std::back_inserter(omrData));
-        aTxtList.emplace_back("omr", omrData.data(), omrData.size());
+        Mdns::Publisher::DecodeTxtData(txtList, mOtMeshCoPTxtValues.data(), mOtMeshCoPTxtValues.size());
+        for (auto &entry : txtList)
+        {
+            if (entry.mKey == "xa")
+            {
+                memcpy(mExtAddress.m8, entry.mValue.data(), sizeof(otExtAddress));
```
superwhd, line 407:
```suggestion
                memcpy(mExtAddress.m8, entry.mValue.data(), sizeof(mExtAddress.m8));
```
Not sure if this will be more readable. Feel free to share your thoughts :).

src/border_agent/border_agent.cpp
```diff
@@ -403,33 +377,44 @@ void BorderAgent::HandleMdnsState(Mdns::Publisher::State aState)
...
     {
-        std::vector<uint8_t> omrData;
+        Mdns::Publisher::TxtList txtList;
 
-        omrData.reserve(1 + OT_IP6_PREFIX_SIZE);
-        omrData.push_back(omrPrefix.mLength);
-        std::copy(omrPrefix.mPrefix.mFields.m8, omrPrefix.mPrefix.mFields.m8 + (omrPrefix.mLength + 7) / 8,
-                  std::back_inserter(omrData));
-        aTxtList.emplace_back("omr", omrData.data(), omrData.size());
+        Mdns::Publisher::DecodeTxtData(txtList, mOtMeshCoPTxtValues.data(), mOtMeshCoPTxtValues.size());
```
superwhd, line 402: Maybe log the error.

src/host/thread_host.hpp
```diff
@@ -249,6 +250,14 @@ class ThreadHost : virtual public NetworkProperties
      */
     virtual void AddThreadEnabledStateChangedCallback(ThreadEnabledStateCallback aCallback) = 0;
 
+    /**
+     * This method sets a callback that will be invoked when there are any changes on the MeshCoP service from
+     * Thread core.
+     *
+     * @param[in] aCallback  The callback function.
+     */
+    virtual void BorderAgentSetMeshCoPServiceChangedCallback(BorderAgentMeshCoPServiceChangedCallback aCallback) = 0;
```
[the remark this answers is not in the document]
Irving-cl: > The name with BorderAgent as the prefix feels a bit like C-style. Will it be better to rename it like SetBorderAgentMeshCoPServiceChangedCallback?

Our OT APIs have this style. But it makes sense to me to use a different style in ThreadHost.

> Will there be a border agent interface in the future? I'm a bit concerned there may be too many methods in the host class.

Do you mean adding a virtual class for border agent? I think that isn't necessary for now. But later when it's required, we can add it. And we let NcpHost / RcpHost implement the interface. Thoughts?

src/host/rcp_host.cpp
```diff
@@ -782,6 +786,29 @@ void RcpHost::UpdateThreadEnabledState(ThreadEnabledState aState)
...
+    static_cast<RcpHost *>(aContext)->HandleMeshCoPServiceChanged();
+}
+
+void RcpHost::HandleMeshCoPServiceChanged(void)
+{
+    if (mBorderAgentMeshCoPServiceChangedCallback)
+    {
+        otBorderAgentMeshCoPServiceTxtData txtData;
+
+        SuccessOrDie(otBorderAgentGetMeshCoPServiceTxtData(mInstance, &txtData),
```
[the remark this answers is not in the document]
Irving-cl: Currently the only possible error is that the MeshCoP TXT data buffer isn't enough. In this case the executable has to be updated. 

Or we just don't do the following when the return is a non-zero error and emit a warning. Thoughts?

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -2078,6 +2077,7 @@ void DBusThreadObjectRcp::DeactivateEphemeralKeyModeHandler(DBusRequest &aReques
         VerifyOrExit(!retain_active_session);
         break;
     case OT_BORDER_AGENT_STATE_DISABLED:
+        error = OT_ERROR_NOT_CAPABLE;
```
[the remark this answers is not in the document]
Irving-cl: I just try to keep the behavior unchanged. I'm not sure but I guess the reason is that we have a concept of 'Epskc Feature enabled' used for feature flag. When the feature is disabled, we cannot activate or deactivate Epskc.

src/agent/application.cpp
Irving-cl: Updated.

src/border_agent/border_agent.cpp
Irving-cl: Agree. Using the same thing in a line is better.

src/border_agent/border_agent.cpp
```diff
@@ -403,33 +377,44 @@ void BorderAgent::HandleMdnsState(Mdns::Publisher::State aState)
...
-        omrData.reserve(1 + OT_IP6_PREFIX_SIZE);
-        omrData.push_back(omrPrefix.mLength);
-        std::copy(omrPrefix.mPrefix.mFields.m8, omrPrefix.mPrefix.mFields.m8 + (omrPrefix.mLength + 7) / 8,
-                  std::back_inserter(omrData));
-        aTxtList.emplace_back("omr", omrData.data(), omrData.size());
+        for (auto &entry : txtList)
+        {
+            if (entry.mKey == "xa")
+            {
+                memcpy(mExtAddress.m8, entry.mValue.data(), sizeof(mExtAddress.m8));
```
superwhd, line 406: What if the field is missing? Then it shouldn't be considered as initialized.

src/border_agent/border_agent.cpp
```diff
@@ -579,51 +506,28 @@ void BorderAgent::PublishMeshCopService(void)
...
+    // advertise the status of the border router. One can learn the thread interface status from `sb` entry so it
+    // doesn't have to send requests to the dummy port when border agent is not running.
+    port = mBaIsActive ? mMeshCoPUdpPort : kBorderAgentServiceDummyPort;
 
-    AppendVendorTxtEntries(mMeshCopTxtUpdate, txtList);
+    error = Mdns::Publisher::EncodeTxtData(txtList, txtData);
+    assert(error == OTBR_ERROR_NONE);
 
-    if (otBorderAgentIsActive(instance))
+    // If `txtList` is empty, `txtData` will have one byte 0x00 as implemented in ` Mdns::Publisher::EncodeTxtData`.
```
superwhd, line 521: This sounds hacky. Is there a more elegant way to handle this?

src/host/thread_host.hpp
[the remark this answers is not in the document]
superwhd: > Do you mean adding a virtual class for border agent? 

Yes.

I agree it's unnecessary for now. Just asked out of curiousity.

src/host/rcp_host.cpp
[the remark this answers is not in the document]
superwhd: I'd prefer just logging it.

src/border_agent/border_agent.cpp
Irving-cl: Moved the setting initialized into the if block.

src/border_agent/border_agent.cpp
Irving-cl: Updated `Mdns::Publisher::EncodeTxtData` and remove the check here.

Irving-cl: Updated to a more elegant way: only call `EncodeTxtData` when `txtList` is not empty. Then we will absolutely get a non-empty `txtData`.

src/border_agent/border_agent.cpp
```diff
@@ -403,33 +377,44 @@ void BorderAgent::HandleMdnsState(Mdns::Publisher::State aState)
...
+        Mdns::Publisher::TxtList txtList;
+        otbrError                error =
+            Mdns::Publisher::DecodeTxtData(txtList, mOtMeshCoPTxtValues.data(), mOtMeshCoPTxtValues.size());
+
+        otbrLogResult(error, "Result of decoding MeshCoP TXT data from OT");
+        SuccessOrExit(error);
+
+        for (auto &entry : txtList)
+        {
+            if (entry.mKey == "xa")
```
abtink, line 404: If I undrtansd this correctly, we are parsing the TXT data to extarct `"xa"` key so we can larn the ext address and then construct the "service instance name" from it?

How about if we directly ask the OT stack (e.g. using `otLinkGetExtendedAddress()`)? Would this work? I guess with this we wont need to parse TXT data anymore?

abtink (approved):
LGTM. 

One question below?

src/border_agent/border_agent.cpp
Irving-cl: Yes exactly. That sounds good to me. 

I'm trying to make RCP/NCP design transparent to the `BorderAgent` class and let this class work for both design. So I tried to not calling OT API here. My idea is that we make a GetExtendedAddress as an interface (asynchronized) and implement it for both RCP & NCP cases.  Then we call it in this class with a callback to receive ExtendedAddress as one of the condition to start the BorderAgent. Thoughts?

abtink: Thanks. The current code (parsing TXT data) LGTM.

sunytt (approved):
LGTM 👍

jwhui (approved):

PR #2732, opened 2025-02-24
[bootstrap] Improve Robustness of mDNSResponder Download and Extraction

yangsong-cnyn:
The previous implementation of the mDNSResponder download and build process was susceptible to several issues:

1.  **Incomplete or Corrupted Downloads:** If the `wget` command failed to download the entire mDNSResponder archive, or if the download was corrupted, the script might continue with an invalid archive.
2.   **Pre-existing Files:** If the `/tmp/mDNSResponder-1790.80.10` directory or `mDNSResponder.tar.gz` file already existed from a previous run, the `tar` command would fail to extract the archive correctly, leading to a mix of old and new files.

The PR addresses the two issues with:
    *   The `tar --overwrite ` command is to overwrite existing files when extracting.
    *   The `wget -O mDNSResponder.tar.gz` command is used to explicitly specify the output file name and overwrite the old one

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -41,6 +41,8 @@ OTBR_MDNS="${OTBR_MDNS:-mDNSResponder}"
 OT_BACKBONE_CI="${OT_BACKBONE_CI:-0}"
 REFERENCE_DEVICE="${REFERENCE_DEVICE:-0}"
 
+MDNS_RESPONDER_PATCH_PATH=$(realpath "$(dirname "$0")"/../third_party/mDNSResponder)
+
 install_packages_apt()
 {
     sudo apt-get update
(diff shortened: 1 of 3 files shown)


script/bootstrap
```diff
@@ -71,7 +71,8 @@ install_packages_apt()
     (MDNS_RESPONDER_SOURCE_NAME=mDNSResponder-1790.80.10 \
         && MDNS_RESPONDER_PATCH_PATH=$(realpath "$(dirname "$0")"/../third_party/mDNSResponder) \
         && cd /tmp \
-        && wget --no-check-certificate https://github.com/apple-oss-distributions/mDNSResponder/archive/refs/tags/$MDNS_RESPONDER_SOURCE_NAME.tar.gz \
+        && wget --no-check-certificate -O $MDNS_RESPONDER_SOURCE_NAME.tar.gz https://github.com/apple-oss-distributions/mDNSResponder/archive/refs/tags/$MDNS_RESPONDER_SOURCE_NAME.tar.gz \
+        && rm -rf $MDNS_RESPONDER_SOURCE_NAME \
         && mkdir -p $MDNS_RESPONDER_SOURCE_NAME \
         && tar xvf $MDNS_RESPONDER_SOURCE_NAME.tar.gz -C $MDNS_RESPONDER_SOURCE_NAME --strip-components=1 \
```
superwhd, line 77: What about adding `--overwrite` option for tar instead of adding a `rm -rf`?

superwhd (commented):
There's also a step to install mDNSResponder in `tests/scripts/bootstrap.sh`. Can we create a common script used by these two places?

script/bootstrap
yangsong-cnyn: Good idea. Will use it instead

script/bootstrap
```diff
@@ -32,6 +32,7 @@
 
 # shellcheck source=script/_initrc
 . "$(dirname "$0")"/_initrc
+. script/_mdns_responder
```
superwhd, line 35: Could you treat it as an executable instead of a library? I mean just reference it when you want to install mDNSResponder.

yangsong-cnyn: updated

PR #2742, opened 2025-02-27
[mdns-avahi] utilize Avahi events for complete result handling

### LINKED ISSUE #2711: `Mdns::Publisher` signals resolved with the first address
We noticed this while investigating the code and some related issues. The background for this stems from the following PRs:

* https://github.com/openthread/ot-br-posix/pull/2710
* https://github.com/openthread/ot-br-posix/pull/2709

In the above PRs, we aimed to fix the issue where the `Publisher` is filtering link-local addresses, and move the filtering to `DiscoveryProxy`. This is done because link-local addresses should now be preferred and used for TREL peers, so we cannot have them filtered by the `Publisher`.

----

The `Mdns::Publisher` implementation (with mDNSResponder) when resolving a host or service seem to invoke the callback indicating "resolved" upon getting the first IPv6 address.

* `PublisherMDnsSd::ServiceInstanceResolution::HandleGetAddrInfoResult()` is the callback which is called, passing the addresses one by one (it has a single `const struct sockaddr *aAddress` input).
* The implementation keep the addresses in a list:
```c++
    if (isAdd)
    {
        mInstanceInfo.AddAddress(address);
    }
    else
    {
        mInstanceInfo.RemoveAddress(address);
    }
    mInstanceInfo.mTtl = aTtl;
```
* But then, at the end of this method, if the list is not empty (meaning as soon as we find one address), it will do:
```c++
    if (!mInstanceInfo.mAddresses.empty() || aErrorCode != kDNSServiceErr_NoError)
    {
        FinishResolution();
    }
```
* where from `FinishResolution`, `OnServiceResolved()` is called.

I think we should aim to improve and fix this behavior.

I think the implemenation with Avahi is also behaving in the same way.

FYI. @superwhd @librasungirl @bukepo @wgtdkp

abtink:
This commit enhances `Mdns::Publisher` using Avahi to utilize the events `AVAHI_BROWSER_ALL_FOR_NOW` which notify that all records have been processed and no further results are expected in the near future.

This improvement prevents premature "resolved" callbacks after receiving only the initial IPv6 address, thereby improving host and service address resolution. The `Publisher` now waits for these events to ensure it gathers all available addresses from the cache and query responses before signaling resolved callback. 

---

Resolves #2711

### DIFF
--- a/src/mdns/mdns.cpp
+++ b/src/mdns/mdns.cpp
@@ -766,7 +766,12 @@ void Publisher::UpdateHostResolutionEmaLatency(const std::string &aHostName, otb
 
 void Publisher::AddAddress(AddressList &aAddressList, const Ip6Address &aAddress)
 {
-    aAddressList.push_back(aAddress);
+    auto it = std::find(aAddressList.begin(), aAddressList.end(), aAddress);
+
+    if (it == aAddressList.end())
+    {
+        aAddressList.push_back(aAddress);
+    }
 }
 
 void Publisher::RemoveAddress(AddressList &aAddressList, const Ip6Address &aAddress)
--- a/src/mdns/mdns_avahi.cpp
+++ b/src/mdns/mdns_avahi.cpp
@@ -1367,19 +1367,35 @@ void PublisherAvahi::ServiceResolver::HandleResolveHostResult(AvahiRecordBrowser
     OTBR_UNUSED_VARIABLE(aRecordBrowser);
     OTBR_UNUSED_VARIABLE(aInterfaceIndex);
     OTBR_UNUSED_VARIABLE(aProtocol);
-    OTBR_UNUSED_VARIABLE(aEvent);
     OTBR_UNUSED_VARIABLE(aClazz);
     OTBR_UNUSED_VARIABLE(aType);
     OTBR_UNUSED_VARIABLE(aFlags);
 
     Ip6Address address;
-    bool       resolved   = false;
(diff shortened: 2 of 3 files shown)


abtink:
Looks like some of the addresses are being reported multiple times:
```
  File "/home/runner/work/ot-br-posix/ot-br-posix/third_party/openthread/repo/tests/scripts/thread-cert/border_router/test_dnssd_server_multi_border_routers.py", line 414, in _assert_dig_result_matches
    self.assertEqual(len(dig_result['ANSWER']), len(expected_result['ANSWER']), dig_result)
AssertionError: 2 != 1 : {'QUESTION': [('host1.default.service.arpa.', 'IN', 'AAAA')], 'ANSWER': [('host1.default.service.arpa.', 10, 'IN', 'AAAA', 'fd2d:8910:ab50:1:185c:6ff:2499:20d7'), ('host1.default.service.arpa.', 10, 'IN', 'AAAA', 'fd2d:8910:ab50:1:185c:6ff:2499:20d7')], 'ADDITIONAL': [], 'opcode': 'QUERY', 'status': 'NOERROR', 'id': '57566'}
```

The dig_result contains the same address record twice:
```
[('host1.default.service.arpa.', 10, 'IN', 'AAAA', 'fd2d:8910:ab50:1:185c:6ff:2499:20d7'), 
 ('host1.default.service.arpa.', 10, 'IN', 'AAAA', 'fd2d:8910:ab50:1:185c:6ff:2499:20d7')]
```

PR #2743, opened 2025-02-28
[pd] support of multiple prefixes publishing and deprecating

yangsong-cnyn:
Add support of multiple prefixes handling in both 

- dhcpcd.enter-hook (Publishing Prefixes):

  - This script is triggered when dhcpcd gets new or updated IPv6 PD information.
  - It now iterates through multiple new_dhcp6_ia_pd*_prefix* environment variables (where * is a number), allowing for multiple prefixes.
  - For each prefix, it extracts the prefix, length, valid lifetime (vltime), and preferred lifetime (pltime). If lifetime is zero, it means the prefix is deprecated.
  - It then constructs a radvd.conf file that includes multiple prefix blocks, each with its own lifetimes.
  - It reloads or restarts radvd to apply the new configuration.

- dhcpcd.exit-hook (Deprecating Prefixes):

  - This script is triggered when dhcpcd releases or expires IPv6 PD information.
  - It now iterates through multiple old_dhcp6_ia_pd*_prefix* environment variables to find prefixes that are no longer valid.
  - For each deprecated prefix, it sets the AdvPreferredLifetime and AdvValidLifetime to 0 in radvd.conf.
  - It reloads or restarts radvd to apply the updated configuration.

### DIFF
--- a/script/reference-device/dhcpcd.enter-hook
+++ b/script/reference-device/dhcpcd.enter-hook
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 #
 #  Copyright (c) 2024, The OpenThread Authors.
 #  All rights reserved.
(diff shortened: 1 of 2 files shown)


bukepo (commented):
Any new tests added to cover the changes?

EskoDijk:
@yangsong-cnyn @bukepo  I'm trying to understand what the purpose is of this PR and of the function that it is modifying.

If a prefix gets delegated to a BR, that alone won't have any impact on the RAs being sent on the AIL. So no reason to change anything in the RAs based on DHCPv6-PD events alone.

The only requirement is that the RA must advertise all on-mesh+routable prefixes on the AIL, using RIO. But this is already a requirement even if DHCPv6-PD is not used or was not successful. So there should be standard code already doing this.

(This PR looks like it wants to modify the PIO. And per the Thread spec, the AIL on-link prefix that a BR creates using PIO is always a ULA prefix, not dependent on DHCPv6-PD in any way.)

jwhui:
@EskoDijk , this PR is probably no longer relevant, as we are switching to using a DHCPv6 Client implemented within OT core.

This PR was an attempt to leverage dhcpcd and communicate the PD prefix using RA message format. Those RAs would never be sent on the AIL.

EskoDijk:
> @EskoDijk , this PR is probably no longer relevant, as we are switching to using a DHCPv6 Client implemented within OT core.
> 
> This PR was an attempt to leverage dhcpcd and communicate the PD prefix using RA message format. Those RAs would never be sent on the AIL.

Ok good to know! It seems to encode the PD prefix(es) in the radvd config format which kind of suggested radvd would use it.

jwhui:
> Ok good to know! It seems to encode the PD prefix(es) in the radvd config format which kind of suggested radvd would use it.

Yes, the radvd config was specifying the wpan (i.e. Thread) interface.

PR #2753, opened 2025-03-06
[srp-server] remove the `OTBR_ENABLE_SRP_ADVERTISING_PROXY` guard when enabling SRP server

superwhd:
Previously the `otSrpServerSetAutoEnableMode` call in `RcpHost` is guarded by `OTBR_ENABLE_SRP_ADVERTISING_PROXY`. When we're using the OT Advertising Proxy, we still want the SRP server to automatically get enabled but the macro `OTBR_ENABLE_SRP_ADVERTISING_PROXY` isn't defined at that time. Hence we should remove this macro guard.


abtink (approved):

Irving-cl (approved):
LGTM 👍

jwhui (approved):

PR #2756, opened 2025-03-11
[border agent] refactor dbus meshcop txt update

Irving-cl:
This PR refactors the MeshCoP TXT update method in BorderAgent.

This is a DBUS method for vendors to update their vendor MeshCoP TXT values. The current call stack is:
```
DBus Handler -> ThreadHelper MeshCoP update callback -> BorderAgent Handle MeshCoP update
```
ThreadHelper is totally unnecessary here. So the new stack is simpler:
```
DBus Handler -> BorderAgent Handle MeshCoP update
```

Another intention is to unbind the BorderAgent module from `ThreadHelper` so that it can works in NCP design.

### DIFF
--- a/src/border_agent/border_agent.hpp
+++ b/src/border_agent/border_agent.hpp
@@ -169,17 +169,26 @@ public:
      */
     void AddEphemeralKeyChangedCallback(EphemeralKeyChangedCallback aCallback);
 
+#if OTBR_ENABLE_DBUS_SERVER
+    /**
+     * This method handles update of vendor MeshCoP TXT entries.
+     *
+     * @param[in] aUpdate  A map of vendor MeshCoP TXT entries.
+     */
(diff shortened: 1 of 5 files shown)


src/border_agent/border_agent.hpp
```diff
@@ -163,17 +163,26 @@ class BorderAgent : public Mdns::StateObserver, private NonCopyable
      */
     void AddEphemeralKeyChangedCallback(EphemeralKeyChangedCallback aCallback);
 
+#if OTBR_ENABLE_DBUS_SERVER
+    /**
+     * This method handles update of vendor MeshCoP Txt entries.
+     *
+     * @param[in] aUpdate  A map of vendor MeshCoP Txt entries.
```
superwhd, line 170:
```suggestion
     * This method handles update of vendor MeshCoP TXT entries.
     *
     * @param[in] aUpdate  A map of vendor MeshCoP TXT entries.
```

jwhui (approved):

PR #2758, opened 2025-03-11
[rest] add mleid ip address API

OmegaRelay:
We'd like to add a small PR to add functionality to retrieve the current mleid IP address of the border router over rest API. This PR adds a new endpoint `/node/ip-address/mleid` which only accepts GET requests and responds with the mleid ip address string.


wgtdkp:
Hi @OmegaRelay What's the specific use case for ML-EID? Why not returning the OMR address which is externally routable or all the addresses?

OmegaRelay:
Hi @wgtdkp, I only added support for MLEID as that's what we need in our application for network service discovery. I would be happy to expand it to support all ip addresses if that's preferred?

PR #2761, opened 2025-03-13
[test] fix flaky unit test of test_netif

Irving-cl:
This PR fixes a flaky unit test in test netif (`WpanIfRecvIp6PacketCorrectly_AfterReceivingFromNetif`).

The cause is that if we call `recvfrom` right after `Netif::Ip6Receive`, it is possible that the system hasn't passed the message and the program hangs on `recvfrom`. Then the program will hang forever.

I've seen the failure in CI for many times. I've verified the test locally for 10 times with this change.

### DIFF
--- a/tests/gtest/test_netif.cpp
+++ b/tests/gtest/test_netif.cpp
@@ -46,6 +46,7 @@
 #include <sys/select.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <thread>
 #include <vector>
 
 #ifdef __linux__
@@ -464,6 +465,14 @@ TEST(Netif, WpanIfStateChangesCorrectly_AfterSettingNetifState)
     netif.Deinit();
 }
 
+void receiveTask(int aSockFd, uint8_t *aRecvBuf, struct sockaddr_in6 *aListenAddr)
+{
+    socklen_t   len = sizeof(*aListenAddr);
+    int         n = recvfrom(aSockFd, (char *)aRecvBuf, kMaxIp6Size, MSG_WAITALL, (struct sockaddr *)aListenAddr, &len);
+    std::string udpPayload(reinterpret_cast<const char *>(aRecvBuf), n);
+    EXPECT_EQ(udpPayload, "Hello Otbr Netif!");
+}
+
 TEST(Netif, WpanIfRecvIp6PacketCorrectly_AfterReceivingFromNetif)
 {
     otbr::Netif netif(sDefaultNetifDependencies);
@@ -511,13 +520,12 @@ TEST(Netif, WpanIfRecvIp6PacketCorrectly_AfterReceivingFromNetif)
                                  0xc3, 0x0c, 0x87, 0xd3, 0x00, 0x01, 0xed, 0x1c, 0x0c, 0x91, 0xcc, 0xb6, 0x57,
                                  0x8b, 0xe7, 0x08, 0x30, 0x39, 0x00, 0x19, 0x36, 0x81, 0x48, 0x65, 0x6c, 0x6c,
                                  0x6f, 0x20, 0x4f, 0x74, 0x62, 0x72, 0x20, 0x4e, 0x65, 0x74, 0x69, 0x66, 0x21};
+
+    std::thread recvThread(receiveTask, sockFd, recvBuf, &listenAddr);
+
     netif.Ip6Receive(udpPacket, sizeof(udpPacket));
 
-    socklen_t   len = sizeof(listenAddr);
-    int         n   = recvfrom(sockFd, (char *)recvBuf, kMaxIp6Size, MSG_WAITALL, (struct sockaddr *)&listenAddr, &len);
-    std::string udpPayload(reinterpret_cast<const char *>(recvBuf), n);
-    EXPECT_EQ(udpPayload, "Hello Otbr Netif!");
-
+    recvThread.join();
     close(sockFd);
     netif.Deinit();
 }


yangsong-cnyn (approved):
LGTM!

tests/gtest/test_netif.cpp
```diff
@@ -511,12 +511,38 @@ TEST(Netif, WpanIfRecvIp6PacketCorrectly_AfterReceivingFromNetif)
...
+    bool received = false;
+    while (true)
+    {
+        fd_set         readFdSet;
+        struct timeval timeout = {0, 100};
+        FD_ZERO(&readFdSet);
+        FD_SET(sockFd, &readFdSet);
+
+        int rval = select(sockFd + 1, &readFdSet, nullptr, nullptr, &timeout);
+        if (rval < 0)
```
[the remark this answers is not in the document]
Irving-cl: Generally, we use test marcos to check whether the check points of the test. If this system call fails, we think something went wrong (not the check points in the test).

tests/gtest/test_netif.cpp
```diff
@@ -511,12 +511,38 @@ TEST(Netif, WpanIfRecvIp6PacketCorrectly_AfterReceivingFromNetif)
...
+            int n = recvfrom(sockFd, (char *)recvBuf, kMaxIp6Size, MSG_WAITALL, (struct sockaddr *)&listenAddr, &len);
+            std::string udpPayload(reinterpret_cast<const char *>(recvBuf), n);
+            EXPECT_EQ(udpPayload, "Hello Otbr Netif!");
+
+            break;
+        }
+
+        if (!received)
+        {
+            netif.Ip6Receive(udpPacket, sizeof(udpPacket));
```
bukepo, line 542: if this is needed, it indicates the packet is lost, right?

I don't think it's the test's fault. The test expects receiveIp6 ensures the packet to be correctly injected into the netif when it is called, while the implementation doesn't gurantee that.

So what is the desired behavior of receiveIp6?

Irving-cl: I think the original model has a race condition and it is a problem of the test:
1. `netif::Ip6Receive` is called first but at this moment there is no socket listening for the message
2. before `recvfrom` is called, the kernel has dropped the message because at this moment no one is listening to it
3. Then we call `recvfrom` and we will never receive that message

But I just realized that my previous implementation doesn't actually solved this issue. I call select first and then call `netif::Ip6Receive`. But when `netif::Ip6Receive` is called, `select` is not called. So the fd still not being listened.

So now I've changed to use a thread model:
* Start a thread to call `recvfrom` first.
* Then call `netif::Ip6Receive`. Thus we ensure the message is being listened at this moment.

Thoughts?

bukepo: we may bind the udp socket first, then inject the packet, then recvfrom, The packet should not be dropped, right? i guess select is not necessary here.

Irving-cl: The original code already binds the udp socket first. But the race condition still happens.
```
    if (bind(sockFd, (const struct sockaddr *)&listenAddr, sizeof(listenAddr)) < 0)
    {
        perror("bind failed");
        exit(EXIT_FAILURE);
    }
    // Udp Packet
    // Ip6 source: fd2a:c30c:87d3:1:ed1c:c91:ccb6:578a
    // Ip6 destination: fd2a:c30c:87d3:1:ed1c:c91:ccb6:578b
    // Udp destination port: 12345
    // Udp payload: "Hello Otbr Netif!"
    const uint8_t udpPacket[] = {0x60, 0x0e, 0xea, 0x69, 0x00, 0x19, 0x11, 0x40, 0xfd, 0x2a, 0xc3, 0x0c, 0x87,
                                 0xd3, 0x00, 0x01, 0xed, 0x1c, 0x0c, 0x91, 0xcc, 0xb6, 0x57, 0x8a, 0xfd, 0x2a,
                                 0xc3, 0x0c, 0x87, 0xd3, 0x00, 0x01, 0xed, 0x1c, 0x0c, 0x91, 0xcc, 0xb6, 0x57,
                                 0x8b, 0xe7, 0x08, 0x30, 0x39, 0x00, 0x19, 0x36, 0x81, 0x48, 0x65, 0x6c, 0x6c,
                                 0x6f, 0x20, 0x4f, 0x74, 0x62, 0x72, 0x20, 0x4e, 0x65, 0x74, 0x69, 0x66, 0x21};
    netif.Ip6Receive(udpPacket, sizeof(udpPacket));

    socklen_t   len = sizeof(listenAddr);
    int         n   = recvfrom(sockFd, (char *)recvBuf, kMaxIp6Size, MSG_WAITALL, (struct sockaddr *)&listenAddr, &len);
    std::string udpPayload(reinterpret_cast<const char *>(recvBuf), n);
    EXPECT_EQ(udpPayload, "Hello Otbr Netif!");
```

 I've removed `select` here because `select` is also incorrect. Now I use a thread to call `recvfrom` first to handle the race condition.

bukepo: Is the `write()` in `netif.Ip6Receive()` successful when the test fails?

tests/gtest/test_netif.cpp
```diff
@@ -511,13 +520,12 @@ TEST(Netif, WpanIfRecvIp6PacketCorrectly_AfterReceivingFromNetif)
...
                                  0x6f, 0x20, 0x4f, 0x74, 0x62, 0x72, 0x20, 0x4e, 0x65, 0x74, 0x69, 0x66, 0x21};
-    netif.Ip6Receive(udpPacket, sizeof(udpPacket));
 
-    socklen_t   len = sizeof(listenAddr);
-    int         n   = recvfrom(sockFd, (char *)recvBuf, kMaxIp6Size, MSG_WAITALL, (struct sockaddr *)&listenAddr, &len);
-    std::string udpPayload(reinterpret_cast<const char *>(recvBuf), n);
-    EXPECT_EQ(udpPayload, "Hello Otbr Netif!");
+    std::thread recvThread(receiveTask, sockFd, recvBuf, &listenAddr);
+
+    netif.Ip6Receive(udpPacket, sizeof(udpPacket));
```
bukepo, line 526: I'm wondering whether this method should return `void` if it's not guaranteed to be successful.

tests/gtest/test_netif.cpp
```diff
@@ -464,6 +465,14 @@ TEST(Netif, WpanIfStateChangesCorrectly_AfterSettingNetifState)
     netif.Deinit();
 }
 
+void receiveTask(int aSockFd, uint8_t *aRecvBuf, struct sockaddr_in6 *aListenAddr)
```
bukepo, line 468: I think sleep after bind is simpler than opening a thread.

bukepo (approved):

tests/gtest/test_netif.cpp
Irving-cl: As we experimented, when the test case failed, the write was successful. Currently we don't have any way to handle the failed case. So for now keep it as is.

tests/gtest/test_netif.cpp
Irving-cl: As tested, in failed runs, `write` was successful.

tests/gtest/test_netif.cpp
Irving-cl: As discussed, we used the thread solution for now. We will further investigate the underlying mechanisum later.

PR #2763, opened 2025-03-13
[srp-server] refine build flags for SRP server and Advertising Proxy

superwhd:
This PR introduces a macro `OTBR_ENABLE_SRP_SERVER` to indicate whether the SRP server feature is enabled in OT. In this way we can use this macro instead of `OTBR_ENABLE_SRP_ADVERTISING_PROXY` to guard the code block that depends on the `otSrpServer*` API.

This PR also introduces a CMake option `OTBR_OT_SRP_ADV_PROXY` to configure whether to use the OT core Advertising Proxy.


etc/cmake/options.cmake
```diff
@@ -88,7 +88,24 @@ if (OTBR_SRP_ADVERTISING_PROXY)
...
 
+option(OTBR_USE_OT_SRP_ADV_PROXY "Enable OT core Advertising Proxy" OFF)
+
+if (OTBR_USE_OT_SRP_ADV_PROXY AND OTBR_SRP_ADVERTISING_PROXY)
+    message(FATAL_ERROR "Only one Advertising Proxy can be enabled.")
+endif()
+
+option(OTBR_SRP_SERVER "Enable SRP server" OFF)
+if (OTBR_USE_OT_SRP_ADV_PROXY OR OTBR_SRP_ADVERTISING_PROXY)
+    set(OTBR_SRP_SERVER ON CACHE STRING "Enable SRP server" FORCE)
```
Irving-cl, line 99: I think 'OTBR_SRP_SERVER' here is a 'derived value' so we probably should not provide a way to configure it. It should be only enabled when ADVERTISTING_PROXY (either OT impl or ot-br-posix impl) is enabled. So what about we calculate this derived value in a header in ot-br-posix and use the value in sources by including that header.

As I checked, it seems there is no appropriate header for now. I suggest we add a new header called `derived-config.h`. I believe this can also be used in other scenarios.

Irving-cl (approved):
LGTM 👍

etc/cmake/options.cmake
```diff
@@ -88,6 +88,15 @@ if (OTBR_SRP_ADVERTISING_PROXY)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_SRP_ADVERTISING_PROXY=1)
 endif()
 
+option(OTBR_USE_OT_SRP_ADV_PROXY "Enable OT core Advertising Proxy" OFF)
```
jwhui, line 91: nit: we don't use the "_USE_" prefix in any other config macro. Should we rename to `OTBR_OT_SRP_ADV_PROXY`?

jwhui (approved):

Issue #2731, opened 2025-02-24
./script/setup fails with " error: ‘FastWriter’ is deprecated: Use StreamWriterBuilder instead"


olalonde:
**Describe the bug** A clear and concise description of what the bug is.

With `WEB_GUI=1` on rasperry pi 4.

Fails to build with `wpan_service.cpp:58:33: error: ‘FastWriter’ is deprecated: Use StreamWriterBuilder instead [-Werror=deprecated-declarations]`.

**To Reproduce** Information to reproduce the behavior, including:

1. Git commit id: `23149c321741f7800bd551713924b50380e828a0 `
2. IEEE 802.15.4 hardware platform `nrf52840dongle`
3. Build steps:

```shell
rpi@rpihub:~/code/ot-br-posix $ export WEB_GUI=1
rpi@rpihub:~/code/ot-br-posix $ export INFRA_IF_NAME=eth0
rpi@rpihub:~/code/ot-br-posix $ ./script/bootstrap
...
rpi@rpihub:~/code/ot-br-posix $ ./script/setup
```

4. Network topology

**Expected behavior** A clear and concise description of what you expected to happen.

**Console/log output** If applicable, add console/log output to help explain your problem.

```shell
[534/538] Building CXX object src/web/CMakeFiles/otbr-web.dir/web-service/wpan_service.cpp.o
FAILED: src/web/CMakeFiles/otbr-web.dir/web-service/wpan_service.cpp.o
[...]

PR #2766, opened 2025-03-17
[web] fix the build error caused by deprecated readers and writers

superwhd:
Fixes https://github.com/openthread/ot-br-posix/issues/2731.


superwhd:
I was able to reproduce #2731 when I manually added `__attribute__((deprecated))` to those classes. The PR should be able to fix the issue.

jwhui (approved):

PR #2768, opened 2025-03-19
[posix] refactor netif module

Irving-cl:
This PR refactors the `otbr::Netif` module to move it from `NcpHost` to the `Application`.

The intention behind this PR:
1. We will soon introduce a new module `MulticastRoutingManager` which will need to get the Thread wpan interface index. While we can set a copy to `MulticastRoutingManager`, it's better to have a SSOT. And the `Netif` module will also be able to accessed from other modules in `Application` later.
2. Currently the `Netif` module is only used for NCP. But we will use it for the RCP design as well later and replace the netif module in openthread posix.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -75,6 +75,14 @@ Application::Application(Host::ThreadHost  &aHost,
     {
         CreateRcpMode(aRestListenAddress, aRestListenPort);
     }
+    else if (mHost.GetCoprocessorType() == OT_COPROCESSOR_NCP)
+    {
+        CreateNcpMode();
+    }
+    else
+    {
+        DieNow("Unknown Co-processor type!");
+    }
 }
 
 void Application::Init(void)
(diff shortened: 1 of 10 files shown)


src/agent/application.cpp
```diff
@@ -315,10 +323,20 @@ void Application::DeinitRcpMode(void)
 #endif
 }
 
+void Application::CreateNcpMode(void)
+{
+    otbr::Host::NcpHost &ncpHost = static_cast<otbr::Host::NcpHost &>(mHost);
```
zhanglongxia, line 328: nit: Could we add a common method `GetNcpHost()` to get the ` otbr::Host::NcpHost`.

src/host/ncp_host.hpp
```diff
@@ -87,6 +88,7 @@ class NcpHost : public MainloopProcessor,
      * @param[in]   aInterfaceName          A string of the NCP interface name.
      * @param[in]   aBackboneInterfaceName  A string of the backbone interface name.
      * @param[in]   aDryRun                 TRUE to indicate dry-run mode. FALSE otherwise.
+     * @param[in]   aNetif                  A reference to the netif module.
```
zhanglongxia, line 91: nit: remove extra argument `aNetif `

zhanglongxia (approved):
Looks great! Couple of smaller suggestions

src/agent/application.cpp
Irving-cl: I prefer not adding this because the `mHost` can be RcpHost or NcpHost. Calling the method in wrong cases will cause a fatal error. So let's try to avoid the cast as possible. Finally we want to make the RcpHost/NcpHost transparent to the code here.

yangsong-cnyn (approved):
LGTM!

src/host/ncp_host.cpp
```diff
@@ -148,7 +139,6 @@ void NcpHost::Init(void)
 void NcpHost::Deinit(void)
 {
     mNcpSpinel.Deinit();
-    mNetif.Deinit();
```
jwhui, line 151: Do we need to call `Deinit()` somewhere else?

Irving-cl: Updated! Added it in `Application::DeinitNcpMode`.

jwhui (approved):

PR #2773, opened 2025-03-24
[ncp] Spinel support for CLI daemon command send and receive

yangsong-cnyn:
This change adds support for sending and receiving CLI commands via the Spinel interface.

Key changes:

-   Adds `InputCommandLine` method to `NcpSpinel` to send CLI commands to the NCP.
-   Adds `CliDaemonOutputCallback` to `NcpSpinel` to register a callback for CLI output.
-   Integrates `CliDaemon` into `NcpHost` for managing CLI interactions.

This allows for CLI commands to be sent to and received from the NCP.


yangsong-cnyn:
Changes to openthread repo is just for CI tests. Will remove it after https://github.com/openthread/openthread/pull/11344 is approved and merged.

src/host/ncp_spinel.hpp
```diff
@@ -374,6 +393,7 @@ class NcpSpinel : public InfraIf::Dependencies
     otError ParseIp6AddressTable(const uint8_t *aBuf, uint16_t aLength, std::vector<Ip6AddressInfo> &aAddressTable);
     otError ParseIp6MulticastAddresses(const uint8_t *aBuf, uint16_t aLen, std::vector<Ip6Address> &aAddressList);
     otError ParseIp6StreamNet(const uint8_t *aBuf, uint16_t aLen, const uint8_t *&aData, uint16_t &aDataLen);
+    otError ParseStreamCliOutput(const uint8_t *aBuf, uint16_t aLen, const char *&aOutput);
```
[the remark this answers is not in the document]
yangsong-cnyn: @Irving-cl , do you think we should change their return value to `otbrError`?

src/host/posix/cli_daemon.cpp
```diff
@@ -71,6 +71,39 @@ std::string CliDaemon::GetSocketFilename(const std::string &aNetIfName, const ch
...
+               sizeof(kTruncatedMsg));
+    }
+
+    VerifyOrExit(mSessionSocket != -1);
+
+#ifdef __linux__
+    // Don't die on SIGPIPE
+    ret = send(mSessionSocket, aOutput, length, MSG_NOSIGNAL);
+#else
+    ret = static_cast<int>(write(mSessionSocket, aOutput, length));
```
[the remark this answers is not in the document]
yangsong-cnyn: test added

src/host/ncp_spinel.hpp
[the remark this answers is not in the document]
Irving-cl: @superwhd 

Because these implementation is mostly about using `ot::Encoder` and `ot::Decoder` which returns `otError`. I suggest we first keep as it. Let's do a clean-up in a dedicated PR.

src/host/ncp_spinel.hpp
```diff
@@ -208,6 +210,16 @@ class NcpSpinel : public InfraIf::Dependencies
      */
     otbrError Ip6MulAddrUpdateSubscription(const otIp6Address &aAddress, bool aIsAdded);
 
+    /**
+     * This method sends a command line input to the NCP.
```
Irving-cl, line 214:
```suggestion
     * This method sends a cli command line to the NCP.
```

src/host/posix/cli_daemon.hpp
```diff
@@ -49,14 +49,15 @@ class CliDaemon
     public:
         virtual ~Dependencies(void) = default;
 
-        virtual otbrError InputCommandLine(const uint8_t *aBuf, uint16_t aLength);
+        virtual otbrError InputCommandLine(const char *aLine);
```
Irving-cl, line 52: I think keeping the length is better. After we read from the socket and get the length, we directly call this with the length, instead of setting the last byte to `\0` which seems complex. And the construction of `std::string` in `Dependencies::InputCommandLIne` is more efficient. (Though not a big factor)

Thoughts?

tests/gtest/test_cli_daemon.cpp
```diff
@@ -160,4 +162,54 @@ TEST(CliDaemon, InputCommandLineCorrectly_AfterReveivingOnSessionSocket)
...
+        FD_ZERO(&context.mWriteFdSet);
+        FD_ZERO(&context.mErrorFdSet);
+
+        cliDaemon.UpdateFdSet(context);
+        int rval = select(context.mMaxFd + 1, &context.mReadFdSet, &context.mWriteFdSet, &context.mErrorFdSet,
+                          &context.mTimeout);
+        ASSERT_GE(rval, 0) << "select failed, error: " << strerror(errno);
+
+        cliDaemon.Process(context);
+        cliDaemon.HandleCommandOutput(output);
```
Irving-cl, line 201: The test case is weird. It's not a complete flow. The current flow:
* client socket connect
* Then updateFdSet, select and Process. (However we did nothing in this step).
* Then we manually give an output (`"sample output"`) to `cliDaemon` by `HandleCommandOutput`.
* And last we check the data received on the socket.

So in this process we only tested  posix send (in `HandleCommandOutput`) and receive.

We should test like this instead:
* client socket connect
* Input a command through the socket.
* Let CliDaemon handle the command.
    * Implement a different `CliDaemon::Dependencies` which call `HandleCommandOutput` to return some output.
    * During this step, we run a while loop and do `Update`, `select` and `Process` until the client socket got an output.
* Compare the output with the expected output.

src/host/posix/cli_daemon.hpp
yangsong-cnyn: While passing the length might seem more efficient initially, the `otCliInputLine` function expects a null-terminated string.

Therefore, we need to ensure the string is null-terminated at some point before it's processed by `otCliInputLine`. Doing this directly after reading from the socket seems like a straightforward place.

Additionally, having a null-terminated string allows us to conveniently use Spinel::Encoder::WriteUtf8 when sending the command via Spinel.

tests/gtest/test_cli_daemon.cpp
```diff
@@ -160,4 +163,89 @@ TEST(CliDaemon, InputCommandLineCorrectly_AfterReveivingOnSessionSocket)
...
+        context.mMaxFd   = -1;
+        context.mTimeout = {100, 0};
+        FD_ZERO(&context.mReadFdSet);
+        FD_ZERO(&context.mWriteFdSet);
+        FD_ZERO(&context.mErrorFdSet);
+
+        cliDaemon.UpdateFdSet(context);
+
+        FD_SET(clientSocket, &context.mReadFdSet);
+        context.mMaxFd = std::max(context.mMaxFd, clientSocket);
```
Irving-cl, line 228: We can use `context.AddToReadSet(clientSocket)`. 

https://github.com/openthread/ot-br-posix/blob/main/src/common/mainloop.hpp#L58

Irving-cl (approved):
LGTM 👍  Thanks.

third_party/openthread/repo
superwhd, line 1: Is this submodule bump intentional? In general module bump should be done by dependabot.

src/host/ncp_spinel.hpp
```diff
@@ -208,6 +210,16 @@ class NcpSpinel
      */
     otbrError Ip6MulAddrUpdateSubscription(const otIp6Address &aAddress, bool aIsAdded);
 
+    /**
+     * This method sends a cli command line to the NCP.
```
superwhd, line 214:
```suggestion
     * This method sends a CLI command line to the NCP.
```

tests/gtest/test_cli_daemon.cpp
```diff
@@ -160,4 +163,88 @@ TEST(CliDaemon, InputCommandLineCorrectly_AfterReveivingOnSessionSocket)
...
+
+        if (mCliDaemonInstance != nullptr)
+        {
+            mCliDaemonInstance->HandleCommandOutput(kTestOutput);
+        }
+        return OTBR_ERROR_NONE;
+    }
+};
+
+TEST(CliDaemon, HandleCommandOutputCorrectly_AfterReveivingOnSessionSocket)
```
superwhd, line 184: Ideally there should also be a test case where the output is truncated.

yangsong-cnyn: added

superwhd (approved):
LGTM :+1:

src/host/posix/cli_daemon.cpp
```diff
@@ -71,6 +71,39 @@ std::string CliDaemon::GetSocketFilename(const std::string &aNetIfName, const ch
...
+    size_t length = strlen(aOutput);
+
+    VerifyOrExit(mSessionSocket != -1);
+
+    static_assert(sizeof(kTruncatedMsg) < kCliMaxLineLength, "OTBR_CONFIG_CLI_MAX_LINE_LENGTH is too short!");
+
+    if (length >= kCliMaxLineLength)
+    {
+        length = kCliMaxLineLength - 1;
+        memcpy(const_cast<char *>(aOutput) + kCliMaxLineLength - sizeof(kTruncatedMsg), kTruncatedMsg,
```
abtink, line 86: I think we are assuming that since we get `aOutput` from the NCP spinel decoder (which points to a buffer we own in another module), we can cast and remove `const` from `aOutput` and overwrite its content here? This should work fine for now, but this could cause issues if things change and `aOutput` should be treated as a true `const char *` (i.e., its content should not be changed).

Perhaps we should copy the content into a new string to make the code safer. I believe we can even use `std::string` here within `ot-br-posix`.

abtink (approved):
Looks good.

One suggestion below.

hemanth-silabs:
@jwhui Can you have a look at this PR? Seems like it has all the approvals. This would be helpful for any NCP support.

jwhui (approved):

PR #2775, opened 2025-03-25
[dbus] refactor dbus module to unbind it with border agent feature

Irving-cl:
This PR refactors the DBusAgent module to unbind it with the BORDER_AGENT_ENABLE flag as well as simply the code.

Currently the `mDBusAgent` is guarded with `#if OTBR_ENABLE_DBUS_SERVER && OTBR_ENABLE_BORDER_AGENT`. `OTBR_ENABLE_BORDER_AGENT` shouldn't be a precondition to enable the DBus server. 

We have the current code because the DBusAgent relies on the BorderAgent when `OTBR_ENABLE_BORDER_AGENT` is enabled. However the code to inject the `BorderAgent` to DBusAgent is cumbersome: We have to write two versions of constructor or init for many classes. So as of now we only have one constructor which always accepts a `BorderAgent`.

This PR introduces a helper class `DBus::DependentComponents`. The constructor and init method will always accept this class. And we control the fields in this class by using flags like `OTBR_ENABLE_BORDER_AGENT`. Thus we don't need to write two versions of constructors and can enable DBusAgent without having to enable BorderAgent.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -70,8 +70,8 @@ Application::Application(Host::ThreadHost  &aHost,
 #if OTBR_ENABLE_BORDER_AGENT
     , mBorderAgent(*mPublisher)
 #endif
-#if OTBR_ENABLE_DBUS_SERVER && OTBR_ENABLE_BORDER_AGENT
-    , mDBusAgent(MakeUnique<DBus::DBusAgent>(mHost, *mPublisher))
+#if OTBR_ENABLE_DBUS_SERVER
+    , mDBusAgent(MakeDBusDependentComponents())
 #endif
 {
     if (mHost.GetCoprocessorType() == OT_COPROCESSOR_RCP)
@@ -104,6 +104,10 @@ void Application::Init(void)
         DieNow("Unknown coprocessor type!");
         break;
     }
+
+#if OTBR_ENABLE_DBUS_SERVER
+    mDBusAgent.Init();
+#endif
 
     otbrLogInfo("Co-processor version: %s", mHost.GetCoprocessorVersion());
 }
@@ -230,8 +234,9 @@ void Application::CreateRcpMode(const std::string &aRestListenAddress, int aRest
     mVendorServer = vendor::VendorServer::newInstance(*this);
 #endif
 
-    OT_UNUSED_VARIABLE(aRestListenAddress);
-    OT_UNUSED_VARIABLE(aRestListenPort);
+    OTBR_UNUSED_VARIABLE(rcpHost);
+    OTBR_UNUSED_VARIABLE(aRestListenAddress);
+    OTBR_UNUSED_VARIABLE(aRestListenPort);
 }
 
 void Application::InitRcpMode(void)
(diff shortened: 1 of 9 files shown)


src/dbus/server/dbus_object.hpp
```diff
@@ -47,17 +47,30 @@
...
 namespace DBus {
 
+class DependentComponents
+{
+public:
+    Host::ThreadHost &mHost;
+    Mdns::Publisher  &mPublisher;
+#if OTBR_ENABLE_BORDER_AGENT
+    otbr::BorderAgent &mBorderAgent;
+#endif
```
superwhd, line 71: This could make the implementation diverging. Could we always enable border agent?

@bukepo

Irving-cl: What does diverging mean? It makes sense to me to always enable border agent. But then I think we'd better remove the flag 'OTBR_ENABLE_BORDER_AGENT'. Thoughts?

superwhd: By diverge I meant there'll always be two variants of code (enable/disable BA) being used.

> But then I think we'd better remove the flag 'OTBR_ENABLE_BORDER_AGENT'.

Yes. That's also what I think we should do eventually. In this way there's no risk of diverging.

Irving-cl: That sounds good to me. Maybe let's do it in a seperate PR?

superwhd: > That sounds good to me. Maybe let's do it in a seperate PR?

SGTM

src/agent/application.cpp
```diff
@@ -376,4 +375,18 @@ void Application::SetBorderAgentOnInitState(void)
...
+#if OTBR_ENABLE_DBUS_SERVER
+DBus::DependentComponents Application::MakeDBusDependentComponents(void)
+{
+    return DBus::DependentComponents
+    {
+        mHost, *mPublisher
+#if OTBR_ENABLE_BORDER_AGENT
+            ,
+            mBorderAgent
+#endif
```
superwhd, line 387: nit: I think C++ 11 should already support trailing coma. Does it work if you put the comma after `*mPublisher`?

src/dbus/server/dbus_agent.cpp
```diff
@@ -46,14 +46,13 @@ namespace DBus {
 const struct timeval           DBusAgent::kPollTimeout = {0, 0};
 constexpr std::chrono::seconds DBusAgent::kDBusWaitAllowance;
 
-DBusAgent::DBusAgent(otbr::Host::ThreadHost &aHost, Mdns::Publisher &aPublisher)
-    : mInterfaceName(aHost.GetInterfaceName())
-    , mHost(aHost)
-    , mPublisher(aPublisher)
+DBusAgent::DBusAgent(const DependentComponents &aDeps)
+    : mInterfaceName(aDeps.mHost.GetInterfaceName())
+    , mDeps(std::move(aDeps))
```
superwhd, line 51: The `std::move` here doesn't make it use move constructor because there's a `const` qualifier for `aDeps`.

mia1yang (approved):

src/agent/application.cpp
```diff
@@ -376,4 +375,17 @@ void Application::SetBorderAgentOnInitState(void)
...
 #endif
 
+#if OTBR_ENABLE_DBUS_SERVER
+DBus::DependentComponents Application::MakeDBusDependentComponents(void)
+{
+    return DBus::DependentComponents
+    {
+        mHost, *mPublisher,
+#if OTBR_ENABLE_BORDER_AGENT
+            mBorderAgent
```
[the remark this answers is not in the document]
Irving-cl: clang-format made it like this. I'm not quite familiar how to change it. Let's enhance it in a later PR.

jwhui (approved):

PR #2800, opened 2025-04-11
[infra-if] use `std::string` for infra network interface names

superwhd:
This is a minor refactor CL which lets the code prefer `std::string` over C-style strings for convenience and safety.


Irving-cl (approved):
LGTM 👍

jwhui (approved):

PR #2801, opened 2025-04-11
[bbr] add multicast routing manager

Irving-cl:
This PR adds the MulticastRoutingManager module for backbone router multicast forwarding feature.

The implementation follows the [multicast module](https://github.com/openthread/openthread/blob/main/src/posix/platform/multicast_routing.hpp) in Openthread posix.

In this PR the module isn't integrated into application. A unit test is added to test it. It will then be first used for NCP mode to enable the multicast forwarding feature.

### DIFF
--- a/src/common/types.cpp
+++ b/src/common/types.cpp
@@ -67,6 +67,30 @@ Ip6Address Ip6Address::ToSolicitedNodeMulticastAddress(void) const
     return ma;
 }
 
+uint8_t Ip6Address::GetScope(void) const
+{
+    uint8_t rval;
+
+    if (IsMulticast())
+    {
+        rval = m8[1] & 0xf;
+    }
+    else if (IsLinkLocal())
+    {
+        rval = kLinkLocalScope;
+    }
+    else if (IsLoopback())
+    {
+        rval = kNodeLocalScope;
+    }
+    else
+    {
+        rval = kGlobalScope;
+    }
+
+    return rval;
+}
+
 void Ip6Address::CopyTo(struct sockaddr_in6 &aSockAddr) const
 {
     memset(&aSockAddr, 0, sizeof(aSockAddr));
--- a/src/common/types.hpp
+++ b/src/common/types.hpp
@@ -109,6 +109,18 @@ static constexpr char kLinkLocalAllNodesMulticastAddress[] = "ff02::01";
 class Ip6Address
 {
 public:
+    // IPv6 Address Scopes
+    static constexpr uint8_t kNodeLocalScope      = 0;  ///< Node-Local scope
+    static constexpr uint8_t kInterfaceLocalScope = 1;  ///< Interface-Local scope
+    static constexpr uint8_t kLinkLocalScope      = 2;  ///< Link-Local scope
+    static constexpr uint8_t kRealmLocalScope     = 3;  ///< Realm-Local scope
+    static constexpr uint8_t kAdminLocalScope     = 4;  ///< Admin-Local scope
+    static constexpr uint8_t kSiteLocalScope      = 5;  ///< Site-Local scope
+    static constexpr uint8_t kOrgLocalScope       = 8;  ///< Organization-Local scope
+    static constexpr uint8_t kGlobalScope         = 14; ///< Global scope
+
+    static constexpr uint8_t kBitsPerByte = 8;
+
     /**
      * Default constructor.
      */
@@ -230,6 +242,13 @@ public:
      * @retval FALSE  If the Ip6 address is not the Loopback Address.
      */
     bool IsLoopback(void) const { return (m32[0] == 0 && m32[1] == 0 && m32[2] == 0 && m32[3] == htobe32(1)); }
+
+    /**
+     * Returns the IPv6 address scope.
+     *
+     * @returns The IPv6 address scope.
+     */
+    uint8_t GetScope(void) const;
 
     /**
      * This function returns the wellknown Link Local All Nodes Multicast Address (ff02::1).
--- a/src/host/posix/multicast_routing_manager.cpp
+++ b/src/host/posix/multicast_routing_manager.cpp
@@ -1,0 +1,583 @@
+/*
+ *  Copyright (c) 2025, The OpenThread Authors.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
+ *  1. Redistributions of source code must retain the above copyright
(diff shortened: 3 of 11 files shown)


teheather070 (approved):

src/host/posix/multicast_routing_manager.cpp
```diff
@@ -0,0 +1,565 @@
...
+        // MLR.
+        if (HasMulticastListener(aGroupAddr))
+        {
+            forwardMif = kMifIndexThread;
+        }
+    }
+    else
+    {
+        VerifyOrExit(!aSrcAddr.IsLinkLocal(), error = OTBR_ERROR_NONE);
+        // TODO: VerifyOrExit(aSrcAddr.GetPrefix() != AsCoreType(otThreadGetMeshLocalPrefix(gInstance)), error =
```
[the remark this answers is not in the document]
Irving-cl: Good point! I'll need to add a way to get the MeshLocal Prefix in NCP mode first.

sunytt (approved):
LGTM 👍

src/common/types.hpp
```diff
@@ -231,6 +241,25 @@ class Ip6Address
...
+     * This function returns whether the address matches a given network prefix.
+     *
+     * This function only matches the prefix in unit of bytes, not bits.
+     *
+     * @param[in] aPrefix  The network prefix to check.
+     * @param[in] aLength  The length of the prefix to match (in bytes).
+     *
+     * @returns Whether this Ip6 address matches the prefix.
+     */
+    bool MatchPrefix(const uint8_t *aPrefix, uint8_t aLength) const { return memcmp(m8, aPrefix, aLength) == 0; }
```
abtink, line 261: This may become confusing since prefix lengths are often specified in bits?

The similar method in the OT core `Ip6::Address` uses the bit length, so perhaps it would be better to harmonize this (to avoid incorrect use?)

[ot::Ip6::Address::MatchesPrefix()](https://github.com/openthread/openthread/blob/08a8767fcd8599178c40244c16935227c766ed1d/src/core/net/ip6_address.hpp#L775-L784)

src/host/posix/multicast_routing_manager.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+private:
+    enum
+    {
+        kUsPerSecond                              = 1000000, //< Microseoncs per second.
+        kMulticastForwardingCacheExpireTimeout    = 300, //< Expire timeout of Multicast Forwarding Cache (in seconds)
+        kMulticastForwardingCacheExpiringInterval = 60,  //< Expire interval of Multicast Forwarding Cache (in seconds)
+        kMulticastMaxListeners                    = 75,  //< The max number of Multicast listeners
+        kMulticastForwardingCacheTableSize        = kMulticastMaxListeners * 10, //< The max size of MFC table.
+    };
```
abtink, line 69: I suggest using `static constexpr` constants instead of `enum`. The advantage is that the constant's type is well-defined.

We used `enum` early on before OpenThread moved to `C++11` (where `constexpr` is first introduced).

abtink (approved):
Looks good overall.
Two smaller suggestions below.

src/common/types.hpp
Irving-cl: Updated! Used similiar implementation in OT and added a simple unit test.

src/host/posix/multicast_routing_manager.cpp
```diff
@@ -0,0 +1,568 @@
...
+    {
+        error = OTBR_ERROR_ERRNO;
+    }
+
+    otbrLogResult(error, "%s: address %s %s", __FUNCTION__, aAddress.ToString().c_str(), (isAdd ? "Added" : "Removed"));
+}
+
+bool MulticastRoutingManager::HasMulticastListener(const Ip6Address &aAddress) const
+{
+    return mMulticastListeners.find(aAddress) != mMulticastListeners.end();
```
superwhd, line 185: This is rather a large PR and I'd suggest not to introduce this enhancement in this PR. Better to reuse the existing implementation to ensure the correctness.

src/host/posix/multicast_routing_manager.hpp
```diff
@@ -0,0 +1,138 @@
...
+                                     const Host::NetworkProperties &aNetworkProperties);
+
+    void Deinit(void) { FinalizeMulticastRouterSock(); }
+    bool IsEnabled(void) const { return mMulticastRouterSock >= 0; }
+    void HandleStateChange(otBackboneRouterState aState);
+    void HandleBackboneMulticastListenerEvent(otBackboneRouterMulticastListenerEvent aEvent,
+                                              const Ip6Address                      &aAddress);
+
+private:
+    static constexpr uint32_t kUsPerSecond = 1000000; //< Microseoncs per second.
```
superwhd, line 62:
```suggestion
    static constexpr uint32_t kUsPerSecond = 1000000; //< Microseconds per second.
```

tests/gtest/test_backbone_multicast_routing.cpp
```diff
@@ -0,0 +1,210 @@
...
+        exit(EXIT_FAILURE);
+    }
+    while (fgets(buffer.data(), buffer.size(), pipe.get()) != nullptr)
+    {
+        result += buffer.data();
+    }
+    return result;
+}
+
+std::vector<std::string> GetMcastRoutingTable(void)
```
superwhd, line 71:
```suggestion
std::vector<std::string> GetMulticastRoutingTable(void)
```

src/common/types.cpp
```diff
@@ -67,6 +67,61 @@ Ip6Address Ip6Address::ToSolicitedNodeMulticastAddress(void) const
     return ma;
 }
 
+uint8_t Ip6Address::GetScope(void) const
```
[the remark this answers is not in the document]
Irving-cl: Sounds good. Let's do it in a seperate PR because we need to update the OT API.

src/common/types.cpp
```diff
@@ -67,6 +67,61 @@ Ip6Address Ip6Address::ToSolicitedNodeMulticastAddress(void) const
...
+    {
+        rval = kGlobalScope;
+    }
+
+    return rval;
+}
+
+bool Ip6Address::MatchPrefix(const uint8_t *aPrefix, uint8_t aLength) const
+{
+    uint8_t matchedLength = 0;
```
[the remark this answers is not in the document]
Irving-cl: Updated to use existing OT API `otIp6PrefixMatch`.

superwhd (approved):
LGTM 👍  with 2 minor suggestions

jwhui (approved):

PR #2805, opened 2025-04-15
[dbus] add TrelInfo to introspect.xml

jdswensen:
There is currently a handler for TrelInfo, but no introspect entry.

Example fetching information structure:
```
dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:TrelInfo
method return time=1744738821.325205 sender=:1.76 -> destination=:1.81 serial=19 reply_serial=2
   variant       struct {
         boolean true
         uint16 0
         struct {
            uint64 0
            uint64 0
            uint64 0
            uint64 0
            uint64 0
         }
      }
```

### DIFF
--- a/src/dbus/server/introspect.xml
+++ b/src/dbus/server/introspect.xml
@@ -960,6 +960,25 @@
       <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="false"/>
     </property>
 
+    <!-- TrelInfo: Information about the TREL link
+    <literallayout>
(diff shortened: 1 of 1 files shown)


superwhd (approved):
LGTM 👍

jwhui (approved):
Thanks! 👍🏼

PR #2806, opened 2025-04-16
[host] add network property for mesh local prefix

Irving-cl:
This PR adds "otMeshLocalPrefix" to "NetworkProperties" interface.

The intention is to make mesh local prefix available in both NCP & RCP modes. The Multicast Routihng Manager that is under development needs to use the mesh local prefix for check.

In NCP mode, the mesh local prefix is obtained from the notification of `SPINEL_PROP_IPV6_ML_ADDR`. Current NCP alreadys updates this property but it doesn't updates `SPINEL_PROP_IPV6_ML_PREFIX` to the host. So `SPINEL_PROP_IPV6_ML_ADDR` is used.

I've verified locally that in NCP mode the mesh local prefix can be obtained by the interface correctly.

### DIFF
--- a/src/host/ncp_host.cpp
+++ b/src/host/ncp_host.cpp
@@ -45,10 +45,15 @@ namespace Host {
 
 // =============================== NcpNetworkProperties ===============================
 
+constexpr otMeshLocalPrefix kMeshLocalPrefixInit = {
+    {0xfd, 0xde, 0xad, 0x00, 0xbe, 0xef, 0x00, 0x00},
+};
+
 NcpNetworkProperties::NcpNetworkProperties(void)
     : mDeviceRole(OT_DEVICE_ROLE_DISABLED)
 {
     memset(&mDatasetActiveTlvs, 0, sizeof(mDatasetActiveTlvs));
+    SetMeshLocalPrefix(kMeshLocalPrefixInit);
 }
 
 otDeviceRole NcpNetworkProperties::GetDeviceRole(void) const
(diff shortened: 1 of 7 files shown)


src/host/ncp_host.cpp
```diff
@@ -91,6 +96,16 @@ void NcpNetworkProperties::GetDatasetPendingTlvs(otOperationalDatasetTlvs &aData
     OTBR_UNUSED_VARIABLE(aDatasetTlvs);
 }
 
+void NcpNetworkProperties::SetMeshLocalPrefix(const otIp6NetworkPrefix &aMeshLocalPrefix)
+{
+    memcpy(mMeshLocalPrefix.m8, aMeshLocalPrefix.m8, sizeof(mMeshLocalPrefix.m8));
```
abtink, line 101: Do we need to somehow sync/pass this to NCP? (or is it only tracking the value from NCP?)

src/host/ncp_host.hpp
```diff
@@ -58,19 +58,22 @@ class NcpNetworkProperties : virtual public NetworkProperties, public PropsObser
...
 
 private:
     // PropsObserver methods
     void SetDeviceRole(otDeviceRole aRole) override;
     void SetDatasetActiveTlvs(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs) override;
+    void SetMeshLocalPrefix(const otIp6NetworkPrefix &aMeshLocalPrefix) override;
 
     otDeviceRole             mDeviceRole;
     otOperationalDatasetTlvs mDatasetActiveTlvs;
+    otIp6NetworkPrefix       mMeshLocalPrefix;
```
abtink, line 76: This works. But how about using `otMeshLocalPrefix` here instead of `otIp6NetworkPrefix`? I think these two are `typedef`ed as equivalent, but `otMeshLocalPrefix` expresses the intention better?

abtink (approved):
Looks good. 👍 

One small suggestion and a question below. Looks good as is.

src/host/ncp_host.cpp
Irving-cl: It is only tracking the value from NCP. The set method is only called from the `NcpSpinel` to set the value from NCP :)

jwhui (approved):

PR #2824, opened 2025-04-28
[backbone-router] integrate multicast forwarding feature in NCP mode

Irving-cl:
This PR integrates the bbr multicast forwarding feature in NCP mode.

This PR also adds a expect test to test multicast ping on the infrastructure interface can be received by the Thread device.


src/host/ncp_host.cpp
```diff
@@ -250,6 +250,23 @@ void NcpHost::AddThreadEnabledStateChangedCallback(ThreadEnabledStateCallback aC
...
+
+void NcpHost::SetBackboneRouterMulticastListenerCallback(BackboneRouterMulticastListenerCallback aCallback)
+{
+    mNcpSpinel.SetBackboneRouterMulticastListenerCallback(aCallback);
+}
+
+void NcpHost::SetBackboneRouterStateChangedCallback(BackboneRouterStateChangedCallback aCallback)
+{
+    mNcpSpinel.SetBackboneRouterStateChangedCallback(aCallback);
+}
```
superwhd, line 267: For the std::function objects, either:
- Pass by value and use move constructor
or
- Pass by const reference

src/host/ncp_spinel.cpp
```diff
@@ -676,6 +686,15 @@ void NcpSpinel::HandleValueInserted(spinel_prop_key_t aKey, const uint8_t *aBuff
         break;
     }
 #endif // OTBR_ENABLE_SRP_ADVERTISING_PROXY
+    case SPINEL_PROP_BACKBONE_ROUTER_MULTICAST_LISTENER:
+    {
+        const otIp6Address *addr;
+
+        VerifyOrExit(decoder.ReadIp6Address(addr) == OT_ERROR_NONE, error = OTBR_ERROR_PARSE);
```
superwhd, line 693: nit: `SuccessOrExit`

src/host/ncp_spinel.cpp
```diff
@@ -750,6 +769,15 @@ void NcpSpinel::HandleValueRemoved(spinel_prop_key_t aKey, const uint8_t *aBuffe
...
```
superwhd, line 776: Is it calling this one? https://github.com/openthread/openthread/blob/819938d05dae7393d19f2c02a8e5f0aa99900efb/src/lib/spinel/spinel_decoder.hpp#L245

src/host/rcp_host.cpp
```diff
@@ -289,6 +289,10 @@ void RcpHost::Init(void)
 
     otBorderAgentSetMeshCoPServiceChangedCallback(mInstance, RcpHost::HandleMeshCoPServiceChanged, this);
     otBorderAgentEphemeralKeySetCallback(mInstance, RcpHost::HandleEpskcStateChanged, this);
+#if OTBR_ENABLE_BACKBONE_ROUTER
+    otBackboneRouterSetMulticastListenerCallback(mInstance, RcpHost::HandleBackboneRouterMulticastListenerChanged,
+                                                 this);
```
superwhd, line 294: So this PR modifies both RCP and NCP. The PR description should clarify this.

src/host/rcp_host.cpp
```diff
@@ -846,6 +875,25 @@ void RcpHost::HandleEpskcStateChanged(void)
...
+                                                           const otIp6Address                    *aAddress)
+{
+    static_cast<RcpHost *>(aContext)->HandleBackboneRouterMulticastListenerChanged(aEvent, aAddress);
+}
+
+void RcpHost::HandleBackboneRouterMulticastListenerChanged(otBackboneRouterMulticastListenerEvent aEvent,
+                                                           const otIp6Address                    *aAddress)
+{
+    VerifyOrExit(mBackboneRouterMulticastListenerCallback != nullptr);
+    mBackboneRouterMulticastListenerCallback(aEvent, Ip6Address(*aAddress));
```
superwhd, line 890: SafeInvoke?

src/agent/application.cpp
```diff
@@ -370,6 +373,17 @@ void Application::InitNcpMode(void)
         });
     SetBorderAgentOnInitState();
 #endif
+#if OTBR_ENABLE_BACKBONE_ROUTER
+    mHost.SetBackboneRouterStateChangedCallback(
+        [this](otBackboneRouterState aState) { mMulticastRoutingManager->HandleStateChange(aState); });
+    mHost.SetBackboneRouterMulticastListenerCallback(
+        [this](otBackboneRouterMulticastListenerEvent aEvent, const Ip6Address &aAddress) {
+            mMulticastRoutingManager->HandleBackboneMulticastListenerEvent(aEvent, aAddress);
+        });
```
superwhd, line 382: I'm wondering if we should add the same registrations for RCP host?

src/host/ncp_spinel.cpp
Irving-cl: It's a little tricky here. Because the type of error is 'otbrError'. So I used `VerifyOrExit` here to convert the otError to otbrError flexibly.

src/host/ncp_spinel.cpp
Irving-cl: It's calling this one: https://github.com/openthread/openthread/blob/819938d05dae7393d19f2c02a8e5f0aa99900efb/src/lib/spinel/spinel_decoder.hpp#L261

The two are identical.

src/host/rcp_host.cpp
Irving-cl: I actually didn't update the implementation of RCP. That needs update the posix platform in openthread. Since the implementation for this ThreadHost API in RCP is short, I put it in this PR BTW.

src/host/rcp_host.cpp
Irving-cl: SafeInvoke is currently defined for type `AsyncResultReceiver`. Other callback type cannot be used with SafeInvoke. So we can only use `VerifyOrExit` here.

src/agent/application.cpp
Irving-cl: In this PR the multicast forwarding in RCP mode is unchanged. We will have a few dedicated PRs (including changes in openthread posix platform) to replace the implementation.

Irving-cl:
Pending on an issue found: https://github.com/openthread/openthread/issues/11464

tests/scripts/ncp_mode
```diff
@@ -195,7 +196,7 @@ setup_infraif()
 {
     if ! ip link show backbone1 >/dev/null 2>&1; then
         echo "Creating backbone1 with Docker..."
-        docker network create --driver bridge --ipv6 --subnet 9101::/64 -o "com.docker.network.bridge.name"="backbone1" backbone1
+        docker network create --driver bridge -o "com.docker.network.bridge.name"="backbone1" backbone1
```
superwhd, line 199: I'm wondering about the purpose of having `9101::/64` in the old code. Don't we need it any more?

Irving-cl: As I recall, it was added because otherwise will get an error when specifying '--ipv6'. I just found without '--ipv6' all the functionality can still work. 

But with the subnet '9101::/.64', when doling the multicast ping, source address with this prefix will be preferred instead of the on-link address. And the ping fails. So I remove it here.

superwhd: Thanks! I also think we should prefer the on-link prefix over `9101::/64` in tests.

superwhd (approved):
LGTM 👍

abtink (approved):
LGTM. Thanks. 👍

jwhui (approved):

PR #2838, opened 2025-05-08
[ncp] fix the docker network for ncp integration tests

Irving-cl:
This PR fixes the CI failure of ncp_border_routing.

I found that after removing `--ipv6` the otbr in docker cannot get the OMR address. But in my local run, it can work without `--ipv6` option.

This PR adds the original `--ipv6` option back and delete the 9101::1/64 address so that the ping tests through backbone1 interface can use the on-link address.


superwhd (approved):
LGTM 👍

jwhui (approved):

PR #2846, opened 2025-05-09
[dnssd-platform] fix registering services for local host

abtink:
This commit updates `DnssdPlatform::RegisterService()` so it can correctly handle registering a service for the local host. The `otPlatDnssdRegisterService()` indicates this by setting the `mHostName` field in the passed-in `otPlatDnssdService` structure to null. As documented: "The `mHostName` field specifies the host name of the service if it is not NULL. Otherwise, if it is NULL, it indicates that this service is for the device itself and leaves the host name selection to DNS-SD platform."

The `Mdns::Publisher::PublishService()` expects `std::string` as input and assumes an empty host string to indicate the service is for the local host.


jwhui (approved):
Thanks! 👍🏼

PR #2847, opened 2025-05-09
[scripts] remove Network Manager support

jwhui:


bukepo (approved):

Samonitari:
Excuse me for asking, but why it has been removed?

For the reference: I am running otbr-agent on openSUSE MicroOS host, and although I was not running the scripts as they assume debland (to my dismay, although understandably), it provided some reference.
Maybe some docs instead could substitute, in case you are no longer officially support/test devices with nm: ignoring nm completely - especially if you've already mostly done the work - seems like a bad move to me...

jwhui:
> Excuse me for asking, but why it has been removed?

Network Manager interferes with protocols necessary to implement compliant Thread Border Routers. As a result, it is not something we've supported for quite some time:
https://github.com/openthread/ot-br-posix/blob/19fe408163894bd764b2a77667aa2c203e28988d/script/_initrc#L140-L145

If someone wants to contribute support for Network Manager that is compliant with the [Thread Specification](https://www.threadgroup.org/Resources#specifications), we'd be happy to review it.

PR #2848, opened 2025-05-11
[dnssd-platform] check `OTBR_ENABLE_DNSSD_PLAT` config

abtink:
This commit adds `#if` checks for `OTBR_ENABLE_DNSSD_PLAT` to the `dnssd` modules, ensuring the `otPlatDnssd` APIs are not provided unless this feature is enabled.

The `OTBR_ENABLE_DNSSD_PLAT` option controls the corresponding OpenThread core `OT_PLATFORM_DNSSD` configuration, which indicates that `otPlatDnssd` APIs are provided by the platform layer.

### DIFF
--- a/src/host/posix/dnssd.hpp
+++ b/src/host/posix/dnssd.hpp
@@ -36,6 +36,8 @@
 
 #include "openthread-br/config.h"
 
+#if OTBR_ENABLE_DNSSD_PLAT
+
 #include <functional>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


src/host/posix/dnssd.hpp
```diff
@@ -31,11 +31,13 @@
  *   This file includes definitions for implementing OpenThread DNS-SD platform APIs.
  */
 
-#ifndef OTBR_AGENT_POSIX_DNSSD_HPP_
-#define OTBR_AGENT_POSIX_DNSSD_HPP_
+#ifndef OTBR_POSIX_DNSSD_PLAT_HPP_
```
superwhd, line 34: I'm not sure if we should add the `PLAT` into this macro because it's not in the name of the header file.

superwhd (approved):
LGTM with one minor question

PR #2888, opened 2025-06-05
[host] add IsInitialized method

Irving-cl:
This PR adds an API 'IsInitialized' to `ThreadHost`.

The intention is to provide a method for the application side to know whether ThreadHost APIs are ready to call. An immediate use case is add checks in Android OtDaemonServer binder methods.

### DIFF
--- a/src/host/ncp_host.cpp
+++ b/src/host/ncp_host.cpp
@@ -109,7 +109,8 @@ const otMeshLocalPrefix *NcpNetworkProperties::GetMeshLocalPrefix(void) const
 // ===================================== NcpHost ======================================
 
 NcpHost::NcpHost(const char *aInterfaceName, const char *aBackboneInterfaceName, bool aDryRun)
-    : mSpinelDriver(*static_cast<ot::Spinel::SpinelDriver *>(otSysGetSpinelDriver()))
+    : mIsInitialized(false)
+    , mSpinelDriver(*static_cast<ot::Spinel::SpinelDriver *>(otSysGetSpinelDriver()))
     , mCliDaemon(mNcpSpinel)
 {
     memset(&mConfig, 0, sizeof(mConfig));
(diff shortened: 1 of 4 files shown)


src/host/ncp_host.hpp
```diff
@@ -169,6 +175,7 @@ class NcpHost : public MainloopProcessor,
                             const uint8_t    *aData,
                             uint16_t          aDataLen) override;
 
+    std::atomic<bool>         mIsInitialized;
```
wgtdkp: nit: do you expect this field being accessed from multiple threads simultaneously? For Android, all ot-daemon components runs on a single thread.

wgtdkp (approved):

src/host/ncp_host.hpp
wgtdkp: Looks like "atomic" was suggested by Gemini :)

Irving-cl: As discussed, remove the usage of `atomic` here. Thanks!

src/host/ncp_host.cpp
```diff
@@ -109,7 +109,8 @@ const otMeshLocalPrefix *NcpNetworkProperties::GetMeshLocalPrefix(void) const
 // ===================================== NcpHost ======================================
 
 NcpHost::NcpHost(const char *aInterfaceName, const char *aBackboneInterfaceName, bool aDryRun)
-    : mSpinelDriver(*static_cast<ot::Spinel::SpinelDriver *>(otSysGetSpinelDriver()))
+    : mIsInitialized(false)
+    , mSpinelDriver(*static_cast<ot::Spinel::SpinelDriver *>(otSysGetSpinelDriver()))
     , mCliDaemon(mNcpSpinel)
```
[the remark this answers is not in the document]
Irving-cl: This is not true. They already match the declaration order. Reordering this will cause compile failure.

PR #2908, opened 2025-06-25
[thread-helper] move `ThreadHelper` to `host/`

superwhd:
`ThreadHelper` is a class with many responsibilities and dependencies. It's currently in `utils/` which contians a bunch of low level components. Moving `ThreadHelper` to `host` can make the `utils` module more self-contained and more friendly for build systems.


Irving-cl (approved):
LGTM 👍 

It makes more sense to put it under `host`.

superwhd:
> **src/host/thread_helper.cpp:29**
> 
> * [nitpick] Consider updating the log tag from 'UTILS' to better reflect its new location in 'host', such as 'HOST', to avoid potential confusion.

I've updated the log tag.

PR #2928, opened 2025-07-10
[cmake] add `OTBR_MDNS_USE_OT_CORE` build option

jwhui:
Will enable TREL support in a separate PR.


third_party/openthread/CMakeLists.txt
```diff
@@ -62,21 +61,35 @@ set(OT_NAT64_TRANSLATOR ${OTBR_NAT64} CACHE STRING "enable NAT64 translator" FOR
...
 set(OT_UPTIME ON CACHE STRING "enable uptime" FORCE)
 
-if (OTBR_DNSSD_PLAT)
-    set(OT_BORDER_AGENT_SERVICE_NAME ${OTBR_MESHCOP_SERVICE_INSTANCE_NAME} CACHE STRING "set the border agent service base name" FORCE)
+if (OTBR_MDNS_USE_OT_CORE)
+    set(OT_DNSSD_DISCOVERY_PROXY ON CACHE STRING "enable DNS-SD Discovery Proxy" FORCE)
+    set(OT_DNSSD_SERVER ON CACHE STRING "enable DNS-SD server support" FORCE)
+    set(OT_MDNS ON CACHE STRING "OpenThread mDNS" FORCE)
+    set(OT_SRP_ADV_PROXY ON CACHE STRING "enable SRP Advertising Proxy" FORCE)
+    set(OT_TREL_MANAGE_DNSSD ON CACHE STRING "TREL to manage DNSSD and peer discovery" FORCE)
```
abtink, line 83: One combination to consider is `OTBR_DNSSD_DISCOVERY_PROXY` along with `OTBR_MDNS_USE_OT_CORE`.

I think we want `OTBR_DNSSD_DISCOVERY_PROXY` to be explicitly off when `OTBR_MDNS_USE_OT_CORE` is enabled. Otherwise, the OTBR code will enable its own discovery proxy function by registering `otDnssdQuerySetCallbacks()`. In this case, the OT `Dnssd::Server` will use these functions. Basically, if the platform indicates it wants to act as a discovery proxy by registering callbacks, it will take precedence over the native discovery proxy, even if the feature is enabled.

Perhaps we can add a check here that if `OTBR_DNSSD_DISCOVERY_PROXY` is enabled, an error message is emitted for an inconsistent configuration.

third_party/openthread/CMakeLists.txt
```diff
@@ -62,21 +61,35 @@ set(OT_NAT64_TRANSLATOR ${OTBR_NAT64} CACHE STRING "enable NAT64 translator" FOR
...
 set(OT_UDP_FORWARD OFF CACHE STRING "disable udp forward" FORCE)
 set(OT_UPTIME ON CACHE STRING "enable uptime" FORCE)
 
-if (OTBR_DNSSD_PLAT)
-    set(OT_BORDER_AGENT_SERVICE_NAME ${OTBR_MESHCOP_SERVICE_INSTANCE_NAME} CACHE STRING "set the border agent service base name" FORCE)
+if (OTBR_MDNS_USE_OT_CORE)
+    set(OT_DNSSD_DISCOVERY_PROXY ON CACHE STRING "enable DNS-SD Discovery Proxy" FORCE)
+    set(OT_DNSSD_SERVER ON CACHE STRING "enable DNS-SD server support" FORCE)
+    set(OT_MDNS ON CACHE STRING "OpenThread mDNS" FORCE)
+    set(OT_SRP_ADV_PROXY ON CACHE STRING "enable SRP Advertising Proxy" FORCE)
```
abtink, line 82: To be safe maybe we can also do `set(OT_PLATFORM_DNSSD OFF CACHE STRING "enable platform DNSSD" FORCE)` (this should be off by default)

abtink: Also another invalid combination would be if both `OTBR_DNSSD_PLAT` and `OTBR_MDNS_USE_OT_CORE` are enabled. Maybe we can add a check for this?

---

**Edit**: Ignore both comment above please. 

We already do the same check in OpenThread core  - which is even a better place to check for this 😃 
https://github.com/openthread/openthread/blob/d2d644d470193ad37871431fce38666ef44bcce1/src/core/net/dnssd.hpp#L39-L50

abtink (approved):
Looks great. Thanks @jwhui

PR #2941, opened 2025-07-16
[border-agent] add "vendor TXT Data changed" callback

abtink:
This commit adds a callback to `otbr::BorderAgent` to signal changes to the vendor-specific TXT data.

This callback is used when `OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE` is disabled, meaning the OpenThread core is responsible for advertising the Border Agent's MeshCoP service. The new mechanism allows `otbr::BorderAgent` to notify the `ThreadHost`(e.g., `RcpHost` or `NcpHost`) to then pass the updated vendor TXT data to the OpenThread stack.

To support this, a new virtual method `SetBorderAgentVendorTxtData()` is added to the `ThreadHost` interface. An implementation is provided for `RcpHost`, with an empty implementation for `NcpHost` to be added in the future.


src/host/rcp_host.cpp
```diff
@@ -886,6 +886,18 @@ void RcpHost::AddEphemeralKeyStateChangedCallback(EphemeralKeyStateChangedCallba
     mEphemeralKeyStateChangedCallbacks.push_back(aCallback);
 }
 
+void RcpHost::SetBorderAgentVendorTxtData(const std::vector<uint8_t> &aVendorTxtData)
+{
+    VerifyOrExit(mInstance != nullptr);
+
+#if OTBR_ENABLE_BORDER_AGENT && !OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
```
Irving-cl: I suggest we guard this for the entire `ThreadHost::SetBorderAgentVendorTxtData` interface and the implementations in RcpHost and NcpHost. In this way we know that this method is only required when `OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE` is false and we are using OT's implementation.

And this also avoids entering an empty function when `OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE` is true. Thoughts?

Irving-cl (commented):
Overall LGTM. Thanks!

src/host/rcp_host.cpp
abtink: Good point. I updated the code based on this.

Irving-cl (approved):
LGTM 👍

PR #2962, opened 2025-07-23
[dnssd] implement Dnssd Browser/Resolver APIs

Irving-cl:
Following https://github.com/openthread/ot-br-posix/pull/2042 and https://github.com/openthread/ot-br-posix/pull/2664, this PR implements other `otPlatDnssd` platform APIs in ot-br-posix. 

The PR implements all the dnssd platform APIs for browse/resolve except for `RecordQuerier` because currently the mdns publisher doesn't support record query. I plan to support this later.

The background for this PR is that I plan to use the natvie OT discovery proxy and remove the current Discovery Proxy module in ot-br-posix. I've noticed that OT now even supports native mDNS. But to support DNSSD runtime selection (OPENTHREAD_CONFIG_PLATFORM_DNSSD_ALLOW_RUN_TIME_SELECTION), we will also need an implementation on the posix platform.

The implementation in this PR basically follows the old PR https://github.com/openthread/ot-br-posix/pull/2042. But there are some major differences in the dnssd platform API definition since then. The core idea of the implementation for browse/resolve is that we maintain a map of query type to a map of netif index to a vector of callback. Whenever the dnssd gets some result, it returns the result to all the callbacks that are interested to it.

```
Map_BrowsersTable
|  DnsName  |  Map_BrowserCallbackTable  |
|  DnsName  |  Map_BrowserCallbackTable  |
|  DnsName  |  Map_BrowserCallbackTable  |

Map_BrowserCallbackTable
|  NetifIndex  |   Vector_Callbacks   |
|  NetifIndex  |   Vector_Callbacks   |
|  NetifIndex  |   Vector_Callbacks   |
```

As a contrast, the old implementation doesn't need a map of map because there is only one callback. But now we have a callback for each browse/resolve action.

### DIFF
--- a/src/common/types.hpp
+++ b/src/common/types.hpp
@@ -311,6 +311,13 @@ public:
      */
     void CopyFrom(const struct in6_addr &aIn6Addr);
 
+    /**
+     * This method copies the Ip6 address to an `otIp6Address` structure.
+     *
+     * @param[out] aAddress  The `otIp6Address` structure to copy the Ip6 address to.
+     */
+    void CopyTo(otIp6Address &aAddress) const;
+
     union
     {
         uint8_t  m8[16];
--- a/src/host/posix/dnssd.cpp
+++ b/src/host/posix/dnssd.cpp
@@ -36,17 +36,17 @@
 #include "host/posix/dnssd.hpp"
 
 #if OTBR_ENABLE_DNSSD_PLAT
-
 #include <string>
+#include <vector>
 
 #include <openthread/platform/dnssd.h>
 
 #include "common/code_utils.hpp"
 #include "common/logging.hpp"
 #include "common/types.hpp"
+#include "utils/dns_utils.hpp"
 
-static otbr::DnssdPlatform::RegisterCallback MakeRegisterCallback(otInstance                 *aInstance,
-                                                                  otPlatDnssdRegisterCallback aCallback)
+otbr::DnssdPlatform::RegisterCallback MakeRegisterCallback(otInstance *aInstance, otPlatDnssdRegisterCallback aCallback)
 {
     return [aInstance, aCallback](otPlatDnssdRequestId aRequestId, otError aError) {
         if (aCallback)
@@ -113,62 +113,62 @@ extern "C" void otPlatDnssdUnregisterKey(otInstance                 *aInstance,
 
 extern "C" void otPlatDnssdStartBrowser(otInstance *aInstance, const otPlatDnssdBrowser *aBrowser)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aBrowser);
+    otbr::DnssdPlatform::Get().StartServiceBrowser(
+        *aBrowser, MakeUnique<otbr::DnssdPlatform::OtBrowseCallback>(aInstance, aBrowser->mCallback));
 }
 
 extern "C" void otPlatDnssdStopBrowser(otInstance *aInstance, const otPlatDnssdBrowser *aBrowser)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aBrowser);
+    otbr::DnssdPlatform::Get().StopServiceBrowser(
+        *aBrowser, otbr::DnssdPlatform::OtBrowseCallback(aInstance, aBrowser->mCallback));
 }
 
 extern "C" void otPlatDnssdStartSrvResolver(otInstance *aInstance, const otPlatDnssdSrvResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StartServiceResolver(
+        *aResolver, MakeUnique<otbr::DnssdPlatform::OtSrvCallback>(aInstance, aResolver->mCallback));
 }
 
 extern "C" void otPlatDnssdStopSrvResolver(otInstance *aInstance, const otPlatDnssdSrvResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StopServiceResolver(*aResolver,
+                                                   otbr::DnssdPlatform::OtSrvCallback(aInstance, aResolver->mCallback));
 }
 
 extern "C" void otPlatDnssdStartTxtResolver(otInstance *aInstance, const otPlatDnssdTxtResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StartTxtResolver(
+        *aResolver, MakeUnique<otbr::DnssdPlatform::OtTxtCallback>(aInstance, aResolver->mCallback));
 }
 
 extern "C" void otPlatDnssdStopTxtResolver(otInstance *aInstance, const otPlatDnssdTxtResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StopTxtResolver(*aResolver,
+                                               otbr::DnssdPlatform::OtTxtCallback(aInstance, aResolver->mCallback));
 }
 
 extern "C" void otPlatDnssdStartIp6AddressResolver(otInstance *aInstance, const otPlatDnssdAddressResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StartIp6AddressResolver(
+        *aResolver, MakeUnique<otbr::DnssdPlatform::OtAddressCallback>(aInstance, aResolver->mCallback));
 }
 
 extern "C" void otPlatDnssdStopIp6AddressResolver(otInstance *aInstance, const otPlatDnssdAddressResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StopIp6AddressResolver(
+        *aResolver, otbr::DnssdPlatform::OtAddressCallback(aInstance, aResolver->mCallback));
 }
 
 void otPlatDnssdStartIp4AddressResolver(otInstance *aInstance, const otPlatDnssdAddressResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StartIp4AddressResolver(
+        *aResolver, MakeUnique<otbr::DnssdPlatform::OtAddressCallback>(aInstance, aResolver->mCallback));
 }
 
 void otPlatDnssdStopIp4AddressResolver(otInstance *aInstance, const otPlatDnssdAddressResolver *aResolver)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+    otbr::DnssdPlatform::Get().StopIp4AddressResolver(
+        *aResolver, otbr::DnssdPlatform::OtAddressCallback(aInstance, aResolver->mCallback));
 }
 
 void otPlatDnssdStartRecordQuerier(otInstance *aInstance, const otPlatDnssdRecordQuerier *aQuerier)
@@ -187,13 +187,16 @@ void otPlatDnssdStopRecordQuerier(otInstance *aInstance, const otPlatDnssdRecord
 
 namespace otbr {
 
-DnssdPlatform *DnssdPlatform::sDnssdPlatform = nullptr;
+DnssdPlatform            *DnssdPlatform::sDnssdPlatform = nullptr;
+static constexpr uint64_t kInvalidSubscriberId          = 0;
 
 DnssdPlatform::DnssdPlatform(Mdns::Publisher &aPublisher)
     : mPublisher(aPublisher)
     , mState(kStateStopped)
     , mRunning(false)
+    , mInvokingCallbacks(false)
     , mPublisherState(Mdns::Publisher::State::kIdle)
+    , mSubscriberId(kInvalidSubscriberId)
 {
     sDnssdPlatform = this;
 }
@@ -222,13 +225,15 @@ void DnssdPlatform::UpdateState(void)
     {
         VerifyOrExit(mState != kStateReady);
 
-        mState = kStateReady;
+        mState        = kStateReady;
+        mSubscriberId = mPublisher.AddSubscriptionCallbacks(HandleDiscoveredService, HandleDiscoveredHost);
     }
     else
     {
         VerifyOrExit(mState != kStateStopped);
 
         mState = kStateStopped;
+        mPublisher.RemoveSubscriptionCallbacks(mSubscriberId);
     }
 
     if (mStateChangeCallback)
@@ -330,6 +335,160 @@ void DnssdPlatform::UnregisterKey(const Key &aKey, RequestId aRequestId, Registe
     mPublisher.UnpublishKey(KeyNameFor(aKey), MakePublisherCallback(aRequestId, aCallback));
 }
 
+void DnssdPlatform::StartServiceBrowser(const Browser &aBrowser, BrowseCallbackPtr aCallbackPtr)
+{
+    DnsServiceType fullServiceType(aBrowser.mServiceType, aBrowser.mSubTypeLabel);
+    auto          &entryList         = mServiceBrowsersMap[fullServiceType];
+    bool           needsSubscription = !entryList.HasSubscribedOnInfraIf(aBrowser.mInfraIfIndex);
+
+    VerifyOrExit(!mInvokingCallbacks);
+
+    entryList.AddIfAbsent(aBrowser.mInfraIfIndex, std::move(aCallbackPtr));
+
+    if (needsSubscription)
+    {
+        mPublisher.SubscribeService(fullServiceType.ToString(), "");
+    }
+
+exit:
+    return;
+}
+
+void DnssdPlatform::StopServiceBrowser(const Browser &aBrowser, const BrowseCallback &aCallback)
+{
+    DnsServiceType fullServiceType(aBrowser.mServiceType, aBrowser.mSubTypeLabel);
+    auto           iter = mServiceBrowsersMap.find(fullServiceType);
+
+    if (iter != mServiceBrowsersMap.end())
+    {
+        auto &entryList = iter->second;
+
+        entryList.MarkAsDeleted(aBrowser.mInfraIfIndex, aCallback);
+
+        if (!entryList.HasAnyValidCallbacks())
+        {
+            mPublisher.UnsubscribeService(fullServiceType.ToString(), "");
+        }
+
+        if (!mInvokingCallbacks)
+        {
+            entryList.CleanUpDeletedEntries();
+            if (entryList.IsEmpty())
+            {
+                mServiceBrowsersMap.erase(iter);
+            }
+        }
+    }
+}
+
+void DnssdPlatform::StartServiceResolver(const SrvResolver &aSrvResolver, SrvCallbackPtr aCallbackPtr)
+{
+    auto &entryList = mServiceResolversMap[DnsServiceName(aSrvResolver.mServiceInstance, aSrvResolver.mServiceType)];
+    bool  needsSubscription = !entryList.HasSubscribedOnInfraIf(aSrvResolver.mInfraIfIndex);
+
+    VerifyOrExit(!mInvokingCallbacks);
+
+    entryList.AddIfAbsent(aSrvResolver.mInfraIfIndex, std::move(aCallbackPtr));
+
+    if (needsSubscription)
+    {
+        mPublisher.SubscribeService(aSrvResolver.mServiceType, aSrvResolver.mServiceInstance);
+    }
+
+exit:
+    return;
+}
+
+void DnssdPlatform::StopServiceResolver(const SrvResolver &aSrvResolver, const SrvCallback &aCallback)
+{
+    auto iter = mServiceResolversMap.find(DnsServiceName(aSrvResolver.mServiceInstance, aSrvResolver.mServiceType));
+
+    if (iter != mServiceResolversMap.end())
+    {
+        auto &entryList = iter->second;
+
+        entryList.MarkAsDeleted(aSrvResolver.mInfraIfIndex, aCallback);
+
+        if (!entryList.HasAnyValidCallbacks())
+        {
+            mPublisher.UnsubscribeService(aSrvResolver.mServiceType, aSrvResolver.mServiceInstance);
+        }
+
+        if (!mInvokingCallbacks)
+        {
+            entryList.CleanUpDeletedEntries();
+            if (entryList.IsEmpty())
+            {
+                mServiceResolversMap.erase(iter);
+            }
+        }
+    }
+}
+
+void DnssdPlatform::StartTxtResolver(const TxtResolver &aTxtResolver, TxtCallbackPtr aCallbackPtr)
+{
+    auto &entryList = mTxtResolversMap[DnsServiceName(aTxtResolver.mServiceInstance, aTxtResolver.mServiceType)];
+    bool  needsSubscription = !entryList.HasSubscribedOnInfraIf(aTxtResolver.mInfraIfIndex);
+
+    VerifyOrExit(!mInvokingCallbacks);
+
+    entryList.AddIfAbsent(aTxtResolver.mInfraIfIndex, std::move(aCallbackPtr));
+
+    if (needsSubscription)
+    {
+        mPublisher.SubscribeService(aTxtResolver.mServiceType, aTxtResolver.mServiceInstance);
+    }
+
+exit:
+    return;
+}
+
+void DnssdPlatform::StopTxtResolver(const TxtResolver &aTxtResolver, const TxtCallback &aCallback)
+{
+    auto iter = mTxtResolversMap.find(DnsServiceName(aTxtResolver.mServiceInstance, aTxtResolver.mServiceType));
+
+    if (iter != mTxtResolversMap.end())
+    {
+        auto &entryList = iter->second;
+
+        entryList.MarkAsDeleted(aTxtResolver.mInfraIfIndex, aCallback);
+
+        if (!entryList.HasAnyValidCallbacks())
+        {
+            mPublisher.UnsubscribeService(aTxtResolver.mServiceType, aTxtResolver.mServiceInstance);
+        }
+
+        if (!mInvokingCallbacks)
+        {
+            entryList.CleanUpDeletedEntries();
+            if (entryList.IsEmpty())
+            {
+                mTxtResolversMap.erase(iter);
+            }
+        }
+    }
+}
+
+void DnssdPlatform::StartIp6AddressResolver(const AddressResolver &aAddressResolver, AddressCallbackPtr aCallbackPtr)
+{
+    StartAddressResolver(aAddressResolver, std::move(aCallbackPtr), mIp6AddrResolversMap);
+}
+
+void DnssdPlatform::StopIp6AddressResolver(const AddressResolver &aAddressResolver, const AddressCallback &aCallback)
+{
+    StopAddressResolver(aAddressResolver, aCallback, mIp6AddrResolversMap);
+}
+
+void DnssdPlatform::StartIp4AddressResolver(const AddressResolver &aAddressResolver, AddressCallbackPtr aCallbackPtr)
+{
+    StartAddressResolver(aAddressResolver, std::move(aCallbackPtr), mIp4AddrResolversMap);
+}
+
+void DnssdPlatform::StopIp4AddressResolver(const AddressResolver &aAddressResolver, const AddressCallback &aCallback)
+{
+    StopAddressResolver(aAddressResolver, aCallback, mIp4AddrResolversMap);
+}
+
 void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
 {
     if (mPublisherState != aState)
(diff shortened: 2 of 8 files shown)


src/host/posix/dnssd.cpp
```diff
@@ -339,6 +463,186 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+    for (auto callback : mTxtResolversTable[serviceName].GetCallbacks(txtResult.mInfraIfIndex))
+    {
+        callback(txtResult);
+    }
+
+exit:
+    return;
+}
+
+void DnssdPlatform::ProcessAddrResolvers(const std::string                                 &aHostName,
```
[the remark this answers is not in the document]
Irving-cl: I think our DNSSD platform APIs have such contracts where pointers are only valid for the duration of the callback. Is it true? @abtink

src/common/types.hpp
```diff
@@ -313,6 +313,11 @@ class Ip6Address
      */
     void CopyFrom(const struct in6_addr &aIn6Addr);
 
+    /**
+     * This method returns a const reference to type otIp6Address.
+     */
+    const otIp6Address &AsOtIp6Address(void) const { return *reinterpret_cast<const otIp6Address *>(m8); }
```
superwhd: I'm not sure if this will work. I saw `otIp6Address` marked with `OT_TOOL_PACKED_FIELD`. I don't know whether this will change the layout of the struct.

src/host/posix/dnssd.cpp
```diff
@@ -113,71 +114,84 @@ extern "C" void otPlatDnssdUnregisterKey(otInstance                 *aInstance,
...
 extern "C" void otPlatDnssdStopIp6AddressResolver(otInstance *aInstance, const otPlatDnssdAddressResolver *aResolver)
 {
     OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aResolver);
+
+    otbr::DnssdPlatform::Get().StopIp6AddressResolver(*aResolver);
 }
 
-void otPlatDnssdStartIp4AddressResolver(otInstance *aInstance, const otPlatDnssdAddressResolver *aResolver)
+extern "C" void otPlatDnssdStartIp4AddressResolver(otInstance *aInstance, const otPlatDnssdAddressResolver *aResolver)
```
superwhd: Just curious: this will still work even if you don't have mark it as `extern "C"` in the .cpp file, right?

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,136 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
-    State                    mState;
-    bool                     mRunning;
-    Mdns::Publisher::State   mPublisherState;
-    DnssdStateChangeCallback mStateChangeCallback;
+    Mdns::Publisher                                     &mPublisher;
+    State                                                mState;
+    bool                                                 mRunning;
+    Mdns::Publisher::State                               mPublisherState;
+    DnssdStateChangeCallback                             mStateChangeCallback;
+    uint64_t                                             mSubscriberId;
```
superwhd: Better to explicitly initialize it with an invalid ID value so that you won't accidentally reference to an uninitialized value.

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,136 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+        DnsName mType;
+    };
+
+    template <typename CallbackType> class CallbackTable
+    {
+    public:
+        void Add(uint64_t aInterfaceIndex, CallbackType aCallback)
+        {
+            mCallbackTable[aInterfaceIndex].push_back(aCallback);
+        }
```
superwhd: 1. The `InterfaceIndex` may be ambiguous. Better to be consistent with `otPlat` APIs to name it as `InfraIfIndex`. 
2. I suppose we only need to support one infra netif at any moment. Do we need to introduce such a map?

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,136 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+        bool Matches(uint64_t aInterfaceIndex) { return mCallbackTable.find(aInterfaceIndex) != mCallbackTable.end(); }
+
+        bool IsEmpty(void) { return mCallbackTable.empty(); }
+
+        const std::vector<CallbackType> &GetCallbacks(uint64_t aInterfaceIndex)
+        {
+            static std::vector<CallbackType> sEmptyResult;
+
+            auto *ptr = &sEmptyResult;
+            if (mCallbackTable.find(aInterfaceIndex) != mCallbackTable.end())
```
superwhd: nit: You can store the iterator and derive the return value from it.

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,136 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
 
 private:
     static constexpr State kStateReady   = OT_PLAT_DNSSD_READY;
     static constexpr State kStateStopped = OT_PLAT_DNSSD_STOPPED;
 
+    class DnsName
+    {
+    public:
+        DnsName(std::string aName)
+            : mName(aName)
```
superwhd:
```suggestion
            : mName(std::move(aName))
```

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,136 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+    private:
+        std::string mName;
+    };
+
+    class DnsServiceName
+    {
+    public:
+        DnsServiceName(std::string aInstance, std::string aType)
+            : mInstance(aInstance)
+            , mType(aType)
```
superwhd:
```suggestion
            : mInstance(std::move(aInstance))
            , mType(std::move(aType))
```

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,136 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+        DnsName mInstance;
+        DnsName mType;
+    };
+
+    template <typename CallbackType> class CallbackTable
+    {
+    public:
+        void Add(uint64_t aInterfaceIndex, CallbackType aCallback)
+        {
+            mCallbackTable[aInterfaceIndex].push_back(aCallback);
```
superwhd:
```suggestion
            mCallbackTable[aInterfaceIndex].push_back(std::move(aCallback));
```

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,128 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+        bool IsEmpty(void) { return mCallbackTable.empty(); }
+
+        const std::vector<CalllbackType> &GetCallbacks(uint64_t aInterfaceIndex)
+        {
+            return mCallbackTable[aInterfaceIndex];
+        }
+
+    private:
+        std::map<uint64_t, std::vector<CalllbackType>> mCallbackTable;
+    };
```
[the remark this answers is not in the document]
Irving-cl: 1. Currently there isn't an identifier for each Browser when stopping them. (Like a RequestId we used for register service) Is it guaranteed that the Browser object keeps the same callback when calling `Dnssd::Browser`? If so, we can iterate the vector and find the callback and remove it. Thoughts? @abtink 

2. Updated.
3. Updated.

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +469,186 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+{
+    auto &callbackTable = aResolversTable[DnsName(aAddressResolver.mHostName)];
+    bool  wasEmpty      = callbackTable.IsEmpty();
+
+    callbackTable.Add(aAddressResolver.mInfraIfIndex, aCallback);
+
+    if (wasEmpty)
+    {
+        mPublisher.SubscribeHost(aAddressResolver.mHostName);
+    }
```
superwhd: I'm not so sure whether the logic is correct. Don't you want to subscribe to the host if you've subscribed to it on a different interface?

src/host/posix/dnssd.cpp
```diff
@@ -330,6 +340,120 @@
...
+        mServiceResolversTable[DnsServiceName(aSrvResolver.mServiceInstance, aSrvResolver.mServiceType)];
+    bool wasEmpty = callbackTable.IsEmpty();
+
+    callbackTable.Add(aSrvResolver.mInfraIfIndex, aCallback);
+
+    if (wasEmpty)
+    {
+        mPublisher.SubscribeService(aSrvResolver.mServiceType, aSrvResolver.mServiceInstance);
+    }
+}
```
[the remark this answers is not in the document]
Irving-cl: There are still some differences regarding the parameters used. For example, in `StartServiceBrowser` we use "" as the instance name while in `StopServiceResolver` we use `aSrvResolver.mServiceInstance`. To adapt them we need to introduce extra template specialization, which is not worthwhile.

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +469,186 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+    srvResult.mServiceInstance = instanceName.c_str();
+    srvResult.mServiceType     = aType.c_str();
+    srvResult.mHostName        = hostName.c_str();
+    srvResult.mPort            = aInfo.mPort;
+    srvResult.mPriority        = aInfo.mPriority;
+    srvResult.mWeight          = aInfo.mWeight;
+    srvResult.mTtl             = aInfo.mTtl;
+    srvResult.mInfraIfIndex    = aInfo.mNetifIndex;
+
+    for (auto callback : it->second.GetCallbacks(srvResult.mInfraIfIndex))
```
superwhd:
```suggestion
    for (const auto &callback : it->second.GetCallbacks(srvResult.mInfraIfIndex))
```

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +469,186 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+
+    instanceName = DnsUtils::UnescapeInstanceName(aInfo.mName);
+
+    result.mServiceType     = aType.c_str();
+    result.mSubTypeLabel    = nullptr;
+    result.mServiceInstance = instanceName.c_str();
+    result.mTtl             = aInfo.mTtl;
+    result.mInfraIfIndex    = aInfo.mNetifIndex;
+
+    for (auto callback : it->second.GetCallbacks(result.mInfraIfIndex))
```
superwhd:
```suggestion
    for (const auto &callback : it->second.GetCallbacks(result.mInfraIfIndex))
```

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +469,186 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+    VerifyOrExit(it != mTxtResolversTable.end() && it->second.Matches(aInfo.mNetifIndex));
+
+    txtResult.mServiceInstance = instanceName.c_str();
+    txtResult.mServiceType     = aType.c_str();
+    txtResult.mTxtData         = aInfo.mTxtData.data();
+    txtResult.mTxtDataLength   = aInfo.mTxtData.size();
+    txtResult.mTtl             = aInfo.mTtl;
+    txtResult.mInfraIfIndex    = aInfo.mNetifIndex;
+
+    for (auto callback : it->second.GetCallbacks(txtResult.mInfraIfIndex))
```
superwhd:
```suggestion
    for (const auto &callback : it->second.GetCallbacks(txtResult.mInfraIfIndex))
```

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +469,186 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+    result.mHostName = aHostName.c_str();
+    for (const auto &addr : aInfo.mAddresses)
+    {
+        addressAndTtls.push_back({addr.AsOtIp6Address(), aInfo.mTtl});
+    }
+    result.mAddresses       = addressAndTtls.data();
+    result.mAddressesLength = addressAndTtls.size();
+    result.mInfraIfIndex    = aInfo.mNetifIndex;
+
+    for (auto callback : aResolversTable[aHostName].GetCallbacks(result.mInfraIfIndex))
```
superwhd:
```suggestion
    for (const auto &callback : aResolversTable[aHostName].GetCallbacks(result.mInfraIfIndex))
```

src/host/posix/dnssd.cpp
Irving-cl: As I tried, yes. I noticed the previous ones have it so I added it as well. But thinking on this, the plat dnssd API is only called by OT core. So there isn't any c/c++ interoperability. I think I can remove them and remove those in above functions as well. Thoughts?

src/common/types.hpp
Irving-cl: Updated to use a different approach.

src/host/posix/dnssd.hpp
Irving-cl: Add initialization for it in the constructor.

src/host/posix/dnssd.hpp
Irving-cl: 1. Updated.
2. The current dnssd platform APIs are designed in a way to support multiple infra if. (Each query has a infra if index bound to it) And the original implementation in https://github.com/openthread/ot-br-posix/pull/2042 also supports multiple infra if. So maybe let's keep it.

src/host/posix/dnssd.cpp
Irving-cl: Good catch! I added a check to make sure the subscription will be done if it's a different interface.

src/host/posix/dnssd.cpp
```diff
@@ -330,6 +348,120 @@ void DnssdPlatform::UnregisterKey(const Key &aKey, RequestId aRequestId, Registe
...
+void DnssdPlatform::StartServiceBrowser(const Browser &aBrowser, const BrowseCallback &aCallback)
+{
+    auto &callbackTable = mServiceBrowsersTable[DnsName(aBrowser.mServiceType)];
+    bool  wasEmpty      = callbackTable.IsEmpty() || !callbackTable.Matches(aBrowser.mInfraIfIndex);
+
+    callbackTable.Add(aBrowser.mInfraIfIndex, aCallback);
+
+    if (wasEmpty)
+    {
+        mPublisher.SubscribeService(aBrowser.mServiceType, "");
```
abtink: Looks like we are not yet handling the sub-type browse case? (i.e., when `aBrowser.mSubTypeLabel` is not null)?

src/host/posix/dnssd.cpp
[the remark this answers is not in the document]
abtink: Yes. Correct. Using stack allocation is perfectly fine.

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +471,189 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+    result.mServiceType     = aType.c_str();
+    result.mSubTypeLabel    = nullptr;
+    result.mServiceInstance = instanceName.c_str();
+    result.mTtl             = aInfo.mTtl;
+    result.mInfraIfIndex    = aInfo.mNetifIndex;
+
+    for (const auto &callback : it->second.GetCallbacks(result.mInfraIfIndex))
+    {
+        callback(result);
+    }
```
abtink: I think we may need to make this more robust/safe:

* We iterate over a list of callbacks and invoke them sequentially with the new `result`.
* What if a callback itself makes changes to the callback list (e.g., by calling `otPlatDnssd` APIs to start/stop resolvers/browsers and add/remove callbacks)?
* This could lead to modifying the list while iterating over it, causing iterator invalidation issues.

----

One way to address this is to have `GetCallbacks()` return a copy of the list instead of a `const &`.
* While this is simpler, it has a downside: if the callback list changes, we might still use the stale list (e.g., invoking a callback that has already been removed).

----

A more complex, but safer, way is as follows:
* We track when we are invoking callbacks using an `mInvokingCallbacks` boolean flag.
* In the `Stop{Item}()` methods, we behave differently based on the state of `mInvokingCallbacks`.
    * Instead of immediately removing an entry from the list, we keep it but mark it as deleted (e.g., by setting its function pointer to `nullptr` or adding a new `mRemoved` flag).
    * We create a `CleanUpRemovedCallbacks()` method to permanently remove the entries marked for deletion from all callback tables.
    * In `Stop{Item}()`, if `!mInvokingCallbacks`, we can safely call `CleanUpRemovedCallbacks()`; otherwise, we don't.
* In the `Process{Item}()` methods (where callbacks are invoked), we first set `mInvokingCallbacks` to true. 
* Then, we iterate through the list and only invoke callbacks that have not been marked as removed.
* Finally, after the iteration is complete, we clear `mInvokingCallbacks` and call `CleanUpRemovedCallbacks()`.

There may be other ways to solve this as well.

src/host/posix/dnssd.hpp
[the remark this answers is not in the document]
abtink: I think the comment 1 may be legitimate. Will add comment in the code around `StopServiceBrowser`

src/host/posix/dnssd.cpp
```diff
@@ -330,6 +348,120 @@ void DnssdPlatform::UnregisterKey(const Key &aKey, RequestId aRequestId, Registe
...
+        auto &callbackTable = it->second;
+
+        callbackTable.Remove(aBrowser.mInfraIfIndex);
+
+        if (callbackTable.IsEmpty())
+        {
+            mServiceBrowsersTable.erase(it);
+            mPublisher.UnsubscribeService(aBrowser.mServiceType, "");
+        }
+    }
```
abtink, line 381: Follow-up from https://github.com/openthread/ot-br-posix/pull/2962#discussion_r2224568436

The `otPlatDnssdBrowser` is defined as:

```c
typedef struct otPlatDnssdBrowser
{
    const char               *mServiceType;   ///< The service type (e.g., "_mt._udp"). MUST NOT include domain name.
    const char               *mSubTypeLabel;  ///< The sub-type label if Browse for sub-type, NULL otherwise.
    uint32_t                  mInfraIfIndex;  ///< The infrastructure network interface index.
    otPlatDnssdBrowseCallback mCallback;       ///< The callback to report result.
} otPlatDnssdBrowser;
```

Essentially, a browser is considered the same if all its fields match: service type, sub-type, interface index, and callback pointer.

Multiple browsers can be started for the same service type (and sub-type), provided they use a different `mCallback` or `mInfraIfIndex`. They can be stopped separately from one another; each browser is its own entity.

-----

So, the logic here should be as follows:

1.  First, we look up the `mServiceBrowsersTable`, which is a map associating a `DNSName` (the service type) with a "callback table".
2.  The `callbackTable` itself is a map from an interface index to a list of callbacks (a `std::vector<CallbackType>`).
3.  We find the entry in this table associated with the `aBrowser.mInfraIfIndex`.
4.  Finally, we search its list of callbacks and remove the one matching `aBrowser.mCallback`.

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +128,138 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+        bool operator<(const DnsName &aOther) const
+        {
+            return StringUtils::ToLowercase(mName) < StringUtils::ToLowercase(aOther.mName);
+        }
+
+    private:
+        std::string mName;
+    };
+
+    class DnsServiceName
```
abtink, line 290: Similar to `DnsServiceName`, perhaps we could define a new class (e.g., `DnsServiceType`) to represent a service type with an optional sub-type. This class would basically contain two `DnsName` objects, where the sub-type can be empty or null.

We could then use this new `DnsServiceType` class as the key in `mServiceBrowsersTable`, mapping it to the `callbacktable`.

This should also allow us to handle sub-type Browse, though we will probably need to construct the full sub-type name from its pieces before passing it to the `ot-br` `mPublisher`. I assume the `mPublisher` requires the format `"<sub-type-label>._sub.<service_type>"` and can then do sub-type browse?

src/host/posix/dnssd.cpp
```diff
@@ -56,6 +56,11 @@ static otbr::DnssdPlatform::RegisterCallback MakeRegisterCallback(otInstance
     };
 }
 
+template <typename T> otbr::DnssdPlatform::DiscoveryCallbackIdentifier MakeDiscoveryCallbackIdentifier(T *aCallback)
```
superwhd: This function doesn't 'make' a new ID. Maybe we can rename it as `ToDiscoveryCallbackId()` for short?

src/host/posix/dnssd.cpp
```diff
@@ -113,62 +118,80 @@ extern "C" void otPlatDnssdUnregisterKey(otInstance                 *aInstance,
 
 extern "C" void otPlatDnssdStartBrowser(otInstance *aInstance, const otPlatDnssdBrowser *aBrowser)
 {
-    OTBR_UNUSED_VARIABLE(aInstance);
-    OTBR_UNUSED_VARIABLE(aBrowser);
+    otbr::DnssdPlatform::Get().StartServiceBrowser(
+        *aBrowser, MakeDiscoveryCallbackIdentifier(aBrowser->mCallback),
+        [aInstance, aBrowser](const otbr::DnssdPlatform::BrowseResult &aResult) {
+            aBrowser->mCallback(aInstance, &aResult);
```
superwhd: I haven't looked into the OT implementation in details. I wonder if `aBrowser` is guaranteed to be available in this lambda? Same for the `aResolver` in the following code.

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +129,243 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
 
-    Mdns::Publisher         &mPublisher;
-    State                    mState;
-    bool                     mRunning;
-    Mdns::Publisher::State   mPublisherState;
-    DnssdStateChangeCallback mStateChangeCallback;
+    Mdns::Publisher                                               &mPublisher;
+    State                                                          mState;
+    bool                                                           mRunning;
+    bool                                                           mInvokingCallbacks;
```
superwhd: I saw this used for blocking the future discoveries. Could you explain why it's needed?

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +534,209 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+}
+
+void DnssdPlatform::StartAddressResolver(const AddressResolver                                    &aAddressResolver,
+                                         const AddressCallback                                    &aCallback,
+                                         const DiscoveryCallbackIdentifier                        &aId,
+                                         std::map<DnsName, InfraIfCallbackTable<AddressCallback>> &aResolversTable)
+{
+    auto &callbackTable     = aResolversTable[DnsName(aAddressResolver.mHostName)];
+    bool  needsSubscription = !callbackTable.HasAnyValidCallbacks(aAddressResolver.mInfraIfIndex);
+    VerifyOrExit(!mInvokingCallbacks);
```
superwhd, line 652:
```suggestion

    VerifyOrExit(!mInvokingCallbacks);
```
nit

src/host/posix/dnssd.cpp
Irving-cl: Updated.

src/host/posix/dnssd.cpp
Irving-cl: Good catch! I updated to capture the callback (pointer) instead.

src/host/posix/dnssd.hpp
Irving-cl: The main intention of this is to support the lazy deletion as mentioned in above the [comment](https://github.com/openthread/ot-br-posix/pull/2962#discussion_r2229864405).

But I just came up with another issue: adding callback (invoking Start... methods during callback invoking) also brings the "modifying during traversal issue" and it is even more complicated. To support adding callbacks during table traversal, we have to change this model to a event driven model (adding a task runner to post task). 

Considering it's really not a meaningful use case that the user starts a discovery in the callback of a previous discovery, I think we can simply forbid it. The user code can just start the discoveries. We can make this in the comment of these methods and simply discard the callback if callbacks are being invoked. Thoughts?

src/host/posix/dnssd.cpp
Irving-cl: Added `DnsServiceType` and `ToString` method to prepend subType and  `_.sub.`

src/host/posix/dnssd.cpp
Irving-cl: I implemented the lasy deletion basically following the comments. Thanks!

But I just came up with another issue: adding to the data structure during iteration is more difficult and complicated. I think we have to change to a event driven model to support that. However, I think that may be not worthwhile. If the user code wants to discover something, it can simply calls the `Dnssd::Start` method and doesn't need to do that in the callback of previous discovery operation.

So my proposal is that we simply don't support adding callbacks while iteration. Thoughts?

src/host/posix/dnssd.cpp
Irving-cl: I added a `DiscoveryCallbackIdentifier` at the Dnssd class API layer to identify the calls. It's similar to `RequestId` but the Dnssd platform API user doesn't need to maintain the Ids.

`DiscoveryCallbackIdentifier` is directly converted from the otPlatDnssd callbacks. 

**Why not directly using the otPlatDnssd callbacks as the identifier?**
 It's ok for the RCP design. But for NCP design, I expect the callback (function pointer)'s value will still be passed to the host. To make sure the function pointer from the NCP can be stored, we'd better use uint64_t.

src/host/posix/dnssd.hpp
Irving-cl: Updated!

src/host/posix/dnssd.hpp
superwhd: I'm not so sure whether there is a use case. Could there be a browse of services triggering following up resolutions of records?

Irving-cl: It's not a problem if we really have such use cases. But we just need to be clear on the API's capability.

Since the caller of these dnssd platform APIs are OT core dnssd, the callback is also implemented in OT core. We can post a task to do trailing resolution. Thus the `Start` won't be directly invoked in the callback.

src/host/posix/dnssd.cpp
```diff
@@ -339,6 +536,210 @@ void DnssdPlatform::HandleMdnsState(Mdns::Publisher::State aState)
...
+
+    VerifyOrExit(mState == kStateReady);
+
+    VerifyOrExit(it != aResolversTable.end());
+
+    result.mHostName = aHostName.c_str();
+    for (const auto &addr : aInfo.mAddresses)
+    {
+        otIp6Address ip6Addr;
+        addr.CopyTo(ip6Addr);
```
superwhd:
```suggestion
        otIp6Address ip6Addr;

        addr.CopyTo(ip6Addr);
```
nit

superwhd (approved):
LGTM 👍

src/host/posix/dnssd.hpp
```diff
@@ -112,25 +129,243 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+                }
+            }
+        }
+
+        bool IsEmpty(void) { return mInfraIfCallbackTable.empty(); }
+
+    private:
+        typedef std::map<DiscoveryCallbackIdentifier, CallbackType> CallbackMap;
+
+        std::map<uint64_t, CallbackMap> mInfraIfCallbackTable;
```
abtink: This works, but it feels a bit complex. The current approach of casting a callback function pointer to a `uint64_t` just to use it as a map key is a bit confusing.

I have a suggestion: what if we change the data structure to something more direct? An OT browser or resolver is always tied to a DNS name, a `mCallback`, and an `mInfraIfIndex`.

```
typedef struct otPlatDnssdAddressResolver
{
    const char                *mHostName;     ///< The host name (e.g., "myhost"). MUST NOT contain domain name.
    uint32_t                   mInfraIfIndex; ///< The infrastructure network interface index.
    otPlatDnssdAddressCallback mCallback;     ///< The callback to report result.
} otP


We could use a map from a `DnsName` (or `DnsServiceType`, etc.) to a `std::vector` of structs that track the callback. This new struct could store and track both the `mCallback` and `mInfraIfIndex` together.

Then, when we need to stop a browser or resolver, we can simply look up the DNS name to find the `std::vector`, search that list to find the matching entry, and then remove it.

What do you think of this idea?

File: src/host/posix/dnssd.cpp [hunk 27]
```diff
@@ -330,6 +360,173 @@ void DnssdPlatform::UnregisterKey(const Key &aKey, RequestId aRequestId, Registe
...
+    bool           needsSubscription = !callbackTable.HasAnyValidCallbacks(aBrowser.mInfraIfIndex);
+
+    VerifyOrExit(!mInvokingCallbacks);
+
+    callbackTable.Add(aBrowser.mInfraIfIndex, aId, aCallback);
+
+    if (needsSubscription)
+    {
+        mPublisher.SubscribeService(fullServiceType.ToString(), "");
+    }
```
abtink, line 351: I just noticed a potential issue with how the `Publisher` works.

We can start multiple `Browser` instances for the same service type, each with a different callback and at different times.

While an initial `Browser` is running, it may discover several results and invoke its callback. When a second `Browser` for the same service name starts, the expectation is that the previously cached results from the underlying mDNS will also be reported to this new browser via its callback.

I'm not sure this will work with the current model, where `mPublisher.SubscribeService()` is called only once when the first browser starts.

I think that the underlying Avahi or mDNSResponder raw APIs should allow starting multiple browsers/resolvers without a problem. However, the `Publisher` abstraction and its methods appear to add a restriction. We seem to have implicitly assumed there will be only one interested browser/resolver for a given name at a time, which is why `Publisher` provide the "Subscribe" methods.

Thread's current use of mDNS mostly follows this pattern (e.g., the Discovery Proxy shares the same resolver/browser for multiple queries of the same DNS name), but this behavior may not be guaranteed in the future.

I know addressing this may be a much bigger change. On one hand, I think we can assume this one-browser/resolver limitation for now. On the other hand, I feel this may lead to some subtle bugs or issues in the future.

abtink (commented):
Thanks Li. Some new questions/suggestions below:

src/host/posix/dnssd.hpp
Irving-cl: I updated the implementation and used a vector to hold all dnssd browse/resolvers. I also defined the new struct that stores both the `mCallback` and `mInfraIfIndex` so that we can track them simply.

In the meantime, I introduced an abstraction layer for this struct to make the code in this class can work for not only RCP design, but for NCP design. Because for RCP design, we can simply compare the otPlatDnssdCallback to identify the request. However, for NCP, we will still NCP send the pointer values of the callback to the host and when the host gets an answer, it sends the pointer value of the callback as well as the result back to NCP. But the callback invoked on host is actually the same one.

So I introduced `DnssdCallback` and two child classes `OtDnssdCallback` and `StdDnssdCallback`. They will have a different comparison implementations so that the find/add/remove code here can work for both NCP&RCP design.

src/host/posix/dnssd.cpp
Irving-cl: This is indeed a problem. The current model doesn't work for such case. If we doesn't change `Publisher`, then we need to add cache in this module which I think is too complex I think we can change the interface and implementation of `Publisher`to resolve this issue later. Thoughts?

abtink: Yes. We either need to make changes to the `Publisher` methods or we can directly interact with mDNSResponder/Avahi APIs in this module?

Irving-cl: That sounds good! I'll do it later.

PR #2963, opened 2025-07-23
[cmake] enable `OTBR_DNS_UPSTREAM_QUERY` by default

jwhui:


etc/cmake/options.cmake
```diff
@@ -171,7 +171,12 @@ else()
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_VENDOR_INFRA_LINK_SELECT=0)
 endif()
 
-option(OTBR_DNS_UPSTREAM_QUERY "Allow sending DNS queries to upstream" OFF)
+set(OTBR_DNS_UPSTREAM_QUERY_DEFAULT OFF)
+if (OTBR_NAT64)
```
abtink: Why check `OTBR_NAT64` here? I feel they are somehow unrelated feature "nat64" vs "DNS upstream query"?

jwhui: Good point. I changed it to depend on `OTBR_BORDER_ROUTING`.

PR #2972, opened 2025-07-28
[utils] move dns_utils from common to utils

Irving-cl:
This PR merges the code in `common/dns_utils` into `utils/dns_utils`.

There are two `dns_utils` sources in the code which is confusing. `dns_utils` is a more 'utils' thing so it should be put under the utils directory.


src/utils/dns_utils.cpp
```diff
@@ -36,6 +36,114 @@ namespace otbr {
...
+    {
+        transportPos = fullName.rfind("._tcp.");
+    }
+
+    if (transportPos == std::string::npos)
+    {
+        // host.domain or domain
+        size_t dotPos = fullName.find_first_of('.');
+
+        assert(dotPos != std::string::npos);
```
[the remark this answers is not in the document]
Irving-cl: This function doesn't return an error so `assert` is used. This PR is a refactor and won't change the behavior.

src/utils/dns_utils.cpp
```diff
@@ -36,6 +36,114 @@ namespace otbr {
...
+    {
+        fullName += '.';
+    }
+
+    transportPos = fullName.rfind("._udp.");
+
+    if (transportPos == std::string::npos)
+    {
+        transportPos = fullName.rfind("._tcp.");
+    }
```
abtink, line 60: I have never been a fan of this `SplitFullDnsName` helper method.

Its logic is brittle because it infers the name type by parsing and checking for specific substrings (e.g., `_udp` or `_tcp`).  A more generic implementation that doesn't rely on such methods would be more robust.

abtink: This is unrelated to this PR (which just refactors this method) 😄

Irving-cl: Make senses! I can update this later.

Irving-cl:
@jwhui This PR is ready to merge :)

jwhui (approved):

PR #2975, opened 2025-07-30
[utils] move telemetry retriever code from utils to host directory

Irving-cl:
This PR moves the location of the telemetry_retriever_border_agent from `utils` to `host` directory.

I just realized that it doesn't make sense to put such sources under `utils` directory. As I see, the code under `utils` directory shouldn't be closely binded with any features. That says, these code shouldn't be guarded with any feature macros. I put the sources under `host` because currently they are used by the thread_helper which is located under the `host` directory.

Looking at the bigger picture, I'm trying to make the lib `utils` more general and can be directly linked by other components (with different configurations) in this project. This will enable us to write more flexible unit tests. For example, we write some feature code and test code for this feature. We can make a small test target only having the feature code and test code and link it with `utils` lib (and probably the `common` lib).


src/host/CMakeLists.txt
```diff
@@ -41,6 +41,8 @@ add_library(otbr-host
     thread_helper.hpp
     thread_host.cpp
     thread_host.hpp
+    telemetry_retriever_border_agent.cpp
+    telemetry_retriever_border_agent.hpp
```
jwhui: Keep alphabetical order?

Irving-cl: Updated!

src/host/telemetry_retriever_border_agent.hpp
```diff
@@ -34,10 +34,11 @@
 #include <openthread/history_tracker.h>
 #include <openthread/instance.h>
 
+#include "common/logging.hpp"
```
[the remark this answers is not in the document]
Irving-cl: This is to resolve a temporary include issue on MacOS.

jwhui (approved):

PR #2990, opened 2025-08-04
[format] upgrade clang-format to 19

Irving-cl:
This PR updates the clang-format version from 16 to 19.

We just updated the version from 14 to 16. But very unfortunately clang-format-16 was just deprecated recently. So this time we update the version to the latest available one. There are newer releases than 19 (like 20) but as I tested, `sudo apt-get install -y clang-format-19` can work while 20 doesn't work.


superwhd (approved):
LGTM 👍

tests/scripts/bootstrap.sh
```diff
@@ -118,7 +118,7 @@ case "$(uname)" in
         fi
 
         if [ "$BUILD_TARGET" == pretty-check ]; then
-            sudo apt-get install -y clang-format-16 shellcheck
+            sudo apt-get install -y clang-format-19 shellcheck
```
abtink, line 121: Can you please update `install_packages_brew()` to also check and install `19` instead of `16`?

abtink: Please ignore my comment. This PR is for `ot-br-posix` (where we do not install clang-format under `brew`), not the `openthread` repo as I had mistakenly assumed.

Irving-cl: Yeah I'll update brew when updating that in openthread :)

Irving-cl:
@jwhui This PR is ready.

PR #3003, opened 2025-08-08
[mdns] fix ServiceInstanceResolutions leak in mdnssd

mtilchen:
### Background 

While testing TREL in our BR implementation we started seeing forced exits due to "too many open files".
Closer investigation showed a file descriptor leak that was growing by `n` descriptors every five minutes
where `n` is equal to the number of discoverable `_trel._udp` services on the network. These file descriptors
were found to be open sockets to `mdnsResponder` at `/var/run/mdnsd`. 

The root cause is due to accumulation of `ServiceInstanceResolution` objects in the `mResolvingInstances`
vector in `PublisherMDnsSd::ServiceSubscription` that are never removed. As a result the `DNSServiceRef`
handles owned by the `ServiceInstanceResolution` are never deallocated. I believe this issue may have
been introduced in [this PR](https://github.com/openthread/ot-br-posix/pull/2148) when the
`RemoveInstanceResolution` was taken out. In this fix we simply take ownership of the object after removing
it from the vector and keep it in scope so downstream handlers can reference it if needed.

### Steps to reproduce

- Linux system with `dns-sd` and `mdnsResponder` on board.
- At least one other Border Router on the network with TREL enabled (advertising `_trel._udp`)
- Relevant CMake config used:
    ```
    -DOTBR_TREL:BOOL=ON
   -DOTBR_NAT64:BOOL=ON
   -DOTBR_DUA_ROUTING:BOOL=ON
   -DOTBR_DHCP6_PD:BOOL=ON
   -DOTBR_MDNS:STRING=mDNSResponder
  ```
 - Every 5 minutes:
   ```
   ls /proc/$(pidof otbr-agent)/fd | wc -l
   ....
   netstat | grep /var/run/mdnsd | wc -l
   ```
- Observe counts from above commands increasing by the number of discoverable `_trel._udp` services.

### Testing
I and some colleagues have tested this patch on several systems and have found it prevents the leak.

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -50,6 +50,7 @@
 #include "common/logging.hpp"
 #include "common/time.hpp"
 #include "utils/dns_utils.hpp"
+#include "utils/string_utils.hpp"
 
 namespace otbr {
 
@@ -1194,6 +1195,7 @@ void PublisherMDnsSd::ServiceSubscription::HandleBrowseResult(DNSServiceRef     
     else
     {
         mPublisher.OnServiceRemoved(aInterfaceIndex, mType, aInstanceName);
+        Remove(aInterfaceIndex, aInstanceName, aType, aDomain);
     }
 
 exit:
@@ -1218,9 +1220,31 @@ void PublisherMDnsSd::ServiceSubscription::Resolve(uint32_t           aInterface
                                                    const std::string &aType,
                                                    const std::string &aDomain)
 {
+    Remove(aInterfaceIndex, aInstanceName, aType, aDomain);
+
     mResolvingInstances.push_back(
         std::make_shared<ServiceInstanceResolution>(*this, aInstanceName, aType, aDomain, aInterfaceIndex));
     mResolvingInstances.back()->Resolve();
+}
+
+void PublisherMDnsSd::ServiceSubscription::Remove(uint32_t           aInterfaceIndex,
+                                                  const std::string &aInstanceName,
+                                                  const std::string &aType,
+                                                  const std::string &aDomain)
+{
+    otbrLogDebug("Removing service instance resolution for %s %s inf %u ", aInstanceName.c_str(), aType.c_str(),
+                 aInterfaceIndex);
+
+    const auto it = std::find_if(mResolvingInstances.begin(), mResolvingInstances.end(),
+                                 [aInterfaceIndex, &aInstanceName, &aType,
+                                  &aDomain](const std::shared_ptr<ServiceInstanceResolution> &aServiceResolution) {
+                                     return aServiceResolution->Matches(aInterfaceIndex, aInstanceName, aType, aDomain);
+                                 });
+
+    if (it != mResolvingInstances.end())
+    {
+        mResolvingInstances.erase(it);
+    }
 }
 
 void PublisherMDnsSd::ServiceSubscription::UpdateAll(MainloopContext &aMainloop) const
@@ -1242,6 +1266,16 @@ void PublisherMDnsSd::ServiceSubscription::ProcessAll(const MainloopContext     
     {
         instance->Process(aMainloop, aReadyServices);
     }
+}
+
+bool PublisherMDnsSd::ServiceInstanceResolution::Matches(uint32_t           aInterfaceIndex,
+                                                         const std::string &aInstanceName,
+                                                         const std::string &aType,
+                                                         const std::string &aDomain) const
+{
+    return (mNetifIndex == aInterfaceIndex) && otbr::StringUtils::EqualCaseInsensitive(mInstanceName, aInstanceName) &&
+           otbr::StringUtils::EqualCaseInsensitive(mType, aType) &&
+           otbr::StringUtils::EqualCaseInsensitive(mDomain, aDomain);
 }
 
 void PublisherMDnsSd::ServiceInstanceResolution::Release(void)
--- a/src/mdns/mdns_mdnssd.hpp
+++ b/src/mdns/mdns_mdnssd.hpp
@@ -230,6 +230,10 @@ private:
         {
         }
 
+        bool      Matches(uint32_t           aInterfaceIndex,
+                          const std::string &aInstanceName,
+                          const std::string &aType,
+                          const std::string &aDomain) const;
         void      Release(void);
         void      Resolve(void);
         otbrError GetAddrInfo(uint32_t aInterfaceIndex);
@@ -293,6 +297,10 @@ private:
                      const std::string &aInstanceName,
                      const std::string &aType,
                      const std::string &aDomain);
+        void Remove(uint32_t           aNetifIndex,
+                    const std::string &aInstanceName,
+                    const std::string &aType,
+                    const std::string &aDomain);
         void UpdateAll(MainloopContext &aMainloop) const;
         void ProcessAll(const MainloopContext &aMainloop, std::vector<DNSServiceRef> &aReadyServices) const;
 


src/mdns/mdns_mdnssd.cpp
```diff
@@ -1240,6 +1240,21 @@ void PublisherMDnsSd::ServiceSubscription::ProcessAll(const MainloopContext
...
+{
+    for (auto it = mResolvingInstances.begin(); it != mResolvingInstances.end(); ++it)
+    {
+        if (it->get() == aResolution)
+        {
+            auto ptr = std::move(*it);
+            mResolvingInstances.erase(it);
+            return ptr;
+        }
+    }
```
Irving-cl:
```suggestion
     auto iter = std::find(mResolvingInstances.begin(), mResolvingInstances.end(), aResolution);
     if (iter != mResolvingInstances.end())
     {
            auto ptr = std::move(*iter);
            mResolvingInstances.erase(iter);
            return ptr;
     }
     
     return nullptr;
```
Use std::find? (No strong preference)

Irving-cl (approved):
LGTM from me. 

Need review from @superwhd

src/mdns/mdns_mdnssd.cpp
superwhd: IIUC, it may not be able to call `==` operator between a raw pointer and a smart pointer.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1240,6 +1240,21 @@ void PublisherMDnsSd::ServiceSubscription::ProcessAll(const MainloopContext
...
+std::shared_ptr<PublisherMDnsSd::ServiceInstanceResolution>
+PublisherMDnsSd::ServiceSubscription::ExtractResolution(const ServiceInstanceResolution * aResolution)
+{
+    for (auto it = mResolvingInstances.begin(); it != mResolvingInstances.end(); ++it)
+    {
+        if (it->get() == aResolution)
+        {
+            auto ptr = std::move(*it);
+            mResolvingInstances.erase(it);
+            return ptr;
```
superwhd: nit
```suggestion
            auto ptr = std::move(*it);

            mResolvingInstances.erase(it);
            return ptr;
```

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1412,9 +1427,19 @@ void PublisherMDnsSd::ServiceInstanceResolution::HandleGetAddrInfoResult(DNSServ
                                                                 aInstance->GetAddrInfo(aInterfaceIndex);
                                                             });
     }
-    else if ((!mInstanceInfo.mAddresses.empty() && !moreComing) || aErrorCode != kDNSServiceErr_NoError)
+    else
     {
-        FinishResolution();
+        std::shared_ptr<ServiceInstanceResolution> resolution {};
+        if (!moreComing)
```
superwhd: nit
```suggestion
        std::shared_ptr<ServiceInstanceResolution> resolution {};

        if (!moreComing)
```

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1412,9 +1427,19 @@ void PublisherMDnsSd::ServiceInstanceResolution::HandleGetAddrInfoResult(DNSServ
...
+        {
+            // Remove ourself from the vector of open resolutions but stay in scope long enough for handlers.
+            // mSubscription might be freed after invoking FinishResolution so this needs to be done here.
+            resolution = mSubscription->ExtractResolution(this);
+        }
+        if ((!mInstanceInfo.mAddresses.empty() && !moreComing) || aErrorCode != kDNSServiceErr_NoError)
+        {
+            FinishResolution();
```
superwhd: I wonder if you also need to change the code here similarly: https://github.com/openthread/ot-br-posix/blob/ca3e7e6d90db1e9f83dc5c74aa084db856b89386/src/mdns/mdns_mdnssd.cpp#L1326

mtilchen: I was also suspicious of this. However, a few lines up you will see a call to `DeallocateServiceRef()`. During the next phase (address resolution) the `DNSServiceRef` gets allocated again, which was the leaking allocation. I believe this code can remain as-is.

mtilchen:
@abtink (and others) thank you for the guidance and context in your reviews. After a fresh look I think I have a much better handle on the design now. I updated my changes to keep the object graph intact in favor of deallocating the `DNSServiceRef` after it is no longer needed after address resolution and guarding against duplicate entries in the `mResolvingInstances` vector.

> We also need to handle cases where an instance is removed by stopping its service resolution and clearing it from mResolvingInstances. This logic appears to be missing in the code today.

I elected to not implement it this way (going with a guard in the `Resolve` method solely) because I was not sure if there were other things that might be interested in knowing that a discovered service was in a "removed" state at some point in the future. If and when the browse callback adds that same service again the existing object is updated with the new information. If folks are in favor of removing the related `ServiceInstanceResolution` object from  `mResolvingInstances` immediately I can add that as well.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1218,9 +1218,31 @@ void PublisherMDnsSd::ServiceSubscription::Resolve(uint32_t           aInterface
...
+        mResolvingInstances.push_back(
+            std::make_shared<ServiceInstanceResolution>(*this, aInstanceName, aType, aDomain, aInterfaceIndex));
+        mResolvingInstances.back()->Resolve();
+    }
+    else
+    {
+        const auto& aResolvingInstance = *it;
+
+        otbrLogDebug("Resolving previously added service %s %s inf %u is_removed %s",
+                      mInstanceName.c_str(), mType.c_str(), aInterfaceIndex,
```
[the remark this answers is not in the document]
mtilchen: fixed in bcd92a89

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1218,11 +1219,64 @@ void PublisherMDnsSd::ServiceSubscription::Resolve(uint32_t           aInterface
...
+
+    if (it == mResolvingInstances.end())
+    {
+        mResolvingInstances.push_back(
+            std::make_shared<ServiceInstanceResolution>(*this, aInstanceName, aType, aDomain, aInterfaceIndex));
+        mResolvingInstances.back()->Resolve();
+    }
+    else
+    {
+        const auto &aResolvingInstance = *it;
```
abtink: rename `aResolvingInstance -> resolvingInstance` (the prefix `a` is used for input args to a method/function)

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1218,11 +1219,64 @@ void PublisherMDnsSd::ServiceSubscription::Resolve(uint32_t           aInterface
...
-    mResolvingInstances.push_back(
-        std::make_shared<ServiceInstanceResolution>(*this, aInstanceName, aType, aDomain, aInterfaceIndex));
-    mResolvingInstances.back()->Resolve();
+    // Look for an existing service that may have been marked as removed to prevent
+    // adding a duplicate ServiceInstanceResolution
+    const auto it = std::find_if(mResolvingInstances.begin(), mResolvingInstances.end(),
+                                 [aInterfaceIndex, &aInstanceName, &aType,
+                                  &aDomain](const std::shared_ptr<ServiceInstanceResolution> &aServiceResolution) {
+                                     return aServiceResolution->mNetifIndex == aInterfaceIndex &&
+                                            aServiceResolution->mInstanceName == aInstanceName &&
```
abtink: Ideally, we should do a case-insensitive name check here for DNS names, but it may be fine to keep it as is for now. I noticed we don't seem to do this in other places in the class, so perhaps this can be fixed or enhanced later in its own PR.

Thoughts?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1218,11 +1219,64 @@ void PublisherMDnsSd::ServiceSubscription::Resolve(uint32_t           aInterface
...
+
+    bool removed = false;
+
+    std::remove_if(mResolvingInstances.begin(), mResolvingInstances.end(),
+                   [&removed, aInterfaceIndex, &aInstanceName,
+                    &aType, &aDomain](const std::shared_ptr<ServiceInstanceResolution> &aServiceResolution) {
+                       const auto found = aServiceResolution->mNetifIndex == aInterfaceIndex &&
+                                          aServiceResolution->mInstanceName == aInstanceName &&
+                                          aServiceResolution->mType == aType &&
+                                          aServiceResolution->mDomain == aDomain;
```
abtink: Can consider adding a method (e.g. `Matches()`) on `ServiceInstanceResolution` class to perform this check. This way we can use it here here in `Remove()` and also in `Resolve()`.

```c++
bool PublisherMDnsSd::ServiceInstanceResolution::Matches(uint32_t           aInterfaceIndex,
                                                         const std::string &aInstanceName,
                                                         const std::string &aType,
                                                         const std::string &aDomain) const
{
    return (mNetifIndex == aInterfaceIndex) && 
           otbr::StringUtils::EqualCaseInsensitive(mInstanceName, aInstanceName) &&
           otbr::StringUtils::EqualCaseInsensitive(mType, aType) &&
           otbr::StringUtils::EqualCaseInsensitive(mDomain, aDomain);
}
```

abtink: e.g. in `Resolve()` we can do:
```c++
    const auto it = std::find_if(mResolvingInstances.begin(), mResolvingInstances.end(),
                                 [aInterfaceIndex, &aInstanceName, &aType,
                                  &aDomain](const std::shared_ptr<ServiceInstanceResolution> &aServiceResolution) {
                                     return aServiceResolution->Matches(aInterfaceIndex, aInstanceName, aType, aDomain);
                                 });
```

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1416,9 +1470,16 @@ void PublisherMDnsSd::ServiceInstanceResolution::HandleGetAddrInfoResult(DNSServ
                                                                 aInstance->GetAddrInfo(aInterfaceIndex);
                                                             });
     }
-    else if ((!mInstanceInfo.mAddresses.empty() && !moreComing) || aErrorCode != kDNSServiceErr_NoError)
+    else
     {
-        FinishResolution();
+        if (!moreComing || aErrorCode != kDNSServiceErr_NoError)
+        {
+            DeallocateServiceRef();
```
abtink: Do we still need to make this change here? I think we can keep the previous implementation.

With the new code in `Resolve()`, we now detect and handle duplicates. `Release()` is called first, which deallocates the service reference before `Resolve()` is called, correct?

abtink (commented):
Thanks @mtilchen  for all the updates. Looks good overall.

Some suggestions below:

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1218,11 +1219,64 @@ void PublisherMDnsSd::ServiceSubscription::Resolve(uint32_t           aInterface
...
+    const auto it = std::find_if(mResolvingInstances.begin(), mResolvingInstances.end(),
+                                 [aInterfaceIndex, &aInstanceName, &aType,
+                                  &aDomain](const std::shared_ptr<ServiceInstanceResolution> &aServiceResolution) {
+                                     return aServiceResolution->mNetifIndex == aInterfaceIndex &&
+                                            aServiceResolution->mInstanceName == aInstanceName &&
+                                            aServiceResolution->mType == aType &&
+                                            aServiceResolution->mDomain == aDomain;
+                                 });
+
+    if (it == mResolvingInstances.end())
```
abtink: I have an idea that may make the PR change even smaller.

In `Resolve()`, we could first call `Remove(aInterfaceIndex, aInstance, aType, ...)` to clear any previously matching entries (if any). After that, we can proceed as before by starting a new `ServiceInstanceResolution`:

```c++
    mResolvingInstances.push_back(
        std::make_shared<ServiceInstanceResolution>(*this, aInstanceName, aType, aDomain, aInterfaceIndex));
    mResolvingInstances.back()->Resolve();
```

We would then need to change `Remove()` so that it no longer logs an error or warning if the entry to be removed does not exist. This would prevent extra logs when it is called from `Resolve()`.

Essentially, `Remove()` would attempt to remove a matching `ServiceInstanceResolution` if found; otherwise, it would do nothing.

What do you think?

abtink (commented):
Another idea:

src/mdns/mdns_mdnssd.cpp
mtilchen: Yes, I think that makes it more succinct and straightforward. I will update it to use this approach.

src/mdns/mdns_mdnssd.cpp
mtilchen: Thanks. I was actually going to ask about case-sensitivity so I am glad you brought it up. I will incorporate this.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -213,8 +214,8 @@ static const char *DNSErrorToString(DNSServiceErrorType aError)
     case kDNSServiceErr_DefunctConnection:
         return "Defunct Connection";
 
-    case kDNSServiceErr_StaleData:
-        return "Stale Data";
+    // case kDNSServiceErr_StaleData:
+        // return "Stale Data";
```
superwhd: I wonder why this is commented out? Will it work if we just keep it?

mtilchen: This was mistakenly include during a rebase. Thanks for pointing it out.

src/mdns/mdns_mdnssd.cpp
mtilchen: It does deallocate the service reference but the call to `Resolve()` re-allocates it. I think we should leave it in because otherwise the underlying connection to `mdnsResponder` stays open until the `ServiceInstanceResolution` instance is removed. Said instance may stay in the vector substantially longer than the point in time where the address resolution completes. Unless I am missing something if `moreComing` is `false` I don't see any reason to keep the service reference open. Freeing the resource at the specific time it is no longer needed seems like the way to go?

abtink: > otherwise the underlying connection to mdnsResponder stays open until the ServiceInstanceResolution instance is removed

I believe keeping the service reference allocated and the connection open is the correct and desired behavior.

The `ServiceRef` is effectively owned by the `ServiceInstanceResolution` instance, and they go hand-in-hand. As long as the `ServiceInstanceResolution` object exists, it makes sense for it to maintain its `ServiceRef`.

We want a **long-lived service resolver** so that we are notified of any subsequent changes. For example, if the device advertising this service later changes its TXT data record, we want to receive that update.

The `moreComing` flag only indicates if the callback will be invoked again immediately; it does not mean there won't be more answers later. Can refer to the documentation here: [mDNSResponder/dns_sd.h#L243-L257](https://github.com/apple-oss-distributions/mDNSResponder/blob/d89f8d1d0e001b810d6c055aa2a57b768bcf9aa2/mDNSShared/dns_sd.h#L243-L257)

Additionally, the `FinishResolution()` method does not end the resolution process itself but merely signals the completion of the current discovery round. Perhaps its name is not ideal for conveying its actual function.

mtilchen: Thank you. This makes sense and I suspect it is probably why this test is failing (not picking up the address change):
https://github.com/mtilchen/ot-br-posix/blob/fb76359fe3a2ded0d0b38656c8dccf7c4074a5ee/tests/gtest/test_mdns_subscribe.cpp#L322-L324

Given the intent of having a long-lived resolver that will pick up changes over time is it a problem that `GetAddrInfo()` reuses the same `DNSServiceRef` as the `Resolve()` method? The former using `DNSServiceGetAddrInfo` to initialize the service ref and the latter using `DNSServiceResolve`? In other words, will this only allow us to observe address changes over time (where we would also like to observe TXT record changes like you point out) ? I read the docs in `dns_sd.h` and it was not 100% clear what to expect with regard to changes in the service beyond just address changes.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -1218,11 +1220,33 @@ void PublisherMDnsSd::ServiceSubscription::Resolve(uint32_t           aInterface
...
         std::make_shared<ServiceInstanceResolution>(*this, aInstanceName, aType, aDomain, aInterfaceIndex));
     mResolvingInstances.back()->Resolve();
 }
 
+void PublisherMDnsSd::ServiceSubscription::Remove(uint32_t           aInterfaceIndex,
+                                                  const std::string &aInstanceName,
+                                                  const std::string &aType,
+                                                  const std::string &aDomain)
+{
+    otbrLogDebug("Removing service instance resolution for %s %s inf %u ", aInstanceName.c_str(), aType.c_str(),
```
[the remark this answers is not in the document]
mtilchen: Following convention in other logs here.

src/mdns/mdns_mdnssd.cpp
abtink: Yes. Indeed. The reusing of `mServiceRef` does not seem correct. We should  probably use a separate ref for the host address resolution process.

abtink (approved):
Thanks for making all the changes. Looks great. 👍

src/mdns/mdns_mdnssd.cpp
mtilchen: I will look at addressing this in a follow up PR as bandwidth permits. 👍🏻

abtink: Thanks

PR #3004, opened 2025-08-08
Enable Multi AIL Detection

jdswensen:
These changes perform the following:
- Allow the multi AIL detection feature to be enabled/disabled from the command line while compiling
- Adds a DBus API to fetch the Multi AIL detection status

Example of detection being enabled and fetching the property:
```
dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:MultiAilDetected
method return time=1754693794.210054 sender=:1.4271 -> destination=:1.5952 serial=12 reply_serial=2
   variant       boolean true
```

Example of detection being disabled and fetching the property:
```
dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:MultiAilDetected
Error io.openthread.Error.NotImplemented: io.openthread.Error.NotImplemented
```

This change depends on the PR in Openthread here: https://github.com/openthread/openthread/pull/11800


src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1486,6 +1488,25 @@ otError DBusThreadObjectRcp::GetTrelInfoHandler(DBusMessageIter &aIter)
...
 }
 
+otError DBusThreadObjectRcp::GetMultiAilDetectedHandler(DBusMessageIter &aIter)
+{
+#if OTBR_ENABLE_MULTI_AIL
+    auto           instance = mHost.GetInstance();
+    otError        error    = OT_ERROR_NONE;
+    bool otMultiAilDetected = otBorderRoutingIsMultiAilDetected(instance);
+
+    SuccessOrExit(DBusMessageEncodeToVariant(&aIter, otMultiAilDetected), error = OT_ERROR_INVALID_ARGS);
```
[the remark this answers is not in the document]
jdswensen: I think 1 might be a false positive. I've addressed 2-4.

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1486,6 +1488,22 @@ otError DBusThreadObjectRcp::GetTrelInfoHandler(DBusMessageIter &aIter)
 #endif // OTBR_ENABLE_TREL
 }
 
+otError DBusThreadObjectRcp::GetMultiAilDetectedHandler(DBusMessageIter &aIter)
+{
+#if OTBR_ENABLE_MULTI_AIL
+    otError        error    = OT_ERROR_NONE;
+    const bool isDetected = otBorderRoutingIsMultiAilDetected(mHost.GetInstance());
```
jwhui: To fix pretty check:
```suggestion
    otError    error      = OT_ERROR_NONE;
    const bool isDetected = otBorderRoutingIsMultiAilDetected(mHost.GetInstance());
```

etc/cmake/options.cmake
```diff
@@ -250,3 +250,13 @@ if (OTBR_DNSSD_PLAT)
...
 endif()
+
+set(OTBR_MULTI_AIL_DEFAULT OFF)
+if (OTBR_BORDER_ROUTING)
+    set(OTBR_MULTI_AIL_DEFAULT ON)
+endif()
+
+option(OTBR_MULTI_AIL "Enable multi-ail detection support" ${OTBR_MULTI_AIL_DEFAULT})
+if(OTBR_MULTI_AIL)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_MULTI_AIL=1)
```
superwhd, line 261: nit: I saw that other options set the macro to 0 when the according CMake option is off. 

I don't have a strong opinion on this.

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1486,6 +1488,22 @@ otError DBusThreadObjectRcp::GetTrelInfoHandler(DBusMessageIter &aIter)
 #endif // OTBR_ENABLE_TREL
 }
 
+otError DBusThreadObjectRcp::GetMultiAilDetectedHandler(DBusMessageIter &aIter)
+{
+#if OTBR_ENABLE_MULTI_AIL
```
superwhd, line 1493: Is it possible that we only keep the branch where `OTBR_ENABLE_MULTI_AIL` is enabled? 

IMHO the code is more maintainable if we have fewer build time branches. cc @bukepo

superwhd (commented):
LGTM overall. Thanks!

src/dbus/server/dbus_thread_object_rcp.cpp
jdswensen: In that case, I think there could be a potential configuration/compile issue if the feature was ever disabled in openthread. We would need to always force the feature to be included by default with every ot-br-posix build. That wouldn't necessarily be bad in my opinion, but I was just following the existing pattern for features that can be optionally disabled in OT. I'm happy to make any changes if others feel there is a better pattern to follow.

src/dbus/server/dbus_thread_object_rcp.cpp
jdswensen: Done!

etc/cmake/options.cmake
jdswensen: I added the else branch of the statement.

jwhui (approved):

PR #3006, opened 2025-08-11
[telemetry] refactor telemetry code

Irving-cl:
This PR refactors the telemetry code.

The PR moves the telemetry code in `ThreadHelper` into a seperate modole `Telemetry` for these reasons:
1. Centralizes the code for Telemetry uploading.
2. Decouple `Telemetry` with `ThreadHelper` because `Telemetry` code depends on `otbr-proto` while many modules depend on `ThreadHelper` but not `Telemetry`. Such dependencies makes these modules depend on `otbr-proto` as well, which is not ideal.
3. We can gradually remove `ThreadHelper`.


src/host/telemetry/telemetry.cpp
```diff
@@ -0,0 +1,987 @@
...
+            auto otTrelCounters = otTrelGetCounters(mInstance);
+            auto trelCounters   = trelInfo->mutable_counters();
+
+            trelInfo->set_is_trel_enabled(otTrelIsEnabled(mInstance));
+            trelInfo->set_num_trel_peers(otTrelGetNumberOfPeers(mInstance));
+
+            trelCounters->set_trel_tx_packets(otTrelCounters->mTxPackets);
+            trelCounters->set_trel_tx_bytes(otTrelCounters->mTxBytes);
+            trelCounters->set_trel_tx_packets_failed(otTrelCounters->mTxFailure);
+            trelCounters->set_tre_rx_packets(otTrelCounters->mRxPackets);
```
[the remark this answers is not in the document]
Irving-cl: This is indeed a typo but was defined in a wrong way in an early time. We cannot change it because the proto has been in use.

src/host/telemetry/telemetry.cpp
```diff
@@ -0,0 +1,987 @@
...
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "telemetry.hpp"
```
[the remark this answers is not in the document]
Irving-cl: Added the includes in other lines.

src/host/telemetry/telemetry.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+class TelemetryRetriever
+{
+public:
+    /**
+     * Constructor.
+     *
+     * @param[in] aInstance  A pointer to the OT instance.
+     */
+    explicit TelemetryRetriever(otInstance *aInstance);
```
superwhd, line 54: Since you're passing an OT instance into it, is this specific to RCP?

src/host/telemetry/telemetry_retriever_border_agent.hpp
superwhd, line 1: I may missed quite some background here. What's the point of making this as a separate class/file?

src/host/telemetry/telemetry.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include "openthread-br/config.h"
+
+#if OTBR_ENABLE_TELEMETRY_DATA_API
+
+#include <openthread/instance.h>
+
+#include "mdns/mdns.hpp"
+#include "proto/thread_telemetry.pb.h"
+
+#include "telemetry_retriever_border_agent.hpp"
```
superwhd: Shall we include the full path?

src/host/telemetry/telemetry.hpp
```diff
@@ -0,0 +1,99 @@
...
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef OTBR_TELEMETRY_HPP_
+#define OTBR_TELEMETRY_HPP_
```
superwhd: Correct me if I'm wrong. 
```suggestion
#ifndef OTBR_AGENT_TELEMETRY_HPP_
#define OTBR_AGENT_TELEMETRY_HPP_
```

src/host/telemetry/telemetry_retriever_border_agent.hpp
```diff
@@ -26,6 +26,11 @@
  *    POSSIBILITY OF SUCH DAMAGE.
  */
 
+#ifndef OTBR_TELEMETRY_RETRIEVER_BORDER_AGENT_HPP_
+#define OTBR_TELEMETRY_RETRIEVER_BORDER_AGENT_HPP_
```
superwhd: Correct me if I'm wrong.
```suggestion
#ifndef OTBR_AGENT_TELEMETRY_RETRIEVER_BORDER_AGENT_HPP_
#define OTBR_AGENT_TELEMETRY_RETRIEVER_BORDER_AGENT_HPP_
```

src/host/telemetry/telemetry.hpp
Irving-cl: Yes, currently it is.

src/host/telemetry/telemetry_retriever_border_agent.hpp
Irving-cl: The history is: when I first add it, I want the code to be a standalone part so that I can write unit test for so I didn't add the code in `thread_helper`. That way we will need to depend on `ThreadHelper` when writing unit test which is not ideal.

For our future direction, I think we can break the telemetry uploading code into small pieces (for those having complicated uploading logic) so that the owners can better maintain them.

morningboata (commented):
Thanks for refactor the code to make it more module like. 

Just check, this will not impact the DBus interface behavior and its tests (https://github.com/openthread/ot-br-posix/blob/main/tests/dbus/test_dbus_client.cpp), right?

Irving-cl:
> Thanks for refactor the code to make it more module like.
> 
> Just check, this will not impact the DBus interface behavior and its tests (https://github.com/openthread/ot-br-posix/blob/main/tests/dbus/test_dbus_client.cpp), right?

@morningboata 

This won't impact the DBus interface behavior and its tests. The DBus client actually doesn't rely the telemetry code. It only depends on the proto which isn't changed in this PR.

src/host/telemetry/telemetry.hpp
Irving-cl: Updated.

src/host/telemetry/telemetry.hpp
superwhd: What's the plan of expanding it to cover NCP?

superwhd (approved):
LGTM 👍 

One minor question.

src/host/telemetry/telemetry.hpp
Irving-cl: No plan yet. I considered it before. But the implementation in NCP mode will be very different. (We can not use OT APIs here to fetch the telemetry) We will probably use a different class to fetch the telemetry from NCP.

src/host/telemetry/telemetry.cpp
```diff
@@ -0,0 +1,1004 @@
...
+        }
+
+        int8_t rssi = otPlatRadioGetRssi(mInstance);
+
+        wpanTopoFull->set_instant_rssi(rssi);
+
+        const otExtendedPanId *extPanId = otThreadGetExtendedPanId(mInstance);
+        uint64_t               extPanIdVal;
+
+        extPanIdVal = ConvertOpenThreadUint64(extPanId->m8);
```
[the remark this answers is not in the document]
Irving-cl: It's in `code_utils.hpp` and it is included.

src/host/telemetry/telemetry.cpp
```diff
@@ -0,0 +1,1004 @@
...
+    otExternalRouteConfig config;
+
+    while (otNetDataGetNextRoute(mInstance, &iterator, &config) == OT_ERROR_NONE)
+    {
+        if (!config.mStable || config.mRloc16 != rloc16)
+        {
+            continue;
+        }
+
+        prefix.Set(config.mPrefix);
```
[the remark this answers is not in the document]
Irving-cl: Included `common/type.hpp`.

src/host/telemetry/telemetry.hpp
```diff
@@ -0,0 +1,99 @@
...
+     * This method populates the telemetry data with best effort. The best effort means, for a given
+     * telemetry, if its retrieval has error, it is left unpopulated and the process continues to
+     * retrieve the remaining telemetries instead of the immediately return. The error code
+     * OT_ERRROR_FAILED will be returned if there is one or more error(s) happened in the process.
+     *
+     * @param[in] aPublisher     The Mdns::Publisher to provide MDNS telemetry if it is not `nullptr`.
+     * @param[in] telemetryData  The telemetry data to be populated.
+     *
+     * @retval OTBR_ERROR_NONE  There is no error happened in the process.
+     * @retval OT_ERRROR_FAILED There is one or more error(s) happened in the process.
```
jwhui: Are we really mixing return types here?

Irving-cl: No, it uses otError. I've fixed it.

PR #3012, opened 2025-08-13
[scripts] introduce systemd-networkd for network management

yangsong-cnyn:
This PR introduces support for `systemd-networkd` to manage the host's network configuration. This is necessary to support newer OS versions like Debian 12 on reference devices, where `systemd-networkd` is a common network manager.

A new script, `script/_systemd_networkd`, is added. It is controlled by the `SYSTEMD_NETWORKD` build flag.

The script also Includes a check for `REFERENCE_DEVICE` as we enable OTBR's built-in DHCPv6 PD client on reference device. If OTBR's built-in DHCPv6 PD client is active, it disables `systemd-networkd`'s client to prevent port conflicts.


script/_systemd_networkd
```diff
@@ -0,0 +1,121 @@
...
+[Match]
+Name=${interface}
+
+[Network]
+DHCP=yes
+IPv6AcceptRA=yes
+EOF
+        # If OpenThread's built-in DHCPv6 PD client is enabled, we must
+        # disable systemd-networkd's client to avoid port conflicts.
+        if with OT_BORDER_ROUTING_DHCP6_PD_CLIENT; then
```
superwhd: I doubt if this will work as what you expect. The `OT_BORDER_ROUTING_DHCP6_PD_CLIENT` is a CMake build option while you're checking it as a variable in bash.

script/_systemd_networkd
```diff
@@ -0,0 +1,121 @@
...
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+ETH_INTERFACE="eth0"
+ETH_NETWORK_FILE="/etc/systemd/network/10-${ETH_INTERFACE}.network"
+
+WIFI_INTERFACE="wlan0"
+WIFI_NETWORK_FILE="/etc/systemd/network/25-${WIFI_INTERFACE}.network"
```
superwhd: What are the numbers 10 and 25 representing?

script/_systemd_networkd
```diff
@@ -0,0 +1,121 @@
...
+
+    # Create the network configuration file for the network interface
+    {
+        cat <<EOF
+[Match]
+Name=${interface}
+
+[Network]
+DHCP=yes
+IPv6AcceptRA=yes
```
superwhd: Could you elaborate why you need this? Will there be anything broken without it?

script/_systemd_networkd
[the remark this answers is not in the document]
superwhd: This is reasonable. Hard coding the network interface name may not be a good idea. Also, I don't see the necessity of manipulating both WiFi and Ethernet interfaces at the same time.

yangsong-cnyn: it's just random selected priority number.

script/_systemd_networkd
yangsong-cnyn: though 'DHCP=yes' and 'IPv6AcceptRA=yes' are default, better to have them here for readability.

superwhd: I think OTBR should touch such settings as little as possible. This is to achieve better compatibility with other software running on that device. 

Unless such settings are proven critical to enable PD, I'd suggest not to change them.

script/_systemd_networkd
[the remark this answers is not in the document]
superwhd: I don't think this is addressed. My point is that you should change the config for the desired infra network interface. IIRC the infra network interface name is available as a bash variable in such scripts.

script/_systemd_networkd
yangsong-cnyn: Sure. removed

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+
+INFRA_IF_NAME="${INFRA_IF_NAME:-eth0}"
```
superwhd: Do you need to create a new variable?

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+    {
+        cat <<EOF
+[Match]
+Name=${interface}
+
+EOF
+        # we enable Openthread's built-in DHCPv6 PD client on reference
+        # device. If the built-in client is enabled, we must disable
+        # systemd-networkd's client to avoid port conflicts.
+        if with REFERENCE_DEVICE; then
```
superwhd: Do you need to guard it here? I think the guard is already done for `SYSTEMD_NETWORKD` at other places so you can make `create_network_file` as simple as creating the file without checking it.

I think we no longer need to guard anything by `REFERENCE_DEVICE` in this PR.

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+
+INFRA_IF_NAME="${INFRA_IF_NAME:-eth0}"
+NETWORK_FILE="/etc/systemd/network/10-${INFRA_IF_NAME}.network"
```
superwhd: Please add a comment about the magic number.

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+    } | sudo tee "${network_file}" >/dev/null
+}
+
+systemd_networkd_install()
+{
+    with SYSTEMD_NETWORKD || return 0
+
+    if ! have systemctl; then
+        echo "This script requires systemctl!"
+        return 0
```
superwhd: I think it should give an error.

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+    sudo systemctl stop NetworkManager || true
+    sudo systemctl disable NetworkManager || true
+
+    sudo systemctl daemon-reload
+
+    create_network_file "${INFRA_IF_NAME}" "${NETWORK_FILE}"
+
+    # Start and enable systemd-networkd
+    sudo systemctl enable systemd-networkd || die "Failed to enable systemd-networkd."
+    sudo systemctl restart systemd-networkd || die "Failed to restart systemd-networkd."
```
superwhd, line 68: Is restart necessary?

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+    # Remove the network configuration file
+    sudo rm -f "${NETWORK_FILE}"
+
+    sudo systemctl daemon-reload
+
+    # Restore original services
+    #
+    # Preferentially restore dhcpcd if it is present on the system.
+    # Otherwise, restore NetworkManager. This avoids enabling two conflicting
+    # network managers.
```
superwhd: Why prefer dhcpcd over NetworkManager?

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+
+INFRA_IF_NAME="${INFRA_IF_NAME:-eth0}"
+NETWORK_FILE="/etc/systemd/network/10-${INFRA_IF_NAME}.network"
+
+create_network_file()
+{
+    local interface="$1"
+    local network_file="$2"
```
superwhd: These may not be necessary given you already have the global variables.

script/_systemd_networkd
```diff
@@ -0,0 +1,115 @@
...
+    sudo systemctl enable systemd-networkd || die "Failed to enable systemd-networkd."
+    sudo systemctl restart systemd-networkd || die "Failed to restart systemd-networkd."
+}
+
+systemd_networkd_uninstall()
+{
+    with SYSTEMD_NETWORKD || return 0
+
+    if ! have systemctl; then
+        echo "This script requires systemctl!"
```
superwhd: I think it should give an error.

script/_systemd_networkd
yangsong-cnyn: I am going to remove dhcpcd stuff, as on new Debian OS, NetworkManager is the standard tool and dhcpcd is no longer a default package.

script/_systemd_networkd.sh
superwhd, line 1: Is this file added by mistake?

script/_systemd_networkd
```diff
@@ -0,0 +1,91 @@
...
+#  POSSIBILITY OF SUCH DAMAGE.
+
+# The '00-' prefix gives this configuration a high priority. For a given
+# interface, systemd-networkd applies the first matching file it finds in
+# lexical order.
+NETWORK_FILE="/etc/systemd/network/00-${INFRA_IF_NAME}.network"
+
+create_network_file()
+{
+    # Disable systemd-networkd's client to avoid port conflicts with Openthread's built-in DHCPv6 PD client
```
superwhd: It's probably better to move this comment into the file you're going to write.

script/_systemd_networkd
```diff
@@ -0,0 +1,91 @@
...
+    # Stop NetworkManager service to let systemd-networkd take over
+    sudo systemctl stop NetworkManager || true
+    sudo systemctl disable NetworkManager || true
+
+    sudo systemctl daemon-reload
+
+    create_network_file
+
+    sudo systemctl enable systemd-networkd || die "Failed to enable systemd-networkd."
+    sudo systemctl start systemd-networkd || die "Failed to start systemd-networkd."
```
superwhd: Could you confirm that if networkd was already running before the `start`, it can use the fresh config at this moment?

script/_systemd_networkd
```diff
@@ -0,0 +1,91 @@
...
+#  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+#  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+
```
superwhd, line 29: Add a summary for this file.

script/_systemd_networkd
```diff
@@ -0,0 +1,91 @@
...
+# interface, systemd-networkd applies the first matching file it finds in
+# lexical order.
+NETWORK_FILE="/etc/systemd/network/00-${INFRA_IF_NAME}.network"
+
+create_network_file()
+{
+    # Disable systemd-networkd's client to avoid port conflicts with Openthread's built-in DHCPv6 PD client
+    sudo tee ${network_file} >/dev/null <<EOF
+[Match]
+Name=${interface}
```
superwhd: Wait... Does this really work? IIUC the lower-case variables are never defined.

script/_systemd_networkd
yangsong-cnyn: OK, better to use restart here. Is documented that networkd should automatically detect and apply new configuration files when they are added to /etc/systemd/network/, but using restart guarantees that it will use the new configuration immediately

script/_systemd_networkd
yangsong-cnyn: summary comment added

script/_systemd_networkd.sh
yangsong-cnyn: yes

script/_systemd_networkd
yangsong-cnyn: It's documented that networkd should automatically detect and apply new configuration files when they are added to /etc/systemd/network/, but using restart guarantees that it will use the new configuration immediately

script/_systemd_networkd
```diff
@@ -0,0 +1,96 @@
...
+#  POSSIBILITY OF SUCH DAMAGE.
+#
+#  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+
+# This script manages systemd-networkd configuration for the reference device
+# on Debian 12. It disables the DHCPv6 client in systemd-networkd to prevent
+# conflicts with OpenThread's DHCPv6 PD client when DHCPv6 PD is
+# used for prefix delegation.
```
superwhd:
```suggestion
# This script manages systemd-networkd configuration for the reference device
# on Debian 12. It disables the DHCPv6 client in systemd-networkd to prevent
# conflicts with OpenThread's DHCPv6 PD client.
```

superwhd (approved):

script/_systemd_networkd
```diff
@@ -0,0 +1,95 @@
...
+[IPv6AcceptRA]
+# Disable systemd-networkd's client to avoid port conflicts with Openthread's built-in DHCPv6 PD client
+DHCPv6Client=no
+EOF
+
+}
+
+systemd_networkd_install()
+{
+    with SYSTEMD_NETWORKD || return 0
```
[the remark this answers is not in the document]
yangsong-cnyn: no need to do so as SYSTEMD_NETWORKD is set only with REFERENCE_DEVICE already

librasungirl (approved):

script/_systemd_networkd
```diff
@@ -0,0 +1,91 @@
...
+NETWORK_FILE="/etc/systemd/network/00-${INFRA_IF_NAME}.network"
+
+create_network_file()
+{
+    sudo tee "${NETWORK_FILE}" >/dev/null <<EOF
+[Match]
+Name=${INFRA_IF_NAME}
+
+[IPv6AcceptRA]
+# Disable systemd-networkd's client to avoid port conflicts with OpenThread's built-in DHCPv6 PD client
```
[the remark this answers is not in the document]
yangsong-cnyn: no need to mention "This is only necessary when OTBR's DHCPv6 PD client is active". it is implied

bukepo (approved):

script/_systemd_networkd
```diff
@@ -0,0 +1,91 @@
...
+        die "This script requires systemctl!"
+    fi
+
+    # Stop NetworkManager service to let systemd-networkd take over
+    sudo systemctl stop NetworkManager || true
+    sudo systemctl disable NetworkManager || true
+
+    sudo systemctl daemon-reload
+
+    create_network_file
```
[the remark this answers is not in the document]
yangsong-cnyn: the guard is already done for SYSTEMD_NETWORKD at reference release script

PR #3028, opened 2025-08-25
[ncp] add DBus interface to set the host power mode

hemanth-silabs:
The NCP monitors the host's power state using the SPINEL_PROP_HOST_POWER_STATE property. Previously, the host was unable to modify this property on the NCP. This commit adds support for updating the host power state property on the NCP. As a result, the NCP can now accurately track the host's power state and wake the host when necessary.


src/host/thread_host.hpp
```diff
@@ -321,6 +321,14 @@ class ThreadHost : virtual public NetworkProperties, public UdpProxy::Dependenci
...
 #endif // OTBR_ENABLE_BACKBONE_ROUTER
 
+    /**
+     * This method updates the Host Power state on NCP to indicate that the host is in a low power mode.
+     *
+     * @param[in] aEnabled  Whether to enable or disable the low power mode.
+     * @param[in] aReceiver A receiver to get the async result of this operation.
+     */
+    virtual void Sleep(bool aEnabled, const AsyncResultReceiver aReceiver) = 0;
+
```
Irving-cl: We can just add `Sleep` method for `NcpHost` instead of `ThreadHost` because only `NcpHost` has this method.

src/host/rcp_host.cpp
```diff
@@ -901,6 +901,15 @@ void RcpHost::SetUdpForwardToHostCallback(UdpForwardToHostCallback aCallback)
     OTBR_UNUSED_VARIABLE(aCallback);
 }
 
+void RcpHost::Sleep(bool aEnabled, const AsyncResultReceiver aReceiver)
```
Irving-cl: We can remove the method for `RcpHost`.

src/host/ncp_host.cpp
```diff
@@ -253,6 +253,23 @@ void NcpHost::AddThreadEnabledStateChangedCallback(ThreadEnabledStateCallback aC
...
+{
+    otError              error    = OT_ERROR_NONE;
+    std::string          errorMsg = "Success";
+    spinel_power_state_t state    = aEnabled ? SPINEL_POWER_STATE_LOW_POWER : SPINEL_POWER_STATE_ONLINE;
+
+    VerifyOrExit((error = SetHostPowerState(state)) == OT_ERROR_NONE, errorMsg = "Error setting host power state");
+
+exit:
+    mTaskRunner.Post([aReceiver, error, errorMsg](void) { aReceiver(error, errorMsg); });
+}
```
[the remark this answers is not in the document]
Irving-cl: @hemanth-silabs 

We do need a callback in `NcpSpinel::SetHostPowerState` and pass the callback from here to ensure we get the result after the power state is correctly set. With the current implementation, `aReceiver` will be invoked regardless of the result of the setting POWER_STATE on NCP.

You may refer to some implementation like `NcpSpinel::ThreadDetach`:
```
    otError      error        = OT_ERROR_NONE;
    EncodingFunc encodingFunc = [](ot::Spinel::Encoder &) { return OT_ERROR_NONE; };

    VerifyOrExit(mThreadDetachGracefullyTask == nullptr, error = OT_ERROR_BUSY);

    SuccessOrExit(error = SetProperty(SPINEL_PROP_NET_LEAVE_GRACEFULLY, encodingFunc));
    mThreadDetachGracefullyTask = aAsyncTask;

exit:
    if (error != OT_ERROR_NONE)
    {
        mTaskRunner.Post([aAsyncTask, error](void) { aAsyncTask->SetResult(error, "Failed to detach gracefully!"); });
    }
```
Note that we store the AsyncTask here as `mThreadDetachGracefullyTask`. And in NcpSpinel, we invoked the callback of this task when the response for `SPINEL_PROP_NET_LEAVE_GRACEFULLY` is received from the NCP. So the similar thing should be applied for HOST_POWER.

src/host/ncp_host.hpp
```diff
@@ -128,6 +128,8 @@ class NcpHost : public MainloopProcessor,
 #if OTBR_ENABLE_BORDER_AGENT && !OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
     void SetBorderAgentVendorTxtData(const std::vector<uint8_t> &aVendorTxtData) override;
 #endif
+    void    Sleep(bool aEnabled, const AsyncResultReceiver aReceiver) override;
+    otError SetHostPowerState(spinel_power_state_t aState);
```
Irving-cl: There is no need to add such method because it simply calls `NcpSpinel::SetHostPowerState`. And it also introduces extra dependency on lib spinel which is not preferred.

src/host/ncp_spinel.hpp
```diff
@@ -404,6 +404,14 @@ class NcpSpinel : public CliDaemon::Dependencies
         mBackboneRouterMulticastListenerCallback = aCallback;
     }
 
+    /**
+     * This method sets the Host Power state on the NCPk.
+     *
+     * @param[in] aState  The Host Power state.
+     */
+
+    otError SetHostPowerState(spinel_power_state_t aState);
```
Irving-cl: As commented above, we should add an asyncTask here to return the async result when receiving a response from the host.

Irving-cl (commented):
@hemanth-silabs 

Thanks for the contribution! I have a few comments.

BTW, what would be your system design like? Because as I learnt, previously this model is used on RTOS. When it comes to posix system, which part will interact with otbr-agent and send the `Sleep` command? And will the system queries otbr-agent first on whether it's ok to sleep now?

hemanth-silabs:
@Irving-cl can you review again? I have incorporated all the suggestions.

src/host/ncp_spinel.hpp
```diff
@@ -404,6 +404,14 @@ class NcpSpinel : public CliDaemon::Dependencies
         mBackboneRouterMulticastListenerCallback = aCallback;
     }
 
+    /**
+     * This method sets the Host Power state on the NCP.
+     *
+     * @param[in] aState  The Host Power state.
+     */
+
```
Irving-cl:
```suggestion
```

Irving-cl (commented):
Overall LGTM. Thanks!

Could you add a simple expect test to verify that the dbus command to set host power state can get a response successfully?

You can refer to this simple test case: https://github.com/openthread/ot-br-posix/blob/main/tests/scripts/expect/ncp_version.exp

Just add a new test called `ncp_xxx.exp` under the same directory. And our CI will run it automatically.

To run it locally, you can build your otbr first. And then run this under the directory of `ot-br-posix`:
```
sudo top_builddir={YOUR_OTBR_BUILD_DIR} bash ./tests/scripts/ncp_mode expect tests/scripts/expect/ncp_xxx.exp
```

src/host/ncp_spinel.hpp
hemanth-silabs: @Irving-cl I have added a new test for this. Please have a look.

tests/scripts/expect/ncp_host_power_mode.exp
```diff
@@ -0,0 +1,71 @@
...
+set HOST_POWER_STATE_ONLINE     "4"
+
+
+# Step 1. Start otbr-agent with a NCP
+spawn_node 1 otbr $::env(EXP_OT_NCP_PATH)
+sleep 1
+
+# Step 2. Send dbus command to set host power state on NCP.
+spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.SetHostPowerState "byte:${HOST_POWER_STATE_OFFLINE}"
+expect eof
```
Irving-cl, line 45: I think there are something as response. What's the output after successfully sending the SetHostPowerState dbus command? Let's expect some keyword in the response instead of eof.

tests/scripts/expect/ncp_host_power_mode.exp
```diff
@@ -0,0 +1,71 @@
...
+sleep 1
+
+spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.SetHostPowerState "byte:${HOST_POWER_STATE_RESERVED}"
+expect eof
+
+
+# Shut down otbr-agent
+switch_node 1
+send "\x04"
+expect eof
```
Irving-cl: Add an empty new line at the end.
```suggestion
expect eof

```

tests/scripts/expect/ncp_host_power_mode.exp
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/expect -f
+#
+#  Copyright (c) 2024, The OpenThread Authors.
```
Irving-cl:
```suggestion
#  Copyright (c) 2025, The OpenThread Authors.
```

tests/scripts/expect/ncp_host_power_mode.exp
hemanth-silabs: I was looking for something as well. But this is the response I get:
```
sudo dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.SetHostPowerState "byte:3"
method return time=1756300080.777905 sender=:1.53 -> destination=:1.54 serial=4 reply_serial=2
```
There is nothing I can interpret to be the status.

I do see something on the logs:
```
2025-08-27T09:08:00.778381-04:00 raspi-Hemanth otbr-agent[3663]: [INFO]-DBUS----: Replied to io.openthread.BorderRouter.SetHostPowerState with result io.openthread.Error.OK
2025-08-27T09:08:00.779574-04:00 raspi-Hemanth otbr-agent[3663]: [INFO]-NcpSpin-: Set Host Power result: OK
```
But that is no use here.

hemanth-silabs:
@jwhui the errors don't seem to be related to the code changes here. Is it possible to restart the failing checks?

jwhui (approved):

PR #3038, opened 2025-09-06
[epskc] fix build option for epskc

Irving-cl:
This PR fixes the usage of cmake option for epskc.

There is  `OTBR_ENABLE_EPSKC` but it's not used at all. All the existing code are guarded by `OTBR_BORDER_AGENT`. This PR replaces `OTBR_BORDER_AGENT` by `OTBR_ENABLE_EPSKC` for epskc code. The PR also sets the default value of `OTBR_EPSKC` to be the same as `OTBR_BORDER_AGENT` and add a check to ensure `OTBR_BORDER_AGENT` must be enabled if `OTBR_EPSKC` is enabled. Thus we can only use `OTBR_ENABLE_EPSKC` to guard epskc code (without `&& OTBR_ENABLE_BORDER_AGENT`.


etc/cmake/options.cmake
```diff
@@ -165,8 +165,11 @@ if(OTBR_TREL_DNSSD)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_TREL_DNSSD=1)
 endif()
 
-option(OTBR_EPSKC "Enable ephemeral PSKc" ON)
+option(OTBR_EPSKC "Enable ephemeral PSKc" ${OTBR_BORDER_AGENT})
 if (OTBR_EPSKC)
+    if(NOT OTBR_BORDER_AGENT)
+        message(FATAL_ERROR "OTBR_EPSKC requires OTBR_BORDER_AGENT to be enabled.")
+    endif()
```
mia1yang, line 172: Can the code be executed? move to else？

Irving-cl: Yes I have verified that it can be executed.

```
option(OTBR_EPSKC "Enable ephemeral PSKc" ${OTBR_BORDER_AGENT})
```
If we explicitly set `OTBR_EPSKC`, then this line won't take effect. So this check prevents us from explicitly setting `OTBR_EPSKC` to ON and `OTBR_BORDER_AGENT` to OFF.

jwhui (approved):

PR #3048, opened 2025-09-15
[dnssd] update the implementation of browse/resolve methods

Irving-cl:
This PR enhances the implementation of dnssd discovery proxy methods and fixes the issue of starting browse/resolve in the callback of these methods.

The general idea is:
1. The callback tables will all be updated immediately when doing start/stop browse/resolve operations. Unlike the current implementation: A delete operation will be done in a lazy way. And an add operation is simply discarded if invoking callbacks.
2. When invoking callbacks, we will make a copy of all the callbacks and iterate over the copy. So the copy won't change during the invoke process.
3. To control what's to subscribe and unsubscribe via mPublisher, 3 std::set members are introduced to track what are currently being subscribed. Each time when the callback maps change, a task will posted to update the subscriptions based on the difference between the callback maps and the set.

This way the implementation has better readability and can support flexible usages. I have also verified the implementation with `OT_DNSSD_DISCOVERY_PROXY` and it can work (a ot-cli-ftd use dns client to discover the meshcop service of a BR):
```
> dns config fdaa:4ba1:6f87:81ab:a0b2:690d:624:9496 53 3000 1 0
Done
> dns browse _meshcop._udp.default.service.arpa.
DNS browse response for _meshcop._udp.default.service.arpa.
OpenThread BorderRouter #D613
    Port:49154, Priority:0, Weight:0, TTL:10
    Host:irvingcl2.default.service.arpa.
    HostAddress:2401:fa00:41:23:ceec:3330:6811:c4fd TTL:10
    TXT:[vn=4f70656e546872656164, mn=426f72646572526f75746572, id=cf530e58b510a700870052402bc220e0, rv=31, nn=4f70656e5468726561642d31313831, xp=a6adf9b8a4f8ac2a, tv=312e342e30, xa=3a6faf7a1ce8d613, sb=00000e31, pt=378765e9, at=0000000000010000, omr=40fd07744b849f0001] TTL:10

Done
```
This change is required to support this because the OT discovery proxy has some usages of adding new subscriptions in the callback of the previous dnssd operation.


src/host/posix/dnssd.cpp
```diff
@@ -672,21 +608,144 @@ void DnssdPlatform::StopAddressResolver(const AddressResolver
...
+void DnssdPlatform::ExecuteServiceSubscriptionUpdate(void)
+{
+    mServiceSubscriptionUpdateTaskPosted = false;
+
+    // Unsubscribe services (DnsServiceType) that are stale.
+    for (auto iter = mServiceTypeSubscriptions.begin(); iter != mServiceTypeSubscriptions.end();)
+    {
+        if (mServiceBrowsersMap.find(*iter) == mServiceBrowsersMap.end())
+        {
+            mPublisher.UnsubscribeService(iter->ToString(), "");
```
superwhd: Here and elsewhere, better to document the second argument.

src/host/posix/dnssd.cpp
```diff
@@ -672,21 +608,144 @@ void DnssdPlatform::StopAddressResolver(const AddressResolver
...
+    {
+        const DnsName &dnsName = entry.first;
+
+        if (mHostSubscriptions.find(dnsName) == mHostSubscriptions.end())
+        {
+            mHostSubscriptions.insert(dnsName);
+            mPublisher.SubscribeHost(dnsName.GetName());
+        }
+    }
+    for (auto &entry : mIp6AddrResolversMap)
```
superwhd: Seems to be a copy-paste mistake? Shall it be IPv4?

src/host/posix/dnssd.cpp
```diff
@@ -672,21 +608,144 @@ void DnssdPlatform::StopAddressResolver(const AddressResolver
...
+    {
+        const DnsName &dnsName = entry.first;
+
+        if (mHostSubscriptions.find(dnsName) == mHostSubscriptions.end())
+        {
+            mHostSubscriptions.insert(dnsName);
+            mPublisher.SubscribeHost(dnsName.GetName());
+        }
+    }
```
superwhd: Probably you can merge these two loops into one to make it more concise.

src/host/posix/dnssd.hpp
```diff
@@ -391,10 +367,16 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
+            IteratorType iter;
+
+            for (iter = mEntries.begin(); iter != mEntries.end(); iter++)
+            {
+                if (iter->first == aInfraIfIndex && *(iter->second) == aCallback)
+                {
+                    break;
+                }
+            }
+            return iter;
```
superwhd: Just curious, why didn't this continue using `std::find_if`?

src/host/posix/dnssd.cpp
```diff
@@ -672,21 +608,144 @@ void DnssdPlatform::StopAddressResolver(const AddressResolver
...
+
+    // Subscribe to services (DnsServiceName) that haven't been subscribed.
+    for (const auto &entry : mServiceResolversMap)
+    {
+        const DnsServiceName &serviceName = entry.first;
+
+        if (mServiceNameSubscriptions.find(serviceName) == mServiceNameSubscriptions.end())
+        {
+            mServiceNameSubscriptions.insert(serviceName);
+            mPublisher.SubscribeService(serviceName.GetType(), serviceName.GetInstance());
```
superwhd, line 657: I have an impression that the `mdns::Publisher` doesn't handle the case well when subscribing to the same thing multiple times. Do we have any dependency on this?

superwhd (commented):
Had a glance. Will look into more details later.

src/host/posix/dnssd.cpp
Irving-cl: I've merged the 2 maps into one. Currently there is no difference to resolve Ipv6/Ipv4 host. The underlying to call mPublisher are exactly the same.

src/host/posix/dnssd.cpp
Irving-cl: I've merged the 2 maps into one. Currently there is no difference to resolve Ipv6/Ipv4 host. The underlying to call mPublisher are exactly the same.

src/host/posix/dnssd.hpp
Irving-cl: I found the original code is less readable.

src/host/posix/dnssd.cpp
Irving-cl: The `mServiceNameSubscriptions` should guarantee that the same thing is only subcribed by `mdns::Publisher` once. Because we always find whether `mServiceNameSubscriptions` contains the name first. If it doesn't contain, we added it to the set and call `mdns::Publisher` to subscribe.

src/host/posix/dnssd.hpp
```diff
@@ -342,47 +333,32 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
-                    ++iter;
+                    copyCallbacks.push_back(entry.second);
                 }
             }
-        }
 
-        void InvokeAllCallbacks(uint64_t aInfraIfIndex, CallbackResultType &aResult)
-        {
-            for (auto &entry : mEntries)
+            for (CallbackPtrType callback : copyCallbacks)
```
superwhd: nit: Could this be a const reference?

src/host/posix/dnssd.hpp
```diff
@@ -306,33 +308,22 @@ class DnssdPlatform : public Mdns::StateObserver, private NonCopyable
...
 
-    template <typename T> struct FirstFunctionArg;
-
-    template <typename R, typename Arg, typename... Rest> struct FirstFunctionArg<std::function<R(Arg, Rest...)>>
-    {
-        using Type = Arg;
-    };
-
-    // RequestType MUST be a std::pair<uint64_t, std::unique_ptr<CallbackType>>
+    // RequestType MUST be a std::pair<uint64_t, std::shared_ptr<CallbackType>>
```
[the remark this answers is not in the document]
Irving-cl: All occurrences of `unique_ptr` have been replaced by `shared_ptr`.

PR #3066, opened 2025-09-25
[ncp] add EPSKC support to NCP

hemanth-silabs:
This commit adds EPSKC suppport to NCP.
    1. Adds new DBus interfaces
    2. Updates EPSKC related property handling on host.
    3. Adds new UDP Proxy for EPSKC messages.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -70,6 +70,9 @@ Application::Application(Host::ThreadHost  &aHost,
     , mBorderAgent(*mPublisher)
 #endif
     , mBorderAgentUdpProxy(mHost)
+#if OTBR_ENABLE_EPSKC
+    , mEphemeralKeyUdpProxy(mHost)
+#endif
 #endif
 #if OTBR_ENABLE_DBUS_SERVER
     , mDBusAgent(MakeDBusDependentComponents())
@@ -91,9 +94,13 @@ Application::Application(Host::ThreadHost  &aHost,
 
 void Application::Init(const std::string &aRestListenAddress, int aRestListenPort)
 {
+    CoprocessorType type;
+
     mHost.Init();
 
-    switch (mHost.GetCoprocessorType())
+    type = mHost.GetCoprocessorType();
+
+    switch (type)
     {
     case OT_COPROCESSOR_RCP:
         InitRcpMode(aRestListenAddress, aRestListenPort);
@@ -110,7 +117,8 @@ void Application::Init(const std::string &aRestListenAddress, int aRestListenPor
     mDBusAgent.Init();
 #endif
 
-    otbrLogInfo("Co-processor version: %s", mHost.GetCoprocessorVersion());
+    otbrLogInfo("%s Co-processor version: %s", type == OT_COPROCESSOR_RCP ? "Radio" : "Network",
+                mHost.GetCoprocessorVersion());
 }
 
 void Application::Deinit(void)
@@ -389,7 +397,30 @@ void Application::InitNcpMode(void)
         {
             mBorderAgentUdpProxy.SendToPeer(aUdpPayload, aLength, aPeerAddr, aPeerPort);
         }
+#if OTBR_ENABLE_EPSKC
+        else if (aLocalPort == mEphemeralKeyUdpProxy.GetThreadPort())
+        {
+            mEphemeralKeyUdpProxy.SendToPeer(aUdpPayload, aLength, aPeerAddr, aPeerPort);
+        }
+#endif // OTBR_ENABLE_EPSKC
     });
+#if OTBR_ENABLE_EPSKC
+    mHost.AddEphemeralKeyStateChangedCallback([this](otBorderAgentEphemeralKeyState aState, uint16_t aPort) {
+        if (aState == OT_BORDER_AGENT_STATE_STARTED)
+        {
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Active on port %d", aPort);
+            mEphemeralKeyUdpProxy.Start(aPort);
+        }
+        else if (aState == OT_BORDER_AGENT_STATE_STOPPED || aState == OT_BORDER_AGENT_STATE_DISABLED)
+        {
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Inactive");
+            mEphemeralKeyUdpProxy.Stop();
+        }
+#if OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
+        mBorderAgent.HandleEpskcStateChanged(aState, mEphemeralKeyUdpProxy.GetHostPort());
+#endif // OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
+    });
+#endif // OTBR_ENABLE_EPSKC
     SetBorderAgentOnInitState();
 #endif
 #if OTBR_ENABLE_BACKBONE_ROUTER
--- a/src/dbus/server/dbus_thread_object_ncp.cpp
+++ b/src/dbus/server/dbus_thread_object_ncp.cpp
@@ -31,7 +31,6 @@
 #include "common/api_strings.hpp"
 #include "common/byteswap.hpp"
 #include "common/code_utils.hpp"
-#include "dbus/common/constants.hpp"
 #include "dbus/server/dbus_agent.hpp"
 #include "host/thread_helper.hpp"
 
@@ -46,6 +45,9 @@ DBusThreadObjectNcp::DBusThreadObjectNcp(DBusConnection            &aConnection,
                                          const DependentComponents &aDeps)
     : DBusObject(&aConnection, OTBR_DBUS_OBJECT_PREFIX + aInterfaceName)
     , mHost(static_cast<Host::NcpHost &>(aDeps.mHost))
+#if OTBR_ENABLE_BORDER_AGENT
+    , mBorderAgent(aDeps.mBorderAgent)
+#endif
 {
 }
 
@@ -66,6 +68,14 @@ otbrError DBusThreadObjectNcp::Init(void)
                    std::bind(&DBusThreadObjectNcp::ScheduleMigrationHandler, this, _1));
     RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_HOST_POWER_STATE_METHOD,
                    std::bind(&DBusThreadObjectNcp::HostPowerStateHandler, this, _1));
+#if OTBR_ENABLE_EPSKC
+    RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_PROPERTY_EPHEMERAL_KEY_ENABLED,
+                   std::bind(&DBusThreadObjectNcp::EnableEphemeralKeyModeHandler, this, _1));
+    RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_ACTIVATE_EPHEMERAL_KEY_MODE_METHOD,
+                   std::bind(&DBusThreadObjectNcp::ActivateEphemeralKeyModeHandler, this, _1));
+    RegisterMethod(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_DEACTIVATE_EPHEMERAL_KEY_MODE_METHOD,
+                   std::bind(&DBusThreadObjectNcp::DeactivateEphemeralKeyModeHandler, this, _1));
+#endif
 
     SuccessOrExit(error = Signal(OTBR_DBUS_THREAD_INTERFACE, OTBR_DBUS_SIGNAL_READY, std::make_tuple()));
 exit:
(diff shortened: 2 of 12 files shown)


src/dbus/common/constants.hpp
```diff
@@ -65,6 +65,7 @@
 #define OTBR_DBUS_GET_PROPERTIES_METHOD "GetProperties"
 #define OTBR_DBUS_LEAVE_NETWORK_METHOD "LeaveNetwork"
 #define OTBR_DBUS_SET_NAT64_ENABLED_METHOD "SetNat64Enabled"
+#define OTBR_DBUS_ENABLE_EPHEMERAL_KEY_MODE_METHOD "EnableEphemeralKeyMode"
```
Irving-cl:
```suggestion
#define OTBR_DBUS_SET_EPHEMERAL_KEY_ENABLED_METHOD "SetEphemeralKeyEnabled"
```

src/agent/application.cpp
```diff
@@ -383,6 +389,18 @@ void Application::InitNcpMode(void)
             OTBR_UNUSED_VARIABLE(aLength);
 #endif
         });
+    mHost.AddEphemeralKeyStateChangedCallback([this](otBorderAgentEphemeralKeyState aState, uint16_t aPort) {
+        if (aState == OT_BORDER_AGENT_STATE_STARTED)
+        {
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Active on port %d", aPort);
+            mEphemeralKeyUdpProxy.Start(aPort);
+        }
+        else if(aState == OT_BORDER_AGENT_STATE_STOPPED)
```
Irving-cl:
```suggestion
        else if (aState == OT_BORDER_AGENT_STATE_STOPPED || aState == OT_BORDER_AGENT_STATE_DISABLED)
```

src/agent/application.cpp
```diff
@@ -65,11 +65,15 @@ Application::Application(Host::ThreadHost  &aHost,
 #if OTBR_ENABLE_DNSSD_PLAT
     , mDnssdPlatform(*mPublisher)
 #endif
+
```
Irving-cl: No need to add an empty line.

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -26,15 +26,17 @@
...
+#include <openthread/border_router.h>
 
 #include "common/api_strings.hpp"
 #include "common/byteswap.hpp"
 #include "common/code_utils.hpp"
-#include "dbus/common/constants.hpp"
 #include "dbus/server/dbus_agent.hpp"
 #include "host/thread_helper.hpp"
 
+#include "dbus_thread_object_ncp.hpp"
```
Irving-cl: Though not documented, the general include guideline is that, in the cpp sources (implementation), put its corresponding definition header at the first. So let's still put "dbus_thread_object_ncp.hpp" at the top.

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -182,5 +192,77 @@ void DBusThreadObjectNcp::HostPowerStateHandler(DBusRequest &aRequest)
...
+    }
+}
+
+void DBusThreadObjectNcp::ActivateEphemeralKeyModeHandler(DBusRequest &aRequest)
+{
+    otError              error = OT_ERROR_NONE;
+    std::vector<uint8_t> data;
+    uint32_t             lifetime = 0;
+    uint16_t             port     = 0;
+    std::string          ePskc;
```
Irving-cl: For RCP the client doesn't need to generate the ePSKc. It's generated by otbr-agent. We have a static method `CreateEphemeralKey` in `BorderAgent`.

src/host/ncp_host.cpp
```diff
@@ -263,6 +263,47 @@ void NcpHost::SetHostPowerState(uint8_t aState, const AsyncResultReceiver &aRece
...
+    AsyncTaskPtr task;
+    auto errorHandler = [aReceiver](otError aError, const std::string &aErrorInfo) { aReceiver(aError, aErrorInfo); };
+
+    task = std::make_shared<AsyncTask>(errorHandler);
+    task->First([this, aRetainActiveSession](AsyncTaskPtr aNext) {
+        mNcpSpinel.DeactivateEphemeralKey(aRetainActiveSession, std::move(aNext));
+    });
+    task->Run();
+}
+#endif
```
Irving-cl:
```suggestion
#endif // OTBR_ENABLE_EPSKC
```

src/host/ncp_spinel.cpp
```diff
@@ -292,6 +292,14 @@ void NcpSpinel::SetBorderAgentMeshCoPServiceChangedCallback(const BorderAgentMes
     SuccessOrDie(GetProperty(SPINEL_PROP_BORDER_AGENT_MESHCOP_SERVICE_STATE), "Failed to get MeshCoP Service State");
 }
 
+void NcpSpinel::AddEphemeralKeyStateChangedCallback(const EphemeralKeyStateChangedCallback &aCallback)
+{
+    mEphemeralKeyStateChangedCallback = aCallback;
+
+    // Get the Ephemeral Key state to have an initial value.
+    SuccessOrDie(GetProperty(SPINEL_PROP_BORDER_AGENT_EPHEMERAL_KEY_STATE), "Failed to get Ephemeral Key State");
```
Irving-cl: We don't need to get it once in this method and shouldn't call synchronized spinel methods (like GetProperty, which causes `two select` scenario) in the NcpSpinel.  Because after initialization it can only be "Stopped" or "Disabled" and the host doesn't care whether it's stopped or disabled.

src/host/ncp_spinel.hpp
```diff
@@ -413,6 +423,40 @@ class NcpSpinel : public CliDaemon::Dependencies
...
+    /**
+     * Deactivates the Ephemeral Key on the NCP.
+     *
+     * This method instructs the NCP to deactivate the currently active ephemeral key.
+     *
+     * @param[in] aRetainActiveSession  TRUE to retain the active session, FALSE to terminate it.
+     * @param[in] aAsyncTask            The async task pointer to receive the result of this operation.
+     */
+    void DeactivateEphemeralKey(bool aRetainActiveSession, AsyncTaskPtr aAsyncTask);
+#endif
```
Irving-cl:
```suggestion
#endif // OTBR_ENABLE_EPSKC
```

src/host/ncp_spinel.hpp
```diff
@@ -413,6 +423,40 @@ class NcpSpinel : public CliDaemon::Dependencies
...
+
+    /**
+     * Activates the Ephemeral Key on the NCP.
+     *
+     * This method instructs the NCP to activate an ephemeral key for secure operations.
+     *
+     * @param[in] aPskc         The null-terminated PSKc string to use as the ephemeral key.
+     * @param[in] aDurationMilli The duration (in milliseconds) for which the ephemeral key should be active.
+     * @param[in] aPort         The UDP port on which the ephemeral key is to be used.
+     * @param[in] aAsyncTask    The async task pointer to receive the result of this operation.
```
Irving-cl:
```suggestion
     * @param[in] aPskc               The null-terminated PSKc string to use as the ephemeral key.
     * @param[in] aDurationMilli  The duration (in milliseconds) for which the ephemeral key should be active.
     * @param[in] aPort                The UDP port on which the ephemeral key is to be used.
     * @param[in] aAsyncTask     The async task pointer to receive the result of this operation.
```

src/host/ncp_spinel.cpp
```diff
@@ -1377,6 +1426,80 @@ void NcpSpinel::SetHostPowerState(uint8_t aState, AsyncTaskPtr aAsyncTask)
...
 
+#if OTBR_ENABLE_EPSKC
+void NcpSpinel::EnableEphemeralKey(bool aEnable, AsyncTaskPtr aAsyncTask)
+{
+    otError      error        = OT_ERROR_NONE;
+    EncodingFunc encodingFunc = [aEnable](ot::Spinel::Encoder &aEncoder) { return aEncoder.WriteBool(aEnable); };
+
+    VerifyOrExit(mEphemeralKeyTask == nullptr, error = OT_ERROR_BUSY);
+
+    otbrLogInfo("In EnableEphemeralKey aEnable: %s", (aEnable ? "true" : "false"));
```
Irving-cl: This seems like a log during development?

src/agent/application.cpp
```diff
@@ -383,6 +389,18 @@ void Application::InitNcpMode(void)
...
+        if (aState == OT_BORDER_AGENT_STATE_STARTED)
+        {
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Active on port %d", aPort);
+            mEphemeralKeyUdpProxy.Start(aPort);
+        }
+        else if(aState == OT_BORDER_AGENT_STATE_STOPPED)
+        {
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Inactive");
+            mEphemeralKeyUdpProxy.Stop();
+        }
```
Irving-cl, line 418: Should call `mBorderAgent.HandleEpskcStateChanged` here to update the meshcop-e service.

src/agent/application.cpp
```diff
@@ -389,7 +395,30 @@ void Application::InitNcpMode(void)
...
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Active on port %d", aPort);
+            mEphemeralKeyUdpProxy.Start(aPort);
+        }
+        else if (aState == OT_BORDER_AGENT_STATE_STOPPED || aState == OT_BORDER_AGENT_STATE_DISABLED)
+        {
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Inactive");
+            mEphemeralKeyUdpProxy.Stop();
+#if OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
+            mBorderAgent.HandleEpskcStateChanged(aState, aPort);
+#endif // OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
```
Irving-cl, line 421: This should be invoked in both cases? When state is `STARTED`, this should also be called to publish the meshcop-e service.

src/host/ncp_spinel.cpp
```diff
@@ -1377,6 +1423,78 @@ void NcpSpinel::SetHostPowerState(uint8_t aState, AsyncTaskPtr aAsyncTask)
...
+    mEphemeralKeyTask = aAsyncTask;
+
+exit:
+    if (error != OT_ERROR_NONE)
+    {
+        mTaskRunner.Post(
+            [aAsyncTask, error](void) { aAsyncTask->SetResult(error, "Failed to deactivate ephemeral key!"); });
+    }
+}
+#endif
```
Irving-cl:
```suggestion
#endif // OTBR_ENABLE_EPSKC
```

Irving-cl (commented):
Overall LGTM 👍  Thanks.

Could you add a simple expect case just to verify that the meshcop-e service is published correctly after activating the ePSKc mode? I think you can refer to this border agent case and it will be easy to implement: https://github.com/openthread/ot-br-posix/blob/main/tests/scripts/expect/ncp_border_agent.exp#L77  

I understand that currently we have a gap to get the host side ePSKc port by cli command. So let's only verify the meshcop-e service first. We can add full commission test later.

src/host/ncp_spinel.cpp
```diff
@@ -818,6 +836,19 @@ otbrError NcpSpinel::HandleResponseForPropGet(spinel_tid_t      aTid,
...
+        uint8_t             state;
+        uint16_t            port;
+        ot::Spinel::Decoder decoder;
+
+        decoder.Init(aData, aLength);
+        SuccessOrExit(decoder.ReadUint8(state), error = OTBR_ERROR_PARSE);
+        SuccessOrExit(decoder.ReadUint16(port), error = OTBR_ERROR_PARSE);
+        SafeInvoke(mEphemeralKeyStateChangedCallback, static_cast<otBorderAgentEphemeralKeyState>(state), port);
+        break;
+    }
```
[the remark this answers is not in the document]
hemanth-silabs: Is this needed? I am following the current coding conventions in the same file, like the one for SPINEL_PROP_BORDER_AGENT_MESHCOP_SERVICE_STATE.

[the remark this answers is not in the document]
Irving-cl: We actually don't need this in `HandleResponseForPropGet`. Let's remove it from `HandleResponseForPropGet`.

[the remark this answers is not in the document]
hemanth-silabs: Done

src/agent/application.cpp
```diff
@@ -93,7 +96,9 @@ void Application::Init(const std::string &aRestListenAddress, int aRestListenPor
 {
     mHost.Init();
 
-    switch (mHost.GetCoprocessorType())
+    CoprocessorType type = mHost.GetCoprocessorType();
```
abtink: Suggest moving this to after `{`. We generally aim to place all local variable definitions at the top of the scope (start of `{}`).

src/agent/application.cpp
```diff
@@ -389,7 +395,30 @@ void Application::InitNcpMode(void)
...
+            mEphemeralKeyUdpProxy.Start(aPort);
+        }
+        else if (aState == OT_BORDER_AGENT_STATE_STOPPED || aState == OT_BORDER_AGENT_STATE_DISABLED)
+        {
+            otbrLogInfo("Border Agent Ephemeral Key State Changed: Inactive");
+            mEphemeralKeyUdpProxy.Stop();
+        }
+#if OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
+        mBorderAgent.HandleEpskcStateChanged(aState, mEphemeralKeyUdpProxy.GetHostPort());
+#endif // OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
```
abtink, line 421: Same here.

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -27,11 +27,12 @@
  */
 
 #include "dbus_thread_object_ncp.hpp"
+#include <openthread/border_agent.h>
+#include <openthread/border_router.h>
```
abtink: Can we include these headers (public OT headers) in this file?  Do we need them?
Not sure if on NCP they are always accessible?

src/host/ncp_host.cpp
```diff
@@ -142,6 +142,7 @@ void NcpHost::Init(void)
     mNcpSpinel.SrpServerSetEnabled(/* aEnabled */ true);
 #endif
 #endif
+    otPlatCryptoRandomInit();
```
abtink: Why do we need this call? I am not sure we have access to this on the NCP host.

The `otPlat*` APIs are intended for OT stack use. In the NCP architecture, the OT stack would be part of the NCP firmware. On the NCP host (which interacts with the NCP), we should not be able to see or access any of the `otPlat*` APIs.

abtink (commented):
Thanks. Some questions/suggestions below.

src/host/ncp_host.cpp
Irving-cl: Currently this is used by `BorderAgent::CreateEphemeralKey`. When the dbus call comes to activate ePSKc mode, we generate the ephemeral key for the app side. OT itself isn't responsible for generating the key. The ephemeral key generation is added in the dbus call of ot-br-posix for convenience. (As I understand it, this helps the app side to activate ePSKc mode easier by saving their effort to generate an ephemeral key) The key generation in RCP design depends on the platform crypto. And the platform crypto is initialized during OT initialization in RCP design. So we did the same for NCP design. 

I think we can change the implementation of ephemeral key generation by not using OT platform crypto but a different random engine. Thoughts? I can help work on that.

src/dbus/server/dbus_thread_object_ncp.cpp
hemanth-silabs: @abtink we don't need this header anymore. It was used for a different purpose and I thought it might be OK as `dbus_thread_object_rcp.cpp` includes them as well. I can remove this.

src/host/ncp_host.cpp
hemanth-silabs: @abtink don't we have a POSIX implementation of all otPlat APIs. If we need to use entropy in host what is the best solution as of now?

abtink: We do, but it is intended for use by the OT stack running on a POSIX platform. It is not intended for use by other components and software modules. We should be careful about creating such dependencies, which do not follow good software design principles and can lead to confusion (some folks may wonder and ask why they would need to implement a subset of `otPlat*` APIs for use on an NCP host software module which is not actually running the OT stack but interacts with an NCP).

If required, the NCP host should define its own set of platform/helper functions it wants to use for things like crypto, random number generation, etc. Reusing the same `otPlat*` APIs is not a good idea.

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -182,5 +192,72 @@ void DBusThreadObjectNcp::HostPowerStateHandler(DBusRequest &aRequest)
     }
 }
 
+#if OTBR_ENABLE_EPSKC
+void DBusThreadObjectNcp::EnableEphemeralKeyModeHandler(DBusRequest &aRequest)
+{
+    otError error = OT_ERROR_NONE;
+    bool    enable;
+    auto    args = std::tie(enable);
+    SuccessOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args), error = OT_ERROR_INVALID_ARGS);
```
abtink, line 202: Newline after variable declartion?
```suggestion
    auto    args = std::tie(enable);
    
    SuccessOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args), error = OT_ERROR_INVALID_ARGS);
```

hemanth-silabs: Added..

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -182,5 +192,73 @@ void DBusThreadObjectNcp::HostPowerStateHandler(DBusRequest &aRequest)
...
+    otError     error    = OT_ERROR_NONE;
+    uint32_t    lifetime = 0;
+    std::string ePskc;
+    auto        args = std::tie(lifetime);
+
+    SuccessOrExit(DBusMessageToTuple(*aRequest.GetMessage(), args), error = OT_ERROR_INVALID_ARGS);
+
+    VerifyOrExit(lifetime <= OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT, error = OT_ERROR_INVALID_ARGS);
+    SuccessOrExit(mBorderAgent.CreateEphemeralKey(ePskc), error = OT_ERROR_INVALID_ARGS);
+    otbrLogInfo("Created Ephemeral Key: %s", ePskc.c_str());
```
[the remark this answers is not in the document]
hemanth-silabs: EPSKC is a Border agent feature. There is no scenario where BORDER_AGENT can be disabled but EPSKC is enabled. The code is structured similarly to the code in `dbus_thread_object_rcp.cpp`

src/dbus/server/dbus_thread_object_ncp.cpp
```diff
@@ -182,5 +192,73 @@ void DBusThreadObjectNcp::HostPowerStateHandler(DBusRequest &aRequest)
...
+
+    mHost.ActivateEphemeralKey(ePskc.c_str(), lifetime, OTBR_CONFIG_BORDER_AGENT_MESHCOP_E_UDP_PORT,
```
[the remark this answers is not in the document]
hemanth-silabs: `openthread-br/config.h` is included in `dbus_thread_object_ncp.cpp` through `dbus_thread_object_ncp.hpp`, similar to to its RCP counterpart.

PR #3067, opened 2025-09-26
Toggle Border Agent on Thread role changes

zesonzhang:
This pull request introduces a new callback mechanism to notify components when the Thread device role changes, and integrates it with the Border Agent to enable or disable its functionality based on the current role. The changes also include updates to the callback management in `RcpHost`, the addition of relevant tests, and supporting code refactoring.

**Callback mechanism for Thread role changes**

* Added `ThreadRoleChangedCallback` type to `ThreadHost` and implemented storage and management of these callbacks in `RcpHost`, including methods to add and invoke them when the device role changes. (`src/host/thread_host.hpp` [[1]](diffhunk://#diff-a2c6f7e8ab49cb7375cb652daf97dbcd4f028c642f997fec7177edeceb2726d6R134) `src/host/rcp_host.hpp` [[2]](diffhunk://#diff-3a93981dd30498b6cd5a81ab5764d1c5e1a8bbe6f2cd6b748739527151fb77b7R215) [[3]](diffhunk://#diff-3a93981dd30498b6cd5a81ab5764d1c5e1a8bbe6f2cd6b748739527151fb77b7R293) `src/host/rcp_host.cpp` [[4]](diffhunk://#diff-a2ce80157c7cca8257562e1cb13fd395c39f725e5a2bb14e9b0163bc25a48104R435-R439) [[5]](diffhunk://#diff-a2ce80157c7cca8257562e1cb13fd395c39f725e5a2bb14e9b0163bc25a48104L367-R381) [[6]](diffhunk://#diff-a2ce80157c7cca8257562e1cb13fd395c39f725e5a2bb14e9b0163bc25a48104R348)

**Integration with Border Agent**

* Implemented `BorderAgent::HandleThreadRoleChanged` to enable or disable the Border Agent depending on whether the device is attached, using a new static helper `IsAttached`. Registered this callback with `RcpHost` during initialization. (`src/border_agent/border_agent.hpp` [[1]](diffhunk://#diff-696b10f62cc91c69588189efe44cc10f08f904165cbf646c3ceb35a8a3d78877R172-R178) [[2]](diffhunk://#diff-696b10f62cc91c69588189efe44cc10f08f904165cbf646c3ceb35a8a3d78877R225) `src/border_agent/border_agent.cpp` [[3]](diffhunk://#diff-2ca9c1d72f7c52036524f45dde89c1edee382351f37c6b472565a7523b0e49d2R212-R216) [[4]](diffhunk://#diff-2ca9c1d72f7c52036524f45dde89c1edee382351f37c6b472565a7523b0e49d2R335-R340) `src/agent/application.cpp` [[5]](diffhunk://#diff-d6ab38e80ae0bd95c36dd0ba778c7232cb63861d7df073062bf6276744e9cfbfR272)

**Testing**

* Added a new unit test `ThreadRoleChangedCallbackInvoked` to verify that the callback is invoked correctly when the device role changes, ensuring the mechanism works as intended. (`tests/gtest/test_rcp_host_api.cpp` [tests/gtest/test_rcp_host_api.cppR438-R486](diffhunk://#diff-07f1c9f6ffca51398e07d7a1a52bae48cd37021642229fc7f92c32ccd53f67a1R438-R486))

**Code maintenance**

* Cleared the new callback vector in `RcpHost::Deinit` to prevent resource leaks. (`src/host/rcp_host.cpp` [src/host/rcp_host.cppR348](diffhunk://#diff-a2ce80157c7cca8257562e1cb13fd395c39f725e5a2bb14e9b0163bc25a48104R348))
* Minor whitespace cleanup in test code for clarity. (`tests/gtest/test_rcp_host_api.cpp` [tests/gtest/test_rcp_host_api.cppL59](diffhunk://#diff-07f1c9f6ffca51398e07d7a1a52bae48cd37021642229fc7f92c32ccd53f67a1L59))


tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +435,45 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+    otDeviceRole lastRole        = OT_DEVICE_ROLE_DISABLED;
+    auto         roleCb          = [&callbackInvoked, &lastRole](otDeviceRole aRole) {
+        lastRole        = aRole;
+        callbackInvoked = true;
+    };
+    host.AddThreadRoleChangedCallback(roleCb);
+
+    otInstance *instance = ot::FakePlatform::CurrentInstance();
+
+    // Cast 1. check CB is invoked when enabling Thread
```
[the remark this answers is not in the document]
zesonzhang: fixed. thanks

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +443,49 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+    auto         roleCb          = [&callbackInvoked, &lastRole](otDeviceRole aRole) {
+        lastRole        = aRole;
+        callbackInvoked = true;
+    };
+    host.AddThreadRoleChangedCallback(roleCb);
+
+    otInstance *instance = ot::FakePlatform::CurrentInstance();
+
+    // Case 1. check CB is invoked when enabling Thread
+    OT_UNUSED_VARIABLE(otIp6SetEnabled(instance, true));
```
bukepo:
```suggestion
    ASSERT_EQ(otIp6SetEnabled(instance, true), OT_ERROR_NONE);
```

src/host/rcp_host.cpp
```diff
@@ -364,10 +365,18 @@ void RcpHost::HandleStateChanged(otChangedFlags aFlags)
...
+        otDeviceRole role = GetDeviceRole();
+        for (auto &callback : mThreadRoleChangedCallbacks)
+        {
+            callback(role);
+        }
+        if (IsAttached() && mJoinReceiver != nullptr)
+        {
+            otbrLogInfo("Join succeeded");
+            SafeInvokeAndClear(mJoinReceiver, OT_ERROR_NONE, "Join succeeded");
+        }
```
Irving-cl, line 380:
```suggestion

        for (auto &callback : mThreadRoleChangedCallbacks)
        {
            callback(role);
        }

        if (IsAttached() && mJoinReceiver != nullptr)
        {
            otbrLogInfo("Join succeeded");
            SafeInvokeAndClear(mJoinReceiver, OT_ERROR_NONE, "Join succeeded");
        }
```
Nit: add a blank between blocks

src/border_agent/border_agent.hpp
```diff
@@ -169,6 +169,13 @@ class BorderAgent : private NonCopyable
 
 #endif // OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
 
+    /**
+     * This method handles Thread role changes.
+     *
+     * @param[in] aRole  The new Thread role.
+     */
+    void HandleThreadRoleChanged(otDeviceRole aRole);
```
Irving-cl: It seems unnecessary to add this method. Why not simply invoking `BorderAgent::SetEnabled` in the callback? Regarding the role change log, there are many places that logged role change.

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +443,49 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
 
     host.Deinit();
 }
+
+TEST(RcpHostApi, ThreadEnabledDisableCallbackInvoked)
```
Irving-cl:
```suggestion
TEST(RcpHostApi, ThreadRoleChangedCallbackInvoked)
```

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +443,49 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+    otInstance *instance = ot::FakePlatform::CurrentInstance();
+
+    // Case 1. check CB is invoked when enabling Thread
+    OT_UNUSED_VARIABLE(otIp6SetEnabled(instance, true));
+    OT_UNUSED_VARIABLE(otThreadSetEnabled(instance, true));
+
+    // set aInitialPump to true to make sure the cb is invoked
+    MainloopProcessUntil(
+        mainloop, /* aTimeoutSec */ 1, [&host]() { return host.GetDeviceRole() != OT_DEVICE_ROLE_DISABLED; },
+        /* aInitialPump */ true);
```
Irving-cl: I don't think we need such parameter. The reason why the callback isn't invoked without the parameter is that the condition has been satisfied right after `otThreadSetEnabled(instance, true)` is invoked. The role becomes `Detached` immediately. But the callback won't be invoked because the mainloop isn't run. So, we can simply run `otbr::MainloopManager::GetInstance().Process(aMainloop);`.

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +443,49 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+
+    EXPECT_TRUE(callbackInvoked);
+    EXPECT_NE(lastRole, OT_DEVICE_ROLE_DISABLED);
```
Irving-cl: The check seems unnecessary since we already uses this as condition for `MainloopProcessUntil`. And the test case seems incomplete. We can verify:
1. The node changes from disabled to detached, callback invoked once.
2. The node changes from detached to leader, callback invoked once. (Note this takes some time to happen and we need `MainloopProcessUntil`).
3. The node changes from leader to disabled, callback invoked once.

src/border_agent/border_agent.cpp
```diff
@@ -327,6 +327,12 @@ void BorderAgent::HandleBorderAgentMeshCoPServiceChanged(bool aIsActive, uint16_
 
 #endif // OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
 
+void BorderAgent::HandleThreadRoleChanged(otDeviceRole aRole)
+{
+    otbrLogInfo("Thread role changed to %s", otThreadDeviceRoleToString(aRole));
+    SetEnabled(aRole != OT_DEVICE_ROLE_DISABLED);
```
abtink: I am not sure this is generally correct.  See https://github.com/openthread/ot-br-posix/issues/3068#issuecomment-3348114872

zesonzhang: Thanks Abtin! I updated this function per our discussion in https://github.com/openthread/ot-br-posix/issues/3068. 

Please kindly let me know if there's anything I might miss or miss-understanding. Thanks!

tests/gtest/test_rcp_host_api.cpp
zesonzhang: updated the test name

tests/gtest/test_rcp_host_api.cpp
zesonzhang: removed this parameter

tests/gtest/test_rcp_host_api.cpp
zesonzhang: updated the test, thanks!

src/host/rcp_host.cpp
zesonzhang: Added, thanks

src/border_agent/border_agent.hpp
zesonzhang: Thanks for the feedback! That's a fair point.

My main reason for adding ·HandleThreadRoleChanged()· is to keep the logic encapsulated within the BorderAgent class. This way, the Application class simply acts as a dispatcher, forwarding the role change event, while the BorderAgent itself contains all the logic for how to respond. I find this approach a bit cleaner and easier to maintain. It also makes it simpler to add more complex logic in the future without cluttering the Application class.

Regarding the log, you're right that role changes are logged elsewhere. I agree the extra role-change log doesn’t add much signal—I’ll drop that to avoid redundant logging.

Irving-cl: I'm ok to keep the method.

But from a modular perspective, I strongly recommend to only provide `SetEnabled` method and not introduce the role logic into `BorderAgent`: The `BorderAgent` is only responsible for publishing the meshcop service. It provides the methods like `SetEnabled` and the user of it calls such methods. But `BorderAgent` itself doesn't decide when to enable/disable. The application organizes and controls these modules. This way we centralize the logic of how modules work with each other.

src/border_agent/border_agent.hpp
```diff
@@ -169,13 +169,29 @@ class BorderAgent : private NonCopyable
...
 
+    /**
+     * This method checks if the device role is attached.
+     *
+     * @param[in] aRole The Thread device role.
+     *
+     * @returns true if the device is attached, false otherwise.
+     */
+    static bool IsAttached(otDeviceRole aRole);
+
```
Irving-cl: This method doesn't need to be public. On the other hand, I recommend not adding this method based on the previous comment.

Irving-cl (approved):
LGTM with some suggestions.

src/border_agent/border_agent.hpp
zesonzhang: Thanks for the detailed feedback and for raising this important architectural point. I really appreciate the discussion on modularity.

Just to confirm my understanding: you're suggesting that the BorderAgent should act as a passive component, only exposing a SetEnabled method. The responsibility for deciding when to enable or disable it (based on the Thread device role) should belong to the Application class. This approach would centralize the application's control logic.

My initial thinking was to align this implementation with the BorderAgent in the OpenThread (https://github.com/openthread/openthread/blob/64e064ffc1883b4e54237d379630366e2a3c4bc0/src/core/meshcop/border_agent.cpp#L133-L143), where it self-manages its state based on the device role. I was aiming for consistency with that established pattern, which might be familiar to developers working with the core stack.

However, you've raised that ot-br-posix is an application, not a low-level library, and maintaining a clear separation of concerns at the application layer is crucial for long-term maintainability. Centralizing the control logic in the Application class is indeed a better fit for the overall architecture here.

I agree this is the better approach. I'll go ahead and refactor the code as you suggested. Thanks again for the great discussion!

zesonzhang: Updated according to this discussion.

src/border_agent/border_agent.hpp
zesonzhang: removed. thanks

src/host/rcp_host.cpp
wgtdkp, line 1: Are the changes in this file related to this PR? It's probably better to move them to a separate PR

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +436,93 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+    std::vector<otDeviceRole> observedRoles;
+    host.AddThreadRoleChangedCallback([&observedRoles](otDeviceRole aRole) { observedRoles.push_back(aRole); });
+
+    otInstance              *instance = ot::FakePlatform::CurrentInstance();
+    otOperationalDataset     dataset;
+    otOperationalDatasetTlvs datasetTlvs;
+
+    OT_UNUSED_VARIABLE(otDatasetCreateNewNetwork(instance, &dataset));
+    otDatasetConvertToTlvs(&dataset, &datasetTlvs);
+    OT_UNUSED_VARIABLE(otDatasetSetActiveTlvs(instance, &datasetTlvs));
```
bukepo:
```suggestion
    ASSERT_EQ(otDatasetSetActiveTlvs(instance, &datasetTlvs), OT_ERROR_NONE);
```

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +436,93 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+    host.Init();
+
+    std::vector<otDeviceRole> observedRoles;
+    host.AddThreadRoleChangedCallback([&observedRoles](otDeviceRole aRole) { observedRoles.push_back(aRole); });
+
+    otInstance              *instance = ot::FakePlatform::CurrentInstance();
+    otOperationalDataset     dataset;
+    otOperationalDatasetTlvs datasetTlvs;
+
+    OT_UNUSED_VARIABLE(otDatasetCreateNewNetwork(instance, &dataset));
```
bukepo:
```suggestion
    ASSERT_EQ(otDatasetCreateNewNetwork(instance, &dataset), OT_ERROR_NONE);
```

src/border_agent/border_agent.hpp
abtink: @zesonzhang, note that OT's native Border Agent you referred to in [code](https://github.com/openthread/openthread/blob/64e064ffc1883b4e54237d379630366e2a3c4bc0/src/core/meshcop/border_agent.cpp#L133-L143) does not enable/disable the BA function on role change. It correctly starts or stops it (which are different from enable/disable). Stop/Start controls the DTLS session and whether we can forward/receive traffic to/from the Thread network (if the device is detached, it cannot send/receive traffic to other nodes in the mesh).

zesonzhang: Thanks @abtink for the clarification!

src/host/rcp_host.cpp
zesonzhang: Thank you for your feedback. The changes in rcp_host.cpp are integral to this PR's functionality and are intentionally included.

The core feature requires the Application layer to be notified of Thread role changes, and the modifications in `rcp_host.cpp` introduce the necessary `ThreadRoleChangedCallback` mechanism to enable this.

Furthermore, introducing a new callback mechanism without its first use-case in the same pull request could be confusing. We believe that presenting the mechanism and its usage together makes the PR a more complete and logically contained feature for reviewers.

For these reasons, we consider these changes to be a single, cohesive feature.

src/agent/application.cpp
```diff
@@ -269,6 +269,17 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
...
+    rcpHost.AddThreadRoleChangedCallback([this, &rcpHost](otDeviceRole aRole) {
+        OT_UNUSED_VARIABLE(aRole);
+        // This ensures the Border Agent is started only when the Thread device
+        // successfully attaches to a network. This aligns with the OpenThread
+        // core behavior and is particularly useful when the agent is initially
+        // disabled (e.g., via OTBR_STOP_BORDER_AGENT_ON_INIT).
+        if (rcpHost.IsAttached())
+        {
+            mBorderAgent.SetEnabled(true);
+        }
```
wgtdkp, line 281: Should Border Agent be disabled when the role is detached?

```suggestion
        mBorderAgent.SetEnabled(rcpHost.IsAttached());
```

abtink (approved):

src/agent/application.cpp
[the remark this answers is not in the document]
zesonzhang: per previous discussion, we should not disable border agent when device is not attached

zesonzhang: Per discussion with @abtink , we do not disable BA when detached. See https://github.com/openthread/ot-br-posix/issues/3068

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +436,93 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+    host.Init();
+
+    bool borderAgentEnabled = false;
+    host.AddThreadRoleChangedCallback([&host, &borderAgentEnabled](otDeviceRole aRole) {
+        OT_UNUSED_VARIABLE(aRole);
+        if (host.IsAttached())
+        {
+            borderAgentEnabled = true;
+        }
+    });
```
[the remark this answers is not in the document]
zesonzhang: we only enable border agent when role is attached

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -436,3 +436,93 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
+    ASSERT_EQ(otIp6SetEnabled(instance, true), OT_ERROR_NONE);
+    ASSERT_EQ(otThreadSetEnabled(instance, true), OT_ERROR_NONE);
+
+    EXPECT_FALSE(borderAgentEnabled);
+
+    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 5, [&borderAgentEnabled]() { return borderAgentEnabled; });
+
+    EXPECT_TRUE(host.IsAttached());
+    EXPECT_TRUE(borderAgentEnabled);
+
```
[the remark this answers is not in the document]
zesonzhang: same. we do not disable border agent when thread is not attached

jwhui (approved):

PR #3069, opened 2025-09-28
[ncp] add local port to udp forward callback

Irving-cl:
This PR adds one more parameter `aLocalPort` to `UdpForwardToHostCallback`.

Currently we only use udp proxy for border agent. As we are going to support Epskc and TREL, more UDP proxies will be required. We need to have the local port to find the right UDP proxy to send the UDP packet.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -383,10 +383,13 @@ void Application::InitNcpMode(void)
             OTBR_UNUSED_VARIABLE(aLength);
 #endif
         });
-    mHost.SetUdpForwardToHostCallback(
-        [this](const uint8_t *aUdpPayload, uint16_t aLength, const otIp6Address &aPeerAddr, uint16_t aPeerPort) {
+    mHost.SetUdpForwardToHostCallback([this](const uint8_t *aUdpPayload, uint16_t aLength,
+                                             const otIp6Address &aPeerAddr, uint16_t aPeerPort, uint16_t aLocalPort) {
+        if (aLocalPort == mBorderAgentUdpProxy.GetThreadPort())
+        {
             mBorderAgentUdpProxy.SendToPeer(aUdpPayload, aLength, aPeerAddr, aPeerPort);
-        });
+        }
+    });
     SetBorderAgentOnInitState();
 #endif
 #if OTBR_ENABLE_BACKBONE_ROUTER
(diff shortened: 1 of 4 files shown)


src/agent/application.cpp
```diff
@@ -383,10 +383,13 @@ void Application::InitNcpMode(void)
...
-    mHost.SetUdpForwardToHostCallback(
-        [this](const uint8_t *aUdpPayload, uint16_t aLength, const otIp6Address &aPeerAddr, uint16_t aPeerPort) {
+    mHost.SetUdpForwardToHostCallback([this](const uint8_t *aUdpPayload, uint16_t aLength,
+                                             const otIp6Address &aPeerAddr, uint16_t aPeerPort, uint16_t aLocalPort) {
+        if (aLocalPort == mBorderAgentUdpProxy.GetThreadPort())
+        {
             mBorderAgentUdpProxy.SendToPeer(aUdpPayload, aLength, aPeerAddr, aPeerPort);
-        });
+        }
+    });
```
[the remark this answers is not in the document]
Irving-cl: This is done by clang format. Let's keep the style.

src/host/ncp_spinel.hpp
```diff
@@ -108,7 +108,8 @@ class NcpSpinel : public CliDaemon::Dependencies
     using InfraIfSendIcmp6NdCallback = std::function<void(uint32_t, const otIp6Address &, const uint8_t *, uint16_t)>;
     using BorderAgentMeshCoPServiceChangedCallback = std::function<void(bool, uint16_t, const uint8_t *, uint16_t)>;
     using CliDaemonOutputCallback                  = std::function<void(const char *)>;
-    using UdpForwardSendCallback = std::function<void(const uint8_t *, uint16_t, const otIp6Address &, uint16_t)>;
+    using UdpForwardSendCallback =
+        std::function<void(const uint8_t *, uint16_t, const otIp6Address &, uint16_t, uint16_t)>;
```
bukepo, line 121:
```suggestion
        std::function<void(const uint8_t *, uint16_t, const otIp6Address &, uint16_t, const UdpForward &)>;
```

Irving-cl: I don't think we should use an object reference here. Because the callback is invoked from NcpSpinel when it receives the `UDP_FORWARD_STREAM`. So `NcpSpinel` doesn't have the access to the `UdpProxy` object. If setting the UdpProxy reference into `NcpSpinel`, it introduces extra dependency and initialization step, which seems unnecessary. Thoughts?

bukepo (approved):

hemanth-silabs (approved):
LGTM 👍

abtink (approved):
👍

jwhui (approved):

PR #3091, opened 2025-10-16
[script] Remove systemctl usage inside Docker

db00lean:
While some containers can include the systemctl
package, it is often non-functional or behaves
differently. This commit updates the scripts to
avoid using systemctl inside Docker.

### DIFF
--- a/script/_initrc
+++ b/script/_initrc
@@ -84,7 +84,7 @@ without()
 }
 
 HAVE_SYSTEMCTL=0
-if have systemctl; then
+if have systemctl && without DOCKER; then
     HAVE_SYSTEMCTL=1
 fi
 HAVE_SERVICE=0
--- a/script/_otbr
+++ b/script/_otbr
@@ -50,7 +50,7 @@ readonly REFERENCE_DEVICE
 
 otbr_uninstall()
 {
-    if have systemctl; then
+    if have systemctl && without DOCKER; then
         sudo systemctl stop otbr-web || true
         sudo systemctl stop otbr-agent || true
         sudo systemctl disable otbr-web || true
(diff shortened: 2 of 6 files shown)


jwhui:
@db00lean , thanks for submitting this PR. Out of curiosity, can you provide more detail about how you're using `ot-br-posix` in a Docker environment?

Note that we generally recommend using [`openthread/border-router`](https://github.com/openthread/ot-br-posix/blob/main/etc/docker/border-router/Dockerfile) image, which does not make use of these setup scripts.

bukepo (commented):
My concern for this change is `systemctl` may work in some docker containers. Perhaps we can do some duck test for systemctl if the current environment is DOCKER and disable using it if it fails the duck test.

script/_initrc
```diff
@@ -84,7 +84,7 @@ without()
 }
 
 HAVE_SYSTEMCTL=0
-if have systemctl; then
+if have systemctl && without DOCKER; then
```
jwhui, line 87: Rather than repeating `have systemctl && without DOCKER` across all the individual script files, let's centralize the check here.

For example, define a helper function or rely on `HAVE_SYSTEMCTL`:
```
have_systemctl()
{
    have systemctl && without DOCKER
}
```
(or `if have systemctl && without DOCKER; then HAVE_SYSTEMCTL=1; fi` and use `[[ ${HAVE_SYSTEMCTL} == 1 ]]`).

Also, consider whether checking `[[ -d /run/systemd/system ]]` (or checking whether PID 1 is systemd) is preferable so that container environments without `DOCKER=1` exported are also handled automatically.

script/_otbr
jwhui, line 167: In a Docker environment, this warning is misleading because systemctl may exist and container services are typically managed differently. Can we either guard the warning with if without DOCKER; then or update the message?

jwhui:
@db00lean , can you rebase on the latest main branch and resolve conflicts?

PR #3109, opened 2025-11-03
[config] enable `OT_MESSAGE_USE_HEAP`

abtink:
This allows for message buffers to be allocated from the heap instead of using the pre-allocated message pool. This is the recommended configuration for more capable platforms that are not resource-limited.

Fixes https://github.com/openthread/openthread/issues/12088

### DIFF
--- a/third_party/openthread/CMakeLists.txt
+++ b/third_party/openthread/CMakeLists.txt
@@ -59,6 +59,7 @@ set(OT_LINK_METRICS_MANAGER ${OTBR_LINK_METRICS_TELEMETRY} CACHE STRING "enable 
 set(OT_LOG_LEVEL_DYNAMIC ON CACHE STRING "enable dynamic log level control" FORCE)
 set(OT_MAC_FILTER ON CACHE STRING "enable MAC filter" FORCE)
 set(OT_MESH_DIAG ON CACHE STRING "enable Mesh Diagnostics" FORCE)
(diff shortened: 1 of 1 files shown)


third_party/openthread/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ set(OT_LINK_METRICS_MANAGER ${OTBR_LINK_METRICS_TELEMETRY} CACHE STRING "enable
 set(OT_LOG_LEVEL_DYNAMIC ON CACHE STRING "enable dynamic log level control" FORCE)
 set(OT_MAC_FILTER ON CACHE STRING "enable MAC filter" FORCE)
 set(OT_MESH_DIAG ON CACHE STRING "enable Mesh Diagnostics" FORCE)
+set(OT_MESSAGE_USE_HEAP ON CACHE STRING "use heap for message buffers" FORCE)
```
jwhui, line 62: Should we enable this by default in the core OT repo under `src/posix`?

abtink: I actually considered that first. Was initially think to add it to `openthread-core-posix-config.h`. But then all configs in this header can possibly be overwritten by the corresponding CMake option.

Adding it here in the `ot-br-posix` repo, where we integrate `openthread` itself and set all other `OT` configs, felt like a better and safer approach.

PR #3117, opened 2025-11-07
[border-agent] add MeshCop `vmail` TXT record to indicate multi-AIL indication

yangsong-cnyn:
This PR introduces a `vmail` (vendor multi-AIL) TXT record to the MeshCoP service advertisement. This boolean record indicates if the BR has detected a multi-AIL condition.

This commit also refactors vendor TXT entry management to persist non-standard vendor specific txt entries during updates.

### DIFF
--- a/src/host/rcp_host.cpp
+++ b/src/host/rcp_host.cpp
@@ -339,6 +339,10 @@ void RcpHost::Deinit(void)
 {
     assert(mInstance != nullptr);
 
+#if OTBR_ENABLE_MULTI_AIL
+    otBorderRoutingSetMultiAilCallback(mInstance, nullptr, nullptr);
+#endif
+
     otSysDeinit();
     mInstance = nullptr;
 
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -297,6 +297,14 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
     mHost.AddEphemeralKeyStateChangedCallback([this](otBorderAgentEphemeralKeyState aEpskcState, uint16_t aPort) {
         mBorderAgent.HandleEpskcStateChanged(aEpskcState, aPort);
     });
+#if OTBR_ENABLE_MULTI_AIL
+    otBorderRoutingSetMultiAilCallback(
+        rcpHost.GetInstance(),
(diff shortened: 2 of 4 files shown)


src/host/rcp_host.cpp
```diff
@@ -339,6 +339,10 @@ void RcpHost::Deinit(void)
 {
     assert(mInstance != nullptr);
 
+#if OTBR_ENABLE_MULTI_AIL
+    otBorderRoutingSetMultiAilCallback(mInstance, nullptr, nullptr);
+#endif
```
abtink, line 344: This works, but instead of adding it to `ot-br-posix`, how about adding it to the OT core? We can add it to `BorderAgent::TxtData`. The OT core already helps prepare a good portion of `TXT Data`. This way, it is more common.

Also we dont end up usig the single `MultiAilCallback` for BA TXT data prepreation and have it be available for other use-cases.

yangsong-cnyn: @librasungirl do you think we should add `vmail` entry to OT core at the moment? If adding to OT core, how about removing the vendor tag and using `mail` txt.

bukepo: I think the `vmail` is vendor specific, which should be treated meaningful only when the `vo` equals to some particular value. Is it proper to be included in this project? @abtink

abtink: I dont see any issue (as long as it is optional and it is tied to proper build flags).

bukepo: The `v` prefix in theory is only meaningful to a specific vendor. It may conflict with the current or future vendor specific usage. If we'd put it into core or this project, I'd suggest we use a different prefix, for example `_`(We'd raise a discussion in Thread Group to make this an indication for experimental entries). Thoughts?

abtink: Sounds good.

librasungirl: good suggestions with `_` prefix and adding it to OT core. @yangsong-cnyn

yangsong-cnyn: As we all agree to use `_` prefix instead of vendor-specific `v`. Will close this PR and add `_mail` to meshcop txt in OT core

PR #3122, opened 2025-11-10
[script] refactor DHCPv6 PD client setting to support different client configurations

suveshpratapa:
- Replaces `DHCPV6_PD_REF` with `OTBR_DHCP6_PD_CLIENT`:
   - "openthread": Use OpenThread's native DHCPv6 PD client
   - "dhcpcd": Use external dhcpcd implementation
   - "none": Disable DHCPv6 PD client (default, matches previous behavior for CI and tests)

- Rename `script/_dhcpv6_pd_ref` to `script/_dhcpcd` for clarity


examples/platforms/raspbian/default
```diff
@@ -28,6 +28,5 @@
 #
 
 # shellcheck disable=SC2034
-DHCPV6_PD_REF=0
```
abtink, line 31: Overall this change looks good, but I have some questions about the previous config:

Is this config a public-facing configuration? Is it possible that some projects may already be using this, and this change (removing `DHCPV6_PD_REF`) could now break their build configuration?

If that is possible, I would suggest either keeping it or, if we want to remove it, consider adding a way to detect if `DHCPV6_PD_REF` is being used. We could then add a `warning` or `error` with a message indicating that this is now removed and the project should switch to the new approach of using `OTBR_DHCP6_PD_CLIENT`.

Thoughts?

suveshpratapa: @abtink

Yes, it was a primary flag for OTBR bootstrap and setup. We have had to set this flag for OTBR devices requiring PD client functionality in certification tests. The name `DHCPV6_PD_REF` consequently became a bit misleading since it wasn't just for reference devices, but more like it was a reference PD client implementation that used `dhcpcd` and `radvd`. Now we have a native implementation that can replace it, so I wanted the configuration option to be coherent.

I have added a backward compatibility / migration warning to the user if we detect the `DHCPV6_PD_REF` flag from now on.

For reference, I was using some previous examples of similar settings being removed (for example https://github.com/openthread/ot-br-posix/pull/2842 and https://github.com/openthread/ot-br-posix/pull/2978) where no such warnings were included.  Though now I see they are a bit different since they are completely removing those flags rather than replacing them.

abtink: Thanks.

PR #3132, opened 2025-11-17
[border-agent] apply id from application when using OT mdns

librasungirl:


src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1338,6 +1338,26 @@ void DBusThreadObjectRcp::UpdateMeshCopTxtHandler(DBusRequest &aRequest)
     {
         VerifyOrExit(!update.count(reservedKey), error = OT_ERROR_INVALID_ARGS);
     }
+
+#if !OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
```
abtink: How about if we set the ID regardless? 

There should be harm in having the ID set on OT BA module (even if its TXT data is not used) This way we know the same consistent ID is used.

abtink: We actually have `OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE` (which is enabled by default) and controls whether `otBorderAgentSetId/GetId()` is provided in OT core `BorderAgent` module. So maybe we can check the same config here.

librasungirl: Good suggestion! Update in new patch.

PR #3136, opened 2025-11-20
[dbus] add dbus API to set border agent Id

Irving-cl:
This PR adds a DBUS API to set the Border Agent Id.

This is required for GTBR implementation. The PR also adds a build option `OTBR_BORDER_AGENT_ID` to control whether to enable the Border Agent Id in openthread and the related code in ot-br-posix.

### DIFF
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -54,6 +54,11 @@ if (OTBR_BORDER_AGENT_MESHCOP_SERVICE)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE=1)
 endif()
 
+option(OTBR_BORDER_AGENT_ID "Enable Border Agent ID support" ON)
+if (OTBR_BORDER_AGENT_ID)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_BORDER_AGENT_ID=1)
+endif()
+
 option(OTBR_BACKBONE_ROUTER "Enable Backbone Router" ON)
 if (OTBR_BACKBONE_ROUTER)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_BACKBONE_ROUTER=1)
--- a/src/dbus/client/thread_api_dbus.cpp
+++ b/src/dbus/client/thread_api_dbus.cpp
@@ -893,6 +893,11 @@ ClientError ThreadApiDBus::SetBorderAgentEnabled(bool aEnabled)
     return CallDBusMethodSync(OTBR_DBUS_SET_BORDER_AGENT_ENABLED_METHOD, std::tie(aEnabled));
 }
 
+ClientError ThreadApiDBus::SetBorderAgentId(const std::vector<uint8_t> &aBorderAgentId)
+{
+    return SetProperty(OTBR_DBUS_PROPERTY_BORDER_AGENT_ID, aBorderAgentId);
+}
+
 ClientError ThreadApiDBus::UpdateVendorMeshCopTxtEntries(std::vector<TxtEntry> &aUpdate)
 {
     auto args = std::tie(aUpdate);
--- a/src/dbus/client/thread_api_dbus.hpp
+++ b/src/dbus/client/thread_api_dbus.hpp
@@ -801,6 +801,17 @@ public:
     ClientError SetBorderAgentEnabled(bool aEnabled);
 
     /**
+     * This method sets the 16 byte Border Agent Id.
+     *
+     * @param[in] aBorderAgentId  The Border Agent Id.
+     *
+     * @retval ERROR_NONE  Successfully performed the dbus function call
+     * @retval ERROR_DBUS  dbus encode/decode error
+     * @retval ...         OpenThread defined error value otherwise
+     */
+    ClientError SetBorderAgentId(const std::vector<uint8_t> &aBorderAgentId);
+
+    /**
      * This method sets multiple vendor-specific entries for the TXT record of the MeshCoP service.
      *
      * @note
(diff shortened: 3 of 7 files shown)


src/dbus/client/thread_api_dbus.cpp
```diff
@@ -893,6 +893,10 @@ ClientError ThreadApiDBus::SetBorderAgentEnabled(bool aEnabled)
     return CallDBusMethodSync(OTBR_DBUS_SET_BORDER_AGENT_ENABLED_METHOD, std::tie(aEnabled));
 }
 
+ClientError ThreadApiDBus::SetBorderAgentId(const std::vector<uint8_t> &aBorderAgentId)
+{
+    return SetProperty(OTBR_DBUS_PROPERTY_BORDER_AGENT_ID, aBorderAgentId);
+}
```
zhanglongxia, line 899: nit: add an empty line after this function

zhanglongxia (approved):
LGTM

src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1308,6 +1314,23 @@ otError DBusThreadObjectRcp::SetRadioRegionHandler(DBusMessageIter &aIter)
...
+otError DBusThreadObjectRcp::SetBorderAgentIdHandler(DBusMessageIter &aIter)
+{
+    otError                                        error = OT_ERROR_NONE;
+    std::array<uint8_t, OT_BORDER_AGENT_ID_LENGTH> data;
+    otBorderAgentId                                borderAgentId;
+
+    VerifyOrExit(DBusMessageExtractFromVariant(&aIter, data) == OTBR_ERROR_NONE, error = OT_ERROR_INVALID_ARGS);
+    memcpy(&borderAgentId.mId, &data.front(), sizeof(borderAgentId.mId));
+
+    error = otBorderAgentSetId(mHost.GetInstance(), &borderAgentId);
```
zhanglongxia, line 1327: Will this conflict with the https://github.com/openthread/ot-br-posix/pull/3132?

Irving-cl: I didn't sync code so didn't see https://github.com/openthread/ot-br-posix/pull/3132 yesterday.

https://github.com/openthread/ot-br-posix/pull/3132 can work but it seems a little unnatural to me to set it in  `UpdateMeshCopTxtHandler` because  `UpdateMeshCopTxtHandler` is used to update the external fields of the meshcop txt records while `borderAgentId` is an internal field (though it can be set by an API). The fields set in `UpdateMeshCopTxtHandler` will be taken to integrate into the txt data that provided by OT. But `borderAgentId` shouldn't be handled like that here because it's already in the txt data provided by OT. And in current code, a workaround is taken to do so. So it's like, we avail this API `UpdateMeshCopTxtHandler` and call a OT API here.

So I prefer having a dedicated API to set border agent Id. This PR will ONLY conflict with https://github.com/openthread/ot-br-posix/pull/3132 when both APIs are called to set the border agent Id. If we:
* Set all vendor fields and border agent Id in `UpdateMeshCopTxtHandler`, it's ok
* Set all vendor fields except border agent Id in `UpdateMeshCopTxtHandler` and set border agent id using this dedicated API, it's also ok.

So we can have both PRs.

zhanglongxia: Based on the description of the function [otBorderAgentSetVendorTxtData()](https://github.com/openthread/openthread/blob/main/include/openthread/border_agent.h#L314), the vendor-specific Key-Value TXT data pairs use TXT keys starting with `v`. The function [UpdateMeshCopTxtHandler()](https://github.com/openthread/ot-br-posix/blob/main/src/dbus/server/dbus_thread_object_rcp.cpp#L154) is used to process the vendor  TXT data. The key `id` is not the vendor-specific Key. I'm fine with this PR.

etc/cmake/options.cmake
```diff
@@ -54,6 +54,11 @@ if (OTBR_BORDER_AGENT_MESHCOP_SERVICE)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE=1)
 endif()
 
+option(OTBR_BORDER_AGENT_ID "Enable Border Agent ID support" ON)
```
librasungirl, line 57: is it necessary?

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -800,6 +800,17 @@ class ThreadApiDBus
      */
     ClientError SetBorderAgentEnabled(bool aEnabled);
 
+    /**
+     * This method sets the 16 byte Border Agent Id.
```
librasungirl, line 804: pls add notes as https://github.com/openthread/openthread/blob/main/include/openthread/border_agent.h#L340  and highlight this API should be used with awareness and caution.

src/dbus/server/dbus_thread_object_rcp.cpp
librasungirl: I am not sure about it.

As from spec `An ID to uniquely identify the device hosting the Border Agent among multiple such devices.`, if manufacture want to manage ID themselves, it might also be vendor specific data.

Overall it's fine to add a separate dbus API but it should be well documented as in OT, people should be aware of the requirement of `id` by spec

> The ID MUST remain identical if the device reboots. It MUST NOT persist after a BR factory-reset. It MUST be computed using a random function to minimize the ID collision probability.

etc/cmake/options.cmake
Irving-cl: Since openthread has an option `OT_BORDER_AGENT_ID` to turn on/off Border Agent Id,  I think it's necessary to add an option in ot-br-posix as well to avoid the case where `OT_BORDER_AGENT_ID` is OFF.

BTW, I'm thinking if border agent id is a MUST, why not removing the macro in openthread as well. Thoughts?

librasungirl: It's indeed `MUST` requirement per specification, 
> TXT record inclusion is REQUIRED. The ID MUST remain identical if the device reboots. It MUST NOT persist after a BR factory-reset. It MUST be computed using a random function to minimize the ID collision probability.

But the spec doesn't define who is to maintain it. OT core provides the option to help generate/set/store it and when OT is responsible for meshcop service, OPENTHREAD_CONFIG_BORDER_AGENT_MESHCOP_SERVICE_ENABLE, it will report error if it's not ON.

https://github.com/openthread/openthread/blob/main/src/core/meshcop/border_agent.hpp#L80

For BR,  application can specify it via UpdateMeshCop UpdateMeshCopTxt or use this new dbus API you are introducing if it want to manage the `id`

PR #3143, opened 2025-11-24
dbus: handle disabled border agent ID feature

zesonzhang:
This pull request adds conditional support for the Border Agent ID feature based on the `OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE` configuration flag. The main changes ensure that related API calls and handlers behave correctly depending on whether the feature is enabled or not.

Conditional Border Agent ID support:

* Updated `DBusThreadObjectRcp::GetBorderAgentIdHandler` in `dbus_thread_object_rcp.cpp` to only execute Border Agent ID logic if `OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE` is enabled; otherwise, it returns `OT_ERROR_NOT_IMPLEMENTED`. [[1]](diffhunk://#diff-72a8adb55bd6172c3a74e4255b5036b6b034fa90ae74d5589c874591ce8c8343R1669) [[2]](diffhunk://#diff-72a8adb55bd6172c3a74e4255b5036b6b034fa90ae74d5589c874591ce8c8343R1682-R1685)

Test coverage improvements:

* Modified `test_dbus_client.cpp` to add conditional test assertions for Border Agent ID: tests for successful retrieval when enabled, and for proper error handling when disabled.The GetBorderAgentIdHandler now returns OT_ERROR_NOT_IMPLEMENTED if OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE is disabled.

This avoids compilation errors when the otBorderAgentGetId function is not available.

### DIFF
--- a/src/dbus/server/dbus_thread_object_rcp.cpp
+++ b/src/dbus/server/dbus_thread_object_rcp.cpp
@@ -1666,6 +1666,7 @@ exit:
 
 otError DBusThreadObjectRcp::GetBorderAgentIdHandler(DBusMessageIter &aIter)
 {
+#if OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE
     otBorderAgentId      id;
     std::vector<uint8_t> data;
     otError              error = OT_ERROR_NONE;
@@ -1678,6 +1679,10 @@ otError DBusThreadObjectRcp::GetBorderAgentIdHandler(DBusMessageIter &aIter)
 
 exit:
     return error;
+#else
+    OTBR_UNUSED_VARIABLE(aIter);
+    return OT_ERROR_NOT_IMPLEMENTED;
+#endif
 }
 
 otError DBusThreadObjectRcp::GetOtRcpVersionHandler(DBusMessageIter &aIter)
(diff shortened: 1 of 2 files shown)


src/dbus/server/dbus_thread_object_rcp.cpp
```diff
@@ -1678,6 +1679,10 @@ otError DBusThreadObjectRcp::GetBorderAgentIdHandler(DBusMessageIter &aIter)
 
 exit:
     return error;
+#else
+    OTBR_UNUSED_VARIABLE(aIter);
+    return OT_ERROR_NOT_IMPLEMENTED;
+#endif
```
abtink, line 1685: The change in this PR looks good to me.

One suggestion: if we are adding `#if OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE`, let's add it in other places where `otBorderAgentGet/SetId()` is used.

I recall this recent PR, https://github.com/openthread/ot-br-posix/pull/3132, where `otBorderAgentSetId()` is called. Should we also add the same `#if` check there?

Alternatively, we can require that `ot-br-posix` always enables and uses `OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE` (maybe even force enable it in `posix-config`). Then we would not need these `#if` checks.

Thoughts?

zesonzhang: Thanks for feedback Abtin! 

Had a offline chat with @librasungirl , requires `OPENTHREAD_CONFIG_BORDER_AGENT_ID_ENABLE` enabled in ot-br-posix sounds good, considering `otBorderAgentGetId` is called in multiple places.

abtink: Sounds good. Should this PR be closed then?

zesonzhang: Sure, I will close this. Thanks again!

PR #3146, opened 2025-11-25
[ncp] add TREL support for NCP host

suveshpratapa:
- Bind to TREL UDP port and publish `_trel._udp` service when TREL is enabled.
- Handle incoming TREL/mDNS packets and forward them to NCP using existing UDP forward mechanism.

### DIFF
--- a/src/host/ncp_host.cpp
+++ b/src/host/ncp_host.cpp
@@ -30,12 +30,24 @@
 
 #include "ncp_host.hpp"
 
+#include <errno.h>
+#include <fcntl.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
 #include <memory>
 
 #include <openthread/error.h>
 #include <openthread/thread.h>
 
 #include <openthread/openthread-system.h>
+
+#include "common/code_utils.hpp"
+
+#if OTBR_ENABLE_MDNS
+#include "mdns/mdns.hpp"
+#endif
 
 #include "host/async_task.hpp"
 #include "lib/spinel/spinel_driver.hpp"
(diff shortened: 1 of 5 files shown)


src/host/ncp_host.cpp
```diff
@@ -400,6 +443,312 @@ void NcpHost::InitNetifCallbacks(Netif &aNetif)
...
+    {
+        uint16_t actualPort = ntohs(addr.sin6_port);
+        mTrelSocket.mPort   = actualPort;
+
+        // If actual port differs from requested, update NCP
+        if (actualPort != aPort)
+        {
+            mNcpSpinel.UpdateTrelState(true, actualPort);
+        }
+    }
```
Irving-cl, line 652: This is not ideal because the port originates from NCP but the host needs to send back the port again. This affects the performance and makes error handling very difficult and complex. I understand this is just a temporary solution. But I still recommend using port mapping and the `UdpProxy` provided in ot-br-posix. The BorderAgent does exactly the same thing. With `UdpProxy`, we don't need to implement things like `OpenTrelSocket` again.

I understand your challenge of using port mapping is probably that you want to publish the TrelDnssd from Ncp side as well so you have to know the real port on NCP side. But actually we can still use the TrelDnssd on the host side in ot-br-posix to publish the Trel service. Meanwhile, we use the mdns_socket you added in openthread on NCP side to do the TREL discovery. I think this should work.

Anyway, this is a temporary solution. Overall I think it can work.

Irving-cl (changes requested):
As discussed offline, this is a temporary solution and we won't merge it in main branch. But a temporary solution is needed due to timeline. A stable solution now depends the dnssd discovery on NCP.

PR #3147, opened 2025-12-01
[main] add callback to RealMain for scriptless system configuration

yangsong-cnyn:
This PR refactors `otbr::RealMain` to introduce a callback mechanism, enabling C++ application to perform necessary system configurations before the OTBR agent application starts. 

This change is crucial for deployments where `otbr-agent` is shipped as a self-contained binary without the accompanying shell scripts. To enable platform-specific logic (e.g., system configurations) to be executed by the caller of `otbr::RealMain`, it's necessary to provide a hook before the main application logic begins. This change provides that hook via a callback function, as well as the `MainConfig` struct to hold the command-line arguments.

### DIFF
--- a/src/agent/realmain.hpp
+++ b/src/agent/realmain.hpp
@@ -34,7 +34,11 @@
 #ifndef OTBR_AGENT_REAL_MAIN_HPP_
 #define OTBR_AGENT_REAL_MAIN_HPP_
 
+#include <functional>
 #include <stdint.h>
+#include <vector>
+
+#include "common/logging.hpp"
 
 namespace otbr {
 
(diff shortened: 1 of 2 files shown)


Irving-cl (commented):
As I understand, these system configurations should be done by init.rc script. I think it's a standard practice and I don't see any problems with such practice. What specific operations needs to be done by a callback in the otbr-agent program?

From my perspective, system config operations are easier executed in script and is harder to be executed with c++ code. Plus we need to implement the callback function for each platform which takes more effort than writing the init.rc script. Thoughts? @yangsong-cnyn

src/agent/realmain.hpp
```diff
@@ -47,15 +51,32 @@ namespace otbr {
...
+    bool                      mVerbose;               ///< Whether to enable verbose logging.
+    bool                      mSyslogDisable;         ///< Whether to disable syslog and print to standard out.
+    bool                      mPrintRadioVersion;     ///< Whether to print radio version and exit.
+    bool                      mEnableAutoAttach;      ///< Whether to enable auto attach.
+    const char               *mRestListenAddress;     ///< REST server listen address.
+    int                       mRestListenPort;        ///< REST server listen port.
+    std::vector<const char *> mRadioUrls;             ///< Radio URLs.
+};
+
+using MainConfigCallback = std::function<void(const MainConfig &aConfig)>;
```
zhanglongxia, line 68: How about changing the name to `OtbrConfigCallback` and `OtbrConfig `

yangsong-cnyn: Since we access it as `otbr::MainConfig`, using `OtbrConfig` looks not so nice (i.e., `otbr::OtbrConfig`)

yangsong-cnyn:
> > > As I understand, these system configurations should be done by init.rc script. I think it's a standard practice and I don't see any problems with such practice. What specific operations needs to be done by a callback in the otbr-agent program?
> > > From my perspective, system config operations are easier executed in script and is harder to be executed with c++ code. Plus we need to implement the callback function for each platform which takes more effort than writing the init.rc script. Thoughts? @yangsong-cnyn
> > 
> > 
> > For standard daemons, using initialization scripts is the conventional way to set up the environment.
> > So this callback is designed for the scenario when `otbr-agent` is deployed as a self-contained binary without the surrounding script infrastructure. In this situation, the C++ host application takes full responsibility for setting up the system. The `MainConfigCallback` provides the necessary hook to do this cleanly.
> 
> Thanks. I have another question: What will be the operations intended to do in the callback be like? How are they specific to different platforms? I heard something like "find the infra interface" when talking. Can these operations be done with the same posix implementations? If so we don't need a callback here. @yangsong-cnyn

The callback allows C++ application to perform system configurations that are currently handled by scripts, such as `_nat64`, `_ipforward`, `_border_routing`, etc which are fundamentally Linux pattern.

On other platform like Android it is different. So using callback allows us to specifically make the configurations in the platform's function which calls `otbr::RealMain`.

PR #3170, opened 2025-12-24
[dbus] unregister the dbus object from the dbus library

zhanglongxia:
When running the `factoryreset` command on RaspberryPi, the system generates the following error message: `otbr[2862]: [EMERG]-DBUS----: FAILED src/dbus/server/dbus_agent.cpp:85 - Failed to initialize DBus Agent`

This root cause is that the dbus object was not unregistered from the dbus library when running the `factoryreset` command. When otbr restarts, it registers the dbus object again. So the dbus library generates the error `org.freedesktop.DBus.Error.ObjectPathInUse`.

This commit add `Deinit()` method to unregister the dbus object from the dbus library.

### DIFF
--- a/src/dbus/server/dbus_object.hpp
+++ b/src/dbus/server/dbus_object.hpp
@@ -100,6 +100,13 @@ public:
     virtual otbrError Init(void);
 
     /**
+     * This method de-initializes the d-bus object.
+     *
+     * This method will unregister the object from the d-bus library.
+     */
+    virtual void Deinit(void);
+
+    /**
      * This method registers the method handler.
      *
      * @param[in] aInterfaceName  The interface name.
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -123,6 +123,10 @@ void Application::Init(const std::string &aRestListenAddress, int aRestListenPor
 
 void Application::Deinit(void)
 {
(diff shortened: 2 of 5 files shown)


src/dbus/server/dbus_object.hpp
```diff
@@ -99,6 +99,14 @@ class DBusObject : private NonCopyable
      */
     virtual otbrError Init(void);
 
+    /**
+     * This method de-initializes the d-bus object.
+     *
+     * This method will unregister the object from the d-bus library.
+     *
+     */
+    virtual void Deinit(void);
```
bukepo, line 107:
```suggestion
    void Deinit(void);
```

bukepo (commented):
Is this only seen with pseudo reset?

bukepo (approved):

src/dbus/server/dbus_object.hpp
```diff
@@ -99,6 +99,14 @@ class DBusObject : private NonCopyable
      */
     virtual otbrError Init(void);
 
+    /**
+     * This method de-initializes the d-bus object.
+     *
+     * This method will unregister the object from the d-bus library.
+     *
+     */
```
jwhui, line 106:
```suggestion
     * This method will unregister the object from the d-bus library.
     */
```

jwhui (approved):

PR #3152, opened 2025-12-04
[posix] allow set vendor name and product name at runtime

wtongze:
This commit allows users to pass the vendor name and product name through command line flags `--vendor-name` and `--model-name` when starting the `otbr-agent` if `OTBR_VENDOR_NAME` and `OTBR_PRODUCT_NAME` are not defined at compile time.

These flags have no default value and will exit the program if not set.

When building using `script/cmake-build`, vendor name and model name will not be defined by default. However, when building with `script/setup`, the `OTBR_VENDOR_NAME` and `OTBR_PRODUCT_NAME` will be set to `OpenThread` and `BorderRouter` respectively to preserve backward compatibility.

This PR aims to replace <https://github.com/openthread/openthread/pull/12186>

PR #3179, opened 2026-01-04
[build] set proper vendor name and product name

wtongze:
Currently, ot-br-posix don't set `OT_VENDOR_NAME` and `OT_VENDOR_MODEL` in the underlying openthread if  `OTBR_VENDOR_NAME` and `OTBR_PRODUCT_NAME` are defined. This results in inconsistent vendor name / model name in network diagnostics. (As mentioned in <https://github.com/openthread/ot-br-posix/pull/3152#pullrequestreview-3596663494>)

This commit allows `OTBR_VENDOR_NAME` and `OTBR_PRODUCT_NAME` to be passed to openthread (if both are defined) as `OT_VENDOR_NAME` and `OT_VENDOR_MODEL` in order to have proper vendor name and product name when using `vendor name` and `vendor model` commands in `ot-ctl` to retrive it.

If `OTBR_VENDOR_NAME` and `OTBR_PRODUCT_NAME` aren't both defined at build time, it will enable `OT_NETDIAG_VENDOR_INFO` to set vendor name / product name at runtime.

This commit also allows the `script/setup` to read and set `OTBR_VENDOR_NAME` and `OTBR_PRODUCT_NAME` from environment variables while preserving backward compatibility.

### DIFF
--- a/src/host/thread_host.hpp
+++ b/src/host/thread_host.hpp
@@ -351,7 +351,7 @@ public:
 
 #ifndef OTBR_VENDOR_NAME
     /**
-     * This method sets the netdiag vendor name.
+     * This method sets the vendor name.
      *
      * @param[in] aVendorName   The vendor name.
      *
@@ -365,7 +365,7 @@ public:
 
 #ifndef OTBR_PRODUCT_NAME
     /**
-     * This method sets the netdiag vendor model.
+     * This method sets the vendor model.
      *
      * @param[in] aVendorModel   The vendor model.
      *


third_party/openthread/CMakeLists.txt
```diff
@@ -105,6 +105,13 @@ else()
 endif()
 
 if (OTBR_BORDER_AGENT)
+    if(OTBR_VENDOR_NAME AND OTBR_PRODUCT_NAME)
```
librasungirl: is it necessary to put inside `if (OTBR_BORDER_AGENT)` condition?

wtongze: Good point. I've removed `OTBR_BORDER_AGENT=ON` (`OTBR_ENABLE_BORDER_AGENT`) as the pre-condition for this part and for both `--vendor-name` and `--model-name` options.

src/agent/realmain.cpp
```diff
@@ -420,13 +408,19 @@ static int realmain(int argc, char *argv[])
         snprintf(instanceName, sizeof(instanceName), "%s %s", vendorName, productName);
         app.GetBorderAgent().SetMeshCoPServiceValues(instanceName, productName, vendorName, {}, {});
 #endif
+#endif
+#endif
+
+        app.Init(restListenAddress, restListenPort);
+
+#if !defined(OTBR_VENDOR_NAME) || !defined(OTBR_PRODUCT_NAME)
+        app.GetHost().SetVendorName(vendorName);
```
zhanglongxia: Could we usethe macro `OTBR_VENDOR_NAME ` to guard `app.GetHost().SetVendorName(vendorName);` and use `OTBR_PRODUCT_NAME ` to gurad `app.GetHost().SetVendorModel(productName);` separately?

src/host/rcp_host.cpp
```diff
@@ -911,6 +912,26 @@ void RcpHost::SetBorderAgentVendorTxtData(const std::vector<uint8_t> &aVendorTxt
...
+exit:
+    return;
+}
+#endif
+
+#ifndef OTBR_PRODUCT_NAME
+void RcpHost::SetVendorModel(const char *aVendorModel)
+{
+    VerifyOrExit(mInstance != nullptr);
+    otThreadSetVendorModel(mInstance, aVendorModel);
```
abtink: Both `otThreadSetVendorName` and `otThreadSetVendorModel` return `otError`. Should we also return `error` from these new methods. Also may want to check the error from where these new methods are called?

```c++
/**
 * Set the vendor name string.
 *
 * Requires `OPENTHREAD_CONFIG_NET_DIAG_VENDOR_INFO_SET_API_ENABLE`.
 *
 * @p aVendorName should be UTF8 with max length of 32 chars (`MAX_VENDOR_NAME_TLV_LENGTH`). Maximum length does not
 * include the null `\0` character.
 *
 * @param[in] aInstance       A pointer to an OpenThread instance.
 * @param[in] aVendorName     The vendor name string.
 *
 * @retval OT_ERROR_NONE          Successfully set the vendor name.
 * @retval OT_ERROR_INVALID_ARGS  @p aVendorName is not valid (too long or not UTF8).
 */
otError otThreadSetVendorName(otInstance *aInstance, const char *aVendorName);

/**
 * Set the vendor model string.
 *
 * Requires `OPENTHREAD_CONFIG_NET_DIAG_VENDOR_INFO_SET_API_ENABLE`.
 *
 * @p aVendorModel should be UTF8 with max length of 32 chars (`MAX_VENDOR_MODEL_TLV_LENGTH`). Maximum length does not
 * include the null `\0` character.
 *
 * @param[in] aInstance       A pointer to an OpenThread instance.
 * @param[in] aVendorModel    The vendor model string.
 *
 * @retval OT_ERROR_NONE          Successfully set the vendor model.
 * @retval OT_ERROR_INVALID_ARGS  @p aVendorModel is not valid (too long or not UTF8).
 */
otError otThreadSetVendorModel(otInstance *aInstance, const char *aVendorModel);
```

abtink: Also please check https://github.com/openthread/openthread/pull/12233 which adds additional mandate on vendor name (for a reference device build) - so an improper name can be rejected.

script/_otbr
```diff
@@ -66,6 +66,12 @@ readonly OTBR_DHCP6_PD_CLIENT
 OTBR_MDNS="${OTBR_MDNS:-openthread}"
 readonly OTBR_MDNS
 
+OTBR_VENDOR_NAME="${OTBR_VENDOR_NAME:-OpenThread}"
```
abtink: Regarding https://github.com/openthread/openthread/pull/12233, again we need to make sure the default configuration here does not cause a "build-time" failure.

I do not know if `ot-br-posix` uses `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE`. Ideally, it should not enable this configuration. However, if it does, the vendor name requirement to start with "RD:" is checked and enforced at compile-time. Suggest checking this further.

wtongze: I've added a check in `script/_otbr` so that when `REFERENCE_DEVICE=1` the default vendor name becomes `RD:OpenThread`, otherwise, keep the current default vendor name `OpenThread`. Looks like this is the only place that can cause `OPENTHREAD_CONFIG_REFERENCE_DEVICE_ENABLE` to enable in `ot-br-posix`.

src/host/rcp_host.cpp
wtongze: Done. I've included additional `SuccessOrExit` checks when calling these new methods and for previously added `app.GetBorderAgent().SetMeshCoPServiceValues()` as well in `agent/realmain.cpp`.

src/host/thread_host.hpp
```diff
@@ -349,6 +349,34 @@ class ThreadHost : virtual public NetworkProperties, public UdpProxy::Dependenci
      */
     virtual const char *GetInterfaceName(void) const = 0;
 
+#ifndef OTBR_VENDOR_NAME
+    /**
+     * This method sets the netdiag vendor name.
```
abtink: I suggest not mentioning `netdiag` here. While it is currently used for `netdiag` in the OT core, we want to harmonize the vendor name and model usage across the OT core (e.g., using the same in Joiner or TCAT, etc.).

PR #3230, opened 2026-02-16
[mdns] disable OT_MDNS explicitly when using platform mDNS

martinzi:


third_party/openthread/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ if (OTBR_MDNS STREQUAL "openthread")
     set(OT_SRP_ADV_PROXY ON CACHE STRING "enable SRP Advertising Proxy" FORCE)
     set(OT_TREL_MANAGE_DNSSD ON CACHE STRING "TREL to manage DNSSD and peer discovery" FORCE)
 else()
+    set(OT_MDNS OFF CACHE STRING "disable OpenThread mDNS when using platform mDNS" FORCE)
```
abtink: This is fine to add, but by default it is also disabled in OT core.

abtink (approved):

third_party/openthread/CMakeLists.txt
martinzi: Agree, it is more a minor improvement preventing potential side effects in a development context.

jwhui: The explicit `set(OT_MDNS OFF ... FORCE)` is helpful to prevent stale `CMakeCache.txt` entries when reconfiguring an existing build tree from `OTBR_MDNS=openthread` to `OTBR_MDNS=mDNSResponder` (otherwise both `OT_PLATFORM_DNSSD` and `OT_MDNS` stay enabled in the cache, triggering a compile `#error`).

Minor suggestion: The docstring mentions `"when using platform mDNS"`, but the `else()` branch also applies when `OTBR_MDNS=OFF`. Maybe simplify the docstring to:

```cmake
set(OT_MDNS OFF CACHE STRING "OpenThread mDNS" FORCE)
```

third_party/openthread/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ if (OTBR_MDNS STREQUAL "openthread")
...
     set(OT_DNSSD_SERVER ${OTBR_OT_DISCOVERY_PROXY} CACHE STRING "enable DNS-SD server support" FORCE)
     set(OT_DNSSD_DISCOVERY_PROXY ${OTBR_OT_DISCOVERY_PROXY} CACHE STRING "enable DNS-SD Discovery Proxy" FORCE)
     set(OT_PLATFORM_DNSSD ${OTBR_DNSSD_PLAT} CACHE STRING "enable platform DNSSD" FORCE)
```
jwhui, line 100: While we're cleaning up mDNS cache variables in the `else()` branch, should we also explicitly disable `OT_TREL_MANAGE_DNSSD` here?

third_party/openthread/CMakeLists.txt
martinzi: changed

third_party/openthread/CMakeLists.txt
martinzi: yes, added a line to disable it.

jwhui (approved):
Thanks! 👍🏼

PR #3245, opened 2026-03-02
[netif] verify unicast netlink ACKs and gate cache updates

### LINKED ISSUE #3244: [netif] keep unicast cache aligned with netlink ACK results
**Describe the bug**
`Netif::UpdateIp6UnicastAddresses()` can let OTBR's internal unicast cache drift from kernel reality when a netlink unicast add/remove request fails.

Historically, unicast updates were sent to netlink but OTBR updated `mIp6UnicastAddresses` optimistically. If the kernel rejected a request, OTBR cache could diverge from actual interface state.

**To Reproduce**
1. Git commit id: `e8a3092f` (current `main` at report time).
2. IEEE 802.15.4 hardware platform: any (host-side netif behavior).
3. Build steps:
   - Build and run OTBR on Linux.
   - Trigger IPv6 unicast add/remove updates from Thread state changes.
   - Inject or hit a netlink failure during an address update.
4. Network topology:
   - BR + Thread nodes; issue is host netif synchronization.

**Expected behavior**
OTBR should only mutate unicast cache entries for operations confirmed by matching netlink ACK results, and should avoid indefinite ACK waits.

**Console/log output**
Current code logs sends, but prior behavior did not gate cache mutation on ACK success.

**Additional context**
A fix is prepared that:
- validates matching netlink ACKs by sequence,
- bounds ACK wait with timeout,
- handles idempotent conditions as success (`EEXIST`, `ENOENT`, `EADDRNOTAVAIL`), and
- updates local unicast cache only on successful operations.

binarylogic:
## Summary

This PR hardens Linux unicast netif handling by tying cache mutation to verified netlink outcomes.

## Problem

`UpdateIp6UnicastAddresses()` previously updated `mIp6UnicastAddresses` optimistically, without confirming whether the kernel accepted each unicast add/remove request.  
When RTNETLINK rejected a request, OTBR cache could diverge from kernel state.

## Changes

- Unicast address updates now verify the matching netlink ACK by request sequence.
- ACK wait is bounded by an absolute timeout deadline.
- ACK parsing validates `nlmsgerr` payload length before dereference.
- Transient `recv()` interruptions (`EINTR`, `EAGAIN`, `EWOULDBLOCK`) are retried while waiting.
- Idempotent outcomes are treated as success:
  - add: `EEXIST`
  - remove: `ENOENT`, `EADDRNOTAVAIL`
- `UpdateIp6UnicastAddresses()` now mutates local cache only for operations that succeeded.

## Why this is safe

- Successful paths preserve existing behavior.
- Failure handling is stricter and prevents cache/kernel drift.
- Timeout handling avoids indefinite wait on unrelated netlink traffic.

## Validation

- `clang-format` check (LLVM 19) on modified files.
- `git diff --check` clean.
- Linux CI/runtime validation is still required for full end-to-end netlink path verification.

Fixes #3244

### DIFF
--- a/src/host/posix/netif.cpp
+++ b/src/host/posix/netif.cpp
@@ -154,29 +154,54 @@ void Netif::Deinit(void)
 
 void Netif::UpdateIp6UnicastAddresses(const std::vector<Ip6AddressInfo> &aAddrInfos)
 {
+    std::vector<Ip6AddressInfo> updatedUnicastAddresses = mIp6UnicastAddresses;
+
     // Remove stale addresses
     for (const Ip6AddressInfo &addrInfo : mIp6UnicastAddresses)
     {
         if (std::find(aAddrInfos.begin(), aAddrInfos.end(), addrInfo) == aAddrInfos.end())
         {
+            otbrError error;
+
             otbrLogInfo("Remove address: %s", Ip6Address(addrInfo.mAddress).ToString().c_str());
-            // TODO: Verify success of the addition or deletion in Netlink response.
-            ProcessUnicastAddressChange(addrInfo, false);
+            error = ProcessUnicastAddressChange(addrInfo, false);
+            if (error == OTBR_ERROR_NONE)
+            {
+                updatedUnicastAddresses.erase(
+                    std::remove(updatedUnicastAddresses.begin(), updatedUnicastAddresses.end(), addrInfo),
+                    updatedUnicastAddresses.end());
+            }
+            else
+            {
+                otbrLogWarning("Failed to remove unicast address %s: %s",
+                               Ip6Address(addrInfo.mAddress).ToString().c_str(), otbrErrorString(error));
+            }
         }
     }
 
     // Add new addresses
     for (const Ip6AddressInfo &addrInfo : aAddrInfos)
     {
-        if (std::find(mIp6UnicastAddresses.begin(), mIp6UnicastAddresses.end(), addrInfo) == mIp6UnicastAddresses.end())
+        if (std::find(updatedUnicastAddresses.begin(), updatedUnicastAddresses.end(), addrInfo) ==
+            updatedUnicastAddresses.end())
         {
+            otbrError error;
+
             otbrLogInfo("Add address: %s", Ip6Address(addrInfo.mAddress).ToString().c_str());
-            // TODO: Verify success of the addition or deletion in Netlink response.
-            ProcessUnicastAddressChange(addrInfo, true);
+            error = ProcessUnicastAddressChange(addrInfo, true);
+            if (error == OTBR_ERROR_NONE)
+            {
+                updatedUnicastAddresses.push_back(addrInfo);
+            }
+            else
+            {
+                otbrLogWarning("Failed to add unicast address %s: %s", Ip6Address(addrInfo.mAddress).ToString().c_str(),
+                               otbrErrorString(error));
+            }
         }
     }
 
-    mIp6UnicastAddresses.assign(aAddrInfos.begin(), aAddrInfos.end());
+    mIp6UnicastAddresses = std::move(updatedUnicastAddresses);
 }
 
 otbrError Netif::UpdateIp6MulticastAddresses(const std::vector<Ip6Address> &aAddrs)
--- a/src/host/posix/netif.cpp
+++ b/src/host/posix/netif.cpp
@@ -154,12 +154,18 @@ void Netif::Deinit(void)
 
 void Netif::UpdateIp6UnicastAddresses(const std::vector<Ip6AddressInfo> &aAddrInfos)
 {
-    std::vector<Ip6AddressInfo> updatedUnicastAddresses = mIp6UnicastAddresses;
+    std::vector<Ip6AddressInfo> updatedUnicastAddresses;
+
+    updatedUnicastAddresses.reserve(std::max(mIp6UnicastAddresses.size(), aAddrInfos.size()));
 
     // Remove stale addresses
     for (const Ip6AddressInfo &addrInfo : mIp6UnicastAddresses)
     {
-        if (std::find(aAddrInfos.begin(), aAddrInfos.end(), addrInfo) == aAddrInfos.end())
+        if (std::find(aAddrInfos.begin(), aAddrInfos.end(), addrInfo) != aAddrInfos.end())
+        {
+            updatedUnicastAddresses.push_back(addrInfo);
+        }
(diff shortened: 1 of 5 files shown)


src/host/posix/netif.cpp
```diff
@@ -154,29 +154,54 @@ void Netif::Deinit(void)
...
+                    std::remove(updatedUnicastAddresses.begin(), updatedUnicastAddresses.end(), addrInfo),
+                    updatedUnicastAddresses.end());
+            }
+            else
+            {
+                otbrLogWarning("Failed to remove unicast address %s: %s",
+                               Ip6Address(addrInfo.mAddress).ToString().c_str(), otbrErrorString(error));
+            }
         }
     }
```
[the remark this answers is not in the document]
binarylogic: Applied in f0b76c90.

I kept the same cache correctness behavior but switched stale-removal to a single-pass approach:
- build `updatedUnicastAddresses` by retaining non-stale entries,
- process stale removals once,
- keep entry in cache only if kernel removal fails.

This avoids repeated erase/remove scans while keeping failure semantics explicit.

src/host/posix/netif_linux.cpp
```diff
@@ -232,16 +243,101 @@ void Netif::ProcessUnicastAddressChange(const Ip6AddressInfo &aAddressInfo, bool
...
+            ExitNow(error = OTBR_ERROR_ERRNO);
+        }
+
+        pollTimeoutMs = static_cast<int>(std::chrono::duration_cast<std::chrono::milliseconds>(deadline - now).count());
+
+        pfd.fd      = mNetlinkFd;
+        pfd.events  = POLLIN;
+        pfd.revents = 0;
+
+        pollResult = poll(&pfd, 1, pollTimeoutMs);
```
jwhui: This could block execution for some time. Given that `otbr-agent` and the entire OpenThread stack runs as a single-threaded event loop, this could delay servicing low-layer events. Is this a concern in practice?

binarylogic: Good point. I changed this so unicast address reconciliation queues the netlink changes as a batch and waits for the matching ACKs under a single bounded deadline, instead of allowing one timeout per address operation. That keeps the cache update gated on kernel confirmation, but bounds the possible event-loop delay for the reconciliation pass.

I also tightened the idempotent add handling: EEXIST is only accepted when we are not replacing cached metadata for the same IPv6 address. If a stale same-address entry failed to delete, the add result should not silently move the cache to metadata that the kernel may not have applied.

I kept this synchronous because callers currently expect UpdateIp6UnicastAddresses() to reconcile/cache immediately. Making ACK handling fully async would need a larger pending-state/coalescing design.

src/host/posix/netif_linux.cpp
jwhui, line 147: `mNetlinkFd` is never registered in `MainloopContext` (in `Netif::Update()`) nor drained during mainloop execution (`Netif::Process()`). Consequently, any global link or address changes on the OS will queue unread multicast netlink messages into `mNetlinkFd`'s receive buffer, eventually causing buffer exhaustion (`ENOBUFS`). When `ProcessUnicastAddressChanges()` is called, it may have to parse and discard dozens of accumulated multicast messages before finding its unicast ACKs.

Since `mNetlinkFd` is only used for synchronous request/ACK communication in `Netif`, bind it with `sa.nl_groups = 0` so it does not subscribe to multicast groups:

```suggestion
        sa.nl_groups = 0;
```

src/host/posix/netif_linux.cpp
```diff
@@ -196,52 +199,220 @@ void Netif::SetAddrGenModeToNone(void)
...
+
+            memset(&cacheinfo, 0, sizeof(cacheinfo));
+            cacheinfo.ifa_valid = UINT32_MAX;
+
+            AddRtAttr(&req.nh, sizeof(req), IFA_CACHEINFO, &cacheinfo, sizeof(cacheinfo));
+        }
 
-        memset(&cacheinfo, 0, sizeof(cacheinfo));
-        cacheinfo.ifa_valid = UINT32_MAX;
+        if (send(mNetlinkFd, &req, req.nh.nlmsg_len, 0) == -1)
```
jwhui, line 260: `mNetlinkFd` is configured as non-blocking. The kernel socket send buffer could fill up, causing `send()` to return `-1` with `EAGAIN` or `ENOBUFS`. Should we configure `mNetlinkFd` to be blocking or poll for writeability?

src/host/posix/netif_linux.cpp
```diff
@@ -196,52 +199,220 @@ void Netif::SetAddrGenModeToNone(void)
...
+                        --pendingAckCount;
+                    }
+                }
+                break;
+            }
+
+            union
+            {
+                nlmsghdr mHeader;
+                uint8_t  mBuffer[2048];
```
jwhui, line 339: Do we know that 2048 is sufficient? If the kernel bundles multiple error/ACK headers or multipart messages exceeding 2048 bytes into a single datagram, trailing messages will be silently truncated and lost.

Should we pass the `MSG_TRUNC` flag to `recv()` and explicitly verify that `length <= sizeof(response.mBuffer)`, or increase the receive buffer size to standard kernel page sizing (e.g. 4096 or 8192 bytes).

src/host/posix/netif.cpp
```diff
@@ -154,29 +163,108 @@ void Netif::Deinit(void)
...
+            continue;
+        }
+
+        hasSameCachedAddress =
+            std::find_if(aCachedAddrInfos.begin(), aCachedAddrInfos.end(), [&](const Ip6AddressInfo &aCachedAddrInfo) {
+                return HasSameIp6Address(aCachedAddrInfo, addrInfo);
+            }) != aCachedAddrInfos.end();
+
+        otbrLogInfo("Add address: %s", Ip6Address(addrInfo.mAddress).ToString().c_str());
+        changes.push_back({addrInfo, /* mIsAdded */ true, /* mAllowAlreadyExists */ !hasSameCachedAddress});
```
jwhui, line 216: When replacing metadata for the same IP address (`hasSameCachedAddress == true`), queuing an `RTM_NEWADDR` immediately after an `RTM_DELADDR` in the same batch creates a race condition. If kernel deletion processing is delayed, the subsequent addition can fail with `EEXIST` or disrupt address selection. Consider verifying if `NLM_F_REPLACE` can be used to update cacheinfo/flags in-place without a delete-then-add cycle.

jwhui:
Superseded by #3594

PR #3288, opened 2026-04-07
[web] prevent OT CLI command injection via newline characters

alimezar:
User-controlled strings (pskd, prefix, networkKey, pskc) from the web interface are passed into OpenThreadClient::Execute(), which builds newline-framed commands for the OT daemon Unix socket. A raw CR or LF byte in any of these fields terminates the current command and starts a new one, allowing arbitrary OT CLI command injection.

## Summary

- Reject CR and LF characters before sending commands to the OpenThread daemon.
- Escape spaces, tabs, and backslashes in user-controlled CLI arguments.
- Keep generated PSKc values unchanged.
- Add tests for CR, LF, CRLF, leading and trailing newlines, and argument escaping.

## Testing

- `otbr-gtest-web`

### DIFF
--- a/src/web/web-service/ot_client.cpp
+++ b/src/web/web-service/ot_client.cpp
@@ -161,6 +161,12 @@ char *OpenThreadClient::Execute(const char *aFormat, ...)
         ExitNow();
     }
 
+    if (memchr(&mBuffer[1], '\n', ret) != nullptr || memchr(&mBuffer[1], '\r', ret) != nullptr)
+    {
+        otbrLogErr("OT CLI command contains line break; rejecting to prevent command injection");
+        ExitNow();
+    }
+
     mBuffer[0]       = '\n';
     mBuffer[ret + 1] = '\n';
     ret += 2;
--- a/src/web/web-service/wpan_service.cpp
+++ b/src/web/web-service/wpan_service.cpp
@@ -129,7 +129,8 @@ std::string WpanService::HandleJoinNetworkRequest(const std::string &aJoinReques
     else if (credentialType == CREDENTIAL_TYPE_PSKD)
     {
         VerifyOrExit(client.Execute("ifconfig up") != nullptr, ret = kWpanStatus_JoinFailed);
-        VerifyOrExit(client.Execute("joiner start %s", pskd.c_str()) != nullptr, ret = kWpanStatus_JoinFailed);
+        VerifyOrExit(client.Execute("joiner start %s", escapeOtCliEscapable(pskd).c_str()) != nullptr,
+                     ret = kWpanStatus_JoinFailed);
         VerifyOrExit((rval = client.Read("Join ", 5000)) != nullptr, ret = kWpanStatus_JoinFailed);
         if (strstr(rval, "Join success"))
         {
@@ -154,7 +155,8 @@ std::string WpanService::HandleJoinNetworkRequest(const std::string &aJoinReques
     }
 
     VerifyOrExit(client.Execute("thread start") != nullptr, ret = kWpanStatus_JoinFailed);
-    VerifyOrExit(client.Execute("prefix add %s paso%s", prefix.c_str(), (defaultRoute ? "r" : "")) != nullptr,
+    VerifyOrExit(client.Execute("prefix add %s paso%s", escapeOtCliEscapable(prefix).c_str(),
+                                (defaultRoute ? "r" : "")) != nullptr,
                  ret = kWpanStatus_SetFailed);
 
 exit:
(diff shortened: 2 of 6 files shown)


src/web/web-service/wpan_service.cpp
```diff
@@ -653,10 +662,12 @@ std::string WpanService::escapeOtCliEscapable(const std::string &aArg)
     {
         switch (c)
         {
-        case ' ':
-        case '\t':
         case '\r':
         case '\n':
```
jwhui: Silently stripping `\r` and `\n` here has two unintended side effects:

1. It silently mutates invalid/malicious input instead of failing.
2. It bypasses the primary defense added in `OpenThreadClient::Execute()`, because any string passed through `escapeOtCliEscapable()` will have its newlines removed prior to reaching `Execute()`, so `Execute()` will never detect them or reject the command.

Can we remove `\r` and `\n` cases so they fall through to `default:` without a preceding backslash? That way, if a newline is present in the input, `OpenThreadClient::Execute()` will catch it and properly reject the request.

src/web/web-service/wpan_service.cpp
```diff
@@ -632,10 +639,12 @@ int WpanService::formActiveDataset(otbr::Web::OpenThreadClient &aClient,
     int ret = kWpanStatus_Ok;
 
     VerifyOrExit(aClient.Execute("dataset init new") != nullptr, ret = kWpanStatus_SetFailed);
-    VerifyOrExit(aClient.Execute("dataset networkkey %s", aNetworkKey.c_str()) != nullptr, ret = kWpanStatus_SetFailed);
+    VerifyOrExit(aClient.Execute("dataset networkkey %s", escapeOtCliEscapable(aNetworkKey).c_str()) != nullptr,
+                 ret = kWpanStatus_SetFailed);
     VerifyOrExit(aClient.Execute("dataset networkname %s", escapeOtCliEscapable(aNetworkName).c_str()) != nullptr,
                  ret = kWpanStatus_SetFailed);
-    VerifyOrExit(aClient.Execute("dataset pskc %s", aPskc.c_str()) != nullptr, ret = kWpanStatus_SetFailed);
+    VerifyOrExit(aClient.Execute("dataset pskc %s", escapeOtCliEscapable(aPskc).c_str()) != nullptr,
```
jwhui: `aPskc` is generated by `otbr::Utils::Bytes2Hex(psk.ComputePskc)`, so it is guaranteed to be a hex string and cannot contain spaces or newlines. Wrapping it in `escapeOtCliEscapable()` is redundant.

src/web/web-service/ot_client.cpp
```diff
@@ -161,6 +161,12 @@ char *OpenThreadClient::Execute(const char *aFormat, ...)
         ExitNow();
     }
 
+    if (memchr(&mBuffer[1], '\n', ret) != nullptr || memchr(&mBuffer[1], '\r', ret) != nullptr)
```
jwhui, line 164: Please add unit tests in `tests/gtest/` to verify:

1. `OpenThreadClient::Execute` rejects commands containing `\r` and `\n` and returns `nullptr`.
2. `WpanService::escapeOtCliEscapable` correctly escapes spaces, tabs, and backslashes without altering other characters.

src/web/web-service/wpan_service.cpp
alimezar: Done

src/web/web-service/wpan_service.cpp
alimezar: Done

src/web/web-service/ot_client.cpp
alimezar: Done

tests/gtest/test_web.cpp
```diff
@@ -0,0 +1,71 @@
...
+    int sockets[2];
+
+    ASSERT_EQ(socketpair(AF_UNIX, SOCK_STREAM, 0, sockets), 0);
+
+    {
+        OpenThreadClient client("wpan0");
+
+        SetSocket(client, sockets[0]);
+        EXPECT_EQ(client.Execute("dataset networkname %s", "network\nreset"), nullptr);
+        EXPECT_EQ(client.Execute("dataset networkname %s", "network\rreset"), nullptr);
```
jwhui, line 59: nit: also test CRLF (\r\n) and leading/trailing newlines:
```cpp
        EXPECT_EQ(client.Execute("dataset networkname %s", "network\r\nreset"), nullptr);
        EXPECT_EQ(client.Execute("dataset networkname %s", "\nnetwork"), nullptr);
        EXPECT_EQ(client.Execute("dataset networkname %s", "network\n"), nullptr);
```

alimezar: Done

jwhui:
@alimezar , thanks for addressing the comments! Please squash the commits into a single commit and update the PR description accordingly.

alimezar:
@jwhui Squashed into one commit and updated the PR description. Thank you for your patience and the prompt reviews.

jwhui (approved):
Thanks! 👍🏼

PR #3308, opened 2026-04-17
[rest] add DELETE method for pending dataset endpoint

raman325:
## Summary

Add support for `DELETE` on `/node/dataset/pending` to allow canceling a pending operational dataset (e.g. a scheduled channel migration) before it takes effect.

## Context

While adding pending channel migration visibility to Home Assistant's OTBR integration ([home-assistant/core#168441](https://github.com/home-assistant/core/pull/168441)), we discovered that the REST API has no way to cancel a pending dataset. The `python-otbr-api` library already has a `delete_pending_dataset()` method that sends `DELETE /node/dataset/pending`, but the server returns `405 Method Not Allowed` since the endpoint only registers GET, PUT, and OPTIONS handlers.

Note that the global CORS header (`Access-Control-Allow-Methods`) already advertises `DELETE` as a supported method, so OPTIONS preflight responses already indicate DELETE is allowed — the endpoint just wasn't handling it.

The use case: when a user initiates a channel change via the OTBR, a pending dataset is created with a 5-minute delay. If the user made a mistake, there is currently no REST API method to cancel it — only the OpenThread CLI (`dataset clear`) or D-Bus can do this.

## Changes

- **`rest_web_server.cpp`**: Register a `Delete` handler for the pending dataset endpoint. Add `DeletePendingDataset()` which clears the pending dataset by calling `otDatasetSetPendingTlvs` with a zero-length TLV buffer. Add `kDelete` case to the `Dataset` dispatcher (only for pending; active dataset continues to reject DELETE with 405).
- **`rest_web_server.hpp`**: Declare `DeletePendingDataset`.
- **`openapi.yaml`**: Document the new DELETE method.

## Testing

Tested against a live OTBR (official HA addon connected to a ZBT-2 coordinator) via the Home Assistant websocket API — confirmed the pending dataset is cleared and `get_pending_dataset` returns `None` after deletion.


raman325:
Downstream PR that consumes this endpoint: home-assistant/core#168441 (includes a fallback for firmware without DELETE support).

agners:
> to allow canceling a pending operational dataset (e.g. a scheduled channel migration) before it takes effect.

From what I understand the delay is used to let the pending dataset propagate through the Thread network before it becomes active on the BR side. Not sure how well cancelling this process is supported: E.g. if we'd cancel shortly before it gets applied, a client might not receive the cancellation before the new configuration is applied (if cancellation is sent at all?). I fear this could make devices get stranded on a new but then cancelled configuration :thinking: 

@jwhui any thoughts on this?

jwhui:
> From what I understand the delay is used to let the pending dataset propagate through the Thread network before it becomes active on the BR side. Not sure how well cancelling this process is supported: E.g. if we'd cancel shortly before it gets applied, a client might not receive the cancellation before the new configuration is applied (if cancellation is sent at all?). I fear this could make devices get stranded on a new but then cancelled configuration 🤔

There may be good reasons why a user wants to cancel the operation, even if it may risk stranding nodes. It's hard to say. In this case, I would probably lean towards not having the network impose a limit and leave it to the client to decide whether to check the remaining delay of any existing pending dataset.

agners:
> There may be good reasons why a user wants to cancel the operation, even if it may risk stranding nodes. It's hard to say. In this case, I would probably lean towards not having the network impose a limit and leave it to the client to decide whether to check the remaining delay of any existing pending dataset.

There are definitely good reasons. I guess what I wonder how high are the chances that nodes get stranded. And in any case, we should document the risks here (and clients probably should warn the user before triggering the delete).

I did a bit of digging with the help of Claude, so take this with a grain of salt: There are two fundamental ways how it can go down:
a) BR is Thread leader: Thread nodes see the new pending set in their next MLE advertisement (order of 12–32s) . 
b) BR is not Thread leader: The BR sends the pending dataset via `MGMT_PENDING_SET`. If deleted, the BR will not send an update. It will just not apply the chance and potentially left stranded (while the leader and the rest of the network will apply the new pending dataset after their delay timer run out).

To me it seems the most likely outcome is a split Thread network. Or do I misunderstand something here?

martinzi:
The risk of getting stranded nodes from an operational dataset applied only on a subset of the nodes depends to my understanding primarily on the change that's comprised in the pending operational dataset, e.g. for a channel change existing fallback mechanisms would most likely allow all nodes to recover after a few minutes, but a change of the networkkey could certainly result in stranded nodes.

src/rest/rest_web_server.cpp
```diff
@@ -756,6 +757,28 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
     }
 }
 
+void RestWebServer::DeletePendingDataset(Response &aResponse) const
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    SuccessOrExit(error = RunInMainLoop([this]() {
+                      otOperationalDatasetTlvs datasetTlvs;
```
jwhui, line 765: Should value-initialize `datasetTlvs` with `{}` or `ClearAllBytes(datasetTlvs)` so the stack structure does not contain uninitialized bytes.

src/rest/rest_web_server.cpp
```diff
@@ -154,6 +154,7 @@ RestWebServer::RestWebServer(Host::RcpHost &aHost)
     mServer.Options(OT_REST_RESOURCE_PATH_NODE_DATASET_ACTIVE, MakeHandler(&RestWebServer::DatasetActive));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING, MakeHandler(&RestWebServer::DatasetPending));
     mServer.Put(OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING, MakeHandler(&RestWebServer::DatasetPending));
+    mServer.Delete(OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING, MakeHandler(&RestWebServer::DatasetPending));
```
jwhui, line 157: Since `RestWebServer::Dataset` explicitly handles `kDelete` for `DatasetType::kActive` by returning `405 Method Not Allowed`, should we also register `mServer.Delete(OT_REST_RESOURCE_PATH_NODE_DATASET_ACTIVE, MakeHandler(&RestWebServer::DatasetActive));` here?

Without registering it, `cpp-httplib`'s generic routing error handler will catch `DELETE /node/dataset/active` instead of executing the `405 error branch in 
RestWebServer::Dataset`.

src/rest/openapi.yaml
```diff
@@ -1494,6 +1494,19 @@ paths:
...
         "500":
           $ref: "#/components/responses/InternalServerError"
+    delete:
+      tags:
+        - node
+      summary: Delete the pending operational dataset
+      description: |-
+        Clears the pending operational dataset, canceling any scheduled network
+        configuration changes (e.g. a pending channel migration) before they
+        take effect.
```
jwhui, line 1504: Adding a brief note about the mesh dissemination behavior helps API consumers and UI developers understand the operational implications of canceling a pending dataset.
```suggestion
        Clears the pending operational dataset, canceling any scheduled network
        configuration changes (e.g. a pending channel migration) before they
        take effect.
        Note: Deleting the pending dataset clears local state and stops the local
        delay timer. If the pending dataset has already been disseminated to other
        Thread nodes, deleting it here does not actively recall or roll back the
        dataset on those nodes, which may cause nodes to transition to the new
        configuration when their timers expire.
```

src/rest/rest_web_server.cpp
raman325: I went the other way on this one after merging main, because the `RouteRegistry` that landed since this comment changes the tradeoff.

`DELETE /node/dataset/active` is now already answered as a 405: httplib sets 404 for the unrouted method, and `RoutingErrorHandler` sees that `GetMethods("/node/dataset/active")` returns a non-empty mask that doesn't contain `kDelete`, so it rewrites the status to 405, sets `Allow: GET, PUT, OPTIONS`, and returns `"detail": "method not supported"`. That is the same path `node_state_delete_method_not_allowed_test` already pins for `DELETE /node/state`.

If I register `Delete` for the active dataset, the request routes into `RestWebServer::Dataset` instead, and the 405 branch there calls `ErrorHandler(aResponse, StatusCode::MethodNotAllowed_405)` — no `Allow` header and no detail. So registering it would replace a complete 405 with a less useful one, and would also make `OPTIONS /node/dataset/active` advertise a DELETE that isn't supported.

So I removed the dead `kActive` branch from the `kDelete` case instead and left method negotiation to the route registry, which is where the rest of the server now handles it. Added a comment at the `kDelete` case explaining why the active dataset can't reach it, and a test asserting `DELETE /node/dataset/active` returns 405 with `Allow: GET, PUT, OPTIONS`.

Happy to register it anyway if you'd rather have the route explicit — but then I think the 405 branch in `Dataset` should set the `Allow` header itself so the response doesn't regress.

src/rest/openapi.yaml
```diff
@@ -1544,6 +1544,24 @@ paths:
...
+        Note: Deleting the pending dataset clears local state and stops the local
+        delay timer. If the pending dataset has already been disseminated to other
+        Thread nodes, deleting it here does not actively recall or roll back the
```
agners, line 1557: > If the pending dataset has already been disseminated to other Thread nodes,

Can we say something about the likelyhood of this?

I still wonder if adding DELETE to the pending dataset fits the use case:

> The use case: when a user initiates a channel change via the OTBR, a pending dataset is created with a 5-minute delay. If the user made a mistake, there is currently no REST API method to cancel it — only the OpenThread CLI (dataset clear) or D-Bus can do this.

Since this does not actively recall/roll back, and hence leave stranded node, the safer option seems to send another pending dataset with a higher timestamp and reverted changes, no?

agners: We may add it for other use cases or completeness, but I think if we really want/need an abort feature in Home Assistant, we should make it as safe as possible.

raman325:
@agners you're right, and I dug into the OpenThread side to check rather than argue from the API shape. Marking this as a draft.

Tracing `otDatasetSetPendingTlvs` with a zero-length buffer: it lands in `DatasetManager::SaveLocal`, and in `LocalSave` the `length == 0` path deletes the stored dataset, stops the delay timer, and does `mLocalTimestamp.SetToInvalid()`. What happens next depends entirely on MLE role:

- **BR is leader** — `Restore()` plus `IncrementVersionAndStableVersion()`.
- **BR is child or router** — `SyncLocalWithLeader()`, which guards on `VerifyOrExit(mNetworkTimestamp < mLocalTimestamp, error = kErrorAlready)`. `Timestamp::Compare` ranks an invalid timestamp *below* a valid one, so once the local timestamp is invalid that guard fails and the function exits `kErrorAlready` — **no `MGMT_PENDING_SET` is sent to the leader**.

So in the non-leader case this is a purely local cancel: the BR drops out of a migration that the rest of the network still performs at timer expiry. That's precisely the scenario you sketched back in May, and it's the common case for a Home Assistant BR that isn't the leader. It also answers your "can we say something about the likelihood" question — it isn't a rare race near the deadline, it's the deterministic outcome whenever the BR isn't leader and the dataset has already reached the leader.

Your alternative needs no server change at all: `PendingDataset` in the spec already takes `pendingTimestamp` and `delay`, so a client can PUT a higher-timestamped dataset reverting the change, and that always takes the `SendSetRequest` path because a valid timestamp compares greater than whatever the network holds. Propagating revert beats silent local cancel for the Home Assistant use case, so that's where I'll take the downstream work.

Leaving this as a draft rather than closing it, since @jwhui earlier leaned toward exposing the primitive and letting the client decide. If there's still appetite for `DELETE` as a completeness/parity feature with `dataset clear`, the branch is merged up to main, CI is green, and the docs now carry the dissemination caveat — happy to pick it back up. Otherwise I'm fine closing it.

raman325:
Closing this.

Following up on my last comment: the downstream consumer this was built for (home-assistant/core#168441) was closed back in July, and nothing else depends on the route — so there's no longer a caller waiting on it.

More to the point, the safe direction @agners argued for is already being built, and further along than this: #3553 adds entity tags and `If-Match` to the dataset endpoints, with home-assistant-libs/python-otbr-api#274 and home-assistant/core#178291 on top. That gives a caller an *atomic* replace of exactly the pending dataset it read and showed the user, which covers the cancel-a-mistaken-migration case without the local-only cancel problem I described above — and without the read-then-write race a plain higher-timestamp replace still has.

Thanks @jwhui and @agners for the review time, and @martinzi for the note on how much the stranding risk depends on what the dataset actually changes. The role analysis is written up in the comment above if it's useful evidence for the replace-over-cancel argument on #3553.

PR #3325, opened 2026-05-04
[firewall] opt-in in-process nftables backend via libnftnl/libmnl

### LINKED ISSUE #1675: nftables support
The world is moving away from iptables/ipset to nftables. The latter comes with its own generic set infrastructure. While there are some options that support a hybrid setup (nftables backend with iptables commands), it's probably cleaner to use nftables directly. It's probably also a good time to consider using libmnl or libnftnl or so to talk to the kernel directly rather than calling binaries from C++ code.

Some proof of the claim:
https://firewalld.org/2018/07/nftables-backend
https://git.openwrt.org/08d9f6e3020b4a149b2007b6ed7d684c49af9bbf
https://wiki.debian.org/nftables

LorbusChris:
## Summary

Adds an **opt-in** in-process nftables backend for the OTBR firewall, speaking
netlink directly via libnftnl/libmnl. Addresses #1675, which suggested talking to
the kernel directly via libmnl/libnftnl rather than calling binaries from C++.

**The legacy ipset/ip6tables shell firewall remains the default.** Building with
`-DOTBR_NFTABLES=ON` (default `OFF`) switches the whole firewall in-process:
otbr-agent owns a single `inet otbr` table, including producing the ingress
prefix sets from Thread network data (refreshed on `OT_CHANGED_THREAD_NETDATA`),
replacing the OpenThread posix platform `firewall.cpp` ipset producer
(`OT_FIREWALL` is forced `OFF` for such builds).

```
table inet otbr {
    set ingress_deny_src   { type ipv6_addr; flags interval; }   # on-mesh + mesh-local, agent-produced
    set ingress_allow_dst  { type ipv6_addr; flags interval; }   # on-mesh, agent-produced

    chain forward_ingress  { type filter hook forward     priority filter; ... }
    chain nat_prerouting   { type filter hook prerouting  priority mangle; ... }  # fwmark 0x1001, ipv4 only
    chain nat_postrouting  { type nat    hook postrouting priority srcnat; ... }  # masquerade by mark
    chain nat_forward      { type filter hook forward     priority filter; ... }
}
```

Ingress rule semantics match the legacy ip6tables chain 1:1. Set updates are a
single atomic kernel batch (flush both sets + re-add), so there is never a
half-populated window. Firewall install failures are fatal, matching the legacy
scripts' `die`-on-failure: a border router that cannot install its ingress
filter must not silently forward unfiltered.

## Commits (independently reviewable)

1. **`add INftables interface and libnftnl/libmnl backend`** `INftables` interface + libnftnl/libmnl backend + the
   `OTBR_NFTABLES` flag (default OFF). *Focus: netlink/batch correctness.*
2. **`add FirewallManager policy layer with unit tests`** `FirewallManager` policy layer + gmock unit tests.
   *Focus: table lifecycle, rule/set construction, atomic replace.*
3. **`wire the nftables backend into otbr-agent (opt-in)`** otbr-agent wiring: install on init, produce ingress sets
   from netdata, tear down on shutdown; `OT_FIREWALL=OFF` for nftables
   builds. *Focus: producer parity with OT-core's ipset producer.*
4. **`gate legacy firewall install on the nftables opt-in`** Deployment gating: Docker run/finish, OpenWrt init and
   host `script/setup` skip the legacy firewall/NAT44 when the in-process
   backend is in use, so the two stacks never double-install. *Focus: every
   path defaults to legacy; opt-in is explicit everywhere.*
5. **`build the nftables backend in CI`** A `nftables-check` job that configures
   `-DOTBR_NFTABLES=ON` and runs the test suite. *Focus: the netlink backend is
   actually compiled and exercised in CI.*

## Opting in per platform

- **CMake/host:** `-DOTBR_NFTABLES=ON`, and `OTBR_NFTABLES=1` in the
  environment for `script/setup` (skips the legacy `otbr-firewall` /
  `otbr-nat44` services).
- **Docker:** `--build-arg OTBR_OPTIONS="-DOTBR_NFTABLES=ON"` and run with
  `-e OTBR_NFTABLES=1` (skips the legacy blocks in the s6 scripts).
- **OpenWrt:** `-DOTBR_NFTABLES=ON` plus a manual `DEPENDS` swap in
  `etc/openwrt/openthread-br/Makefile` — drop `+ip6tables +ipset
  +iptables-mod-extra`, add `+libmnl +libnftnl +kmod-nft-core +kmod-nft-nat`.
  The Makefile carries a comment rather than a switch, since the package recipe
  has no visibility of the CMake option. The `otbr-firewall` init script is
  dropped automatically by `src/openwrt/CMakeLists.txt`.

## Validation

The interval-set element encoding (two-element start/`INTERVAL_END`) and the
atomic flush-and-refill batches were validated against live kernels twice
(6.8/nftables 1.0.9 in a VM, 7.1/nftables 1.1.6 in a container), including
verifying that `nft` renders the libnftnl-populated sets as the expected
prefixes. An end-to-end on-device test with a real RCP (join, ingress
filtering, NAT44) is still outstanding.

## Notes for reviewers

- ND-proxy/DUA: upstream removed the DUA routing feature and `nd_proxy.cpp`
  (#3360/#3377) while this PR was in flight. The firewall library still carries
  the ND-proxy NFQUEUE redirect API (`EnableNdProxyRedirect` /
  `dua_prerouting`), currently unconsumed — kept for when ND-proxy returns;
  happy to drop it if preferred.
- With `OTBR_NFTABLES=OFF` the installed firewall/NAT behaviour is unchanged from
  `main` — every new runtime path is `OTBR_ENABLE_NFTABLES`-gated. The *build*
  does differ slightly even when off: `script/bootstrap` and the Dockerfile
  install the libnftnl/libmnl headers unconditionally, and the `otbr-firewall`
  policy layer plus its gmock test binary are always compiled. `nftables_impl.cpp`
  and the libnftnl/libmnl link are what the option actually gates.
- The `nat_prerouting`/`nat_postrouting`/`nat_forward` chains are installed only
  when a backbone/infra interface is configured; that interface is the masquerade
  upstream, matching the legacy `OT_INFRA_IF`.
- Wiring is RCP-mode only. NCP mode is untouched and gets no in-process firewall.
- A `nftables-check` CI job builds with `-DOTBR_NFTABLES=ON` and runs the test
  suite, so `nftables_impl.cpp` is compiled and the firewall tests run in the
  configuration they describe. The `FirewallManager` tests also run in the default
  build, since that target is unconditional.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,6 +105,33 @@ jobs:
     - name: Codecov
       uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
 
+  nftables-check:
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        build_type: ["Debug", "Release"]
+    env:
+      BUILD_TARGET: check
+      OTBR_BUILD_TYPE: ${{ matrix.build_type }}
+      OTBR_MDNS: openthread
+      OTBR_OPTIONS: "-DOTBR_NFTABLES=ON -DOT_FIREWALL=OFF -DBUILD_TESTING=ON"
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
+        submodules: recursive
+        persist-credentials: false
+    - name: Bootstrap
+      run: tests/scripts/bootstrap.sh
+    # Build only: this compiles nftables_impl.cpp, which no other job does.
+    - name: Build
+      run: script/test build
+    # The firewall unit tests mock the backend, so they need no privileges. The
+    # rest of the suite launches otbr-agent, which cannot install nftables rules
+    # without CAP_NET_ADMIN and a usable nf_tables, so it is not run here.
+    - name: Firewall unit tests
+      run: CTEST_OUTPUT_ON_FAILURE=1 ctest --test-dir build/otbr -R 'Firewall|Nftables|MnlNetlink'
+
   script-check:
     runs-on: ubuntu-22.04
     env:
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -86,6 +86,22 @@ if (OTBR_TELEMETRY_DATA_API)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_TELEMETRY_DATA_API=1)
 endif()
 
+# Opt-in, Linux-only: build the in-process nftables firewall backend instead of
+# relying on the legacy ipset/ip6tables shell firewall (which remains the
+# default). Requires libnftnl + libmnl at build time and nf_tables kernel
+# support at runtime.
+option(OTBR_NFTABLES "Enable in-process nftables firewall backend (libnftnl/libmnl)" OFF)
+if (OTBR_NFTABLES)
+    if (NOT CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        message(FATAL_ERROR "OTBR_NFTABLES is only supported on Linux")
+    endif()
+    pkg_check_modules(LIBNFTNL REQUIRED IMPORTED_TARGET libnftnl)
+    pkg_check_modules(LIBMNL REQUIRED IMPORTED_TARGET libmnl)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NFTABLES=1)
+else()
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NFTABLES=0)
+endif()
+
 option(OTBR_OPENWRT "Enable OpenWrt support" OFF)
 if(OTBR_OPENWRT)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_OPENWRT=1)
--- a/etc/docker/border-router/Dockerfile
+++ b/etc/docker/border-router/Dockerfile
@@ -53,6 +53,8 @@ RUN set -x \
            curl \
            git \
            libjsoncpp-dev \
+           libmnl-dev \
+           libnftnl-dev \
            ninja-build \
            nodejs \
            npm \
@@ -136,6 +138,8 @@ RUN set -x \
            iptables \
            iproute2 \
            libjsoncpp25 \
+           libmnl0 \
+           libnftnl11 \
            xz-utils \
     && PLATFORM_SPEC="${TARGETARCH}${TARGETVARIANT:+/$TARGETVARIANT}" \
     && case "${PLATFORM_SPEC}" in \
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -45,10 +45,20 @@
 #endif
 
 #include "agent/application.hpp"
+
 #include "common/code_utils.hpp"
 #include "common/mainloop_manager.hpp"
 #include "host/posix/dnssd.hpp"
 #include "utils/infra_link_selector.hpp"
+
+#if OTBR_ENABLE_NFTABLES
+#include <algorithm>
+#include <string.h>
+#include <vector>
+
+#include <openthread/netdata.h>
+#include <openthread/thread.h>
+#endif
 
 namespace otbr {
 
@@ -297,6 +307,10 @@ void Application::HandleSignal(int aSignal)
 void Application::CreateRcpMode(void)
 {
     otbr::Host::RcpHost &rcpHost = static_cast<otbr::Host::RcpHost &>(mHost);
+#if OTBR_ENABLE_NFTABLES
+    mNftables = MakeUnique<Firewall::Nftables>();
+    mFirewall = MakeUnique<Firewall::FirewallManager>(*mNftables, mInterfaceName);
+#endif
 #if OTBR_ENABLE_BACKBONE_ROUTER
     mBackboneAgent = MakeUnique<BackboneRouter::BackboneAgent>(rcpHost);
 #endif
@@ -386,6 +400,40 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
     });
 #endif
 #endif // OTBR_ENABLE_BORDER_AGENT
+#if OTBR_ENABLE_NFTABLES
+    // The firewall is a security control: if it was built in (OTBR_NFTABLES) but
+    // cannot be installed, abort rather than silently forward traffic unfiltered.
+    // This mirrors the legacy script path, which `die`d when the otbr-firewall /
+    // otbr-nat44 services failed to start.
+    // Each result is taken once: SuccessOrDie() names its argument twice, so a
+    // call passed to it directly runs a second time on the failing path and the
+    // fatal log reports what that second attempt returned.
+    otbrError firewallError;
+
+    firewallError = mNftables->Init();
+    SuccessOrDie(firewallError, "Failed to initialize the nftables firewall!");
+    firewallError = mFirewall->Init();
+    SuccessOrDie(firewallError, "Failed to initialize the firewall manager!");
+    firewallError = mFirewall->EnableIngressFilter();
+    SuccessOrDie(firewallError, "Failed to install the Thread ingress filter!");
+    if (!mBackboneInterfaceName.empty())
+    {
+        firewallError = mFirewall->EnableNat44Masquerade(mBackboneInterfaceName);
+        SuccessOrDie(firewallError, "Failed to install NAT44 masquerade!");
+    }
+    // Populate the ingress allow/deny sets from Thread network data, and keep
+    // them in sync as it changes. This is the in-process replacement for the
+    // OpenThread posix platform firewall's ipset producer.
+    rcpHost.AddThreadStateChangedCallback([this](otChangedFlags aFlags) {
+        // Network data carries the on-mesh prefixes; the active dataset carries
+        // the mesh-local prefix, which the deny set also covers.
+        if (aFlags & (OT_CHANGED_THREAD_NETDATA | OT_CHANGED_ACTIVE_DATASET))
+        {
+            UpdateIngressPrefixes();
+        }
+    });
+    UpdateIngressPrefixes();
+#endif
 #if OTBR_ENABLE_BACKBONE_ROUTER
     mBackboneAgent->Init();
 #endif
@@ -409,8 +457,82 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
 #endif
 }
 
+#if OTBR_ENABLE_NFTABLES
+void Application::UpdateIngressPrefixes(void)
+{
+    otNetworkDataIterator    iterator = OT_NETWORK_DATA_ITERATOR_INIT;
+    otBorderRouterConfig     config;
+    const otMeshLocalPrefix *meshLocal;
+    std::vector<Ip6Prefix>   denySrc;
+    std::vector<Ip6Prefix>   allowDst;
+    otInstance              *instance;
+
+    // Several border routers can advertise the same on-mesh prefix, and duplicate
+    // intervals in a single transaction make the kernel reject the whole batch,
+    // so only distinct prefixes go in. (Declared before the first VerifyOrExit so
+    // no goto crosses its initialisation.)
+    auto addUnique = [](std::vector<Ip6Prefix> &aVec, const Ip6Prefix &aPrefix) {
+        if (std::find(aVec.begin(), aVec.end(), aPrefix) == aVec.end())
+        {
+            aVec.push_back(aPrefix);
+        }
+    };
+
+    // The firewall only exists in RCP mode, so check it before downcasting mHost.
+    VerifyOrExit(mFirewall != nullptr && mFirewall->IsIngressFilterEnabled());
+
+    instance = static_cast<otbr::Host::RcpHost &>(mHost).GetInstance();
+    VerifyOrExit(instance != nullptr);
+
+    // Deny-src and allow-dst both cover every on-mesh prefix.
+    while (otNetDataGetNextOnMeshPrefix(instance, &iterator, &config) == OT_ERROR_NONE)
+    {
+        Ip6Prefix prefix;
+        prefix.Set(config.mPrefix);
+        addUnique(denySrc, prefix);
+        addUnique(allowDst, prefix);
+    }
+
+    // Deny-src additionally covers the mesh-local /64.
+    meshLocal = otThreadGetMeshLocalPrefix(instance);
+    if (meshLocal != nullptr)
+    {
+        Ip6Prefix mlPrefix{};
+        memcpy(mlPrefix.mPrefix.m8, meshLocal->m8, sizeof(meshLocal->m8));
+        mlPrefix.mLength = 64;
+        addUnique(denySrc, mlPrefix);
+    }
+
+    if (mFirewall->ReplaceIngressPrefixes(denySrc, allowDst) != OTBR_ERROR_NONE)
+    {
+        otbrLogWarning("FirewallManager: failed to update ingress prefixes");
+    }
+
+exit:
+    return;
+}
+#endif
+
 void Application::DeinitRcpMode(void)
 {
+#if OTBR_ENABLE_NFTABLES
+    // Tear down the OTBR nftables table while the netlink socket is still open
+    // (mNftables outlives mFirewall by member declaration order).
+    if (mFirewall != nullptr)
+    {
+        otbrError firewallError = mFirewall->Deinit();
+        if (firewallError != OTBR_ERROR_NONE)
+        {
+            otbrLogWarning("FirewallManager: teardown failed (%d)", firewallError);
+        }
+    }
+    // Close the netlink socket too, or a later InitRcpMode() dies in
+    // mNftables->Init() on the still-open socket.
+    if (mNftables != nullptr)
+    {
+        mNftables->Deinit();
+    }
+#endif
 #if OTBR_ENABLE_DNSSD_PLAT
     mDnssdPlatform.Stop();
 #endif
--- a/src/firewall/CMakeLists.txt
+++ b/src/firewall/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-#  Copyright (c) 2020-2021, The OpenThread Authors.
+#  Copyright (c) 2026, The OpenThread Authors.
 #  All rights reserved.
 #
 #  Redistribution and use in source and binary forms, with or without
@@ -26,46 +26,26 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
-add_subdirectory(agent)
-if(OTBR_BORDER_AGENT)
-    add_subdirectory(border_agent)
-endif()
-add_subdirectory(common)
-if(OTBR_DBUS OR OTBR_FEATURE_FLAGS OR OTBR_TELEMETRY_DATA_API)
-    add_subdirectory(proto)
-endif()
-add_subdirectory(host)
+set(OTBR_FIREWALL_SOURCES
+    firewall_manager.cpp
+)
 
-if(OTBR_DBUS)
-    add_subdirectory(dbus)
+if(OTBR_NFTABLES)
+    list(APPEND OTBR_FIREWALL_SOURCES
+        netlink_socket.cpp
+        nftables_impl.cpp
+    )
 endif()
 
-if(OTBR_MDNS)
-    add_subdirectory(mdns)
-    if(NOT OTBR_MDNS STREQUAL "openthread")
-        add_subdirectory(sdp_proxy)
-    endif()
-    if(OTBR_TREL_DNSSD)
-        add_subdirectory(trel_dnssd)
-    endif()
-endif()
+add_library(otbr-firewall ${OTBR_FIREWALL_SOURCES})
 
-add_subdirectory(utils)
+target_link_libraries(otbr-firewall PUBLIC
+    otbr-common
+)
 
-add_subdirectory(firewall)
-
-if(OTBR_OPENWRT)
-    add_subdirectory(openwrt)
-endif()
-
-if(OTBR_WEB)
-    add_subdirectory(web)
-endif()
-
-if(OTBR_REST)
-    add_subdirectory(rest)
-endif()
-
-if (OTBR_BACKBONE_ROUTER)
-    add_subdirectory(backbone_router)
+if(OTBR_NFTABLES)
+    target_link_libraries(otbr-firewall PRIVATE
+        PkgConfig::LIBNFTNL
+        PkgConfig::LIBMNL
+    )
 endif()
--- a/src/firewall/firewall_manager.cpp
+++ b/src/firewall/firewall_manager.cpp
@@ -1,0 +1,362 @@
+/*
+ *    Copyright (c) 2026, The OpenThread Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ *    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#define OTBR_LOG_TAG "FIREWALL"
+
+#include "firewall/firewall_manager.hpp"
+
+#include <string.h>
+#include <utility>
+
+#include "common/logging.hpp"
+
+namespace otbr {
+namespace Firewall {
+
+const char *const  FirewallManager::kTableName           = "otbr";
+const char *const  FirewallManager::kIngressChain        = "forward_ingress";
+const char *const  FirewallManager::kPreroutingChain     = "dua_prerouting";
+const char *const  FirewallManager::kNatPreroutingChain  = "nat_prerouting";
+const char *const  FirewallManager::kNatPostroutingChain = "nat_postrouting";
+const char *const  FirewallManager::kNatForwardChain     = "nat_forward";
+const char *const  FirewallManager::kIngressDenySrcSet   = "ingress_deny_src";
+const char *const  FirewallManager::kIngressAllowDstSet  = "ingress_allow_dst";
+constexpr uint32_t FirewallManager::kNat44Mark;
+
+FirewallManager::FirewallManager(INftables &aNftables, std::string aThreadInterfaceName)
+    : mNftables(aNftables)
+    , mThreadIfName(std::move(aThreadInterfaceName))
+    , mInitialized(false)
+    , mIngressFilterEnabled(false)
+    , mNat44Enabled(false)
+    , mDuaChainCreated(false)
+    , mNdRuleHandle(0)
+{
+}
+
+otbrError FirewallManager::Init(void)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(!mInitialized, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(!mThreadIfName.empty(), error = OTBR_ERROR_INVALID_ARGS);
+
+    // Idempotent reset: drop any leftover OTBR table from a prior run and
+    // recreate it empty, in one transaction so no partial state is visible.
+    SuccessOrExit(error = mNftables.BeginBatch());
+    SuccessOrExit(error = mNftables.DelTable(kTableName));
+    SuccessOrExit(error = mNftables.AddTable(kTableName));
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+    mInitialized = true;
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    otbrLogResult(error, "FirewallManager: %s", __FUNCTION__);
+    return error;
+}
+
+otbrError FirewallManager::Deinit(void)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mInitialized);
+
+    if ((error = mNftables.BeginBatch()) == OTBR_ERROR_NONE)
+    {
+        if ((error = mNftables.DelTable(kTableName)) == OTBR_ERROR_NONE)
+        {
+            error = mNftables.CommitBatch();
+        }
+
+        if (error != OTBR_ERROR_NONE)
+        {
+            mNftables.AbortBatch();
+        }
+    }
+
+    mNdRuleHandle         = 0;
+    mDuaChainCreated      = false;
+    mIngressFilterEnabled = false;
+    mNat44Enabled         = false;
+    mInitialized          = false;
+
+exit:
+    otbrLogResult(error, "FirewallManager: %s", __FUNCTION__);
+    return error;
+}
+
+otbrError FirewallManager::EnableIngressFilter(void)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mInitialized, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(!mIngressFilterEnabled);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+
+    SuccessOrExit(error = mNftables.AddIp6PrefixSet(kTableName, kIngressDenySrcSet));
+    SuccessOrExit(error = mNftables.AddIp6PrefixSet(kTableName, kIngressAllowDstSet));
+
+    SuccessOrExit(error = mNftables.AddChain(kTableName, kIngressChain, Hook::kForward, ChainPriority::kFilter));
+
+    SuccessOrExit(error = mNftables.AddRuleOifnameNeqReturn(kTableName, kIngressChain, mThreadIfName, nullptr));
+    SuccessOrExit(error = mNftables.AddRuleIifPkttypeVerdict(kTableName, kIngressChain, mThreadIfName,
+                                                             PktType::kUnicast, Verdict::kDrop, nullptr));
+    SuccessOrExit(error = mNftables.AddRuleSetLookupVerdict(kTableName, kIngressChain, kIngressDenySrcSet,
+                                                            SetDirection::kSrc, Verdict::kDrop, nullptr));
+    SuccessOrExit(error = mNftables.AddRuleSetLookupVerdict(kTableName, kIngressChain, kIngressAllowDstSet,
+                                                            SetDirection::kDst, Verdict::kAccept, nullptr));
+    SuccessOrExit(
+        error = mNftables.AddRulePkttypeVerdict(kTableName, kIngressChain, PktType::kUnicast, Verdict::kDrop, nullptr));
+    SuccessOrExit(error = mNftables.AddRuleVerdict(kTableName, kIngressChain, Verdict::kAccept, nullptr));
+
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+    mIngressFilterEnabled = true;
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    otbrLogResult(error, "FirewallManager: %s", __FUNCTION__);
+    return error;
+}
+
+otbrError FirewallManager::EnableNat44Masquerade(const std::string &aUpstreamInterfaceName)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mInitialized, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(!aUpstreamInterfaceName.empty(), error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(!mNat44Enabled);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+
+    // Mangle-prerouting: tag packets coming from the Thread interface so the
+    // postrouting chain can MASQUERADE them. Type filter, mangle priority.
+    SuccessOrExit(error = mNftables.AddChain(kTableName, kNatPreroutingChain, Hook::kPrerouting, ChainPriority::kMangle,
+                                             ChainType::kFilter));
+    SuccessOrExit(error =
+                      mNftables.AddRuleIifMark(kTableName, kNatPreroutingChain, mThreadIfName, kNat44Mark, nullptr));
+
+    // Postrouting: source-NAT marked traffic. Type nat, srcnat priority.
+    SuccessOrExit(error = mNftables.AddChain(kTableName, kNatPostroutingChain, Hook::kPostrouting,
+                                             ChainPriority::kSrcNat, ChainType::kNat));
+    SuccessOrExit(error = mNftables.AddRuleMarkMasquerade(kTableName, kNatPostroutingChain, kNat44Mark, nullptr));
+
+    // Forward: accept traffic in either direction on the upstream interface.
+    // Hooked at FORWARD/filter alongside forward_ingress; the iifname/oifname
+    // matches scope each rule cleanly.
+    SuccessOrExit(error = mNftables.AddChain(kTableName, kNatForwardChain, Hook::kForward, ChainPriority::kFilter,
+                                             ChainType::kFilter));
+    SuccessOrExit(error = mNftables.AddRuleOifnameVerdict(kTableName, kNatForwardChain, aUpstreamInterfaceName,
+                                                          Verdict::kAccept, nullptr));
+    SuccessOrExit(error = mNftables.AddRuleIifnameVerdict(kTableName, kNatForwardChain, aUpstreamInterfaceName,
+                                                          Verdict::kAccept, nullptr));
+
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+    mNat44Enabled = true;
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    otbrLogResult(error, "FirewallManager: %s", __FUNCTION__);
+    return error;
+}
+
+otbrError FirewallManager::EnableNdProxyRedirect(const Ip6Prefix   &aDomainPrefix,
+                                                 const std::string &aBackboneInterfaceName,
+                                                 uint16_t           aQueueNum)
+{
+    otbrError error     = OTBR_ERROR_NONE;
+    uint64_t  newHandle = 0;
+
+    VerifyOrExit(mInitialized, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(aDomainPrefix.IsValid(), error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(!aBackboneInterfaceName.empty(), error = OTBR_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+
+    if (!mDuaChainCreated)
+    {
+        SuccessOrExit(error = mNftables.AddChain(kTableName, kPreroutingChain, Hook::kPrerouting, ChainPriority::kRaw));
+    }
+
+    if (mNdRuleHandle != 0)
+    {
+        SuccessOrExit(error = mNftables.DelRule(kTableName, kPreroutingChain, mNdRuleHandle));
+    }
+
+    SuccessOrExit(error = mNftables.AddRuleNdNsRedirect(kTableName, kPreroutingChain, aDomainPrefix,
+                                                        aBackboneInterfaceName, aQueueNum, &newHandle));
+
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+    mDuaChainCreated = true;
+    mNdRuleHandle    = newHandle;
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    otbrLogResult(error, "FirewallManager: %s", __FUNCTION__);
+    return error;
+}
+
+otbrError FirewallManager::DisableNdProxyRedirect(void)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mInitialized, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(mNdRuleHandle != 0);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+    SuccessOrExit(error = mNftables.DelRule(kTableName, kPreroutingChain, mNdRuleHandle));
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+    mNdRuleHandle = 0;
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    otbrLogResult(error, "FirewallManager: %s", __FUNCTION__);
+    return error;
+}
+
+otbrError FirewallManager::AddIngressSetElement(IngressSet aSet, const Ip6Prefix &aPrefix)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mIngressFilterEnabled, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(aPrefix.IsValid(), error = OTBR_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+    SuccessOrExit(error = mNftables.AddSetElement(kTableName, SetName(aSet), aPrefix));
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    return error;
+}
+
+otbrError FirewallManager::DelIngressSetElement(IngressSet aSet, const Ip6Prefix &aPrefix)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mIngressFilterEnabled, error = OTBR_ERROR_INVALID_STATE);
+    VerifyOrExit(aPrefix.IsValid(), error = OTBR_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+    SuccessOrExit(error = mNftables.DelSetElement(kTableName, SetName(aSet), aPrefix));
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    return error;
+}
+
+otbrError FirewallManager::FlushIngressSet(IngressSet aSet)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mIngressFilterEnabled, error = OTBR_ERROR_INVALID_STATE);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+    SuccessOrExit(error = mNftables.FlushSet(kTableName, SetName(aSet)));
+    SuccessOrExit(error = mNftables.CommitBatch());
+
+exit:
+    if (error != OTBR_ERROR_NONE)
+    {
+        mNftables.AbortBatch();
+    }
+    return error;
+}
+
+otbrError FirewallManager::ReplaceIngressPrefixes(const std::vector<Ip6Prefix> &aDenySrc,
+                                                  const std::vector<Ip6Prefix> &aAllowDst)
+{
+    otbrError error = OTBR_ERROR_NONE;
+
+    VerifyOrExit(mIngressFilterEnabled, error = OTBR_ERROR_INVALID_STATE);
+
+    SuccessOrExit(error = mNftables.BeginBatch());
+    SuccessOrExit(error = mNftables.FlushSet(kTableName, kIngressDenySrcSet));
+    SuccessOrExit(error = mNftables.FlushSet(kTableName, kIngressAllowDstSet));
+    for (const Ip6Prefix &prefix : aDenySrc)
+    {
+        VerifyOrExit(prefix.IsValid(), error = OTBR_ERROR_INVALID_ARGS);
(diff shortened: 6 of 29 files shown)


LorbusChris:
Thanks — the batch-leak finding is correct and is now fixed, with a couple of adjustments.

**What was actually broken.** Every `FirewallManager` operation wraps its work in `BeginBatch()`/`CommitBatch()` with `SuccessOrExit` in between, so any intermediate failure jumped to `exit:` without committing. Because `BeginBatch()` guards on `VerifyOrExit(!mInBatch, ...)`, a single mid-batch failure wedged the backend for the lifetime of the process — every later firewall operation returned `OTBR_ERROR_INVALID_STATE` — and leaked `mBatchBuffer`. Fixed by adding `AbortBatch()` and calling it on the error path of all nine functions that open a batch.

One correction: `CommitBatch()` was already cleaning up unconditionally in its own `exit:` block, so a *failed commit* did not leave the backend stuck. Rather than duplicate that logic, `CommitBatch()` now delegates to `AbortBatch()`, and `AbortBatch()` is documented as safe to call when no batch is open, so callers needn't track whether one was started.

**`mNdRuleHandle` desync** — correct, and fixed: the handle is only updated after `CommitBatch()` succeeds, so a rolled-back batch leaves the manager still able to delete the rule installed earlier.

**`memset` on `Ip6Prefix`** — good catch, and it was redundant rather than merely unsafe: `Ip6Prefix` has a default constructor that calls `Clear()`. Dropped the `memset` outright instead of replacing it with `{}`.

**Tests.** Added two regression tests and checked they actually fail without the fix — on a tree with only the behavioural changes reverted: 17 passed / 2 failed; with them: 19/19.

- `FailedOperationAbortsTheBatch` — an intermediate failure must abort and must not commit.
- `EnableNdProxyKeepsRuleHandleWhenCommitFails` — after a failed commit, a later call must still delete the previously installed rule.

**CI.** Also added a `nftables-check` job so `nftables_impl.cpp` is compiled and the firewall tests run in that configuration; previously only the policy layer was covered, since that target builds unconditionally.

All fixes are folded into the commits that introduced the code, so the history stays bisectable.

LorbusChris:
Thanks — all three are addressed.

**`mnl_nlmsg_batch_next()` return value (critical).** Correct, and it was ignored at all ten call sites. The buffer is twice the batch limit, so the message that *crosses* the limit is written safely into the overflow page, but nothing may be added after that — which the code did not enforce. In practice the exposure was bounded, since the number of queued messages comes from Thread network data rather than anything unbounded, but that is an implicit invariant that nothing documents or checks, so it should not be relied on.

Rather than repeat the check ten times, added a small `AdvanceBatch()` helper that every queueing path now goes through, failing with `ENOBUFS`:

```c
VerifyOrExit(mnl_nlmsg_batch_next(mBatch), errno = ENOBUFS, error = OTBR_ERROR_ERRNO);
```

I deliberately did not auto-flush and reset on overflow, which is the other way libmnl supports: that would split one logical change across two kernel transactions and lose the atomicity the batch exists to provide. Failing the whole operation — and, with the abort path added earlier, discarding the batch — keeps it all-or-nothing.

**`GetInstance()` may be null.** Added the guard. Both current call sites run after `Init()`, so this is defensive rather than a live crash, but it costs nothing and matches how the rest of the file is written.

**Empty upstream interface name.** Added `VerifyOrExit(!aUpstreamInterfaceName.empty(), error = OTBR_ERROR_INVALID_ARGS)`. The only caller already guards on `!mBackboneInterfaceName.empty()`, but this is a public method and should validate its own input. Covered by a new test asserting it is rejected before any batch is opened.

Firewall suite is 20/20 with the nftables backend enabled, and all changes are folded into the commits that introduced the code.

LorbusChris:
All six addressed.

**Unchecked expression allocation (high).** Correct. The eleven `Make*` helpers return `nullptr` when `nftnl_expr_alloc()` fails, and all 44 `nftnl_rule_add_expr()` call sites passed the result straight through, where it would have been dereferenced. Rather than repeat the check 44 times, every site now goes through a helper that fails with `ENOMEM`:

```c
otbrError AddExpr(struct nftnl_rule *aRule, struct nftnl_expr *aExpr)
{
    VerifyOrExit(aExpr != nullptr, errno = ENOMEM, error = OTBR_ERROR_ERRNO);
    nftnl_rule_add_expr(aRule, aExpr);
    ...
```

Combined with the abort path added earlier, a failed allocation now unwinds the whole operation and discards the batch instead of building a partial rule.

**Netlink buffer alignment.** Correct, and worth taking seriously here: OpenWrt targets MIPS, which is exactly where an under-aligned buffer cast to `struct nlmsghdr` bites. Added `alignas(struct nlmsghdr)` to all three stack buffers.

**`IfNameToWire`.** Fixed — that site hand-rolled the `memset`/`memcpy` while the other two already used the helper.

**`pkg_check_modules`.** Adopted, and it matches the file, which already does `find_package(PkgConfig)` and `pkg_check_modules(DBUS ...)`. I also wired up `*_INCLUDE_DIRS` and the library dirs, since switching the lookup without consuming them would not actually have helped cross-compilation, which is the point of the change.

One deviation: the suggestion's natural companion, `target_link_directories()`, requires CMake 3.13, while this project sets `cmake_minimum_required(VERSION 3.10.2)`. I used the generator-expression idiom already used for D-Bus in `src/dbus/common/CMakeLists.txt` instead:

```cmake
$<$<BOOL:${LIBNFTNL_LIBRARY_DIRS}>:-L$<JOIN:${LIBNFTNL_LIBRARY_DIRS}," -L">>
```

Verified with `libnftnl` 1.2.4 / `libmnl` 1.0.4 found via pkg-config, full build clean, firewall suite 20/20. All changes folded into the commits that introduced the code.

LorbusChris:
Correct, and thanks — that leak was one I introduced in the previous round. Before `AdvanceBatch()` was added there was no error path after `mnl_nlmsg_batch_start()` succeeded, so `mBatch = nullptr` in the exit block was only ever reached when `mBatch` was already null. Adding the check created a path where a live batch was dropped without `mnl_nlmsg_batch_stop()`.

Fixing it surfaced a second problem on the same exit path, which is the more serious of the two: `VerifyOrExit(!mInBatch, ...)` jumps to that block as well, and the block unconditionally did `free(mBatchBuffer)`. So calling `BeginBatch()` while a batch was already open freed the *live* batch's buffer while leaving `mInBatch` set and `mBatch` dangling — the guard corrupted the in-flight batch instead of rejecting the call.

Both are fixed by cleaning up only what the call itself allocated:

```c
exit:
    // Only clean up what this call allocated. When the failure is that a batch
    // is already open, mInBatch is still set and the buffer belongs to that
    // batch, so it must be left alone.
    if (error != OTBR_ERROR_NONE && !mInBatch)
    {
        AbortBatch();
    }
    return error;
```

`mInBatch` is set only on the success path, so every failure that allocated something sees it clear and goes through `AbortBatch()`, which stops the batch, frees the buffer and clears the pending-handle map. The already-in-a-batch case skips cleanup entirely. `AbortBatch()` is null-safe, so the early failures (no socket, `calloc` failure) are fine too.

One caveat: this is not covered by the unit tests. They drive `FirewallManager` through the mocked `INftables`, so they never enter the real `Nftables::BeginBatch()`, and reaching these paths needs a live netlink socket. Verified by inspection rather than by test; the suite is still 20/20 and the full build is clean.

Folded into the commit that introduces the backend.

LorbusChris:
Three taken, one declined.

**Netlink receive timeout.** Good catch, and the most valuable one here: otbr-agent is single threaded, so a reply that never arrives would hang the whole daemon, and netlink can genuinely drop messages under `ENOBUFS`. Added `SO_RCVTIMEO` (5 s) after the bind. Both receive paths already treat a negative return as an error — `SendStandalone` via `VerifyOrExit(recvLen > 0, ...)` and `CommitBatch` via its `while (... > 0)` loop plus the trailing `recvLen < 0` check — so the timeout turns a hang into a normal failure, which then aborts the batch.

**Empty `mThreadIfName` in `Init()`** and **empty `aBackboneInterfaceName` in `EnableNdProxyRedirect()`.** Both added, matching the check already on `EnableNat44Masquerade()`. Each has a test asserting the call is rejected before any batch is opened.

**`Ip6Prefix mlPrefix{}` — declining this one.** The comment's premise is that `Ip6Prefix` may be a POD whose members would otherwise be uninitialized, but it is not:

```cpp
class Ip6Prefix
{
public:
    Ip6Prefix(void) { Clear(); }
    ...
    void Clear(void) { memset(reinterpret_cast<void *>(this), 0, sizeof(*this)); }
```

`src/common/types.hpp:337,343,396`. The default constructor zeroes the whole object, so `Ip6Prefix mlPrefix;` is already fully initialised and `{}` would not change the generated code. It would also be inconsistent with the surrounding code, which declares these plainly — `Ip6Prefix prefix;` ten lines above in the same function (`src/agent/application.cpp:395`) and again in `src/host/telemetry/telemetry.cpp:875`.

This is the same point raised in an earlier round, where it was resolved by removing a redundant `memset` on that object for the same reason. Happy to add `{}` anyway if you would rather have belt and braces, but as written the code is not relying on uninitialised memory.

Firewall suite is 22/22 with the backend enabled and the full build is clean. As with the previous round, the socket timeout itself is not unit-testable here — the tests drive `FirewallManager` through the mocked `INftables` and never open a real netlink socket. Folded into the commits that introduce the code.

src/agent/application.cpp
```diff
@@ -341,8 +376,61 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
...
+        prefix.Set(config.mPrefix);
+        denySrc.push_back(prefix);
+        allowDst.push_back(prefix);
+    }
+
+    // Deny-src additionally covers the mesh-local /64.
+    meshLocal = otThreadGetMeshLocalPrefix(instance);
+    if (meshLocal != nullptr)
+    {
+        Ip6Prefix mlPrefix; // default-constructed, already cleared
```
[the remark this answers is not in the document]
LorbusChris: Not applying this one: the premise does not hold here. `Ip6Prefix` is not a POD — it has a default constructor that zeroes the whole object:

```cpp
class Ip6Prefix
{
public:
    Ip6Prefix(void) { Clear(); }
    ...
    void Clear(void) { memset(reinterpret_cast<void *>(this), 0, sizeof(*this)); }
```

(`src/common/types.hpp:337`, `:343`, `:396`.) So `Ip6Prefix mlPrefix;` is already fully initialised and `{}` would generate identical code. It would also be inconsistent with the surrounding style, which declares these plainly: `Ip6Prefix prefix;` ten lines above in this same function (`src/agent/application.cpp:395`) and in `src/host/telemetry/telemetry.cpp:875`.

For context, this is the same point raised in an earlier round, where it was resolved by deleting a redundant `memset` on this object for exactly this reason — so adding `{}` now would re-add the redundancy that was removed.

Resolving as won't-fix, but happy to add it if a maintainer would still prefer the explicit form.

LorbusChris:
All three taken, with one adjustment that matters.

**`EINTR` on the netlink receives.** Correct for both `CommitBatch()` and `SendStandalone()`: a signal would have turned into a failed commit, and otbr-agent treats firewall failures as fatal, so a transient signal could have taken the daemon down. The POSIX platform uses timer signals, so this was reachable rather than theoretical.

The adjustment: the retry has to be narrowed to `EINTR` specifically, not to "`recvfrom()` returned -1". The previous round added `SO_RCVTIMEO` to stop a lost netlink ack from hanging the single-threaded daemon, and that timeout surfaces as `EAGAIN`. Retrying on any negative return — including the `do { ... } while (recvLen < 0)` shape — would swallow the timeout and reintroduce exactly the indefinite hang that change was meant to prevent. So:

```c
if (recvLen < 0)
{
    // Retry only when interrupted by a signal. A receive timeout shows up
    // as EAGAIN and must stay an error, or the timeout would never fire.
    VerifyOrExit(errno == EINTR, error = OTBR_ERROR_ERRNO);
    continue;
}
```

and the same `errno == EINTR` condition on `SendStandalone()`'s loop. `EAGAIN` still fails, which aborts the batch as before.

While restructuring that loop I also folded the trailing `if (recvLen < 0 && error == OTBR_ERROR_NONE)` check into it, so there is now one place where a receive failure is classified rather than two.

**`mFirewall != nullptr` on teardown.** Added. It is a `std::unique_ptr` that `CreateRcpMode()` always populates under the same build guard, so this is defensive rather than a live crash, but it costs nothing on the shutdown path.

Firewall suite is 22/22 with the backend enabled, full build clean, and everything folded into the commits that introduce the code.

LorbusChris:
Good catch, taken.

Leaving that loop on a zero-length reply means the kernel never acknowledged the batch — `mnl_cb_run()` never returned `MNL_CB_STOP` — so reporting success would tell the caller rules were installed when they may not have been.

For the record this is not something the previous round introduced. The original loop had the same hole:

```c
while ((recvLen = mnl_socket_recvfrom(mSocket, buf, sizeof(buf))) > 0) { ... }
if (recvLen < 0 && error == OTBR_ERROR_NONE) { error = OTBR_ERROR_ERRNO; }
```

`recvLen == 0` fell out of the loop and the trailing check only looked at `< 0`, so it also returned `OTBR_ERROR_NONE`. Restructuring the loop for `EINTR` preserved that behaviour and happened to put it on one visible line.

One change from the suggestion: it also sets `errno` explicitly rather than relying on whatever a previous syscall left there, since nothing sets `errno` for a zero-length datagram and the log would otherwise report an unrelated error:

```c
// A zero-length reply means the loop ends without the kernel ever having
// acknowledged the batch, so the commit cannot be reported as successful.
VerifyOrExit(recvLen > 0, errno = EPROTO, error = OTBR_ERROR_ERRNO);
```

That matches how the other failure paths in this file are written (`ENOBUFS` on batch overflow, `ENOMEM` on expression allocation).

Firewall suite 22/22, full build clean, folded into the commit that introduces the backend.

LorbusChris:
Two review points plus two CI failures on the branch, all addressed in this push.

**`AddPrefixIncrement` with a length above 128.** The out-of-bounds write cannot actually happen. For `aN > 128`, `bitFromLsb = uint8_t(128 - aN)` wraps to at least 128, so `byteFromLsb >= 16` and `byteIdx = 15 - byteFromLsb` is negative; the loop is guarded by `i >= 0` and its body never runs, so nothing is written:

```c
for (int i = byteIdx; i >= 0 && carry != 0; i--)
```

The underlying point is still fair, though: rather than failing, an out-of-range length would silently yield a wrong interval. So `AttachPrefixInterval()` now rejects it up front with `OTBR_ERROR_INVALID_ARGS`, which turns a silent misbehaviour into something diagnosable. (`Ip6Prefix::IsValid()` already bounds this to 1..128 and `FirewallManager` checks it, so this is defence in depth rather than a live bug.)

**`Ip6Prefix mlPrefix{}`.** Applied. To be accurate about what it changes: nothing functionally. `Ip6Prefix` is not a POD — it has `Ip6Prefix(void) { Clear(); }`, and `Clear()` is `memset(this, 0, sizeof(*this))` (`src/common/types.hpp:337,343,396`), so the object was already fully zeroed and there were no uninitialised bytes for the `memcpy` to leave behind. The explicit form is harmless and makes the intent obvious at the call site, so it is in.

**CI failures on the branch, both mine.**

`nftables-check` failed with `Package 'libnftnl' not found`. The commit message claimed no bootstrap change was needed because `script/bootstrap` installs libnftnl/libmnl — but CI runs `tests/scripts/bootstrap.sh`, which is a different script and did not. My local builds installed the packages by hand, so I had verified the CMake configuration rather than the job. Added `libnftnl-dev` and `libmnl-dev` to `install_common_dependencies()`.

`pretty` failed because `nftables_impl.cpp` no longer matched clang-format: the `alignas(struct nlmsghdr)` declarations widened the alignment column in three functions. Reformatted with clang-format 19.1.7, the version CI installs; that file is the only one that had drifted.

Firewall suite 22/22, full build clean, everything folded into the commits that introduce the code.

LorbusChris:
Both taken, plus a fix to the CI job I added.

**Stale netlink replies.** Correct, and worth noting this became reachable because of the receive timeout added a couple of rounds ago: before that a missing ack blocked forever, so nothing was ever left unread; now a late ack lands in the socket buffer. `CommitBatch()` cannot sequence-check its replies — a batch spans many sequence numbers, which is why it passes `0` to `mnl_cb_run()` — so a stale reply would have been consumed as if it belonged to the next batch. Added a `DrainSocket()` before both sends. `SendStandalone()` does validate its sequence number, so there a stale reply meant a spurious failure rather than silent desync, but draining is right for it too.

**The downcast in `UpdateIngressPrefixes()`.** Moved below the `mFirewall` check. To be accurate about severity: it is not reachable today, since `mFirewall` is only created in `CreateRcpMode()` and both call sites are inside the RCP path, so `mHost` is always an `RcpHost` when this runs. But the cast sat above the guard, which made the function depend on call-site discipline rather than being self-guarding. Reordering costs nothing.

**CI.** The `nftables-check` job I added was failing, twice, both my fault.

First, `Package 'libnftnl' not found`: I had claimed no bootstrap change was needed because `script/bootstrap` installs libnftnl/libmnl, but CI runs `tests/scripts/bootstrap.sh`, which is a different script and did not. Added there now.

Second, the job ran `script/test build check`, and `check` runs the whole ctest suite, part of which launches otbr-agent. With the backend enabled the agent installs nftables rules at startup and dies if it cannot, which on an unprivileged runner it cannot:

```
[EMERG]-APP-----: FAILED application.cpp:339 - -1: Failed to initialize the firewall manager!
```

The job now builds and runs only the firewall unit tests, which mock `INftables` and need no privileges. I have corrected the commit message accordingly: the job gives compile coverage of `nftables_impl.cpp` plus the mocked policy-layer tests, not the backend exercised against a kernel.

Which is the thing I would flag before this merges. Every issue found in the last several rounds — the batch left open on error, the receive timeout, `EINTR`, the zero-length reply, and now stale replies — lives in code that no test reaches, because the suite mocks the backend and never opens a socket. `nftables_impl.cpp` has been validated against a live kernel once, before all of those changes. I am happy to keep fixing what review turns up, but a socket seam that lets these paths be unit-tested, or the on-hardware run, would be worth more than another pass. I have a seam prototyped if that is of interest.

LorbusChris:
Both taken.

**`Deinit()` and `mPendingHandles`.** Correct. Those entries point at caller stack variables — `FirewallManager` passes `&handle` for locals — so keeping them across a `Deinit()`/`Init()` cycle risks a later reply matching a stale sequence number and writing through a dangling pointer.

Rather than adding the missing line, `Deinit()` now calls `AbortBatch()`, which already stops the batch, frees the buffer, clears `mInBatch` *and* clears the map. `Deinit()` was duplicating three of those four steps by hand, which is how the fourth came to be missing, so this removes the duplication as well as the bug.

**`EINTR` in `DrainSocket()`.** Also correct, and it undercut the change it belonged to: a signal would end the drain early and leave behind exactly the stale replies the drain exists to remove. It now continues on `EINTR` and stops on `EAGAIN`, matching `CommitBatch()`.

Worth saying plainly: that hole was avoidable. `EINTR` handling was added to `CommitBatch()` and `SendStandalone()` a few rounds ago, and I wrote a new receive loop without applying the same rule. Thanks for catching it.

Firewall suite 22/22, full build clean, clang-format clean, folded into the commit that introduces the backend. CI on the previous head came back green for `nftables-check` (Debug and Release) and `pretty`.

LorbusChris:
One taken, one declined.

**`MakeBitwiseMask()` bounds.** Taken. The XOR operand is read out of a fixed 16-byte buffer at `aLen` bytes, so it now refuses anything longer. The only caller passes a literal 16, so this is defence in depth on a length-parameterised helper rather than a live bug.

**`Ip6Prefix` initialisation.** Not applying this, and I want to lay out why, because this is the fourth time it has come round and the guidance has now reversed:

- Originally the code was `Ip6Prefix mlPrefix;` followed by a `memset`. Review asked for the `memset` to go, on the grounds that `memset` on a non-POD is unsafe. It was removed.
- Then review asked for `Ip6Prefix mlPrefix{}`, on the grounds that the bare declaration leaves stack garbage. That was applied.
- Now review asks to drop `{}` and add the `memset` back, on the grounds that `{}` may not zero the member.

Those cannot all hold, and in fact none of the three states differ:

```cpp
class Ip6Prefix
{
public:
    Ip6Prefix(void) { Clear(); }
    ...
    void Clear(void) { memset(reinterpret_cast<void *>(this), 0, sizeof(*this)); }
```

`src/common/types.hpp:337`, `:343`, `:396`. `Ip6Prefix` has a user-provided default constructor, so both `Ip6Prefix mlPrefix;` and `Ip6Prefix mlPrefix{}` run it, and it zeroes the entire object including `mPrefix`. There is no uninitialised memory for the `memcpy` to leave behind, and the suggested `memset` would be dead code — the same dead code removed earlier.

Leaving it as `Ip6Prefix mlPrefix{}`. Happy to be overruled by a maintainer, but I would rather not churn the line a third time on reasoning that reverses each round.

Firewall suite 22/22, build clean, clang-format clean, folded into the commits that introduce the code.

src/firewall/nftables_impl.cpp
```diff
@@ -0,0 +1,1324 @@
...
+            if (byteIndex * 8 >= aPrefix.mPrefixLen)
+            {
+                startAddr[byteIndex] = 0;
+            }
+            else if (byteIndex * 8 + 8 > aPrefix.mPrefixLen)
+            {
+                startAddr[byteIndex] &= static_cast<uint8_t>(0xff << (8 - (aPrefix.mPrefixLen - byteIndex * 8)));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
LorbusChris: Agreed, the two blocks were identical. Deduplicated: `ApplyPrefixMask` is now defined above `AttachPrefixInterval` (moved rather than forward-declared — it has no dependencies that made the original placement necessary) and the inline loop is a call to it.

The old guard `if (mPrefixLen < 128) { ... }` and `ApplyPrefixMask`'s `if (aPrefixLen >= 128) return;` are equivalent, so behaviour is unchanged. Firewall gtests still pass.

src/firewall/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+        ${LIBNFTNL_INCLUDE_DIRS}
+        ${LIBMNL_INCLUDE_DIRS}
+    )
+
+    target_link_libraries(otbr-firewall PRIVATE
+        $<$<BOOL:${LIBNFTNL_LIBRARY_DIRS}>:-L$<JOIN:${LIBNFTNL_LIBRARY_DIRS}," -L">>
+        $<$<BOOL:${LIBMNL_LIBRARY_DIRS}>:-L$<JOIN:${LIBMNL_LIBRARY_DIRS}," -L">>
+        ${LIBNFTNL_LIBRARIES}
+        ${LIBMNL_LIBRARIES}
+    )
```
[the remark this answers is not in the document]
LorbusChris: `target_link_directories` was added in CMake 3.13, and this file is governed by the top-level `cmake_minimum_required(VERSION 3.10.2)`, so it isn't available here — it would be an unknown-command error for anyone configuring with 3.10–3.12.

The `-L` generator-expression form is what the tree already uses for pkg-config-provided dependencies: `src/dbus/common/CMakeLists.txt:42` and `src/web/CMakeLists.txt:47` are the same idiom, so this follows existing convention rather than inventing one.

Your second suggestion does work at 3.10 — `pkg_check_modules(... IMPORTED_TARGET ...)` has been available since 3.6, so `PkgConfig::LIBNFTNL` would link fine. I've left it alone only because converting this one target would make it the odd one out; if maintainers want that style, a repo-wide switch seems like the better change.

LorbusChris:
Folded a testability change into the first two commits.

The netlink socket now sits behind an `INetlinkSocket` interface (`MnlNetlinkSocket` in production), so `Nftables` can be constructed against a fake. The error paths raised across the review rounds were previously unreachable from a unit test — they all depend on how the socket behaves, and there was no way to make a real one time out, return short, or fail mid-batch on demand.

Six cases in `otbr-gtest-firewall` now cover them:

- retrying a receive interrupted by a signal, and *not* retrying a genuine timeout
- rejecting a zero-length reply instead of reading past it
- draining stale replies before sending a new batch
- a failed operation not leaving the backend wedged for the process lifetime
- a nested `BeginBatch()` being rejected without tearing down the batch already open

Each was checked by reverting the corresponding fix and confirming the matching test fails; all six do. Counts are 28 tests with `-DOTBR_NFTABLES=ON` and 22 with it off, where the backend cases compile out.

src/firewall/nftables_impl.cpp
```diff
@@ -0,0 +1,1276 @@
...
+#include <libnftnl/set.h>
+#include <libnftnl/table.h>
+
+#include "common/logging.hpp"
+
+namespace otbr {
+namespace Firewall {
+
+Nftables::Nftables(void)
+    : mOwnedSocket(new MnlNetlinkSocket())
```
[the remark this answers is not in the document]
LorbusChris: Applied — `MakeUnique<MnlNetlinkSocket>()`. (`std::make_unique` isn't available here since the project builds as C++11, but `otbr::MakeUnique` in `common/code_utils.hpp` covers it.)

src/firewall/firewall_manager.cpp
```diff
@@ -0,0 +1,366 @@
...
+    SuccessOrExit(error = mNftables.AddRuleOifnameNeqReturn(kTableName, kIngressChain, mThreadIfName, &handle));
+    SuccessOrExit(error = mNftables.AddRuleIifPkttypeVerdict(kTableName, kIngressChain, mThreadIfName,
+                                                             PktType::kUnicast, Verdict::kDrop, &handle));
+    SuccessOrExit(error = mNftables.AddRuleSetLookupVerdict(kTableName, kIngressChain, kIngressDenySrcSet,
+                                                            SetDirection::kSrc, Verdict::kDrop, &handle));
+    SuccessOrExit(error = mNftables.AddRuleSetLookupVerdict(kTableName, kIngressChain, kIngressAllowDstSet,
+                                                            SetDirection::kDst, Verdict::kAccept, &handle));
+    SuccessOrExit(
+        error = mNftables.AddRulePkttypeVerdict(kTableName, kIngressChain, PktType::kUnicast, Verdict::kDrop, &handle));
+    SuccessOrExit(error = mNftables.AddRuleVerdict(kTableName, kIngressChain, Verdict::kAccept, &handle));
```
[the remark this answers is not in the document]
LorbusChris: Applied — the variable was write-only in this path, so all six calls now pass `nullptr`. Confirmed that's a no-op on the wire rather than a behaviour change: `QueueNewRule` sets `NLM_F_ECHO` unconditionally, and `aHandleOut` only decides whether a seq is recorded in `mPendingHandles`, which `HandleEchoReply` already null-checks.

One thing the change needed on top of the diff above: the mock's `ON_CALL` defaults used `SetArgPointee<N>` on those same handle arguments, and `SetArgPointee` writes through the pointer — so with `nullptr` it's undefined behaviour, not a no-op. Dropped the seeding from the nine defaults whose handle is now always null, keeping it for `AddRuleNdNsRedirect`, where the handle is still read into `mNdRuleHandle`.

src/firewall/firewall_manager.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+    // Forward: accept traffic in either direction on the upstream interface.
+    // Hooked at FORWARD/filter alongside forward_ingress; the iifname/oifname
+    // matches scope each rule cleanly.
+    SuccessOrExit(
+        error = mNftables.AddChain(kTableName, kNatForwardChain, Hook::kForward, kPriorityFilter, ChainType::kFilter));
+    SuccessOrExit(error = mNftables.AddRuleOifnameVerdict(kTableName, kNatForwardChain, aUpstreamInterfaceName,
+                                                          Verdict::kAccept, &handle));
+    SuccessOrExit(error = mNftables.AddRuleIifnameVerdict(kTableName, kNatForwardChain, aUpstreamInterfaceName,
+                                                          Verdict::kAccept, &handle));
```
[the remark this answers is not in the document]
LorbusChris: Applied here too — four calls now pass `nullptr` and the variable is gone. Same mock-default caveat as the ingress-filter thread.

src/firewall/netlink_socket.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+uint32_t MnlNetlinkSocket::GetPortId(void) const
+{
+    return mPortId;
+}
+
+ssize_t MnlNetlinkSocket::Send(const void *aBuffer, size_t aLength)
+{
+    return mnl_socket_sendto(mSocket, aBuffer, aLength);
+}
```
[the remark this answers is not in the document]
LorbusChris: Applied, along with the same guard on `Recv` and `Drain`.

Worth being precise about the severity: this isn't reachable through `Nftables` today — `BeginBatch` and the standalone operations all guard on `mSocket.IsOpen()`, and `CommitBatch` requires `mInBatch`, which only a successful `BeginBatch` sets. So it's hardening the seam's contract rather than fixing a live crash.

It is a real dereference though, so I added a test that drives `Send`/`Recv`/`Drain` on a never-opened `MnlNetlinkSocket`. With the guards removed it segfaults (rc=139); with them it returns `-1`/`EBADF` and `Drain` is a no-op.

Written with a single return rather than a `return` before `exit:`, to match the style used elsewhere in the file.

src/firewall/netlink_socket.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+ssize_t MnlNetlinkSocket::Recv(void *aBuffer, size_t aLength)
+{
+    return mnl_socket_recvfrom(mSocket, aBuffer, aLength);
+}
```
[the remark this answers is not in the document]
LorbusChris: Applied — same shape as `Send`, and covered by the same test.

src/firewall/netlink_socket.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+void MnlNetlinkSocket::Drain(void)
+{
+    uint8_t discard[MNL_SOCKET_BUFFER_SIZE];
+
+    while (true)
+    {
+        ssize_t rval = recv(mnl_socket_get_fd(mSocket), discard, sizeof(discard), MSG_DONTWAIT);
```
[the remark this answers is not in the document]
LorbusChris: Applied. `Drain` only needed the one guard, since its existing `VerifyOrExit(rval > 0)` already keeps the `exit:` label live.

LorbusChris:
Friendly ping for a CI approval / first review pass. Two notes that may help planning:

1. This PR is self-contained (nftables backend behind an opt-in flag; legacy iptables path untouched by default).
2. I am about to submit a series adding ubus methods for Matter border-router integration (refactor + pending-dataset support + provision/deprovision methods). The only file both touch is `src/agent/application.cpp`, and the overlap is trivial to rebase in either order, so review order does not matter.

The backend has been running in production on an OpenWrt (Turris Omnia) border router.

src/agent/application.cpp
```diff
@@ -341,8 +376,65 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
...
+    // Tear down the OTBR nftables table while the netlink socket is still open
+    // (mNftables outlives mFirewall by member declaration order).
+    if (mFirewall != nullptr)
+    {
+        otbrError firewallError = mFirewall->Deinit();
+        if (firewallError != OTBR_ERROR_NONE)
+        {
+            otbrLogWarning("FirewallManager: teardown failed (%d)", firewallError);
+        }
+    }
```
jwhui, line 528: Should `mNftables->Deinit()` should be called here after `mFirewall->Deinit()`?

If `mNftables->Deinit()` is not called, the underlying netlink socket remains open. If `InitRcpMode()` is executed again during a daemon lifecycle reset or re-init, `mNftables->Init()` will return `OTBR_ERROR_INVALID_STATE` at `VerifyOrExit(!mSocket.IsOpen())` and trigger `SuccessOrDie`.

etc/docker/border-router/Dockerfile
jwhui, line 134: Do we need to install the runtime packages `libnftnl11` and `libmnl0` (or `libnftnl-dev` / `libmnl-dev`) here?

src/firewall/nftables_impl.cpp
```diff
@@ -0,0 +1,1262 @@
...
+    Nftables *self = static_cast<Nftables *>(aContext);
+
+    // ECHO replies for NFT_MSG_NEWRULE carry the kernel-assigned handle.
+    if ((aNlh->nlmsg_type & 0xff) == NFT_MSG_NEWRULE)
+    {
+        auto it = self->mPendingHandles.find(aNlh->nlmsg_seq);
+        if (it != self->mPendingHandles.end() && it->second != nullptr)
+        {
+            struct nftnl_rule *r = nftnl_rule_alloc();
+            if (r != nullptr && nftnl_rule_nlmsg_parse(aNlh, r) >= 0)
```
jwhui: Should we propagating the parsing error so the transaction commit fails if the handle cannot be retrieved?

If `nftnl_rule_alloc()` fails or `nftnl_rule_nlmsg_parse()` returns an error, returning `MNL_CB_OK` allows `CommitBatch()` to succeed while leaving the out-pointer handle unassigned.

When `FirewallManager::EnableNdProxyRedirect()` receives handle `0`, any subsequent call to update or `DisableNdProxyRedirect()` will skip deleting the installed rule, leaking the rule in the kernel.

src/agent/application.cpp
```diff
@@ -318,6 +330,29 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
...
+    SuccessOrDie(mFirewall->EnableIngressFilter(), "Failed to install the Thread ingress filter!");
+    if (!mBackboneInterfaceName.empty())
+    {
+        SuccessOrDie(mFirewall->EnableNat44Masquerade(mBackboneInterfaceName), "Failed to install NAT44 masquerade!");
+    }
+    // Populate the ingress allow/deny sets from Thread network data, and keep
+    // them in sync as it changes. This is the in-process replacement for the
+    // OpenThread posix platform firewall's ipset producer.
+    rcpHost.AddThreadStateChangedCallback([this](otChangedFlags aFlags) {
+        if (aFlags & OT_CHANGED_THREAD_NETDATA)
```
jwhui: Should we also include `OT_CHANGED_ACTIVE_DATASET` so that changes to the Mesh-Local Prefix are updated immediately in the ingress deny set?

.github/workflows/build.yml
```diff
@@ -105,6 +105,32 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        build_type: ["Debug", "Release"]
+    env:
+      BUILD_TARGET: check
+      OTBR_BUILD_TYPE: ${{ matrix.build_type }}
+      OTBR_MDNS: openthread
+      OTBR_OPTIONS: "-DOTBR_NFTABLES=ON -DOT_FIREWALL=OFF -DBUILD_TESTING=ON"
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
jwhui: We recently updated this yml files due to some new policy updates.

For consistency with the rest of the workflow and to satisfy Zizmor security scanning, please update `actions/checkout` to `@3d3c42e5aac5ba805825da76410c181273ba90b1` and add `persist-credentials: false`.

src/firewall/nftables.hpp
```diff
@@ -0,0 +1,285 @@
...
+    kPriorityMangle = -150,
+    kPriorityFilter = 0,
+    kPrioritySrcNat = 100,
+};
+
+/**
+ * An IPv6 prefix in wire format. Used for set elements and for the
+ * domain-prefix match in the ND-proxy redirect rule.
+ */
+struct Ip6Net
```
jwhui: This appears to duplicate `otbr::Ip6Prefix` (defined in `common/types.hpp`, which is already included in this file).

Using `otbr::Ip6Prefix` directly throughout `INftables` would eliminate the need for `Ip6Net` and helper conversions like `FirewallManager::ToIp6Net()`.

src/firewall/nftables_impl.cpp
```diff
@@ -0,0 +1,1262 @@
...
+    mInBatch     = false;
+    mPendingHandles.clear();
+}
+
+int Nftables::HandleEchoReply(const struct nlmsghdr *aNlh, void *aContext)
+{
+    Nftables *self = static_cast<Nftables *>(aContext);
+
+    // ECHO replies for NFT_MSG_NEWRULE carry the kernel-assigned handle.
+    if ((aNlh->nlmsg_type & 0xff) == NFT_MSG_NEWRULE)
```
jwhui: Use the standard Netfilter macro `NFNL_MSG_TYPE(aNlh->nlmsg_type)` from `<linux/netfilter/nfnetlink.h>` instead of masking manually with `& 0xff`?

src/agent/application.cpp
```diff
@@ -341,8 +376,65 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
...
+    otInstance              *instance;
+
+    // The firewall only exists in RCP mode, so check it before downcasting mHost.
+    VerifyOrExit(mFirewall != nullptr && mFirewall->IsIngressFilterEnabled());
+
+    instance = static_cast<otbr::Host::RcpHost &>(mHost).GetInstance();
+    VerifyOrExit(instance != nullptr);
+
+    // Deny-src and allow-dst both cover every on-mesh prefix.
+    while (otNetDataGetNextOnMeshPrefix(instance, &iterator, &config) == OT_ERROR_NONE)
```
jwhui, line 488: Should we deduplicating prefixes before adding them?

On-mesh prefixes in Thread Network Data can contain duplicate entries (e.g. when multiple border routers advertise the same on-mesh prefix).

If duplicate prefixes are passed to `ReplaceIngressPrefixes()`, adding duplicate or overlapping intervals to an `NFT_SET_INTERVAL` set in the same transaction causes the kernel to fail the entire batch with `EEXIST`/`EINVAL`.

src/agent/application.cpp
LorbusChris: Done, DeinitRcpMode() now closes the netlink socket after firewall teardown

src/firewall/nftables_impl.cpp
LorbusChris: Done, alloc/parse failure now returns MNL_CB_ERROR (errno EBADMSG)

etc/docker/border-router/Dockerfile
LorbusChris: Added libmnl0 and libnftnl11

.github/workflows/build.yml
LorbusChris: Done, matches the other jobs now

src/agent/application.cpp
LorbusChris: Done, the callback now refreshes on OT_CHANGED_ACTIVE_DATASET too.

src/firewall/nftables.hpp
LorbusChris: Indeed, removed.

src/firewall/nftables_impl.cpp
LorbusChris: Done.

src/agent/application.cpp
LorbusChris: Yes, thanks. Fixed so only distinct prefixes go into the sets now.

LorbusChris:
@jwhui thanks for the thorough review! I think I've addressed all points. PTAL.

src/firewall/nftables_impl.cpp
```diff
@@ -0,0 +1,1270 @@
...
+    return nftnl_expr_alloc("masq");
+}
+
+} // namespace
+
+otbrError Nftables::QueueNewRule(struct nftnl_rule *aRule, uint64_t *aHandleOut)
+{
+    struct nlmsghdr *nlh =
+        nftnl_rule_nlmsg_build_hdr(reinterpret_cast<char *>(mnl_nlmsg_batch_current(mBatch)), NFT_MSG_NEWRULE,
+                                   NFPROTO_INET, NLM_F_APPEND | NLM_F_CREATE | NLM_F_ECHO | NLM_F_ACK, mSeq);
```
jwhui: Should we conditionally set `NLM_F_ECHO`?

`NLM_F_ECHO` is only needed when the caller passes a non-null `aHandleOut` to capture the kernel handle. For static rules where `aHandleOut == nullptr` (e.g. `EnableIngressFilter` and `EnableNat44Masquerade`), setting `NLM_F_ECHO` generates extra Netlink echo replies from the kernel that are discarded during commit.

```suggestion
    uint16_t flags = NLM_F_APPEND | NLM_F_CREATE | NLM_F_ACK | (aHandleOut != nullptr ? NLM_F_ECHO : 0);
    struct nlmsghdr *nlh =
        nftnl_rule_nlmsg_build_hdr(reinterpret_cast<char *>(mnl_nlmsg_batch_current(mBatch)), NFT_MSG_NEWRULE,
                                   NFPROTO_INET, flags, mSeq);
```

tests/gtest/test_firewall_manager.cpp
```diff
@@ -0,0 +1,672 @@
...
+    // installed above, i.e. DelRule() is expected for both.
+    EXPECT_CALL(mock, CommitBatch()).WillOnce(Return(OTBR_ERROR_ERRNO)).WillRepeatedly(Return(OTBR_ERROR_NONE));
+    EXPECT_CALL(mock, AbortBatch()).Times(1);
+    EXPECT_CALL(mock, DelRule(_, _, 4242u)).Times(2);
+
+    EXPECT_NE(fw.EnableNdProxyRedirect(domain, "eth0", 88), OTBR_ERROR_NONE);
+    EXPECT_EQ(fw.EnableNdProxyRedirect(domain, "eth0", 88), OTBR_ERROR_NONE);
+}
+
+TEST(FirewallManagerTest, IngressSetOpsRequireEnableIngressFilter)
```
jwhui, line 496: Should we add a test for `FirewallManager::ReplaceIngressPrefixes()`?

This is the main path driven by `otbr-agent` on Thread network-data / dataset updates. Testing that both sets are flushed and refilled in sequence, as well as verifying that invalid prefix arguments abort the batch, would round out the `FirewallManager` test suite.

etc/cmake/options.cmake
```diff
@@ -86,6 +86,19 @@ if (OTBR_TELEMETRY_DATA_API)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_TELEMETRY_DATA_API=1)
 endif()
 
+# Opt-in, Linux-only: build the in-process nftables firewall backend instead of
+# relying on the legacy ipset/ip6tables shell firewall (which remains the
+# default). Requires libnftnl + libmnl at build time and nf_tables kernel
+# support at runtime.
+option(OTBR_NFTABLES "Enable in-process nftables firewall backend (libnftnl/libmnl)" OFF)
+if (OTBR_NFTABLES)
+    pkg_check_modules(LIBNFTNL REQUIRED libnftnl)
```
jwhui: Add a platform check and using `IMPORTED_TARGET` with `pkg_check_modules`?
```
    if (NOT CMAKE_SYSTEM_NAME STREQUAL "Linux")
        message(FATAL_ERROR "OTBR_NFTABLES is only supported on Linux.")
    endif()
```
This allows `src/firewall/CMakeLists.txt` and `tests/gtest/CMakeLists.txt` to simply link `PkgConfig::LIBNFTNL` and `PkgConfig::LIBMNL` directly instead of manually managing include directories and library flags.

src/firewall/nftables.hpp
```diff
@@ -0,0 +1,275 @@
...
+{
+    kUnicast,
+};
+
+/**
+ * Standard nftables chain priorities (see linux/netfilter_ipv4.h /
+ * libnftnl docs). Listed here so callers don't have to pull kernel
+ * headers.
+ */
+enum ChainPriority
```
jwhui: enums above use `enum class`, but this does not. Is this intentional?

src/firewall/nftables_impl.cpp
LorbusChris: Done

tests/gtest/test_firewall_manager.cpp
LorbusChris: Added two: one pins the flush-then-refill order across both sets inside a single batch, the other checks an invalid prefix aborts the batch without a commit.

etc/cmake/options.cmake
LorbusChris: Done

src/firewall/nftables.hpp
LorbusChris: Not intentional, fixed. ChainPriority is an enum class now, with the values renamed to match the other enums (ChainPriority::kFilter like Hook::kForward).

jwhui:
@LorbusChris , thanks for addressing all the previous review comments!

Looking ahead, this in-process nftables backend is significantly cleaner, faster, and more robust than the legacy shell-based ipset/ip6tables approach.

What is your envisioned strategy/roadmap for the legacy firewall support? Specifically:

1. Do you see any blocker to immediately making OTBR_NFTABLES=ON the default for Linux builds?
2. Are there specific legacy targets or kernel constraints where ipset/ip6tables is still strictly needed, or should we plan to deprecate and eventually remove the legacy scripts (script/_firewall, script/_nat64, etc.) and src/posix/platform/firewall.cpp in OpenThread?

LorbusChris:
@jwhui likewise thanks again for the review.

>     1. Do you see any blocker to immediately making OTBR_NFTABLES=ON the default for Linux builds?

I don't see any blockers, but I didn't want to be pushy here proposing that change right away.
 
>     2. Are there specific legacy targets or kernel constraints where ipset/ip6tables is still strictly needed, or should we plan to deprecate and eventually remove the legacy scripts (script/_firewall, script/_nat64, etc.) and src/posix/platform/firewall.cpp in OpenThread?

I'm not aware of any legacy targets but that's obviously not an authoritative answer :) I think making a plan for deprecation of the iptables backend in a future release would be sensible and in line with the broader Linux ecosystem.

jwhui:
@LorbusChris , you mentioned running this on OpenWrt / Turris Omnia - are there specific user communities or downstream projects currently interested in or already testing this?

jwhui:
@LorbusChris , can you rebase this PR?

Also note the conversation in:

- https://github.com/openthread/ot-br-posix/issues/3515

LorbusChris:
@jwhui rebased onto current main.

Thanks for the pointer to #3515, it applies directly to this PR, so the gating commit now accounts for it. As @skirmess pointed out, an accept in the isolated `inet otbr` table isn't terminal for the hook — the host's `filter FORWARD` policy still runs afterwards. So the in-process backend can filter, but it can't authorize forwarding on a Docker host where that policy is DROP. That means the Docker scripts can't simply skip everything when `OTBR_NFTABLES=1`:

- The ipset/ip6tables ingress filter and the iptables NAT44 rules are skipped — the agent owns those now.
- The accepts that authorize forwarding stay in both modes: the IPv4 rules on the infrastructure interface, and the IPv6 egress accept from #3521.
- nftables mode additionally gets an IPv6 accept towards the Thread interface. The legacy ingress chain's final ACCEPT used to provide that as a side effect of living in the host's filter table; the in-process filter can't, and without it inbound traffic would break on Docker hosts the same way #3515 describes for outbound.

`finish` mirrors the same split.

jwhui:
@skirmess , FYI in case you want to review, since it touches some of your recent contributions.

skirmess:
Thanks @jwhui for the ping. I looked at the parts I can speak to — the Docker scripts and the firewall semantics. I have **not** reviewed the netlink/libnftnl backend or the batch handling; I would not add signal there.

## The reasoning in the gating commit is right

> an accept in the isolated `inet otbr` table isn't terminal for the hook — the host's `filter FORWARD` policy still runs afterwards

Agreed, and it is worth spelling out why the split works, because it rests on an asymmetry that is easy to state backwards: at the same hook and priority, a **drop** in any table is final for the packet, while an **accept** only ends evaluation *within that table*. So an in-process backend can decide "no" on its own, but never "yes" — which is exactly why the authorization accepts have to stay in the host's tables in both modes, and why "what the in-process filter drops is gone before this rule matters" holds regardless of table ordering.

## One thing I think is a real problem: the flag can silently disagree with the build

`OTBR_NFTABLES` is a runtime environment variable that has to match a compile-time decision, and nothing checks that it does. The Dockerfile adds the libraries but never sets it, so an image built with `-DOTBR_NFTABLES=ON` still starts with `OTBR_NFTABLES=0` unless the operator remembers `-e OTBR_NFTABLES=1`. `script/_firewall` acknowledges the same requirement for native installs ("should export `OTBR_NFTABLES=1`").

When they disagree in that direction, the result is not a fallback to the legacy firewall — it is a broken one:

1. `-DOTBR_NFTABLES=ON` forces `OT_FIREWALL OFF` (third_party/openthread/CMakeLists.txt), so nothing fills `otbr-ingress-deny-src` / `otbr-ingress-allow-dst` any more;
2. with the env var unset, `run` still installs the legacy ingress chain and creates those ipsets, empty;
3. `-m set --match-set otbr-ingress-allow-dst dst -j ACCEPT` therefore never matches;
4. the next rule is `-m pkttype --pkt-type unicast -j DROP`.

So every unicast packet forwarded *into* the Thread network is dropped, by a chain whose allow-list can no longer be populated. The border router looks healthy — `state` reports leader, the REST API answers, devices attach with good link quality — and inbound traffic is silently gone. That is the same failure signature as #3515, which is what makes me wary of it: this class of bug is expensive to diagnose from the outside precisely because nothing looks wrong.

The mismatch in the other direction (legacy build, env var set) is benign — no firewall is installed at all, which at least fails loudly in testing.

Since the build already knows the answer, could it record it rather than asking the operator to repeat it? For instance install a marker at `-DOTBR_NFTABLES=ON` time and have `run`/`finish`/`_firewall` read that, keeping the env var as an override:

```sh
if [ -n "${OTBR_NFTABLES:-}" ]; then
    :                                   # explicit override wins
elif [ -e /usr/share/otbr/nftables ]; then
    OTBR_NFTABLES=1
else
    OTBR_NFTABLES=0
fi
```

Or, if you would rather not add a file, have `run` refuse to install the legacy chain when the agent reports the in-process backend, so the two cannot drift apart unnoticed.

## Minor

In `finish`, the comment above the shared removal block says the authorization accepts are installed "in both modes". That is true of the egress accept, but `-o "${OT_THREAD_IF}" -j ACCEPT` is installed only under `OTBR_NFTABLES=1`. Removing it unconditionally is right — it cleans up after a mode switch — but the comment reads as if `run` had installed both in either mode.

Everything else in the two scripts matches what I would expect, and thank you for keeping the reasoning in the comments rather than compressing it away.

LorbusChris:
@skirmess Thanks, that's a real bug: the Dockerfile never set `OTBR_NFTABLES`, so an nftables image started straight into the legacy chain with empty sets.

I went with a marker file. A build with `-DOTBR_NFTABLES=ON` now installs `/usr/share/otbr/nftables-backend`, and `run`, `finish`, `script/_firewall` and `script/_nat64` pick that up when `OTBR_NFTABLES` is not set; an explicit value still wins. Folded into the gating commit. The `finish` comment is fixed as well.

jwhui:
@skirmess , would you like to review again?

skirmess:
Looked again, same scope as before: the Docker scripts and the firewall semantics, not the netlink backend.

The marker resolves the mismatch, and I checked the parts that could have quietly undone it rather than taking the mechanism at face value:

- it reaches the image — the builder runs `DESTDIR=/tmp/otbr-install ninja install` and the runtime stage does `COPY --from=builder /tmp/otbr-install/ /`;
- the hardcoded `/usr/share/otbr/nftables-backend` matches where CMake puts it, because both install paths pin the prefix: `-DCMAKE_INSTALL_PREFIX=/usr` in the Dockerfile and in `script/_otbr`. Worth knowing that the path is only correct as long as that holds — `${CMAKE_INSTALL_FULL_DATADIR}` under a `/usr/local` prefix would land somewhere the scripts never look, and the failure would be silent in the same way the original one was;
- all four consumers (`run`, `finish`, `script/_firewall`, `script/_nat64`) resolve it identically, and an explicit `OTBR_NFTABLES` still wins.

## One case left, and it fails the other way round

A presence-based marker cannot express "off", and `otbr_install()` does not uninstall first — it just runs `ninja install` over the previous installation. So:

1. install a build with `-DOTBR_NFTABLES=ON` → marker present;
2. rebuild without the flag and reinstall → the binary no longer has the in-process backend, but the marker is still there, because nothing removes it;
3. `run` reads the marker, skips the legacy ipset/ip6tables setup;
4. the agent installs nothing either.

The result is a border router with **no ingress filter at all** — not over-filtered like the original bug, but unfiltered, which is the worse direction. And it takes exactly the sequence someone trying the opt-in flag would follow: turn it on, look at it, turn it back off.

Smallest fix that fits what is already there: give the marker a value instead of relying on its presence, and install it in both branches, so a later install overwrites it.

```cmake
if(OTBR_NFTABLES)
    install(FILES nftables-backend-on RENAME nftables-backend ...)
else()
    install(FILES nftables-backend-off RENAME nftables-backend ...)
endif()
```

with the scripts reading the value rather than testing `-e`. Docker is unaffected either way, since the image is replaced wholesale; this is about native installs.

Everything else I raised last time is addressed, including the `finish` comment. Thanks for taking the marker idea further than I sketched it — routing `script/_nat64` through the same resolution is a detail I had missed.

LorbusChris:
@skirmess thanks, also fixed the way you sketched it: the marker carries a value and is installed in both branches.

`src/agent/nftables-backend.in` is configured with `1` or `0` and installed unconditionally, so a
reinstall overwrites whatever the previous one left behind. All four consumers read the value
(`grep -qx '1'`) instead of testing for the file. Anything that is not exactly `1` resolves to the
legacy path: a missing file, an empty one, or a marker from an older build that carried only
comments. Every degenerate case therefore errs towards over-filtering rather than towards no filter
at all.

That covers the prefix point too. Under a prefix other than `/usr` the scripts do not find the file
and take the legacy path, so the mismatch now fails in the recoverable direction instead of the
silent one.

I verified the sequence you described: installing a build with `-DOTBR_NFTABLES=ON` writes `1`,
reinstalling a build without the flag writes `0`, and the resolution flips with it.

Going back over the same area turned up more, all in this push:

- `CommitBatch()` stopped reading at the first `MNL_CB_STOP`. The kernel answers every message of a
  batch, one datagram each and in message order, so the acknowledgement of an earlier message
  arrives ahead of the error of a later one, and a commit that failed halfway was reported as
  successful. For the ingress filter that means a border router forwarding into the mesh with
  nothing filtering it, the same failure direction as the marker case. It drains the socket the way
  `nft(8)` does now, keeping the first failure. Two regression tests cover it, both of which fail
  against the old code.
- `firewall_start()` never got the `OTBR_NFTABLES` guard that `nat44_start()` has, so `script/server`
  would either die on a service that was never installed or start the legacy chain alongside the
  in-process one.
- Skipping `nat44_install` also dropped the `filter FORWARD ... ACCEPT` authorizations. The Docker
  run script installs those in both modes, since an accept in the isolated `inet otbr` table cannot
  authorize against the host's forward policy. `nat44_install` now installs them in both modes as
  well and drops only the marking and the masquerade.
- `script/setup` runs `otbr_uninstall`, which deletes the marker, and the firewall and nat44 setup
  before `otbr_install` writes it, so on that path the marker was never readable when the scripts
  consulted it. The cause underneath was that `otbr_install` never passed `-DOTBR_NFTABLES` at all,
  so that path could not build the backend in the first place. It honors the variable now, which
  also makes the build and the services agree within a run without depending on the file.
- The Docker teardown removed the legacy chain, jump and sets only in legacy mode, and an
  nftables-mode start did not clear them either, so state from a container that was killed survived
  the switch. Teardown is unconditional now, and an nftables-mode start clears what a previous run
  left behind.
- CI ran `ctest -R Firewall`, which matched only the 24 `FirewallManagerTest` cases because
  `gtest_discover_tests` registers each case separately. The `NftablesBackendTest` and
  `MnlNetlinkSocketTest` cases, the only ones exercising the netlink code the job exists for, were
  never run. All 33 run now.

Also took the firewall calls in `application.cpp` off `SuccessOrDie`'s argument, since the macro
names it twice and the failing call ran a second time before the fatal log.

skirmess:
Checked the marker change: `nftables-backend.in` configured to `1`/`0`, installed unconditionally, all four consumers reading the value with the reasoning kept in a comment. That closes the case I raised, and the degenerate paths land on the legacy filter, which is the right direction to fall.

One footnote on the prefix: under a non-`/usr` prefix the legacy path is indeed the safer choice, but the symptom is still the original one — an nftables build has `OT_FIREWALL` off, so the legacy chain filters against sets nobody fills and drops unicast into the mesh while everything looks healthy. Safe rather than open, but not self-announcing. Hypothetical as long as both install paths pin `-DCMAKE_INSTALL_PREFIX=/usr`, so I would not act on it.

Everything else in this push is in the netlink and CI area I said I would not review, so treat my silence there as scope, not approval. For what it is worth, the `ctest -R Firewall` finding is the same shape as the bugs this PR keeps turning up: a job that reported success while never running the cases it existed for.

src/firewall/nftables_impl.cpp
```diff
@@ -0,0 +1,1295 @@
...
+        return 0; // PACKET_HOST
+    }
+    return 0;
+}
+
+void IfNameToWire(const std::string &aName, char (&aOut)[IFNAMSIZ])
+{
+    memset(aOut, 0, IFNAMSIZ);
+    memcpy(aOut, aName.c_str(), aName.size());
+}
```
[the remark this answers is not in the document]
LorbusChris: Took the point but not the truncation: capping at IFNAMSIZ-1 would leave a valid name for a different interface, so the rule would be installed against the wrong one rather than not at all. IfNameToWire now returns otbrError and rejects a name that does not fit, and the six callers propagate it instead of each repeating the bound, so a new caller cannot forget the check. Test added for the rejection and for the longest name that still fits.

jwhui (approved):
Thanks! 👍🏼

jwhui:
@LorbusChris , can you help squash commits?

LorbusChris:
> @LorbusChris , can you help squash commits?

@jwhui done. The commit message is rather long now, but so is the contents of the commit. Please let me know if you want that shortened or adapted any other way. Thanks!

PR #3330, opened 2026-05-07
[rest] add REST API version in /.well-known/thread for API discovery

martinzi:
- .well-known/thread resource (RFC 8615) providing semantic API version string and entry point links
- Added OpenAPI schema

### DIFF
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -24,6 +24,43 @@ tags:
     description: Collection of Thread network diagnostic responses.
 paths:
 
+  ### Well-Known ##############################################################
+  #############################################################################
+  /.well-known/thread/br-rest:
+    get:
+      operationId: getWellKnownThread
+      summary: Read API version and entry points
+      description: |
+        Information about the supported API version (see gray tag at the top of this specification), API location (base path), and links to its entry points.
+
+        ## Static
+        API version, location, and entry points are known at compile time.
+
+        ## Link Attributes (`/.well-known/thread/br-rest`)
+        `rel="self" type="application/json"`
+      tags:
+        - Well-Known
+      responses:
+        "200":
+          $ref: "#/components/responses/WellKnownThread200"
+
+    options:
+      operationId: optionsWellKnownThread
+      tags:
+        - Well-Known
+      summary: List of allowed operations.
+      description: |
+        Lists allowed operations.
+      responses:
+        "204":
+          description: List of allowed operations.
+          headers:
+            Allow:
+              description: Allowed operations
+              schema:
+                type: string
+                example: "GET, OPTIONS"
+
   ### Information #############################################################
   #############################################################################
   /api/node:
@@ -2399,6 +2436,46 @@ components:
           description: Joiner expiration time in milliseconds on response and seconds on request 
           default: 60 
 
+    WellKnownThread:
+      type: object
+      required:
+        - api
+        - links
+      properties:
+        api:
+          type: object
+          required:
+            - version
+            - base
+          properties:
+            version:
+              type: string
+              pattern: ^[0-9]+\.[0-9]+(\.[0-9]+)?$
+              description: Semantic version (MAJOR.MINOR.PATCH) of the API
+            base:
+              type: string
+              pattern: ^/[/.a-zA-Z0-9_-]*$
+        links:
+          type: array
+          items:
+            type: object
+            required:
+              - href
+              - rel
+            properties:
+              href:
+                type: string
+                format: uri-reference
+              rel:
+                type: string
+              type:
+                type: array
+                minItems: 1
+                items:
+                  type: string
+              title:
+                type: string
+
   ### Parameters ##############################################################
   #############################################################################
   parameters:
@@ -2516,6 +2593,39 @@ components:
   ### Responses ###############################################################
   #############################################################################
   responses:
+
+    WellKnownThread200:
+      description: |
+        JSON (RFC 8259) providing the API version resource object and links to the API entry points.
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/WellKnownThread"
+          example:
+            api:
+              version: 0.3.0
+              base: /api/
+            links:
+              - href: /.well-known/thread/br-rest
+                rel: self
+                type:
+                  - application/json
+              - href: /api/node
+                rel: node
+                type:
+                  - application/vnd.api+json
+              - href: /api/actions
+                rel: task
+                type:
+                  - application/vnd.api+json
+              - href: /api/devices
+                rel: device
+                type:
+                  - application/vnd.api+json
+              - href: /api/diagnostics
+                rel: diagnostic
+                type:
+                  - application/vnd.api+json
 
     BadRequestError:
       description: |
--- a/src/rest/rest_web_server.cpp
+++ b/src/rest/rest_web_server.cpp
@@ -50,6 +50,7 @@
 #include "rest/rest_devices_coll.hpp"     // Devices Collection
 #include "rest/rest_diagnostics_coll.hpp" // Diagnostics Collection
 #include "rest/services.hpp"
+#include "rest/version.hpp"
 #include "utils/string_utils.hpp"
 
 #include <cJSON.h>
@@ -109,6 +110,8 @@
 
 #define OT_REST_ROUTE_DIAGNOSTICS "/api/diagnostics"
 #define OT_REST_ROUTE_DIAGNOSTICS_ID "/api/diagnostics/:id"
+
+#define OT_REST_ROUTE_WELLKNOWN_THREAD "/.well-known/thread/br-rest"
 
 using std::chrono::duration_cast;
 using std::chrono::microseconds;
@@ -187,6 +190,9 @@ RestWebServer::RestWebServer(Host::RcpHost &aHost)
     mServer.Delete(OT_REST_ROUTE_DIAGNOSTICS_ID,
                    MakeHandlerInMainLoop(&RestWebServer::ApiDiagnosticsItemDeleteHandler));
     mServer.Options(OT_REST_ROUTE_DIAGNOSTICS, MakeHandlerInMainLoop(&RestWebServer::ApiDiagnosticsHandler));
+
+    mServer.Get(OT_REST_ROUTE_WELLKNOWN_THREAD, MakeHandler(&RestWebServer::WellKnownThreadHandler));
+    mServer.Options(OT_REST_ROUTE_WELLKNOWN_THREAD, MakeHandler(&RestWebServer::WellKnownThreadHandler));
 }
 
 RestWebServer::~RestWebServer(void)
@@ -1148,8 +1154,7 @@ void RestWebServer::ApiActionsHandler(const Request &aRequest, Response &aRespon
         aResponse.set_header("Allow", "GET, POST, DELETE, OPTIONS");
         break;
     default:
-        // aResponse.SetAllowMethods(methods);
-        errorDetails = "method not supported";
+        errorDetails = "not supported";
         statusCode   = StatusCode::MethodNotAllowed_405;
         break;
     }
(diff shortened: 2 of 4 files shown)


src/rest/rest_web_server.cpp
```diff
@@ -1806,6 +1811,50 @@ void RestWebServer::ApiDevicesNodeInit()
...
+      "rel": "device",
+      "type": ["application/vnd.api+json"]
+    },
+    {
+      "href": "/api/diagnostics",
+      "rel": "diagnostic",
+      "type": ["application/vnd.api+json"]
+    }
+  ]
+})";
```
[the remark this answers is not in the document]
martinzi: changed as proposed

src/rest/openapi.yaml
```diff
@@ -24,6 +24,26 @@ tags:
...
 paths:
 
+  ### Well-Known ##############################################################
+  #############################################################################
+  /.well-known/thread:
+    get:
+      operationId: getWellKnownThread
+      summary: Read API version and entry points
+      description: |
+        Information about the supported API version (see gray tag at the top of this specification), API location (base path), and links to its entry points.
```
agners, line 34: Worth mentioning RFC 8615 here?

agners (commented):
I realized that the well known URI `thread` got registered a while back at IANA already (see https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml). Do you know for which service/use case that got registered. I wonder if the OTBR REST API should claim `thread`, or if it would rather be better to claim something OTBR specific (maybe `otbr`?).

src/rest/openapi.yaml
martinzi: Likely not much, as RFC 8615 is limited to the URI path. The content is application specific and uses RFC 8288 link attributes with currently specific relation type names.

martinzi:
> I realized that the well known URI `thread` got registered a while back at IANA already (see https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml). Do you know for which service/use case that got registered. I wonder if the OTBR REST API should claim `thread`, or if it would rather be better to claim something OTBR specific (maybe `otbr`?).

The well kown URI `thread` is specified in the meshcop API for CoAP(s) and only defined as a path prefix. So, reusing it here as defined should be acceptable. @EskoDijk What would you recommend?

EskoDijk:
> The well kown URI `thread` is specified in the meshcop API for CoAP(s) and only defined as a path prefix. So, reusing it here as defined should be acceptable. @EskoDijk What would you recommend?

@martinzi It was registered for any (standard) usages related to the Thread specifications. So it's indeed an entry point i.e. parent resource, with specific functions underneath. Original use case was to extend meshcop relaying outside of a Thread mesh e.g. onto a company network, where that extra relay could then forward messages to a commercial Registrar either on-site or in the cloud.

Example: `/.well-known/thread/c/rxr` was defined for sending a relayed packet destined towards a Registrar.

Using this well-known URI is ok as long as the base resource `/.well-known/thread` itself is not used. I would prefer `thread` over others like `otbr`.

agners:
> Using this well-known URI is ok as long as the base resource `/.well-known/thread` itself is not used. I would prefer `thread` over others like `otbr`.

Isn't the current PR do exactly that, respond to a `GET` request to  `/.well-known/thread`?

https://github.com/openthread/ot-br-posix/pull/3330/changes#diff-8e69ee76b1a98fcf833d37948d9b0f9ea06290c526c91d56bf27ff4f47f42394R192

The current implementation doesn't have services there, but just return a json which responds where OTBR REST API is available at. Should we maybe use a sub-resource instead, e.g. `/.well-known/thread/otbr` or  `/.well-known/thread/br-rest`? Matrix (chat system) uses sub-resources for service discovery, e.g. `/.well-known/matrix/server`.

@EskoDijk @martinzi thoughts?

EskoDijk:
> Should we maybe use a sub-resource instead, e.g. `/.well-known/thread/otbr` or `/.well-known/thread/br-rest`? Matrix (chat system) uses sub-resources for service discovery, e.g. `/.well-known/matrix/server`.

Yes, didn't look at the code yet until now - but indeed a sub-resource needs to be used. If we pick a name without "ot" in it, then it becomes easier to generalize this to multiple vendors later on. And standardize it.
Such as `/.well-known/thread/br-rest`.

martinzi:
> > Should we maybe use a sub-resource instead, e.g. `/.well-known/thread/otbr` or `/.well-known/thread/br-rest`? Matrix (chat system) uses sub-resources for service discovery, e.g. `/.well-known/matrix/server`.
> 
> Yes, didn't look at the code yet until now - but indeed a sub-resource needs to be used. If we pick a name without "ot" in it, then it becomes easier to generalize this to multiple vendors later on. And standardize it. Such as `/.well-known/thread/br-rest`.

The idea of this resource path is to provide page wide metadata and/or discovery information. This comprises solely gettable static information. Isn't this part of the `.well-known` concept? It would also not block routing to other services below `.well-known/thread/otherservices`. 
We may move the version information elsewhere, e.g. to `api/version`, but having the resources discoverable seems still useful on the long term.

agners:
> > Yes, didn't look at the code yet until now - but indeed a sub-resource needs to be used. If we pick a name without "ot" in it, then it becomes easier to generalize this to multiple vendors later on. And standardize it. Such as `/.well-known/thread/br-rest`.
> 
> The idea of this resource path is to provide page wide metadata and/or discovery information. This comprises solely gettable static information. Isn't this part of the `.well-known` concept? It would also not block routing to other services below `.well-known/thread/otherservices`. 

I think the idea of using something underneath ``/.well-known` is fine, but just not use the main resource ``/.well-known/thread` for this case. While other things could get merged into the same resource, it would still need coordination since it is underneath the same endpoint.

IMHO, using `/.well-known/thread/br-rest` is perfectly fine. It is just as well known, and then the resource is completely controlled by the REST service.

> We may move the version information elsewhere, e.g. to `api/version`, but having the resources discoverable seems still useful on the long term.

Works for me too, but I think I'd slightly prefer the `/.well-known/thread/br-rest` path. Or offer the same resource on both endpoints :thinking: .

EskoDijk:
@martinzi @agners  A well-known resource can contain anything that the organisation/specification making the registration request defines.  (Including arbitrary depth of sub-resources.)

There's no requirement or expectation that anything discovery-related is hosted on `/.well-known/thread` , for example.  (Although it could!)

Thread originally used sub-resources of its well-known resource for relaying purposes -> no discovery here. The main benefit of well-known resources is that no discovery is required upfront, to still access these resources. While for any resources outside of the well-known space, always discovery is required to learn the names of the resources.

Formally, per IETF, no specification is allowed to reserve/claim URI paths outside of well-known for their own use (e.g. require that `/api/version` is always available for a particular purpose isn't allowed then for a standard).

So something underneath /.well-known/thread looks still best, to avoid both of these issues: 1) name-squatting on URI paths ; and 2) needing to introduce a discovery solution to find URI path names.

And why not the resource `/.well-known/thread` itself?  That's because it is supposed to be hosting either nothing or something that would benefit all of the Thread functionality that lives today under the well-known resource, or that's going to be added in the future to it.   For example it could be a discovery resource of some sort that covers all of Thread functionality of the given device.

So it could be: `/.well-known/thread/br-rest`, or `/.well-known/thread/br-rest/api`, or `/.well-known/thread/br-rest/api/version`, or `/.well-known/thread/br-rest-api/version` or anything similar.

martinzi:
Moved discovery endpoint with version information as is from `/.well-known/thread` to `/.well-known/thread/br-rest`

agners (approved):
LGTM, happy with the solution, thank you both 👍 !

martinzi:
@bukepo Please review this PR.

martinzi:
@jwhui @bukepo Anything left we could do getting this PR merged?

src/rest/rest_web_server.cpp
```diff
@@ -1806,6 +1811,51 @@ void RestWebServer::ApiDevicesNodeInit()
...
+    },
+    {
+      "href": ")" OT_REST_ROUTE_DIAGNOSTICS R"(",
+      "rel": "diagnostic",
+      "type": ["application/vnd.api+json"]
+    }
+  ]
+})";
+
+    aResponse.set_content(kWellKnownThreadJson, "application/json");
```
jwhui:
```suggestion
    aResponse.set_content(kWellKnownThreadJson, OT_REST_CONTENT_TYPE_JSON);
```

src/rest/rest_web_server.cpp
```diff
@@ -185,6 +188,8 @@ RestWebServer::RestWebServer(Host::RcpHost &aHost)
     mServer.Delete(OT_REST_ROUTE_DIAGNOSTICS_ID,
                    MakeHandlerInMainLoop(&RestWebServer::ApiDiagnosticsItemDeleteHandler));
     mServer.Options(OT_REST_ROUTE_DIAGNOSTICS, MakeHandlerInMainLoop(&RestWebServer::ApiDiagnosticsHandler));
+
+    mServer.Get(OT_REST_ROUTE_WELLKNOWN_THREAD, MakeHandler(&RestWebServer::WellKnownThread));
```
jwhui: Should we register an options handler here as well?

src/rest/rest_web_server.cpp
martinzi: Replaced.

src/rest/rest_web_server.cpp
martinzi: Yes, I added options handler.

src/rest/openapi.yaml
```diff
@@ -2399,6 +2419,45 @@ components:
...
+            - version
+            - base
+          properties:
+            version:
+              type: string
+              pattern: ^[0-9]+\.[0-9]+(\.[0-9]+)?$
+              description: Semantic version (MAJOR.MINOR.PATCH) of the API
+            base:
+              type: string
+              pattern: ^/[/.a-zA-Z0-9-_]*$
```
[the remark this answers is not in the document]
martinzi: switched order

src/rest/openapi.yaml
```diff
@@ -2399,6 +2419,45 @@ components:
...
+        links:
+          type: array
+          items:
+            type: object
+            required:
+              - href
```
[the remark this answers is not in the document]
martinzi: added

src/rest/rest_web_server.cpp
```diff
@@ -1813,6 +1816,81 @@ void RestWebServer::ApiDevicesNodeInit()
...
+    static const std::string kWellKnownThreadJson = R"({
+  "api": {
+    "version": ")" OTBR_REST_API_VERSION R"(",
+    "base": "/api/"
+  },
+  "links": [
+    {
+      "href": ")" OT_REST_ROUTE_WELLKNOWN_THREAD R"(",
+      "rel": "self",
+      "type": [OT_REST_CONTENT_TYPE_JSON]
```
jwhui: The preprocessor macro `OT_REST_CONTENT_TYPE_JSON` is not expanded because it is inside a raw string literal. This will result in serving invalid JSON containing `[OT_REST_CONTENT_TYPE_JSON]` at runtime.

To fix this, you need to escape the raw string literal to allow the macro to be expanded.

src/rest/rest_web_server.cpp
```diff
@@ -1813,6 +1816,81 @@ void RestWebServer::ApiDevicesNodeInit()
...
+    },
+    {
+      "href": ")" OT_REST_ROUTE_NODE R"(",
+      "rel": "node",
+      "type": [OT_REST_CONTENT_TYPE_JSONAPI]
```
jwhui: The preprocessor macro `OT_REST_CONTENT_TYPE_JSONAPI` is not expanded because it is inside a raw string literal. Escape the raw string literal to allow macro expansion.

martinzi: Sorry for that, thx.

src/rest/version.hpp
```diff
@@ -0,0 +1,45 @@
+/*
+ *  Copyright (c) 2020, The OpenThread Authors.
```
jwhui: nit:
```suggestion
 *  Copyright (c) 2026, The OpenThread Authors.
```

jwhui (approved):
Thanks! 👍🏼

PR #3336, opened 2026-05-11
[tests] add DNS-SD Discovery Proxy verification to dind_dns_sd.exp

jwhui:
This commit adds verification of the DNS-SD Discovery Proxy capability to the dind_dns_sd.exp test script.

Changes:
- Added a new test section to verify that a Thread device can discover a service published on the infrastructure link via the OTBR's Discovery Proxy.
- The test uses a Python script with zeroconf in the test-client container to publish _http._tcp.local. service.
- The Thread device queries for _http._tcp.default.service.arpa. and successfully browses and resolves the service.

### DIFF
--- a/tests/scripts/expect/dind_dns_sd.exp
+++ b/tests/scripts/expect/dind_dns_sd.exp
@@ -36,20 +36,6 @@ source "tests/scripts/expect/_common.exp"
 
 # Custom start_otbr_docker for our network name
 proc start_otbr_docker_dind {name sim_app sim_id pty} {
-    set format "\{\{range .IPAM.Config\}\}\{\{.Gateway\}\}\n\{\{end\}\}"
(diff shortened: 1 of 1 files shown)


stu-spp:
What if there are two services broadcasting `_http._tcp.default.service.arpa`. Can DNS-SD resolve and return both?

We have this need in production, where we need to be able to resolve multiple `_mqtt._tcp.default.service.arpa`.

jwhui:
> What if there are two services broadcasting _http._tcp.default.service.arpa. Can DNS-SD resolve and return both?

@stu-spp DNS-SD supports this use case, but requires DNS Push to do it. Unfortunately, OpenThread doesn't yet implement DNS Push.

abtink (approved):

PR #3337, opened 2026-05-12
[host] improvements on `ThreadHelper::AttachAllNodesTo`

wtongze:
This commit allows `ThreadHelper` to use `MGMT_ACTIVE_SET` instead of `MGMT_PENDING_SET` if incoming dataset change does not impact Thread network connectivity (like changing Thread network name) to minimize delay in applying new dataset.

### DIFF
--- a/src/host/thread_helper.cpp
+++ b/src/host/thread_helper.cpp
@@ -98,7 +98,7 @@ void ThreadHelper::StateChangedCallback(otChangedFlags aFlags)
                     otbrLogInfo("StateChangedCallback is called during waiting for Mgmt Set Response");
                     ExitNow();
                 }
-                if (mAttachPendingDatasetTlvs.mLength == 0)
+                if (mAttachDatasetTlvs.mLength == 0)
                 {
                     AttachHandler handler = mAttachHandler;
 
@@ -108,16 +108,26 @@ void ThreadHelper::StateChangedCallback(otChangedFlags aFlags)
                 else
                 {
                     otOperationalDataset emptyDataset = {};
-                    otError              error =
-                        otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, mAttachPendingDatasetTlvs.mTlvs,
-                                                    mAttachPendingDatasetTlvs.mLength, MgmtSetResponseHandler, this);
+                    otError              error        = OT_ERROR_NONE;
+
+                    if (mAttachDelayMs == 0)
+                    {
+                        error = otDatasetSendMgmtActiveSet(mInstance, &emptyDataset, mAttachDatasetTlvs.mTlvs,
+                                                           mAttachDatasetTlvs.mLength, MgmtSetResponseHandler, this);
+                    }
+                    else
+                    {
+                        error = otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, mAttachDatasetTlvs.mTlvs,
+                                                            mAttachDatasetTlvs.mLength, MgmtSetResponseHandler, this);
+                    }
+
                     if (error != OT_ERROR_NONE)
                     {
                         AttachHandler handler = mAttachHandler;
 
-                        mAttachHandler            = nullptr;
-                        mAttachPendingDatasetTlvs = {};
-                        mWaitingMgmtSetResponse   = false;
+                        mAttachHandler          = nullptr;
+                        mAttachDatasetTlvs      = {};
+                        mWaitingMgmtSetResponse = false;
                         handler(error, 0);
                     }
                     else
@@ -557,7 +567,9 @@ void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, At
     otError                  error = OT_ERROR_NONE;
     otOperationalDatasetTlvs datasetTlvs;
     otOperationalDataset     emptyDataset{};
-    otDeviceRole             role = mHost->GetDeviceRole();
+    otDeviceRole             role                         = mHost->GetDeviceRole();
+    uint32_t                 attachDelayMs                = 0;
+    bool                     canImpactNetworkConnectivity = true;
 
     if (aHandler == nullptr)
     {
@@ -578,7 +590,12 @@ void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, At
         ExitNow(error = OT_ERROR_NONE);
     }
 
-    SuccessOrExit(error = ProcessDatasetForMigration(datasetTlvs, kDelayTimerMilliseconds));
+    canImpactNetworkConnectivity = otDatasetAffectsConnectivity(mInstance, &datasetTlvs);
+    if (canImpactNetworkConnectivity)
+    {
+        attachDelayMs = kDelayTimerMilliseconds;
+        SuccessOrExit(error = ProcessDatasetForMigration(datasetTlvs, attachDelayMs));
+    }
 
     assert(datasetTlvs.mLength > 0);
 
@@ -601,22 +618,32 @@ void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, At
 
         if (hasActiveDataset)
         {
-            mAttachDelayMs            = kDelayTimerMilliseconds;
-            mAttachPendingDatasetTlvs = datasetTlvs;
+            mAttachDelayMs     = attachDelayMs;
+            mAttachDatasetTlvs = datasetTlvs;
         }
         else
         {
-            mAttachDelayMs            = 0;
-            mAttachPendingDatasetTlvs = {};
+            mAttachDelayMs     = 0;
+            mAttachDatasetTlvs = {};
         }
         mWaitingMgmtSetResponse = false;
         mAttachHandler          = aHandler;
         ExitNow();
     }
 
-    SuccessOrExit(error = otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, datasetTlvs.mTlvs, datasetTlvs.mLength,
-                                                      MgmtSetResponseHandler, this));
-    mAttachDelayMs          = kDelayTimerMilliseconds;
+    if (canImpactNetworkConnectivity)
+    {
+        SuccessOrExit(error = otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, datasetTlvs.mTlvs,
+                                                          datasetTlvs.mLength, MgmtSetResponseHandler, this));
+    }
+    else
+    {
+        otbrLogInfo("No impact on Thread network connectivity, use MGMT_ACTIVE_SET");
+        SuccessOrExit(error = otDatasetSendMgmtActiveSet(mInstance, &emptyDataset, datasetTlvs.mTlvs,
+                                                         datasetTlvs.mLength, MgmtSetResponseHandler, this));
+    }
+
+    mAttachDelayMs          = attachDelayMs;
     mAttachHandler          = aHandler;
     mWaitingMgmtSetResponse = true;
 
@@ -643,8 +670,8 @@ void ThreadHelper::MgmtSetResponseHandler(otError aResult)
     if (mAttachHandler == nullptr)
     {
         otbrLogWarning("mAttachHandler is nullptr");
-        mAttachDelayMs            = 0;
-        mAttachPendingDatasetTlvs = {};
+        mAttachDelayMs     = 0;
+        mAttachDatasetTlvs = {};
         ExitNow();
     }
 
@@ -658,11 +685,11 @@ void ThreadHelper::MgmtSetResponseHandler(otError aResult)
         break;
     }
 
-    attachDelayMs             = mAttachDelayMs;
-    handler                   = mAttachHandler;
-    mAttachDelayMs            = 0;
-    mAttachHandler            = nullptr;
-    mAttachPendingDatasetTlvs = {};
+    attachDelayMs      = mAttachDelayMs;
+    handler            = mAttachHandler;
+    mAttachDelayMs     = 0;
+    mAttachHandler     = nullptr;
+    mAttachDatasetTlvs = {};
     if (aResult == OT_ERROR_NONE)
     {
         handler(aResult, attachDelayMs);
--- a/src/host/thread_helper.hpp
+++ b/src/host/thread_helper.hpp
@@ -303,7 +303,7 @@ private:
 
     ResultHandler mDetachGracefullyHandler = nullptr;
 
-    otOperationalDatasetTlvs mAttachPendingDatasetTlvs = {};
+    otOperationalDatasetTlvs mAttachDatasetTlvs = {};
 
     std::random_device mRandomDevice;
 


src/host/thread_helper.cpp
```diff
@@ -532,6 +532,43 @@ bool ThreadHelper::AreDatasetTlvsEqualToLocalDatasetTlvs(const otOperationalData
...
 
+bool ThreadHelper::CanImpactNetworkConnectivity(const otOperationalDatasetTlvs &aDatasetTlvs) const
+{
+    bool                 ret = false;
+    otOperationalDataset dataset;
+    otOperationalDataset localDataset;
+    SuccessOrExit(otDatasetGetActive(mInstance, &localDataset), ret = true);
+    SuccessOrExit(otDatasetParseTlvs(&aDatasetTlvs, &dataset), ret = true);
+
+    VerifyOrExit(!dataset.mComponents.mIsChannelPresent || dataset.mChannel == localDataset.mChannel, ret = true);
```
zhanglongxia: We can call `otDatasetIsValid` to check whether the `aDatasetTlvs` is valid. If it is valid, all the `mIsXXXPresent` should be true. Then we can drop the `mIsXXXPresent` check in this method.

src/host/thread_helper.cpp
```diff
@@ -532,6 +532,43 @@ bool ThreadHelper::AreDatasetTlvsEqualToLocalDatasetTlvs(const otOperationalData
...
+                            sizeof(dataset.mMeshLocalPrefix.m8)) == 0,
+                 ret = true);
+    VerifyOrExit(!dataset.mComponents.mIsSecurityPolicyPresent ||
+                     memcmp(&dataset.mSecurityPolicy, &localDataset.mSecurityPolicy, sizeof(dataset.mSecurityPolicy)) ==
+                         0,
+                 ret = true);
+    VerifyOrExit(!dataset.mComponents.mIsChannelMaskPresent || dataset.mChannelMask == localDataset.mChannelMask,
+                 ret = true);
+    VerifyOrExit(!dataset.mComponents.mIsWakeupChannelPresent || dataset.mWakeupChannel == localDataset.mWakeupChannel,
+                 ret = true);
```
zhanglongxia: I think there is no need to check the `mWakeupChannel` here. It is not in the active dataset in the official thread spec.

src/host/thread_helper.cpp
```diff
@@ -532,6 +532,43 @@ bool ThreadHelper::AreDatasetTlvsEqualToLocalDatasetTlvs(const otOperationalData
...
 }
 
+bool ThreadHelper::CanImpactNetworkConnectivity(const otOperationalDatasetTlvs &aDatasetTlvs) const
+{
+    bool                 ret = false;
+    otOperationalDataset dataset;
+    otOperationalDataset localDataset;
+    SuccessOrExit(otDatasetGetActive(mInstance, &localDataset), ret = true);
+    SuccessOrExit(otDatasetParseTlvs(&aDatasetTlvs, &dataset), ret = true);
+
```
zhanglongxia: Add notes here to clarify which member of the active dataset won't affect the network connectivity.

src/host/thread_helper.cpp
```diff
@@ -532,6 +532,48 @@ bool ThreadHelper::AreDatasetTlvsEqualToLocalDatasetTlvs(const otOperationalData
...
+    VerifyOrExit(memcmp(dataset.mNetworkKey.m8, localDataset.mNetworkKey.m8, sizeof(dataset.mNetworkKey.m8)) == 0,
+                 ret = true);
+    VerifyOrExit(memcmp(dataset.mPskc.m8, localDataset.mPskc.m8, sizeof(dataset.mPskc.m8)) == 0, ret = true);
+    VerifyOrExit(
+        memcmp(dataset.mMeshLocalPrefix.m8, localDataset.mMeshLocalPrefix.m8, sizeof(dataset.mMeshLocalPrefix.m8)) == 0,
+        ret = true);
+    VerifyOrExit(memcmp(&dataset.mSecurityPolicy, &localDataset.mSecurityPolicy, sizeof(dataset.mSecurityPolicy)) == 0,
+                 ret = true);
+    VerifyOrExit(dataset.mChannelMask == localDataset.mChannelMask, ret = true);
+    VerifyOrExit(dataset.mWakeupChannel == localDataset.mWakeupChannel, ret = true);
```
zhanglongxia: The description said that the `WakeupChannel` won't affect the connectivity, the code here checks the `WakeupChannel`. I think we should remove the `WakeupChannel` check.

wtongze: Done.

zhanglongxia (approved):
LGTM

src/host/thread_helper.cpp
```diff
@@ -614,9 +661,19 @@ void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, At
...
+    {
+        SuccessOrExit(error = otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, datasetTlvs.mTlvs,
+                                                          datasetTlvs.mLength, MgmtSetResponseHandler, this));
+    }
+    else
+    {
+        otbrLogInfo("No impact on Thread network connectivity, use MGMT_ACTIVE_SET");
+        SuccessOrExit(error = otDatasetSendMgmtActiveSet(mInstance, &emptyDataset, datasetTlvs.mTlvs,
+                                                         datasetTlvs.mLength, MgmtSetResponseHandler, this));
+    }
```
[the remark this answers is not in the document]
bukepo: +1

Or we just set the active dataset and start Thread if the new dataset doesn't break connectivity.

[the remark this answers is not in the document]
wtongze: Done. I've updated `StateChangedCallback`.

bukepo (approved):

bukepo (approved):

src/host/thread_helper.cpp
```diff
@@ -532,6 +544,47 @@ bool ThreadHelper::AreDatasetTlvsEqualToLocalDatasetTlvs(const otOperationalData
...
+ * @retval true   The given dataset can impact the network connectivity.
+ * @retval false  The given dataset does not impact the network connectivity.
+ */
+bool ThreadHelper::CanImpactNetworkConnectivity(const otOperationalDatasetTlvs &aDatasetTlvs) const
+{
+    bool                 ret = false;
+    otOperationalDataset dataset;
+    otOperationalDataset localDataset;
+
+    SuccessOrExit(otDatasetIsValid(&aDatasetTlvs, false), ret = true);
```
abtink: Suggest adding an inline comment to help clarify the boolean argument (helps the reader of code):
```suggestion
    SuccessOrExit(otDatasetIsValid(&aDatasetTlvs, /* aActive */ false), ret = true);
```

src/host/thread_helper.cpp
```diff
@@ -532,6 +544,47 @@ bool ThreadHelper::AreDatasetTlvsEqualToLocalDatasetTlvs(const otOperationalData
...
+    SuccessOrExit(otDatasetParseTlvs(&aDatasetTlvs, &dataset), ret = true);
+    SuccessOrExit(otDatasetGetActive(mInstance, &localDataset), ret = true);
```
abtink: It looks like we are setting `ret = true` on exit for all of the `SuccessOrExit()` and `VerifyOrExit()` statements. 

To simplify the code, how about initializing `ret = true` at the top of the method? Then, if all the checks pass, we can just update the value of `ret` at the very end.

src/host/thread_helper.cpp
```diff
@@ -532,6 +544,47 @@ bool ThreadHelper::AreDatasetTlvsEqualToLocalDatasetTlvs(const otOperationalData
...
+
```
abtink: Although it shouldn't be common, it's possible for `localDataset` to be only partially populated when read (e.g., missing `mChannel` or `mPskc`, where `mIsChannelPresent` is false).

Do we need to handle this edge case here, or can we safely assume that `CanImpactNetworkConnectivity()` won't be evaluated unless the device already has a valid active dataset?

abtink (commented):
Thanks @wtongze for the PR.

This reminds me of some existing behavior in the OpenThread core `DatasetManager` modules. In `DatasetManager::ProcessSetOrReplaceRequest()`, there is already similar logic that checks whether a new Dataset affects connectivity or the network key:
https://github.com/openthread/openthread/blob/29bb6f634a747aebe06fc47e8c3b0bcd9a3ec040/src/core/meshcop/dataset_manager_ftd.cpp#L86-L118

If it does affect connectivity, it is automatically converted to use the pending dataset:
https://github.com/openthread/openthread/blob/29bb6f634a747aebe06fc47e8c3b0bcd9a3ec040/src/core/meshcop/dataset_manager_ftd.cpp#L191-L213

This mechanism currently seems to be used when TMF commands are issued (not via public APIs such as `otDatasetSendMgmtActiveSet`. 

-----

I suggest considering whether to move this entire logic into OpenThread core instead of in `ot-br-posix` —perhaps by exposing it as a new API and reusing some of the existing code. 

Implementing this behavior in OT core makes it available across all platforms and easier to validate, rather than limiting it strictly to `ot-br-posix`.

Regarding the specific checks, I see you are checking the Channel, PAN ID, Extended PAN ID, PSKc, Channel Mask, Security Policy, and Mesh-Local Prefix. The existing OpenThread core code only checks a subset of these. We should harmonize these conditions and ensure they align with the Thread specification. For example, I am not certain if the Extended PAN ID or Channel Mask actually need to be included here.

----

Edit: I did check this a bit, Section "8.7.4. Updating the Active Operational Dataset" of Thread spec covers this and mentions the following TLVs:

> Changes to Active Operational Dataset parameters may either be done directly, in which case they take effect immediately, or may be done using the Pending Operational Dataset, in which case they take place after a delay. The following Active Operational Dataset values affect the ability for neighboring Thread Devices to communicate:
> - Channel
> - Mesh-Local Prefix
> - PAN ID
> - Network Key
> - Security Policy fields:
> - - R bit (when changed from 1 to 0)
> - - NCR bit (when changed from 0 to 1)
> - - VR field (when increasing the value VR)

src/host/thread_helper.cpp
wtongze: I think we can safely assume that `CanImpactNetworkConnectivity()` won't be evaluated unless the device already has a valid active dataset. The primary use case of this PR is to improve the user experience when the user decided to change the name of their existing Thread network.

wtongze:
This PR has been updated to use `otDatasetAffectsConnectivity` API added in https://github.com/openthread/openthread/pull/13134

abtink (approved):
👍

src/host/thread_helper.cpp
jwhui: Should this be `mAttachDelayMs = attachDelayMs`?

Once the stack starts and the device successfully connects, the role changes and StateChangedCallback is triggered. Because mAttachDelayMs is non-zero (300,000 ms), it calls otDatasetSendMgmtPendingSet with a 300-second delay instead of updating via otDatasetSendMgmtActiveSet immediately.

```suggestion
            mAttachDelayMs            = attachDelayMs;
```

wtongze: Good point. I've changed this part.

abtink:
Want to just link to this other PR (and comment) which is also changing `AttachAllNodesTo()`:

- https://github.com/openthread/ot-br-posix/pull/3396#discussion_r3351756037

nand-nor:
Maybe this is already handled elsewhere in the code, but if there are multiple devices updating the dataset fields that do not impact connectivity, for whatever reason, I assume conflicts where the same field is changed will be decided by whichever has the most recent timestamp? Is there any logic that handles if the timestamps are equal? And what if there are conflicts where different fields are changed? Just wondering how the dataset distribution behavior will allow for eventual consistency in the case where theres multiple nodes updating the dataset. Thats pretty straightforward with pending datasets but less so to me in this context

wtongze:
> Maybe this is already handled elsewhere in the code, but if there are multiple devices updating the dataset fields that do not impact connectivity, for whatever reason, I assume conflicts where the same field is changed will be decided by whichever has the most recent timestamp? Is there any logic that handles if the timestamps are equal? And what if there are conflicts where different fields are changed? Just wondering how the dataset distribution behavior will allow for eventual consistency in the case where theres multiple nodes updating the dataset. Thats pretty straightforward with pending datasets but less so to me in this context

AFAIK, a timestamp is associated with the whole operational dataset. The new dataset will be accepted if its timestamp is strictly greater than the local dataset. Eventually consistency is achieved by MLE (Mesh Link Establishment). This PR does not change the underlaying logic but only provides an optimization when applying the new dataset. It's based on Section "8.7.4. Updating the Active Operational Dataset" of Thread spec.

abtink:
> Maybe this is already handled elsewhere in the code, but if there are multiple devices updating the dataset fields that do not impact connectivity, for whatever reason, I assume conflicts where the same field is changed will be decided by whichever has the most recent timestamp? Is there any logic that handles if the timestamps are equal? And what if there are conflicts where different fields are changed? Just wondering how the dataset distribution behavior will allow for eventual consistency in the case where theres multiple nodes updating the dataset. Thats pretty straightforward with pending datasets but less so to me in this context


@nand-nor, here is some related info:

As @wtongze already mentioned, a Leader will reject any MGMT_ACTIVE_SET.req if the provided timestamp is not ahead of its current one.

If we somehow end up with two different Active Operational Datasets with the exact same timestamp value (which can happen if the network is partitioned and the partitions are configured separately), the network will not detect it and will remain in a stalled or bad state.
- If a device receives a dataset announcement with a timestamp equal to its own, it "does nothing." It will not pull or update the new data because the timestamp is not strictly newer.
- Recovering such a network requires manual intervention. A new MGMT_ACTIVE_SET.req with a timestamp that is strictly greater than the conflicting ones must be sent to force the mesh to update.

One thing to mention is that the timestamp includes a 15-bit "Ticks" value. It is recommended to use a random value, which practically eliminates the chance of an accidental timestamp collision.

src/host/thread_helper.cpp
```diff
@@ -108,9 +108,21 @@ void ThreadHelper::StateChangedCallback(otChangedFlags aFlags)
...
                     otOperationalDataset emptyDataset = {};
-                    otError              error =
-                        otDatasetSendMgmtPendingSet(mInstance, &emptyDataset, mAttachPendingDatasetTlvs.mTlvs,
-                                                    mAttachPendingDatasetTlvs.mLength, MgmtSetResponseHandler, this);
+                    otError              error        = OT_ERROR_NONE;
+
+                    if (mAttachDelayMs == 0)
+                    {
+                        error =
+                            otDatasetSendMgmtActiveSet(mInstance, &emptyDataset, mAttachPendingDatasetTlvs.mTlvs,
```
jwhui: Should we rename `mAttachPendingDatasetTlvs` to something like `mAttachDatasetTlvs`?

wtongze: Done.

jwhui (approved):
Thanks! 👍🏼

PR #3364, opened 2026-05-22
[sdp-proxy] fix heap underflow in SRP Advertising Proxy

jwhui:
This commit fixes a heap underflow in the SRP Advertising Proxy due to calling pop_back() on the empty mOutstandingUpdates vector.

When a synchronous mDNS error occurs during
PublishHostAndItsServices, the callback OnMdnsPublishResult erases the outstanding update from the vector synchronously. Since the handler was assuming the update was still at the back of the vector, it would call pop_back() on an empty vector.

This fix checks if the outstanding update exists by searching for it by its update ID before erasing it or notifying the result, which avoids the heap underflow and potential double-reporting.

### DIFF
--- a/src/sdp_proxy/advertising_proxy.cpp
+++ b/src/sdp_proxy/advertising_proxy.cpp
@@ -123,10 +123,17 @@ void AdvertisingProxy::AdvertisingHandler(otSrpServerServiceUpdateId aId,
 
     error = PublishHostAndItsServices(aHost, update);
 
-    if (error != OTBR_ERROR_NONE || update->mCallbackCount == 0)
+    for (auto it = mOutstandingUpdates.begin(); it != mOutstandingUpdates.end(); ++it)
     {
-        mOutstandingUpdates.pop_back();
-        otSrpServerHandleServiceUpdateResult(GetInstance(), aId, OtbrErrorToOtError(error));
+        if (it->mId == aId)
+        {
+            if (error != OTBR_ERROR_NONE || it->mCallbackCount == 0)
+            {
+                mOutstandingUpdates.erase(it);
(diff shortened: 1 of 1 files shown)


src/sdp_proxy/advertising_proxy.cpp
```diff
@@ -123,10 +123,17 @@ void AdvertisingProxy::AdvertisingHandler(otSrpServerServiceUpdateId aId,
...
-    if (error != OTBR_ERROR_NONE || update->mCallbackCount == 0)
+    for (auto it = mOutstandingUpdates.begin(); it != mOutstandingUpdates.end(); ++it)
     {
-        mOutstandingUpdates.pop_back();
-        otSrpServerHandleServiceUpdateResult(GetInstance(), aId, OtbrErrorToOtError(error));
+        if (it->mId == aId)
+        {
+            if (error != OTBR_ERROR_NONE || it->mCallbackCount == 0)
+            {
+                mOutstandingUpdates.erase(it);
```
abtink, line 132: I am just curious if this is a safe and valid pattern: Can we erase an entry from the collection as we iterate over it?

abtink: I found this refernce, I think `earse` can invalidate iterators?
https://en.cppreference.com/cpp/container/vector/erase

jwhui: I think this pattern is safe here because we immediately exit the loop via break right after the erase call.

abtink: I see. Thanks. I missed that. Yes, it should be fine!

abtink (approved):

PR #3394, opened 2026-06-03
[mdns] fix service loss on mDNSResponder daemon restart

yangsong-cnyn:
When using `mDNSResponder`, restarting the mdnsd daemon invalidates the active socket connections.

This PR resolves this by leveraging OT Core's native state machine (`otPlatDnssdStateHandleStateChange`). Upon connection loss, the platform triggers a standard stop-start state transition, prompting OT Core and platform observers (`Border Agent`, `TREL`) to re-advertise their services autonomously.

---

Key Changes:

      - Explicitly initialize `mIsInitialized`, `mMeshCoPUdpPort`, and `mBaIsActive` to `false` in `BorderAgent::ClearState()`, which prevents uninitialized memory values.
      - Use active `fd == -1` checks in `PublisherMDnsSd::Update`, `ServiceRef::Update`, and `ServiceRef::Process`. Defunct sockets now immediately trigger a reconnect.
      - Monitore all retryable connection errors (`IsRetryableError`) in `PublisherMDnsSd::Process`.
      - Add the missing `mStateCallback(State::kIdle)` invocation inside `PublisherMDnsSd::Stop`.


src/mdns/mdns_mdnssd.cpp
```diff
@@ -267,6 +267,39 @@ PublisherMDnsSd::~PublisherMDnsSd(void)
...
+        else
+        {
+            service->Resolve(kDNSServiceInterfaceIndexAny, service->mInstanceName, service->mType, kDomain);
+        }
+    }
+
+    for (const auto &host : mSubscribedHosts)
+    {
+        host->Resolve();
+    }
```
abtink, line 274: While this approach may work, I think there is a cleaner, more idiomatic way to handle this recovery process by leveraging existing OpenThread core capabilities.

Here is my proposal:

From the OpenThread core perspective, we already have the `otPlatDnssdStateHandleStateChange()` callback designed for the platform layer to signal when its platform-provided DNS-SD state changes.

```c++
/**
 * Callback to notify state changes of the DNS-SD platform.
 *
 * The OpenThread stack will call `otPlatDnssdGetState()` (from this callback or later) to get the new state. The
 * platform MUST therefore ensure that the returned state from `otPlatDnssdGetState()` is updated before calling this.
 *
 * When the platform signals a state change to `OT_PLAT_DNSSD_STOPPED` using this callback, all active browsers and
 * resolvers are considered to be stopped, and any previously registered host, service, key entries as removed.
 *
 * @param[in] aInstance The OpenThread instance structure.
 */
extern void otPlatDnssdStateHandleStateChange(otInstance *aInstance);

```

There are two states defined for this: `STOPPED` and `READY`.

```c++
/**
 * Represents the state of the DNS-SD platform.
 */
typedef enum otPlatDnssdState
{
    OT_PLAT_DNSSD_STOPPED, ///< Stopped and unable to register any service or host, or start any browser/resolver.
    OT_PLAT_DNSSD_READY,   ///< Running and ready to register service or host.
} otPlatDnssdState;

```

If we instead signal `OT_PLAT_DNSSD_STOPPED` when the connection is lost, and then signal `OT_PLAT_DNSSD_READY` once the platform layer reconnects, the OpenThread core modules that use the DNS-SD (mDNS) will automatically handle the recovery for us. The core is already built to automatically re-register all information (services, keys, hosts, etc.) and restart any active browsers or resolvers upon transitioning back to `READY`.

Using this existing framework seems like a better architecture because it relies on a single, unified recovery/restore path in the core rather than introducing a new, platform-side registry and reconnection mechanism here.

yangsong-cnyn: Great suggestion. Compare to tracking and re-registering entries within the platform layer, it's definitely better to use the core-driven recovery.

I have made the updates, and also fixed an uninitialized member variables issue.

abtink (approved):
👍

src/mdns/mdns_mdnssd.cpp
```diff
@@ -351,7 +358,15 @@ void PublisherMDnsSd::Update(MainloopContext &aMainloop)
     {
         int fd = DNSServiceRefSockFD(mHostsRef);
 
-        assert(fd != -1);
+        if (fd == -1)
+        {
+            otbrLogWarning("mDNS connection socket is invalid, reconnecting...");
+            mTaskRunner.Post(Milliseconds(0), [this]() {
+                Stop(kStopOnServiceNotRunningError);
+                Start();
```
jwhui: Posting `Stop()` and `Start()` directly on every failed socket will queue multiple reconnect tasks if multiple sockets fail at once (e.g., across `mHostsRef`, `mSubscribedServices`, and `mSubscribedHosts`), causing back-to-back `STOPPED -> READY -> STOPPED -> READY` state transitions that interrupt OpenThread Core's re-registration.

Should we centralize the reconnect logic into a helper method that guards against multiple queued tasks?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -428,11 +446,13 @@ void PublisherMDnsSd::Process(const MainloopContext &aMainloop)
             otbrLog(logLevel, OTBR_LOG_TAG, "DNSServiceProcessResult failed: %s (serviceRef = %p)",
                     DNSErrorToString(error), serviceRef);
         }
-        if (error == kDNSServiceErr_ServiceNotRunning)
+        if (IsRetryableError(error))
         {
-            otbrLogWarning("Need to reconnect to mdnsd");
-            Stop(kStopOnServiceNotRunningError);
-            Start();
+            otbrLogWarning("mDNS error %d, need to reconnect to mdnsd", error);
```
jwhui: Using `DNSErrorToString(error)` will provide better diagnostics than raw error codes.

src/mdns/mdns_mdnssd.cpp
jwhui: `DnssdServiceRegistration::Update` and `DnssdServiceRegistration::Process` currently use `VerifyOrExit(fd != -1)` which silently ignores defunct sockets. If an agent has published services without host registrations or subscriptions, a daemon restart will leave these sockets at `fd == -1` without triggering reconnection. 

Should we update `DnssdServiceRegistration::Update` and `DnssdServiceRegistration::Process` to trigger `GetPublisher().ScheduleReconnect()` when `fd == -1`?

src/mdns/mdns_mdnssd.cpp
yangsong-cnyn: Agree with it. I am going to introduce a member var `mReconnecting` to avoid duplicate tasks

src/border_agent/border_agent.cpp
```diff
@@ -196,6 +196,9 @@ void BorderAgent::ClearState(void)
 
 #if OTBR_ENABLE_BORDER_AGENT_MESHCOP_SERVICE
     mServiceInstanceName.clear();
+    mIsInitialized  = false;
+    mMeshCoPUdpPort = 0;
+    mBaIsActive     = false;
```
jwhui, line 201: Should also clear `mExtAddress`?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -336,6 +344,22 @@ void PublisherMDnsSd::DeallocateHostsRef(void)
...
 
+void PublisherMDnsSd::TriggerReconnect(void)
+{
+    VerifyOrExit(mState == State::kReady);
+    VerifyOrExit(!mReconnecting);
+
+    mReconnecting = true;
+    mTaskRunner.Post(Milliseconds(0), [this]() {
+        mReconnecting = false;
+        Stop(kStopOnServiceNotRunningError);
```
jwhui: If `PublisherMDnsSd::Stop()` is called while a reconnect task is pending in `mTaskRunner`, the lambda would still execute `Start()`, reviving a stopped publisher. We should check `VerifyOrExit(mState == State::kReady)` inside the lambda.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -351,7 +375,12 @@ void PublisherMDnsSd::Update(MainloopContext &aMainloop)
     {
         int fd = DNSServiceRefSockFD(mHostsRef);
 
-        assert(fd != -1);
+        if (fd == -1)
+        {
+            otbrLogWarning("mDNS connection socket is invalid, reconnecting...");
```
jwhui: Moving `otbrLogWarning` inside `TriggerReconnect` (after `VerifyOrExit(!mReconnecting)`) avoids logging the warning multiple times in a single loop when all active sockets fail at once upon daemon restart.

Similar instances below.

jwhui:
@yangsong-cnyn , can you help resolve the GitHub Actions failures?

yangsong-cnyn:
> @yangsong-cnyn , can you help resolve the GitHub Actions failures?

@jwhui All checks have passed!

src/mdns/mdns_mdnssd.cpp
```diff
@@ -336,6 +344,27 @@ void PublisherMDnsSd::DeallocateHostsRef(void)
...
+    otbrLogWarning("mDNS connection socket is invalid (%s), reconnecting...", aReason);
+
+    mReconnecting = true;
+    mTaskRunner.Post(Milliseconds(0), [this]() {
+        if (mState == State::kReady)
+        {
+            mReconnecting = false;
+            Stop(kStopOnServiceNotRunningError);
+            Start();
+        }
```
jwhui, line 366: If `PublisherMDnsSd::Stop()` is called while a reconnect task is queued in `mTaskRunner`, `mState` becomes `kIdle`. When the task executes, the `if (mState == State::kReady)` condition evaluates to `false`, which means `mReconnecting = false;` is never executed.

Because `mReconnecting` stays `true`, if the publisher is subsequently restarted with `Start()`, any future calls to `TriggerReconnect()` will bail out at `VerifyOrExit(!mReconnecting)`, permanently preventing automatic reconnection.

To fix this, please reset `mReconnecting = false;` unconditionally inside the task, and also reset `mReconnecting = false;` in `PublisherMDnsSd::Stop()`:

```suggestion
        mReconnecting = false;
        if (mState == State::kReady)
        {
            Stop(kStopOnServiceNotRunningError);
            Start();
        }
```

src/mdns/mdns_mdnssd.cpp
```diff
@@ -351,7 +380,11 @@ void PublisherMDnsSd::Update(MainloopContext &aMainloop)
     {
         int fd = DNSServiceRefSockFD(mHostsRef);
```
jwhui, line 386: `mHostsRef` is missing the `if (fd == -1)` check here in `PublisherMDnsSd::Process`. If `mHostsRef` socket is invalid (`fd == -1`), passing `-1` to `FD_ISSET` causes undefined behavior (out-of-bounds bit array access).

src/border_agent/border_agent.cpp
```diff
@@ -196,6 +196,10 @@ void BorderAgent::ClearState(void)
...
+    memset(&mExtAddress, 0, sizeof(mExtAddress));
```
jwhui: Also clear `mOtTxtData` by calling `mOtTxtData.clear()`?

src/mdns/mdns_mdnssd.cpp
```diff
@@ -336,6 +350,26 @@ void PublisherMDnsSd::DeallocateHostsRef(void)
     return;
 }
 
+void PublisherMDnsSd::TriggerReconnect(const char *aReason)
+{
+    VerifyOrExit(mState == State::kReady);
+    VerifyOrExit(mReconnectTaskId == 0);
+
+    otbrLogWarning("mDNS connection socket is invalid (%s), reconnecting...", aReason);
```
jwhui: Minor phrasing suggestion: when `aReason` is `"service socket invalid"` or a daemon error string like `"Service Not Running"`, `"mDNS connection socket is invalid (%s)..."` leads to slightly repetitive or mismatched wording (e.g. `mDNS connection socket is invalid (service socket invalid), reconnecting...`).
```suggestion
    otbrLogWarning("mDNS reconnect triggered (%s)...", aReason);
```

jwhui (approved):
Thanks! 👍🏼

PR #3396, opened 2026-06-03
[dbus] Add support for setting pending dataset delay timer for AttachAllNodesTo

nand-nor:
Adds support for extending the input parameters to the `AttachAllNodesTo` dbus API to allow the caller to set the pending dataset delay timer. This is intended for supporting pending dataset changes that involve a network key change on networks with ICD LIT devices (or other sleepy device types) that have idle mode durations that are longer than the common default pending dataset timeout (5 minutes). Alternatively, callers of this API can also elect to supply a value that is shorter than the compile-time default. The addition of this parameter leaves the decision up to the caller, which can be useful to be able to vary depending on the dataset change and the current idle mode duration times of the devices on the network.

### DIFF
--- a/src/dbus/client/thread_api_dbus.hpp
+++ b/src/dbus/client/thread_api_dbus.hpp
@@ -174,6 +174,8 @@ public:
      *                       automatically.
      * @param[in] aDelayMs   The delay timer in milliseconds to embed in the pending dataset.
      *                       Use 0 (default) for 300 seconds (5 minutes). Maximum is 72 hours.
+     *                       When 0, the delay_timer_ms argument is omitted from the DBus call for
+     *                       backward compatibility with older servers.
      *
      * @retval ERROR_NONE              Successfully requested the Thread network migration.
      * @retval ERROR_DBUS              D-Bus encode/decode error.
--- a/src/dbus/client/thread_api_dbus.hpp
+++ b/src/dbus/client/thread_api_dbus.hpp
@@ -169,8 +169,11 @@ public:
      *
      * If the device has already attached to a network, send a request to migrate the existing network.
      *
-     * @param[in] aDataset  The Operational Dataset that contains parameter values of the Thread network to attach
-     *                      to. It must have a valid Delay Timer and Pending Timestamp.
+     * @param[in] aDataset   The Operational Dataset that contains parameter values of the Thread network to attach
+     *                       to. It must be a full dataset. The Delay Timer and Pending Timestamp will be added
+     *                       automatically.
+     * @param[in] aDelayMs   The delay timer in milliseconds to embed in the pending dataset.
+     *                       Use 0 (default) for 300 seconds (5 minutes). Maximum is 72 hours.
      *
      * @retval ERROR_NONE              Successfully requested the Thread network migration.
      * @retval ERROR_DBUS              D-Bus encode/decode error.
@@ -180,7 +183,7 @@ public:
      * @retval OT_ERROR_INVALID_ARGS   Arguments are invalid.
      * @retval OT_ERROR_BUSY           There is an ongoing request.
      */
-    ClientError AttachAllNodesTo(const std::vector<uint8_t> &aDataset);
+    ClientError AttachAllNodesTo(const std::vector<uint8_t> &aDataset, uint32_t aDelayMs = 0);
 
     /**
      * This method performs a factory reset.
--- a/src/host/thread_helper.cpp
+++ b/src/host/thread_helper.cpp
@@ -560,9 +560,11 @@ exit:
     return error;
 }
 
-void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, AttachHandler aHandler)
+void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, uint32_t aDelayMs, AttachHandler aHandler)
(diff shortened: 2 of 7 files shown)


nand-nor:
@abtink @bukepo do you think the approach to this PR, modifying the existing AttachAllNodesTo API, is acceptable? Or do you prefer to add a new dbus API to all for setting the pending dataset delay timer? This option seemed the simplest and is a smaller change set but I can modify it to just add a whole new API if thats the preference.

src/dbus/client/thread_api_dbus.hpp
```diff
@@ -180,7 +185,7 @@ class ThreadApiDBus
      * @retval OT_ERROR_INVALID_ARGS   Arguments are invalid.
      * @retval OT_ERROR_BUSY           There is an ongoing request.
      */
-    ClientError AttachAllNodesTo(const std::vector<uint8_t> &aDataset);
+    ClientError AttachAllNodesTo(const std::vector<uint8_t> &aDataset, uint32_t aDelayMs = 0);
```
jdswensen, line 188: Is it worth creating a new dbus method for this? Something like `AttachAllNodesToWithDelay`. That way, the public dbus API doesn't change for `AttachAllNodesTo` and can remain the same for backwards compatibility. Internally, both functions can share the same logic, but the path that goes through `AttachAllNodesTo` will just use the default value instead of setting one.

jdswensen: Oh, whoops. I missed your question above. 😄 My preference would be for the second function.

nand-nor: I like the option of extending the existing one because I think it forces callers to be more intentional around what they are setting for this value, which I think is important and will become even more important in the future. Especially as more ICD LIT devices are introduced to the market, and users are integrating them and maybe dynamically setting them to act as SIT versus LIT, this allows callers to be dynamic and intentional. That being said I will defer to whatever the maintainers prefer

src/host/thread_helper.cpp
```diff
@@ -550,9 +550,11 @@ otError ThreadHelper::StartThreadStack(void)
     return error;
 }
 
-void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, AttachHandler aHandler)
+void ThreadHelper::AttachAllNodesTo(const std::vector<uint8_t> &aDatasetTlvs, uint32_t aDelayMs, AttachHandler aHandler)
```
abtink, line 563: > do you think the approach to this PR, modifying the existing AttachAllNodesTo API, is acceptable? Or do you prefer to add a new dbus API to all for setting the pending dataset delay timer? This option seemed the simplest and is a smaller change set but I can modify it to just add a whole new API if thats the preference.

@nand-nor , I don’t have a strong preference. Extending the existing API (in backward compatible way) sounds reasonable to me. 

I wanted to bring up one point on this:

There is a recent, unmerged PR (#3337) that also modifies `ThreadHelper::AttachAllNodesTo()`. It changes the method to avoid using the `otDatasetSendMgmtPendingSet` command, opting instead for `otDatasetSendMgmtActiveSet` when the updated dataset does not impact network connectivity. The main goal there is to speed up the dataset distribution in those specific cases.

We should consider how these two PRs interact.

For example, what should happen if a DBus caller explicitly provides a delay value alongside a Dataset that *doesn't* impact connectivity? Should the implementation still force a `PendingSet` to honor the requested delay, or should it ignore the delay value and use `ActiveSet`?

ccing @wtongze

nand-nor: My first impression is that it really only makes sense to specify a longer timeout if/when connectivity is impacted, and Im not sure it makes sense to do so in other cases, so its not something the code should allowed. I like & agree with the intent of the other PR to speed up operations, so if someone wants to put a pending delay of 1 hour for setting the thread network name, maybe they should just get a log warning that it wont be honored, and the logic should drop the delay as you suggest. Also open to discussing this in the TC if needed but in the meantime I will take a deeper look to see if there are other potential conflicts with respect to how these two PRs will work together.

bukepo (commented):
I'm wondering whether we should just make AttachAllNodesTo() to accept active dataset or pending dataset. The latter dataset includes the delay timer which could meet the need here.

I know it may sound weird, but in fact, even if the input is an active dataset, it may be converted into pending dataset. So I think it make sense to allow specify a pending dataset directly here.

wtongze (commented):
I agree that we should find a way to make these two PRs work together. Ideally, I think we should try to speed up the dataset distribution operations whenever it's possible. Issuing a log warning when optimization will be applied instead of honoring manually set delay also sounds good to me.

nand-nor:
So if I am understanding correctly it sounds like we do want to keep this PR. For next steps I can rebase and make PR #3337 the merge target, and then modify HEAD (or add a new commit) to add logic to drop the request pending dataset timer delay, if there is one, & emit a log warning whenever the dataset update will be via MGMT_ACTIVE_SET/whenever optimization will be applied. Then this PR can either merge into #3337 before it goes into main, or the merge target will update to main if/when #3337 gets in first. But this will make the needed ordering or merges more clear.

Or if I am misunderstanding the direction folks want to go, or there is another preferred way to handle this just let me know.

nand-nor:
Since #3337 merged I just rebased and updated HEAD, apologies for the force push

PR #3403, opened 2026-06-08
docker: tag border-router images with CalVer release tags

fmurodov:
The workflow was only tagging images with branch names and sha-* short hashes, so Docker Hub never reflected the monthly CalVer releases (e.g. v2026.06.0) created by monthly-release.yml.

Two changes:
- Trigger the workflow on v* tag pushes (was branch-only)
- Add type=ref,event=tag to the metadata-action tags list in the merge job

### DIFF
--- a/.github/workflows/docker-border-router.yml
+++ b/.github/workflows/docker-border-router.yml
@@ -32,6 +32,8 @@ on:
   push:
     branches-ignore:
       - 'dependabot/**'
+    tags:
+      - 'v*'
   pull_request:
     branches:
       - 'main'
@@ -158,6 +160,7 @@ jobs:
             ${{ env.DOCKERHUB_REPO }}
           tags: |
             type=ref,event=branch
+            type=ref,event=tag
             type=raw,value=latest,enable={{is_default_branch}}
             type=sha,prefix=sha-,format=short
 


.github/workflows/docker-border-router.yml
```diff
@@ -32,6 +32,8 @@ on:
   push:
     branches-ignore:
       - 'dependabot/**'
+    tags:
+      - 'v*'
```
jwhui, line 36: Thanks for submitting this PR.

I don't think this will work as intended. As a built-in security feature to prevent infinite recursive execution loops, GitHub Actions explicitly prevents events triggered by `GITHUB_TOKEN` from triggering subsequent workflows.

As an alternative, I think we can update `monthly-release.yml` to use an api-dispatch workflow run.

Would you like to update this PR? I'm also happy to post a PR if you prefer.

jwhui: I posted a PR here: https://github.com/openthread/ot-br-posix/pull/3436

fmurodov: Sorry. Forgot to update it. But thanks. I'll close this PR then

jwhui:
Superseded by https://github.com/openthread/ot-br-posix/pull/3436

PR #3404, opened 2026-06-08
[tests] add SRP service lease integration test (1_3_SRP_TC_3)

jwhui:
Add integration test for Thread 1.3 SRP Service Instance Lease specification "2.3. [1.3] [CERT] Service Instance Lease – Renewal and Automatic Service/Host Removal" (1_3_SRP_TC_3.md).

This test validates:
- Registration of a service and host with short lease (30s) and longer key lease (600s).
- Querying host/service via unicast DNS-SD on the Thread network.
- Querying host/service via mDNS on the infrastructure link.
- Renewal of the lease by restarting the SRP client before expiry.
- Automatic removal/cleanup of the registered host and service records from the SRP server once the lease expires.

The test utilizes the docker-in-docker framework and is executed by tests/scripts/test_dind_dns_sd.sh.

### DIFF
--- a/tests/scripts/expect/dind_srp_tc_3.exp
+++ b/tests/scripts/expect/dind_srp_tc_3.exp
@@ -1,0 +1,437 @@
+#!/usr/bin/expect -f
+#
+#  Copyright (c) 2026, The OpenThread Authors.
+#  All rights reserved.
+#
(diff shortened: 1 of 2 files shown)


abtink (approved):
👍

PR #3416, opened 2026-06-12
[cmake] enable OT core advertising proxy by default for all mdns

jwhui:
This commit updates the default configuration options to enable OpenThread core's Advertising Proxy (OTBR_OT_SRP_ADV_PROXY) by default whenever any mDNS provider (e.g. mDNSResponder) is configured.

Previously, this default option was only enabled when OTBR_MDNS was set to "openthread". For other mDNS providers, it would fall back to the legacy OTBR-level Advertising Proxy, which does not support publishing SRP KEY records. This resulted in failures during integration tests (like dind_srp_tc_13) when verifying KEY records under mDNSResponder.

Additionally, this commit removes the redundant build matrix job "Border Router with OT Core Advertising Proxy (mDNSResponder)" from the GHA workflow since this is now tested as part of the default "mDNSResponder" workflow.


abtink (approved):
👍

Issue #3382, opened 2026-05-28
[RFC] Unify REST and WebSocket Services in `otbr-agent` via uWebSockets (Replacing `cpp-httplib`)


jwhui:
## 1. Background & Problem Statement

The current OpenThread Border Router (OTBR) architecture serves web services through two distinct layers:
1.  **`otbr-web`**: A standalone daemon serving the legacy AngularJS frontend and translating HTTP requests to CLI commands over a Unix socket.
2.  **`otbr-agent` REST Server**: An embedded REST engine in `otbr-agent` built using `cpp-httplib`.

While `cpp-httplib` has served the REST API well, it introduces two primary architectural challenges:

1.  **Threading Model Mismatch**: `cpp-httplib` uses a synchronous, multi-threaded thread pool. Because the core OpenThread stack is **single-threaded and not thread-safe**, every REST endpoint handler must marshal its queries back to the primary event loop using `RunInMainLoop`. This introduces context-switching overhead, lock contention, and complicates the handler codebase.
2.  **Lack of WebSocket Support**: To support modern, real-time, low-latency user interfaces (such as interactive topology graph renderers and live DTLS commissioning logs), we need push-delivery capabilities. `cpp-httplib` does not natively support WebSockets or Server-Sent Events (SSE).

---

## 2. Proposed Architecture: Unified Single-Threaded API Gateway

I propose replacing `cpp-httplib` with **uWebSockets (uWS)**, a highly optimized, asynchronous, single-threaded HTTP and WebSocket library.

Under this proposal, the new uWS gateway will run **directly inside the primary event loop of `otbr-agent`** by integrating its loop file descriptor into `MainloopContext` (`select`/`poll`).

```
+----------------------------------------------------------------------------+
|                                Browser UI                                  |
+----------------------------------------------------------------------------+
                   |                                   ^
      REST API (HTTP / JSON)                WebSocket Stream (/ws/topology)
                   v                                   |
+----------------------------------------------------------------------------+
|                     otbr-agent (Main Event Loop Thread)                    |
|                                                                            |
|   +--------------------------------------------------------------------+   |
|   |                       uWebSockets Gateway                          |   |
|   +--------------------------------------------------------------------+   |
|                                    |                                       |
|                         Direct C++ Synch Calls                             |
|                                    v                                       |
|   +--------------------------------------------------------------------+   |
|   |                       OpenThread Core Stack                        |   |
|   +--------------------------------------------------------------------+   |
[...]

PR #3432, opened 2026-06-21
[web] support HTTPS reverse proxy for API calls

LJspice:
I'm running the web UI behind my existing reverse proxy and authentication middleware, which is configured to properly pass both the web UI and API over HTTPS and port 443. However, the topology tab fails to load due to mixed content errors, as HTTP is hardcoded, as is port 8081 for API calls.

My reference Traefik config, in case these breadcrumbs help someone else in a similar situation:
```yaml
services:
  otbr:
    labels:
      traefik.enable: "true"
      # Web UI (Port 8080)
      traefik.http.routers.otbr-web.rule: "Host(`otbr.${DOMAIN_NAME}`)"
      traefik.http.routers.otbr-web.middlewares: "authelia@file"
      traefik.http.routers.otbr-web.service: "otbr-web"
      traefik.http.services.otbr-web.loadbalancer.server.url: "http://host.docker.internal:8080"
      # REST API (port 8081)
      traefik.http.routers.otbr-api.rule: "Host(`otbr.${DOMAIN_NAME}`) && PathPrefix(`/api`)"
      traefik.http.routers.otbr-api.middlewares: "authelia@file, otbr-api-strip"
      traefik.http.routers.otbr-api.service: "otbr-api"
      traefik.http.services.otbr-api.loadbalancer.server.url: "http://host.docker.internal:8081"
      # Strip headers as OTBR returns 400 with too many headers or invalid characters in headers
      traefik.http.middlewares.otbr-api-strip.headers.customRequestHeaders.Cookie: ""
      traefik.http.middlewares.otbr-api-strip.headers.customRequestHeaders.Remote-User: ""
      traefik.http.middlewares.otbr-api-strip.headers.customRequestHeaders.Remote-Groups: ""
      traefik.http.middlewares.otbr-api-strip.headers.customRequestHeaders.Remote-Email: ""
      traefik.http.middlewares.otbr-api-strip.headers.customRequestHeaders.Remote-Name: ""
```

---

This commit updates the Web UI to support being served over HTTPS via a reverse proxy.

- Replaces hardcoded 'http://' with 'window.location.protocol' for all REST API calls.
- Modifies formatRestAddr() to conditionally omit the '8081' port when the page is accessed via HTTPS, allowing the proxy's port mapping to handle the request.

These changes ensure the Web UI remains functional when TLS is terminated at a reverse proxy, while maintaining backward compatibility for standard HTTP deployments.

Test results:
image

### DIFF
--- a/src/web/web-service/frontend/res/js/app.js
+++ b/src/web/web-service/frontend/res/js/app.js
@@ -287,7 +287,7 @@
                     console.log(response);
                     $scope.res = response.data.result;
                     if (response.data.result == 'successful') {
-                        var image = "http://api.qrserver.com/v1/create-qr-code/?color=000000&amp;bgcolor=FFFFFF&amp;data=v%3D1%26%26eui%3D" + response.data.eui64 +"%26%26cc%3D" + $scope.thread.pskd +"&amp;qzone=1&amp;margin=0&amp;size=400x400&amp;ecc=L";
+                        var image = "https://api.qrserver.com/v1/create-qr-code/?color=000000&amp;bgcolor=FFFFFF&amp;data=v%3D1%26%26eui%3D" + response.data.eui64 +"%26%26cc%3D" + $scope.thread.pskd +"&amp;qzone=1&amp;margin=0&amp;size=400x400&amp;ecc=L";
                         $scope.showQRCode(event, image);
                     } else {
                         $scope.showQRAlert(event, "sorry, can not generate the QR code.");
@@ -440,6 +440,10 @@
 
         $scope.restServerPort = '8081';
         var formatRestAddr = function(host, port) {
+            // When using HTTPS, omit port (assumes client sees 443 and reverse proxy routes to correct backend port)
+            if (window.location.protocol === 'https:') {
+                return (host.indexOf(':') > -1) ? '[' + host + ']' : host;
+            }
             return (host.indexOf(':') > -1) ? '[' + host + ']:' + port : host + ':' + port;
         };
         $scope.ipAddr = formatRestAddr(window.location.hostname, $scope.restServerPort);
@@ -505,7 +509,7 @@
         $scope.dataInit = async function() {
             let response;
         
-            $http.get('http://' + $scope.ipAddr + '/api/node', {
+            $http.get(window.location.protocol + '//' + $scope.ipAddr + '/api/node', {
                     headers: {
                         'Accept': 'application/json'
                     }
@@ -609,7 +613,7 @@
         // GET request to check action status
         $scope.getActionStatus = async function(action_id) {
 
-            const response = await $http.get('http://' + $scope.ipAddr + '/api/actions/' + action_id, {
+            const response = await $http.get(window.location.protocol + '//' + $scope.ipAddr + '/api/actions/' + action_id, {
                 headers: {
                     'Accept': 'application/vnd.api+json'
                 }
@@ -628,7 +632,7 @@
             console.log("discover network ...");  // Debugging log message
             do {
                 // start discovery task
-                const postResponse = await $http.post('http://' + $scope.ipAddr + '/api/actions', $scope.createRequestBodyUpdateDeviceCollection(deviceCount), {
+                const postResponse = await $http.post(window.location.protocol + '//' + $scope.ipAddr + '/api/actions', $scope.createRequestBodyUpdateDeviceCollection(deviceCount), {
                     headers: {
                         'Content-Type': 'application/vnd.api+json',
                         'Accept': 'application/vnd.api+json'
@@ -670,7 +674,7 @@
 
         // GET device collection
         $scope.fetchDevices = async function () {
-            const response = await $http.get('http://' + $scope.ipAddr + '/api/devices', {
+            const response = await $http.get(window.location.protocol + '//' + $scope.ipAddr + '/api/devices', {
                 headers: {
                     'Accept': 'application/json'
                 }
@@ -689,7 +693,7 @@
 
             do {
                 // Fetch device diagnostics
-                const postResponse = await $http.post('http://' + $scope.ipAddr + '/api/actions', $scope.createRequestBody(device_id), {
+                const postResponse = await $http.post(window.location.protocol + '//' + $scope.ipAddr + '/api/actions', $scope.createRequestBody(device_id), {
                     headers: {
                         'Content-Type': 'application/vnd.api+json',
                         'Accept': 'application/vnd.api+json'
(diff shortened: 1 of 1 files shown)


src/web/web-service/frontend/res/js/app.js
```diff
@@ -440,6 +440,10 @@
 
         $scope.restServerPort = '8081';
         var formatRestAddr = function(host, port) {
+            // When using HTTPS, omit port (assumes client sees 443 and reverse proxy routes to correct backend port)
+            if (window.location.protocol === 'https:') {
+                return (host.indexOf(':') > -1) ? '[' + host + ']' : host;
```
jwhui: If host (which can be assigned from window.location.hostname) is an IPv6 literal, it may already contain square brackets (e.g., `[2001:db8::1]`) in modern browsers. The check `host.indexOf(':') > -1` will still evaluate to `true`, resulting in duplicate brackets like `[[2001:db8::1]]` which is an invalid host format.

We should normalize the host by stripping brackets first before checking for IPv6 and formatting.

src/web/web-service/frontend/res/js/app.js
```diff
@@ -440,6 +440,10 @@
...
+            }
             return (host.indexOf(':') > -1) ? '[' + host + ']:' + port : host + ':' + port;
```
jwhui: This line has the same IPv6 bracket duplication issue. If host already contains brackets, it will return `[[IPv6]]:port`.

Normalize the host here as well.

src/web/web-service/frontend/res/js/app.js
```diff
@@ -440,6 +440,10 @@
 
         $scope.restServerPort = '8081';
         var formatRestAddr = function(host, port) {
+            // When using HTTPS, omit port (assumes client sees 443 and reverse proxy routes to correct backend port)
+            if (window.location.protocol === 'https:') {
```
jwhui: Unconditionally omitting the port when using HTTPS will break setups where the Web UI (and reverse proxy) are hosted on a non-standard port (e.g., `https://otbr.example.com:8443/`). In such cases, omitting the port causes the API requests to default to port 443, resulting in connection failure.

If the target `host` matches the Web UI host (`window.location.hostname`), we should use the Web UI's port. We can do this by using `window.location.host` (which includes the port if non-standard).

LJspice:
Checking in - I believe I've addressed all feedback

src/web/web-service/frontend/res/js/app.js
```diff
@@ -440,7 +440,22 @@
...
+
+            // When using HTTPS, assume reverse proxy places web UI and API on same port
+            if (window.location.protocol === 'https:') {
+                // If the target API host is the same as the Web UI host
+                if (normalizedHost === window.location.hostname.replace(/^\[(.*)\]$/, '$1')) {
+                    // Use the web UI full host string (including non-standard port if present)
+                    return window.location.host;
+                }
+                // Otherwise, assume the proxy handles the port translation for the external host
+                return formattedHost;
```
jwhui: When `window.location.protocol === 'https:'` and `normalizedHost` differs from the Web UI host (`window.location.hostname`), `formatRestAddr` returns `formattedHost` without the `port` parameter.

If `get_rest_api_info` returns a remote external host on a non-standard port, omitting `:port` forces requests to default port 443 (`https://<remote-host>/api/...`), which will fail if the remote host does not translate port 443 to the REST API service.

Consider preserving `port` for non-matching external hosts:

```suggestion
                // Otherwise, preserve the configured port for external hosts
                return formattedHost + ':' + port;
```

src/web/web-service/frontend/res/js/app.js
```diff
@@ -440,7 +440,22 @@
...
+            if (window.location.protocol === 'https:') {
+                // If the target API host is the same as the Web UI host
+                if (normalizedHost === window.location.hostname.replace(/^\[(.*)\]$/, '$1')) {
+                    // Use the web UI full host string (including non-standard port if present)
+                    return window.location.host;
+                }
+                // Otherwise, preserve the configured port for external hosts
+                return formattedHost + ':' + port;
+            }
+            return formattedHost + ':' + port;
```
jwhui: Thanks for addressing the feedback to preserve `:port` for external target hosts!
Since both the fallback inside the `https:` block (line 453) and the outer function return (line 455) execute `return formattedHost + ':' + port;`, we can simplify `formatRestAddr` by combining the conditions:
```javascript
            // When using HTTPS and the target API host matches the Web UI host, use the Web UI host (including port if present)
            if (window.location.protocol === 'https:' && normalizedHost === window.location.hostname.replace(/^$(.*)$$/, '$1')) {
                return window.location.host;
            }
            return formattedHost + ':' + port;

Comment (LJspice) [reply src/web/web-service/frontend/res/js/app.js]: Ah good point - resolved & rebased on main. I did end up keeping the original regex, so it would strip brackets before comparing to the bracket-stripped `normalizedHost`

File: src/web/web-service/frontend/res/js/app.js [hunk 6]
```diff
@@ -440,7 +440,16 @@
...
         $scope.restServerPort = '8081';
         var formatRestAddr = function(host, port) {
-            return (host.indexOf(':') > -1) ? '[' + host + ']:' + port : host + ':' + port;
+            // Remove existing IPv6 brackets if present
+            var normalizedHost = host.replace(/^\[(.*)\]$/, '$1');
+            
+            var formattedHost = (normalizedHost.indexOf(':') > -1) ? '[' + normalizedHost + ']' : normalizedHost;
+
+            // When using HTTPS and the target API host matches the Web UI host, use the web UI host (including port if present)
+            if (window.location.protocol === 'https:' && normalizedHost === window.location.hostname.replace(/^\[(.*)\]$/, '$1')) {
```
jwhui, line 449: Relying on `window.location.protocol === 'https:'` directly addresses the HTTPS mixed-content requirement. Note that if a user deploys a reverse proxy over plain HTTP (e.g. port 80 routing `/` to Web UI and `/api` to REST API), it will fall back to `:8081`.

In a future refactor, migrating REST API calls to relative paths (`/api/...`) would allow single-origin reverse proxies to work identically over both HTTP and HTTPS.

LJspice: Thoughts on how to address it in the scope of this PR? `https` is a definite signal that a reverse proxy is in use, but I'm not clear what signal could be used to reliably indicate reverse proxy with `http` without breaking non-proxied deployments.

I see #3382 would obsolete the changes in this PR, as well as a bunch of existing logic in `app.js`. Is it best tracked by a comment on that issue, and merging this PR in the meantime?

jwhui: I agree. In the current architecture with separate daemons (`otbr-web` on 8080 and `otbr-agent` on 8081), `https:` serves as an unambiguous signal that a reverse proxy is present. Detecting an HTTP reverse proxy client-side without extra backend config or probing risks breaking standalone HTTP deployments.

Given that #3382 will unify the web services and transition to relative paths (`/api/...`), keeping this scoped to HTTPS in this PR is the right approach.

jwhui (approved):
Thanks! 👍🏼

PR #3449, opened 2026-06-28
[web] generate QR codes on webUI client instead of remote API

LJspice:
Adding one client-executed QR-code generator script to replace the `api.qrserver.com` call. It didn't seem great or necessary to send `thread.pskd` out in the existing implementation.

## Demo
with change

### DIFF
--- a/src/web/web-service/frontend/.gitignore
+++ b/src/web/web-service/frontend/.gitignore
@@ -1,2 +1,0 @@
-# Dependency directories
-node_modules/
--- a/src/web/web-service/frontend/.gitignore
+++ b/src/web/web-service/frontend/.gitignore
@@ -1,0 +1,2 @@
+# Dependency directories
+node_modules/
--- a/src/web/web-service/frontend/CMakeLists.txt
+++ b/src/web/web-service/frontend/CMakeLists.txt
@@ -38,6 +38,7 @@ set(NPM_JS_DEPENDENCIES
     ${CMAKE_CURRENT_BINARY_DIR}/node_modules/angular/angular.min.js
     ${CMAKE_CURRENT_BINARY_DIR}/node_modules/d3/d3.min.js
     ${CMAKE_CURRENT_BINARY_DIR}/node_modules/material-design-lite/material.min.js
+    ${CMAKE_CURRENT_BINARY_DIR}/node_modules/qrcode-generator/dist/qrcode.js
 )
 
 add_custom_target(otbr-web-frontend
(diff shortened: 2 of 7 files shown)


src/web/web-service/frontend/res/js/app.js
```diff
@@ -287,11 +287,20 @@
                     console.log(response);
                     $scope.res = response.data.result;
                     if (response.data.result == 'successful') {
-                        var image = "http://api.qrserver.com/v1/create-qr-code/?color=000000&amp;bgcolor=FFFFFF&amp;data=v%3D1%26%26eui%3D" + response.data.eui64 +"%26%26cc%3D" + $scope.thread.pskd +"&amp;qzone=1&amp;margin=0&amp;size=400x400&amp;ecc=L";
-                        $scope.showQRCode(event, image);
+                        try {
+                            var qrData = "v=1&&eui=" + response.data.eui64 + "&&cc=" + $scope.thread.pskd;
```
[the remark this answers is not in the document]
LJspice: `wpan_service.cpp` will not return successful for undefined or null `eui64` values.

`join.dialog.html` will not allow clicking the button until the `pskd` value complies with checks.

hiSandog:
Moving QR generation into the client avoids disclosing the network credential to a third-party service and removes an external availability dependency. It would be worth checking that the generator asset is included in packaged and embedded WebUI builds, not only the npm development path, so the page remains fully offline after installation. A small browser test that blocks all network requests and still produces a QR code would document that privacy property directly.

src/web/web-service/frontend/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Dependency directories
+node_modules/
```
jwhui: In `ot-br-posix`, all ignore patterns are managed centrally in the root `.gitignore` file rather than in sub-directories. Can we move this directly to the repository root `/.gitignore`?

src/web/web-service/frontend/CMakeLists.txt
jwhui, line 48: Because `package-lock.json` is not copied and not listed under `DEPENDS`, CMake builds do not use the lockfile when running `npm install`. Suggest updating this to copy and depend on both `package.json` and `package-lock.json`.

src/web/web-service/frontend/res/js/app.js
```diff
@@ -287,11 +287,20 @@
...
+                            var qr = qrcode(0, 'L');
+                            qr.addData(qrData);
+                            qr.make();
+                            var image = qr.createDataURL(10, 0);
```
jwhui: Adding a small quiet zone (e.g. `2` modules as per ISO/IEC 18004 recommendations) ensures mobile camera scanners and commissioner apps can reliably detect the QR code boundary regardless of container styling.
```suggestion
                            var image = qr.createDataURL(10, 2);
```

jwhui (approved):
Thanks! 👍🏼

PR #3454, opened 2026-07-06
[github-actions] remove MATN test jobs from workflow

jwhui:
This commit removes the MATN test jobs from the border router GitHub Actions workflow because equivalent integration tests now exist under the tests/scripts/expect directory using the docker-in-docker framework.

Specifically, the multicast forwarding and BBR tests are covered by tests/scripts/expect/dind_multicast.exp which runs in the DinD integration suite.

Equivalent tests path:
https://github.com/openthread/ot-br-posix/tree/main/tests/scripts/expect


abtink (approved):
👍

PR #3476, opened 2026-07-20
[utils] add nullptr guards to Bytes2Hex overloads

jwhui:
This commit adds nullptr checks to otbr::Utils::Bytes2Hex overloads in src/utils/hex.cpp using VerifyOrExit and the exit: label convention to handle null pointers safely across the codebase.


abtink (approved):
👍

PR #3478, opened 2026-07-24
[agent] add generic readiness notification fd support

agners:
## Motivation

otbr-agent already notifies systemd (via `sd_notify(READY=1)`) and Upstart (via `SIGSTOP`) when it is ready. Other service supervisors — s6, dinit, and similar — use a simpler, fd-based readiness protocol: the daemon writes a newline to a file descriptor provided by the supervisor (see https://skarnet.org/software/s6/notifywhenup.html).

This PR adds support for that protocol, mirroring the existing Upstart integration: a new CMake option `OTBR_NOTIFY_FD` (default `ON`, like `OTBR_NOTIFY_UPSTART`) compiles the support in, and at runtime the agent writes a newline to the file descriptor given in the `OTBR_NOTIFY_FD` environment variable — at the same point in `Application::Run()` where systemd/Upstart readiness is signalled. Behavior is unchanged when the variable is not set, and no new dependencies are needed.

This allows supervising otbr-agent under s6/s6-overlay (e.g. in containers) with exact, event-driven readiness instead of polling for the control socket: dependent services (`ot-ctl` based configuration, otbr-web) can be started exactly when the agent is initialized.

## Testing

- Built with the option `ON` (default) and `OFF` (code compiled out).
- Ran under s6-overlay (Docker, Debian trixie) with an OpenThread simulation RCP: s6 marks the service ready exactly when the agent logs `Notify readiness on file descriptor 3.`, and dependent services start afterwards.
- Verified no behavior change when `OTBR_NOTIFY_FD` is unset, and that an invalid value only logs a warning.

### DIFF
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -187,6 +187,11 @@ if(OTBR_NOTIFY_UPSTART)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NOTIFY_UPSTART=1)
 endif()
 
+option(OTBR_NOTIFY_FD "Support readiness notification via file descriptor (s6, dinit, ...)." ON)
+if(OTBR_NOTIFY_FD)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NOTIFY_FD=1)
+endif()
+
 set(OTBR_NAT64_DEFAULT OFF)
 if (OTBR_BORDER_ROUTING)
     set(OTBR_NAT64_DEFAULT ON)
--- a/etc/cmake/options.cmake
+++ b/etc/cmake/options.cmake
@@ -187,11 +187,6 @@ if(OTBR_NOTIFY_UPSTART)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NOTIFY_UPSTART=1)
 endif()
 
-option(OTBR_NOTIFY_FD "Support readiness notification via file descriptor (s6, dinit, ...)." ON)
-if(OTBR_NOTIFY_FD)
-    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NOTIFY_FD=1)
-endif()
-
 set(OTBR_NAT64_DEFAULT OFF)
 if (OTBR_BORDER_ROUTING)
     set(OTBR_NAT64_DEFAULT ON)
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -33,14 +33,15 @@
 
 #define OTBR_LOG_TAG "APP"
 
-#ifdef HAVE_LIBSYSTEMD
-#include <systemd/sd-daemon.h>
-#endif
-
 #include <errno.h>
+#include <limits.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+
+#ifdef HAVE_LIBSYSTEMD
+#include <systemd/sd-daemon.h>
+#endif
 
 #include "agent/application.hpp"
 #include "common/code_utils.hpp"
@@ -178,7 +179,12 @@ otbrError Application::Run(void)
     }
 #endif
 
-#if OTBR_ENABLE_NOTIFY_FD
+    // allow quitting elegantly
+    signal(SIGTERM, HandleSignal);
+
+    // avoid exiting on SIGPIPE
+    signal(SIGPIPE, SIG_IGN);
+
     // Generic readiness notification for fd-based supervisors (s6, dinit,
     // ...): write a newline to the file descriptor passed in OTBR_NOTIFY_FD.
     // See https://skarnet.org/software/s6/notifywhenup.html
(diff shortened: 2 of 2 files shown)


etc/cmake/options.cmake
```diff
@@ -187,6 +187,11 @@ if(OTBR_NOTIFY_UPSTART)
     target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NOTIFY_UPSTART=1)
 endif()
 
+option(OTBR_NOTIFY_FD "Support readiness notification via file descriptor (s6, dinit, ...)." ON)
+if(OTBR_NOTIFY_FD)
+    target_compile_definitions(otbr-config INTERFACE OTBR_ENABLE_NOTIFY_FD=1)
+endif()
```
agners: If we want to avoid too many compile time options, this change could also be part of the default build since the code only becomes active if `OTBR_NOTIFY_FD` is set.

jwhui: Makes sense to me. I agree with simplifying the build config.

src/agent/application.cpp
```diff
@@ -37,6 +37,11 @@
 #include <systemd/sd-daemon.h>
 #endif
 
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
```
jwhui, line 40: nit: can we move this above `HAVE_LIBSYSTEMD` above?

src/agent/application.cpp
```diff
@@ -173,6 +178,44 @@ otbrError Application::Run(void)
...
+        const char *notifyFdEnv = getenv("OTBR_NOTIFY_FD");
+
+        if (notifyFdEnv != nullptr)
+        {
+            char *end      = nullptr;
+            long  notifyFd = strtol(notifyFdEnv, &end, 10);
+
+            if (end != notifyFdEnv && *end == '\0' && notifyFd >= 0)
+            {
+                ssize_t rval;
```
jwhui: Suggest adding bounds checking and `errno` validation for `strtol`, and ensure the file descriptor does not point to standard I/O streams (`0`, `1`, or `2`). 
```suggestion
            errno = 0;
            long notifyFdLong = strtol(notifyFdEnv, &end, 10);

            if (end != notifyFdEnv && *end == '\0' && errno == 0 && notifyFdLong >= 3 && notifyFdLong <= INT_MAX)
            {
                int     notifyFd = static_cast<int>(notifyFdLong);
                ssize_t rval;
                // ... use `notifyFd` directly without repeated static_cast
```

jwhui: Suggest adding bounds checking and `errno` validation for `strtol`, and ensure the file descriptor does not point to standard I/O streams (`0`, `1`, or `2`). 
```suggestion
            errno = 0;
            long notifyFdLong = strtol(notifyFdEnv, &end, 10);

            if (end != notifyFdEnv && *end == '\0' && errno == 0 && notifyFdLong >= 3 && notifyFdLong <= INT_MAX)
            {
                int     notifyFd = static_cast<int>(notifyFdLong);
                ssize_t rval;
                // ... use `notifyFd` directly without repeated static_cast
```

src/agent/application.cpp
```diff
@@ -173,6 +178,44 @@ otbrError Application::Run(void)
     }
 #endif
 
+#if OTBR_ENABLE_NOTIFY_FD
```
jwhui: Should we move this after `signal(SIGPIPE, SIG_IGN)` below?

src/agent/application.cpp
```diff
@@ -173,6 +178,44 @@ otbrError Application::Run(void)
...
+                {
+                    rval = write(static_cast<int>(notifyFd), "\n", 1);
+                } while (rval == -1 && errno == EINTR);
+
+                if (rval == -1)
+                {
+                    otbrLogWarning("Failed to notify readiness on file descriptor %ld: %s", notifyFd, strerror(errno));
+                }
+
+                close(static_cast<int>(notifyFd));
```
jwhui: `Application::Run()` can be executed multiple times within the same process lifecycle (e.g., after a software reset handled via `otPlatReset`).

When `notifyFd` is closed during the initial run, the environment variable `OTBR_NOTIFY_FD` remains set. On a subsequent re-run of `Application::Run()`, the code will attempt to write to and close that file descriptor number again. By that time, the OS may have reassigned that file descriptor number to a new socket or DBus connection, causing data corruption and closing an active resource.

Should we call `unsetenv()` after closing the `notifyFd`?

agners: Hm, I wonder how other processes deal with a restarting agent :thinking: . With a process manager in process, I'd expect a restart to just exit and rely on systemd/process manager to start the process again. We actually run some configuration on startup via a one-shot script (after readyness), it appears that is actually not applied when this platform reset triggers :cry: .

I see there is `OTBR_ENABLE_PLATFORM_RESET_EXIT` which is set for Android. But this could also be useful for other platforms, maybe worth turning into a CMake configuration :thinking: .

But in any case, there is only one readyness notification. So, especially for cases where `OTBR_ENABLE_PLATFORM_RESET_EXIT` is not set, `unsetenv()` is the right thing todo here.

jwhui: Thanks for adding `unsetenv("OTBR_NOTIFY_FD")` - that addresses the descriptor reuse issue cleanly for both pseudo-resets and in-place `execvp` runs.

Regarding `otPlatReset()` behavior: the in-place `execvp` approach was historically chosen to support standalone environments and to inject `--auto-attach=0` on software reset. However, you make a great point that in containerized or supervisor-managed environments (systemd, s6, etc.), exiting and letting the supervisor handle restarts and readiness hooks is much cleaner.

Exposing `OTBR_ENABLE_PLATFORM_RESET_EXIT` as a CMake option (e.g., `OTBR_PLATFORM_RESET_EXIT`) makes a lot of sense.

jwhui (approved):
Thanks! 👍🏼

PR #3479, opened 2026-07-26
[web] add ePSKc (Thread 1.4 Credentials Sharing) support to the web GUI

lboue:
## Summary

Adds Thread 1.4 Credentials Sharing / ephemeral key (ePSKc) support to the legacy built-in web GUI (`otbr-web`), which currently has no way to enable, activate, or monitor ePSKc sessions.

- `GET /epskc_status` — current feature `enabled` flag, session `state` (`disabled`/`stopped`/`started`/`connected`/`accepted`), `active` convenience flag, and UDP `port`.
- `POST /epskc_enabled` — enable/disable the feature (`{"enabled": true|false}`).
- `POST /epskc_activate` — generate a 9-digit TAP and start a session (optional `{"lifetime": ms, "port": n}`), returns `{"tap": "...", "port": n}`.
- `POST /epskc_deactivate` — stop the current session.
- New "ePSKc" panel in the AngularJS frontend: enable/disable toggle, optional lifetime input, Activate/Deactivate buttons, and a prominent read-only display of the generated TAP code.

## Implementation notes

This is implemented purely through the existing `OpenThreadClient` CLI-over-socket mechanism (`ba ephemeralkey enable/disable/start/stop/port/generate-tap`) that every other `WpanService` handler already uses to talk to the daemon (`HandleCommission`, `HandleStatusRequest`, etc.) — **no D-Bus, no new dependencies, no CMake changes**. `ba ephemeralkey` is already available in the vendored OpenThread core.

**Known limitation**: `ba ephemeralkey stop` is an unconditional hard stop. Unlike the D-Bus API's `DeactivateEphemeralKeyMode(retain_active_session)`, there's no CLI equivalent for keeping an already-`Accepted` session alive while just closing the activation window. Documenting this as an accepted trade-off of going through the CLI rather than D-Bus for this GUI.

## Test plan

- [x] `script/make-pretty check clang` equivalent: diffed against both a local clang-format (22.1.5) and clang-format-19.1.1 (the version pinned by `script/clang-format`) — clean on all modified C++ files.
- [x] `node --check` clean on `app.js`.
- [x] Full local build verified end-to-end: fetched the pinned submodules (OpenThread core, mbedtls + framework, cJSON, cpp-httplib), configured with `-DOTBR_WEB=ON -DOTBR_BORDER_AGENT=ON -DOTBR_EPSKC=ON`, and built `otbr-web` with `ninja` — clean under `-Wall -Wextra -Werror`. Confirmed `OPENTHREAD_CONFIG_BORDER_AGENT_EPHEMERAL_KEY_ENABLE=1` is set in the build flags.
- [x] Verified the new `WpanService` handler symbols (`HandleGetEpskcStatusRequest`, `HandleSetEpskcEnabledRequest`, `HandleActivateEpskcRequest`, `HandleDeactivateEpskcRequest`) are present in the linked binary via `nm`. (The four matching `WebServer::Handle*(..., void *aUserData)` static overloads were removed per review — they mirrored an `HttpRequestCallback` dispatch path, `WebServer::HandleHttpRequest`, that is declared but never implemented/called anywhere in the file, so they were dead code.)
- [x] Ran the built binary (`otbr-web -p 8080`, no daemon attached): serves `index.html` (200), serves the updated `app.js` with the new ePSKc functions, and `/epskc_status` returns a well-formed JSON error response (`{"error":15,"result":"failed"}`, i.e. `kWpanStatus_Uninitialized`) instead of crashing — confirms the routing/JSON/error-handling path end-to-end.
- [x] Manual end-to-end test against a real `otbr-agent` attached to a simulated `ot-rcp`, run inside a privileged container (for the TUN interface) with a Thread network formed via `ot-ctl`, then `otbr-web` pointed at the same interface. Exercised a real `ba ephemeralkey` session through the REST endpoints:
  - `epskc_status` → `error:0`
  - `epskc_enabled {"enabled":true}` → `error:0`, reflected back on the next `epskc_status`
  - `epskc_activate` with an **empty body** (exercising the optional-fields fix) → `error:0`, real 9-digit TAP (`"tap":"769077288"`) and real UDP port (`"port":49155"`) assigned by the border agent
  - `epskc_status` → `active:true`
  - Cross-checked directly against the daemon with `ot-ctl ba ephemeralkey` → `Started` (independent confirmation outside the web GUI)
  - `epskc_deactivate` → `error:0`; `epskc_status` → `active:false`, `state:"stopped"`
  - `epskc_enabled {"enabled":"yes"}` (wrong type) → correctly rejected with `error:11` against the real daemon too
- [x] Extended `tests/scripts/expect/dind_web.exp` with a Scenario 3 covering `epskc_status → epskc_enabled → epskc_activate (empty body) → epskc_status → epskc_deactivate → epskc_status`, mirroring the existing DinD web GUI test structure. JSON-field assertions use a whitespace-tolerant regex helper (`check_json_field`) since jsoncpp's default writer pretty-prints with `"key" : value`, not the compact form. Verified the matching logic against real server output with `tclsh`; the script itself needs the full DinD Docker harness (`infrastructure-link` network, built `OTBR_DOCKER_IMAGE`) to run end-to-end, which is what CI's `test_dind_dns_sd.sh` provides.
- [x] Added gtest unit coverage (`tests/gtest/test_web.cpp`, `otbr-gtest-web` target) for `HandleGetEpskcStatusRequest`, `HandleSetEpskcEnabledRequest`, `HandleActivateEpskcRequest`, `HandleDeactivateEpskcRequest`, and the new `toLower` helper — no live daemon required: the input-validation branches (malformed JSON, missing/wrong-typed fields, the empty-body case for `HandleActivateEpskcRequest`'s optional fields) run entirely before `Connect()`, and a bogus interface name deterministically exercises the well-formed-error-JSON path when `Connect()` fails. 14/14 pass locally. This addresses Codecov's "0% patch coverage" flag on `wpan_service.cpp`/`web_server.cpp` — the only test previously exercising this code was the `dind_web.exp` Scenario 3 integration test above, which runs in the `Docker Test` CI workflow and isn't uploaded to Codecov, so it never counted toward patch coverage.
- [x] Rebased onto current `main` to pick up the `otbr-gtest-web` test target and `WpanService`/`OpenThreadClient` test-friend hooks added by #3288, which the unit tests above build on; full rebuild (`otbr-web`, `otbr-gtest-unit`, `otbr-gtest-web`) and `clang-format-19` stay clean after the rebase.

## Review

Addressed automated review feedback from gemini-code-assist:
- Use the local `ev` event parameter (or `null` when none is available) instead of the global `event` in the new frontend functions.
- Replaced the inline `onclick` handler on the TAP display with an `ng-click`-bound `selectEpskcTap()` controller function (CSP-friendly).
- Loop over `$scope.menu.length` instead of a hardcoded panel count in `showPanels()`.
- Format the ephemeral key lifetime (`uint32_t`) with `PRIu32` instead of `%u` for portability.

Addressed maintainer review feedback from @jwhui:
- `wpan_service`: validate that `"enabled"` is present and boolean before extracting it in `HandleSetEpskcEnabledRequest`.
- `wpan_service`: allow an empty POST body in `HandleActivateEpskcRequest` (both `lifetime` and `port` are optional), and validate their types when present instead of blindly calling `asUInt()`.
- `web_server`: removed the four new static `Handle*EpskcRequest(..., void *aUserData)` overloads — dead code, see test plan above.
- `frontend/index.html`: closed the `&nbsp` entities with a semicolon; added `min="1"`/`max="10"` to the lifetime input plus a label/validation message noting the router's 2-minute default / 10-minute maximum.
- `frontend/app.js`: replaced `ev.target.disabled` toggling in `activateEpskc`/`deactivateEpskc` (unreliable with `md-button`, since the click can bubble from an inner ripple/span) with a `$scope.isEpskcBusy` flag bound via `ng-disabled`, with error callbacks so it resets on HTTP/network failure.
- `frontend/app.js`: poll `epskc_status` every 4s while the ePSKc panel is visible, so the UI reflects a session that times out or gets accepted/stopped on the device without local user action; polling stops when navigating away or on `$destroy`.

### DIFF
--- a/src/web/web-service/frontend/index.html
+++ b/src/web/web-service/frontend/index.html
@@ -450,6 +450,43 @@
 
 
         </div>
+        <div class="demo-charts mdl-color--white  mdl-cell mdl-cell--12-col mdl-shadow--2dp mdl-grid" ng-show="menu[7].show">
+          <h4>Ephemeral Key (ePSKc) Credential Sharing</h4>
+            <md-content layout-padding flex="100">
+              <div layout="row" layout-align="start center">
+                <span class="status-title">&nbsp;&nbsp;Feature:&nbsp;</span>
+                <md-switch ng-model="epskc.enabled" ng-change="toggleEpskcEnabled()" aria-label="Enable ePSKc">
+                  {{ epskc.enabled ? 'Enabled' : 'Disabled' }}
+                </md-switch>
+              </div>
+              <div>
+                <span class="status-title">&nbsp;&nbsp;Session state:&nbsp;</span>
+                <span class="status-content">{{ epskc.state }}</span>
+              </div>
+
+              <form name="epskcForm" ng-if="!epskc.active">
+                <md-input-container flex="50">
+                  <label>Lifetime (minutes, 2 default, 10 max)</label>
+                  <input name="lifetimeMinutes" type="number" min="1" max="10" ng-model="epskc.lifetimeMinutes" placeholder="use router default">
+                  <div ng-messages="epskcForm.lifetimeMinutes.$error">
+                    <div ng-message-exp="['min', 'max']">Lifetime must be between 1 and 10 minutes.</div>
+                  </div>
+                </md-input-container>
+                <div>
+                  <md-button name="activateEpskcButton" class="mdl-button mdl-js-button mdl-button--raised mdl-button--colored" ng-disabled="!epskc.enabled || isEpskcBusy || epskcForm.$invalid" ng-click="activateEpskc($event)">Activate</md-button>
+                </div>
+              </form>
+
+              <div ng-if="epskc.tap" class="epskc-tap-display">
+                <h3>TAP Code</h3>
+                <input readonly class="epskc-tap-code" ng-value="epskc.tap" ng-click="selectEpskcTap($event)">
+              </div>
+
+              <div ng-if="epskc.active">
+                <md-button name="deactivateEpskcButton" class="mdl-button mdl-js-button mdl-button--raised" ng-disabled="isEpskcBusy" ng-click="deactivateEpskc($event)">Deactivate</md-button>
+              </div>
+            </md-content>
+        </div>
       </div>
     </main>
   </div>
--- a/src/web/web-service/frontend/index.html
+++ b/src/web/web-service/frontend/index.html
@@ -450,6 +450,40 @@
 
 
         </div>
+        <div class="demo-charts mdl-color--white  mdl-cell mdl-cell--12-col mdl-shadow--2dp mdl-grid" ng-show="menu[7].show">
+          <h4>Ephemeral Key (ePSKc) Credential Sharing</h4>
+            <md-content layout-padding flex="100">
+              <div layout="row" layout-align="start center">
+                <span class="status-title">&nbsp&nbspFeature:&nbsp</span>
+                <md-switch ng-model="epskc.enabled" ng-change="toggleEpskcEnabled()" aria-label="Enable ePSKc">
+                  {{ epskc.enabled ? 'Enabled' : 'Disabled' }}
+                </md-switch>
+              </div>
+              <div>
+                <span class="status-title">&nbsp&nbspSession state:&nbsp</span>
+                <span class="status-content">{{ epskc.state }}</span>
+              </div>
+
+              <form name="epskcForm" ng-if="!epskc.active">
+                <md-input-container flex="50">
+                  <label>Lifetime (minutes, optional)</label>
+                  <input name="lifetimeMinutes" type="number" min="0" ng-model="epskc.lifetimeMinutes" placeholder="use router default">
+                </md-input-container>
+                <div>
+                  <md-button name="activateEpskcButton" class="mdl-button mdl-js-button mdl-button--raised mdl-button--colored" ng-disabled="!epskc.enabled" ng-click="activateEpskc($event)">Activate</md-button>
+                </div>
+              </form>
+
+              <div ng-if="epskc.tap" class="epskc-tap-display">
+                <h3>TAP Code</h3>
+                <input readonly class="epskc-tap-code" ng-value="epskc.tap" ng-click="selectEpskcTap($event)">
+              </div>
+
+              <div ng-if="epskc.active">
+                <md-button name="deactivateEpskcButton" class="mdl-button mdl-js-button mdl-button--raised" ng-click="deactivateEpskc($event)">Deactivate</md-button>
+              </div>
+            </md-content>
+        </div>
       </div>
     </main>
   </div>
--- a/src/web/web-service/frontend/res/js/app.js
+++ b/src/web/web-service/frontend/res/js/app.js
@@ -86,6 +86,11 @@
                 icon: 'add_circle_outline',
                 show: false,
             },
+            {
+                title: 'ePSKc',
+                icon: 'vpn_key',
+                show: false,
+            },
 
         ];
 
@@ -108,6 +113,36 @@
         $scope.headerTitle = 'Home';
         $scope.status = [];
 
+        $scope.epskc = {
+            enabled: false,
+            state: 'unknown',
+            active: false,
+            tap: '',
+            port: null,
+            lifetimeMinutes: null,
+        };
+        $scope.isEpskcBusy = false;
+
+        // Polls epskc_status while the ePSKc panel is visible, so the UI reflects a session
+        // that times out or gets accepted/stopped on the device without any local user action.
+        var epskcStatusPoll = null;
+        var startEpskcStatusPoll = function() {
+            if (epskcStatusPoll) {
+                return;
+            }
+            epskcStatusPoll = $interval(function() {
+                $scope.refreshEpskcStatus(/* suppressAlert */ true);
+            }, 4000);
+        };
+        var stopEpskcStatusPoll = function() {
+            if (!epskcStatusPoll) {
+                return;
+            }
+            $interval.cancel(epskcStatusPoll);
+            epskcStatusPoll = null;
+        };
+        $scope.$on('$destroy', stopEpskcStatusPoll);
+
         $scope.isLoading = false;
 
         $scope.showScanAlert = function(ev) {
@@ -124,7 +159,7 @@
         };
         $scope.showPanels = async function(index) {
             $scope.headerTitle = $scope.menu[index].title;
-            for (var i = 0; i < 7; i++) {
+            for (var i = 0; i < $scope.menu.length; i++) {
                 $scope.menu[i].show = false;
             }
             $scope.menu[index].show = true;
@@ -164,6 +199,12 @@
 
                 await $scope.dataInit();
                 await $scope.showTopology();
+            }
+            if (index == 7) {
+                $scope.refreshEpskcStatus();
+                startEpskcStatusPoll();
+            } else {
+                stopEpskcStatusPoll();
             }
         };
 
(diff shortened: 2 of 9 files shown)


src/web/web-service/wpan_service.cpp
```diff
@@ -604,6 +606,177 @@ std::string WpanService::HandleCommission(const std::string &aCommissionRequest)
...
+    Json::StreamWriterBuilder         writerBuilder;
+    std::string                       response;
+    bool                              enabled;
+    int                               ret = kWpanStatus_Ok;
+    otbr::Web::OpenThreadClient       client(mIfName);
+
+    VerifyOrExit(reader->parse(aSetEpskcEnabledRequest.c_str(),
+                               aSetEpskcEnabledRequest.c_str() + aSetEpskcEnabledRequest.size(), &root, nullptr),
+                 ret = kWpanStatus_ParseRequestFailed);
+    enabled = root["enabled"].asBool();
```
jwhui, line 679: Validate that `"enabled"` is present and is a boolean before extracting it?

src/web/web-service/wpan_service.cpp
```diff
@@ -604,6 +606,177 @@ std::string WpanService::HandleCommission(const std::string &aCommissionRequest)
...
+    Json::StreamWriterBuilder         writerBuilder;
+    std::string                       response;
+    std::string                       tap;
+    uint32_t                          lifetime = 0; // 0 == let the router use its own default lifetime
+    uint16_t                          port     = 0; // 0 == let the router use its own default UDP port
+    int                               ret      = kWpanStatus_Ok;
+    otbr::Web::OpenThreadClient       client(mIfName);
+    char                             *rval;
+
+    VerifyOrExit(reader->parse(aActivateEpskcRequest.c_str(),
```
jwhui: Since `lifetime` and `port` are both optional, a client might send a `POST /epskc_activate` request with an empty body (e.g. via `curl -X POST`). `reader->parse` will fail on an empty string.

Should allow an empty body and verifying types when fields are provided?

src/web/web-service/web_server.hpp
```diff
@@ -98,6 +98,10 @@ class WebServer
     static std::string HandleGetAvailableNetworkResponse(const std::string &aGetAvailableNetworkRequest,
                                                          void              *aUserData);
     static std::string HandleCommission(const std::string &aCommissionRequest, void *aUserData);
+    static std::string HandleGetEpskcStatusRequest(const std::string &aGetEpskcStatusRequest, void *aUserData);
+    static std::string HandleSetEpskcEnabledRequest(const std::string &aSetEpskcEnabledRequest, void *aUserData);
+    static std::string HandleActivateEpskcRequest(const std::string &aActivateEpskcRequest, void *aUserData);
+    static std::string HandleDeactivateEpskcRequest(const std::string &aDeactivateEpskcRequest, void *aUserData);
```
jwhui: Are these static methods used?

src/web/web-service/frontend/index.html
```diff
@@ -450,6 +450,40 @@ <h4>Commission</h4>
 
 
         </div>
+        <div class="demo-charts mdl-color--white  mdl-cell mdl-cell--12-col mdl-shadow--2dp mdl-grid" ng-show="menu[7].show">
+          <h4>Ephemeral Key (ePSKc) Credential Sharing</h4>
+            <md-content layout-padding flex="100">
+              <div layout="row" layout-align="start center">
+                <span class="status-title">&nbsp&nbspFeature:&nbsp</span>
```
jwhui: HTML entities should end with a semicolon (i.e. `&nbsp;&nbsp;Feature:&nbsp;` and `&nbsp;&nbsp;Session state:&nbsp;`).

src/web/web-service/frontend/index.html
```diff
@@ -450,6 +450,40 @@ <h4>Commission</h4>
...
+                <md-switch ng-model="epskc.enabled" ng-change="toggleEpskcEnabled()" aria-label="Enable ePSKc">
+                  {{ epskc.enabled ? 'Enabled' : 'Disabled' }}
+                </md-switch>
+              </div>
+              <div>
+                <span class="status-title">&nbsp&nbspSession state:&nbsp</span>
```
jwhui: HTML entities should end with a semicolon (i.e. `&nbsp;&nbsp;Feature:&nbsp;` and `&nbsp;&nbsp;Session state:&nbsp;`).

src/web/web-service/frontend/index.html
```diff
@@ -450,6 +450,40 @@ <h4>Commission</h4>
...
+                <span class="status-content">{{ epskc.state }}</span>
+              </div>
+
+              <form name="epskcForm" ng-if="!epskc.active">
+                <md-input-container flex="50">
+                  <label>Lifetime (minutes, optional)</label>
+                  <input name="lifetimeMinutes" type="number" min="0" ng-model="epskc.lifetimeMinutes" placeholder="use router default">
```
jwhui: OpenThread enforces a maximum ephemeral key lifetime of 10 minutes (`OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT = 10 * 60 * 1000u`) and defaults to 2 minutes. Setting `min="1" max="10"` and mentioning the 2-minute default / 10-minute maximum in the placeholder or label would provide clearer guidance to the user.

src/web/web-service/frontend/res/js/app.js
```diff
@@ -438,6 +455,82 @@
...
+                $scope.refreshEpskcStatus();
+            });
+        };
+
+        $scope.activateEpskc = function(ev) {
+            var data = {};
+            if ($scope.epskc.lifetimeMinutes) {
+                data.lifetime = $scope.epskc.lifetimeMinutes * 60 * 1000;
+            }
+            ev.target.disabled = true;
```
jwhui: In AngularJS / Angular Material, clicking an `<md-button>` frequently triggers the click event on a child element (e.g. inner `<span>` or ripple container `<div>`). As a result, `ev.target` is not guaranteed to be the `<button>` element, meaning `ev.target.disabled = true` will not disable the button or prevent duplicate submissions.

Instead, prefer using a scope variable (e.g., `$scope.isEpskcBusy = true;`) and bind `ng-disabled="isEpskcBusy || !epskc.enabled"` on the button.

Also, consider adding a `.catch()` block (or error callback) to reset `$scope.isEpskcBusy = false` if an HTTP/network failure occurs.

src/web/web-service/frontend/res/js/app.js
```diff
@@ -438,6 +455,82 @@
...
+            $http({
+                method: 'POST',
+                url: 'epskc_activate',
+                data: data,
+            }).then(function(response) {
+                ev.target.disabled = false;
```
jwhui: See https://github.com/openthread/ot-br-posix/pull/3479/changes#r3753886856

src/web/web-service/frontend/res/js/app.js
```diff
@@ -438,6 +455,82 @@
...
+                    $scope.showAlert(ev, 'Activate ePSKc', 'success');
+                } else {
+                    $scope.showAlert(ev, 'Activate ePSKc', 'failed');
+                }
+                $scope.refreshEpskcStatus();
+            });
+        };
+
+        $scope.deactivateEpskc = function(ev) {
+            ev.target.disabled = true;
```
jwhui: See https://github.com/openthread/ot-br-posix/pull/3479/changes#r3753886856

src/web/web-service/frontend/res/js/app.js
```diff
@@ -438,6 +455,82 @@
...
+            $http({
+                method: 'POST',
+                url: 'epskc_deactivate',
+                data: {},
+            }).then(function(response) {
+                ev.target.disabled = false;
```
jwhui: See https://github.com/openthread/ot-br-posix/pull/3479/changes#r3753886856

src/web/web-service/frontend/res/js/app.js
```diff
@@ -438,6 +455,82 @@
             });
         };
 
+        $scope.refreshEpskcStatus = function() {
```
jwhui: When an ePSKc session is active, it will automatically expire on the border router once the timeout passes (or when a commissioner finishes connecting). 

Since `refreshEpskcStatus()` is currently only triggered manually (tab switch, activate/deactivate, or toggle), the UI will remain in the active state displaying the TAP code even after the session has timed out on the device.

Consider starting a periodic `$interval` (e.g. every 3–5 seconds) while `epskc.active` is true or while panel 7 is visible, and canceling it when the panel is left or the session stops.

src/web/web-service/frontend/res/js/app.js
```diff
@@ -447,6 +486,91 @@
...
```
jwhui: Should we suppress modal alerts during status polling so background poll errors don't disrupt the user interface?

Because `refreshEpskcStatus()` is invoked every 4 seconds via `$interval` while viewing the ePSKc panel, calling `$scope.showAlert()` when `response.data.error != 0` will repeatedly trigger modal alert dialogs if the daemon is not running, uninitialized, or temporarily busy.

lboue:
Do I have to squash commits?

jwhui:
@lboue , yes please squash commits.

src/web/web-service/frontend/res/js/app.js
lboue: Addressed in 59735f68 — `refreshEpskcStatus()` is now polled via `$interval` every 4s while panel 7 (ePSKc) is visible: started in `showPanels()` when entering the panel, and canceled when leaving it or on `$destroy`.

That surfaced a side effect: with the poll running, a transient failure (daemon busy/unreachable) would pop `showAlert()` every 4s. Added a `suppressAlert` param to `refreshEpskcStatus()`, passed `true` only from the poll tick, so background polling fails silently while the manual triggers (opening the panel, activate/deactivate/toggle) still alert on failure as before.

jwhui (approved):
Thanks! 👍🏼

PR #3480, opened 2026-07-26
[rest] add endpoints for Border Agent ephemeral key (ePSKc)

lboue:
## Summary

- Adds `GET`/`PUT` `/node/ba-epskc/state` to query and toggle the Border Agent ephemeral key (ePSKc) feature.
- Adds `GET`/`POST`/`DELETE` `/node/ba-epskc/key` to query the ePSKc session status, generate and activate a new ephemeral key (Thread Administration Passcode), and deactivate the current one.
- Documents the new endpoints in `src/rest/openapi.yaml`.
- Adds an integration test (`epskc_test`) to `tests/rest/test_rest.py` covering enable/disable, activation while disabled (409), double activation (409), successful activation, status, and deactivation.

This is a rebase and adaptation of @puddly's original patch (https://github.com/puddly/ot-br-posix/commit/d8bd507e42a3e3a3a8517c7d7f55686cd024a25), which predates ~250 commits of drift on `main`, including the `openthread` submodule bump that moved the ephemeral key API from `otBorderAgentEphemeralKeyGenerateTap()` to the `BorderAgent::CreateEphemeralKey()` helper already used by the D-Bus `ActivateEphemeralKeyMode` method. I opened this on @puddly's behalf since they didn't have time to rebase it themselves; happy to have them take over review/authorship if preferred.

## Test plan

- [x] `cmake -DOTBR_REST=ON -DOT_PLATFORM=simulation ...` + `ninja otbr-agent` builds cleanly with `OTBR_ENABLE_EPSKC=1`.
- [x] `script/clang-format-check` passes (clang-format 19.1.7) on all changed C++ files.
- [x] Manually reproduced `tests/rest/test-rest-server` end-to-end without root, since `sudo`/`wpan0` weren't available in my sandbox: built an OpenThread simulation RCP (`OT_PLATFORM=simulation`), ran `otbr-agent` + `ot-rcp` + `ot-ctl` inside an unprivileged `unshare --user --net --mount` namespace (tmpfs over `/run` for the daemon socket, `--data-path` for settings), formed a Thread network, and ran the real `tests/rest/test_rest.py` against it. All pre-existing checks pass (`200/200` on every endpoint) and the new `epskc_test` passes end-to-end: 409 on activation while disabled, successful activation with a valid 9-digit TAP and port, status transitions (`stopped` → `started`), 409 on double activation, and deactivation back to `stopped`/`disabled`.
- [x] Also validated against real Thread radio hardware (Nabu Casa ZBT-2, Silicon Labs EFR32 RCP firmware `SL-OPENTHREAD/3.0.0.0_GitHub-61e43cffb`) over `spinel+hdlc+uart`, built with the same OpenThread core configuration as the Home Assistant [`openthread_border_router` add-on](https://github.com/home-assistant/addons/tree/master/openthread_border_router) (Thread 1.4, Border Agent, Backbone Router, TREL, `OPENTHREAD_CONFIG_BORDER_AGENT_MESHCOP_SERVICE_BASE_NAME`, etc.). The device formed a real Thread network (became `leader`) and `tests/rest/test_rest.py`, including `epskc_test`, passed end-to-end (`exit=0`) against the real hardware.
- [ ] CI `rest-check` / `check` / other GitHub Actions jobs — currently blocked on `action_required` pending a maintainer approval since this is a first-time external contribution; not something I can trigger myself.

### DIFF
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -1884,6 +1884,92 @@ cJSON *CreateMetaCollection(uint32_t aOffset, uint32_t aLimit, uint32_t aTotal, 
     return meta;
 }
 
+#if OTBR_ENABLE_EPSKC
+static const char *EpskcStateToString(otBorderAgentEphemeralKeyState aState)
+{
+    const char *stateStr;
+
+    switch (aState)
+    {
+    case OT_BORDER_AGENT_STATE_DISABLED:
+        stateStr = "disabled";
+        break;
+    case OT_BORDER_AGENT_STATE_STOPPED:
+        stateStr = "stopped";
+        break;
+    case OT_BORDER_AGENT_STATE_STARTED:
+        stateStr = "started";
+        break;
+    case OT_BORDER_AGENT_STATE_CONNECTED:
+        stateStr = "connected";
+        break;
+    case OT_BORDER_AGENT_STATE_ACCEPTED:
+        stateStr = "accepted";
+        break;
+    default:
+        stateStr = "unknown";
+        break;
+    }
+
+    return stateStr;
+}
+
+std::string EpskcKeyStatus2JsonString(otBorderAgentEphemeralKeyState aState, uint16_t aPort)
+{
+    std::string ret;
+    cJSON      *root = cJSON_CreateObject();
+
+    cJSON_AddStringToObject(root, "state", EpskcStateToString(aState));
+    cJSON_AddNumberToObject(root, "port", aPort);
+
+    ret = Json2String(root);
+    cJSON_Delete(root);
+
+    return ret;
+}
+
+std::string EpskcActivateResult2JsonString(const std::string &aTap, uint16_t aPort)
+{
+    std::string ret;
+    cJSON      *root = cJSON_CreateObject();
+
+    cJSON_AddStringToObject(root, "tap", aTap.c_str());
+    cJSON_AddNumberToObject(root, "port", aPort);
+
+    ret = Json2String(root);
+    cJSON_Delete(root);
+
+    return ret;
+}
+
+bool JsonEpskcActivateParams(const std::string &aJson, uint32_t &aLifetime, uint16_t &aPort)
+{
+    cJSON *root;
+    cJSON *value;
+    bool   ret = true;
+
+    root = cJSON_Parse(aJson.c_str());
+    VerifyOrExit(root != nullptr, ret = false);
+    VerifyOrExit(cJSON_IsObject(root), ret = false);
+
+    value = cJSON_GetObjectItemCaseSensitive(root, "lifetime");
+    if (cJSON_IsNumber(value))
+    {
+        aLifetime = static_cast<uint32_t>(value->valueint);
+    }
+
+    value = cJSON_GetObjectItemCaseSensitive(root, "port");
+    if (cJSON_IsNumber(value))
+    {
+        aPort = static_cast<uint16_t>(value->valueint);
+    }
+
+exit:
+    cJSON_Delete(root);
+    return ret;
+}
+#endif // OTBR_ENABLE_EPSKC
+
 } // namespace Json
 } // namespace rest
 } // namespace otbr
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -1955,13 +1955,15 @@ bool JsonEpskcActivateParams(const std::string &aJson, uint32_t &aLifetime, uint
     value = cJSON_GetObjectItemCaseSensitive(root, "lifetime");
     if (cJSON_IsNumber(value))
     {
-        aLifetime = static_cast<uint32_t>(value->valueint);
+        VerifyOrExit(value->valuedouble >= 0.0 && value->valuedouble <= UINT32_MAX, ret = false);
+        aLifetime = static_cast<uint32_t>(value->valuedouble);
     }
 
     value = cJSON_GetObjectItemCaseSensitive(root, "port");
     if (cJSON_IsNumber(value))
     {
-        aPort = static_cast<uint16_t>(value->valueint);
+        VerifyOrExit(value->valuedouble >= 0.0 && value->valuedouble <= UINT16_MAX, ret = false);
+        aPort = static_cast<uint16_t>(value->valuedouble);
     }
 
 exit:
--- a/src/rest/json.cpp
+++ b/src/rest/json.cpp
@@ -1953,15 +1953,17 @@ bool JsonEpskcActivateParams(const std::string &aJson, uint32_t &aLifetime, uint
     VerifyOrExit(cJSON_IsObject(root), ret = false);
 
     value = cJSON_GetObjectItemCaseSensitive(root, "lifetime");
-    if (cJSON_IsNumber(value))
+    if (value != nullptr)
     {
+        VerifyOrExit(cJSON_IsNumber(value), ret = false);
         VerifyOrExit(value->valuedouble >= 0.0 && value->valuedouble <= UINT32_MAX, ret = false);
         aLifetime = static_cast<uint32_t>(value->valuedouble);
     }
 
     value = cJSON_GetObjectItemCaseSensitive(root, "port");
-    if (cJSON_IsNumber(value))
+    if (value != nullptr)
     {
+        VerifyOrExit(cJSON_IsNumber(value), ret = false);
         VerifyOrExit(value->valuedouble >= 0.0 && value->valuedouble <= UINT16_MAX, ret = false);
         aPort = static_cast<uint16_t>(value->valuedouble);
     }
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -1734,8 +1734,9 @@ paths:
                   description: |-
                     The key lifetime in milliseconds. Defaults to
                     OT_BORDER_AGENT_DEFAULT_EPHEMERAL_KEY_TIMEOUT (2 minutes)
-                    when omitted or zero. Capped at
-                    OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT (10 minutes).
+                    when omitted or zero. Maximum value is
+                    OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT (10 minutes);
+                    values above the maximum are rejected with 400.
                   example: 120000
                 port:
                   type: integer
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -1660,6 +1660,114 @@ paths:
                 type: string
                 description: Coprocessor version string
                 example: "OPENTHREAD/thread-reference-20200818-1740-g33cc75ed3; NRF52840; Jun  2 2022 14:25:49"
+  /node/ba-epskc/state:
+    get:
+      tags:
+        - node
+      summary: Get whether the Border Agent ephemeral key (ePSKc) feature is enabled.
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                type: string
+                description: Whether the ephemeral key feature is enabled.
+                enum: ["enabled", "disabled"]
+                example: "enabled"
+        "404":
+          $ref: "#/components/responses/NotFoundError"
+    put:
+      tags:
+        - node
+      summary: Enable or disable the Border Agent ephemeral key (ePSKc) feature.
+      description: |-
+        The feature must be enabled before an ephemeral key can be activated
+        with a POST request to `/node/ba-epskc/key`.
+      responses:
+        "200":
+          description: Successful operation.
+        "400":
+          $ref: "#/components/responses/BadRequestError"
+        "404":
+          $ref: "#/components/responses/NotFoundError"
+      requestBody:
+        description: New ephemeral key feature state.
+        content:
+          application/json:
+            schema:
+              type: string
+              description: Can be "enable" or "disable".
+              example: "enable"
+  /node/ba-epskc/key:
+    get:
+      tags:
+        - node
+      summary: Get the status of the Border Agent ephemeral key (ePSKc) session.
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/EpskcStatus"
+        "404":
+          $ref: "#/components/responses/NotFoundError"
+    post:
+      tags:
+        - node
+      summary: Generate and activate an ephemeral key (ePSKc).
+      description: |-
+        Generates a new 9-digit Thread Administration Passcode (TAP) and
+        starts listening for an external commissioner candidate to connect
+        using it. The ephemeral key feature must first be enabled with a PUT
+        request to `/node/ba-epskc/state`.
+      requestBody:
+        description: Optional activation parameters. An empty body uses the device defaults.
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                lifetime:
+                  type: integer
+                  description: |-
+                    The key lifetime in milliseconds. Defaults to
+                    OT_BORDER_AGENT_DEFAULT_EPHEMERAL_KEY_TIMEOUT (2 minutes)
+                    when omitted or zero. Capped at
+                    OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT (10 minutes).
+                  example: 120000
+                port:
+                  type: integer
+                  description: The UDP port to use. Defaults to the device's own default when omitted.
+                  example: 49152
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/EpskcActivateResult"
+        "400":
+          $ref: "#/components/responses/BadRequestError"
+        "404":
+          $ref: "#/components/responses/NotFoundError"
+        "409":
+          $ref: "#/components/responses/ConflictError"
+        "500":
+          $ref: "#/components/responses/InternalServerError"
+    delete:
+      tags:
+        - node
+      summary: Deactivate the currently active ephemeral key (ePSKc), if any.
+      description: |-
+        Stops the ephemeral key use and disconnects any commissioner session
+        using it. Has no effect if there is no ephemeral key in use.
+      responses:
+        "200":
+          description: Successful operation.
+        "404":
+          $ref: "#/components/responses/NotFoundError"
 
 ### Components ##############################################################
 #############################################################################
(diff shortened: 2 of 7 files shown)


src/rest/json.cpp
```diff
@@ -1884,6 +1884,92 @@ cJSON *CreateMetaCollection(uint32_t aOffset, uint32_t aLimit, uint32_t aTotal,
...
+    value = cJSON_GetObjectItemCaseSensitive(root, "port");
+    if (cJSON_IsNumber(value))
+    {
+        aPort = static_cast<uint16_t>(value->valueint);
+    }
+
+exit:
+    cJSON_Delete(root);
+    return ret;
+}
```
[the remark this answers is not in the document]
lboue: Fixed [29b1bff](https://github.com/openthread/ot-br-posix/pull/3480/commits/29b1bff92300bf3e288ca2c5ae9494514273d909)

lboue:
### ePSKc REST API demo on real hardware

Ran against a Nabu Casa ZBT-2 (Silicon Labs EFR32 RCP, `SL-OPENTHREAD/3.0.0.0_GitHub-61e43cffb`), `otbr-agent` built from this branch with the same OpenThread core configuration as the Home Assistant [`openthread_border_router`](https://github.com/home-assistant/addons/tree/master/openthread_border_router) add-on (Thread 1.4, Border Agent, `OTBR_ENABLE_EPSKC=1`).

```
### PUT /node/ba-epskc/state  (body: "enable")

HTTP status: 200

### GET /node/ba-epskc/key  (before activation)
{
	"state":	"stopped",
	"port":	0
}
HTTP status: 200

### POST /node/ba-epskc/key  (activation, empty body = defaults)
{
	"tap":	"700855744",
	"port":	49154
}
HTTP status: 200

### GET /node/ba-epskc/key  (after activation)
{
	"state":	"started",
	"port":	49154
}
HTTP status: 200

### DELETE /node/ba-epskc/key  (deactivation)

HTTP status: 200

### PUT /node/ba-epskc/state  (body: "disable")

HTTP status: 200
```

lboue:
I looked into the failing rest-check.

The failure is in RcpHostApi.StateChangesCorrectlyAfterJoin, where the second Join() remains DETACHED instead of reaching LEADER.

I also checked the PR diff: it does not modify RcpHostApi, Join(), Leave(), ThreadDetachGracefully(), or the related GTest.

So this appears unrelated to the changes in this PR. It would be worth checking whether the test also fails on the parent main commit.


The failing rest-check is RcpHostApi.StateChangesCorrectlyAfterJoin, which is outside the code paths modified by this PR.

I opened an issue to track and investigate this failure separately: https://github.com/openthread/ot-br-posix/issues/3507

src/rest/json.cpp
```diff
@@ -1884,6 +1884,94 @@ cJSON *CreateMetaCollection(uint32_t aOffset, uint32_t aLimit, uint32_t aTotal,
...
+{
+    cJSON *root;
+    cJSON *value;
+    bool   ret = true;
+
+    root = cJSON_Parse(aJson.c_str());
+    VerifyOrExit(root != nullptr, ret = false);
+    VerifyOrExit(cJSON_IsObject(root), ret = false);
+
+    value = cJSON_GetObjectItemCaseSensitive(root, "lifetime");
```
jwhui, line 1955: If `lifetime` or `port` is provided in the JSON body with an unexpected type, value is non-null but `cJSON_IsNumber(value)` is `false`. As written, the `if (cJSON_IsNumber(value))` condition simply skips parsing and leaves the default values intact, silently succeeding rather than failing validation.

To ensure invalid types return a 400 Bad Request, we should verify `cJSON_IsNumber` when the field is present.

src/rest/rest_web_server.hpp
```diff
@@ -137,6 +137,16 @@ class RestWebServer : public std::enable_shared_from_this<RestWebServer>
...
     void GetCoprocessorVersion(Response &aResponse) const;
 
+#if OTBR_ENABLE_EPSKC
+    void EpskcState(const Request &aRequest, Response &aResponse);
+    void GetEpskcState(Response &aResponse) const;
+    void SetEpskcState(const Request &aRequest, Response &aResponse);
+    void EpskcKey(const Request &aRequest, Response &aResponse);
+    void GetEpskcKey(Response &aResponse) const;
+    void ActivateEpskcKey(const Request &aRequest, Response &aResponse);
+    void DeactivateEpskcKey(const Request &aRequest, Response &aResponse);
```
jwhui: For consistency with existing handler methods:
```suggestion
    void EpskcState(const Request &aRequest, Response &aResponse) const;
    void GetEpskcState(Response &aResponse) const;
    void SetEpskcState(const Request &aRequest, Response &aResponse) const;
    void EpskcKey(const Request &aRequest, Response &aResponse) const;
    void GetEpskcKey(Response &aResponse) const;
    void ActivateEpskcKey(const Request &aRequest, Response &aResponse) const;
    void DeactivateEpskcKey(const Request &aRequest, Response &aResponse) const;
```

src/rest/rest_web_server.cpp
```diff
@@ -1066,6 +1084,165 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+{
+    otbrError   error = OTBR_ERROR_NONE;
+    std::string body;
+
+    VerifyOrExit(Json::JsonString2String(aRequest.body, body), error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(body == "enable" || body == "disable", error = OTBR_ERROR_INVALID_ARGS);
+
+    RunInMainLoop([this, &body]() {
+        otBorderAgentEphemeralKeySetEnabled(GetInstance(), body == "enable");
+        return OTBR_ERROR_NONE;
```
jwhui: The lambda can return void directly without `return OTBR_ERROR_NONE;`.

src/rest/rest_web_server.cpp
```diff
@@ -1066,6 +1084,165 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+    }
+}
+
+void RestWebServer::DeactivateEpskcKey(const Request &aRequest, Response &aResponse)
+{
+    OT_UNUSED_VARIABLE(aRequest);
+
+    RunInMainLoop([this]() {
+        otBorderAgentEphemeralKeyStop(GetInstance());
+        return OTBR_ERROR_NONE;
```
jwhui: The lambda can return void directly without `return OTBR_ERROR_NONE;`.

tests/rest/test_rest.py
```diff
@@ -485,6 +485,62 @@ def error_test(thread_num):
...
+    put_state("enable")
+    assert get_state() == "enabled"
+
+    status = get_key_status()
+    assert status["state"] == "stopped"
+
+    req = urllib.request.Request(key_url, data=b'{}', method='POST')
+    req.add_header('Content-Type', 'application/json')
+    data = json.loads(urllib.request.urlopen(req).read())
+    assert len(data["tap"]) == 9 and data["tap"].isdigit() and data["port"] > 0
```
jwhui: Suggest also adding test cases to `epskc_test()` that pass custom parameters (e.g., `{"lifetime": 60000, "port": 54321}`) to verify that custom `lifetime` and `port` are parsed and applied, as well as invalid payloads (e.g. `{"lifetime": 700000}` exceeding max timeout or non-numeric types) to verify HTTP 400 Bad Request responses.

src/rest/rest_web_server.cpp
```diff
@@ -1066,6 +1084,165 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+    std::string tap;
+    uint16_t    resultPort = 0;
+
+    if (!aRequest.body.empty())
+    {
+        VerifyOrExit(Json::JsonEpskcActivateParams(aRequest.body, lifetime, port), error = OTBR_ERROR_INVALID_ARGS);
+    }
+    VerifyOrExit(lifetime <= OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT, error = OTBR_ERROR_INVALID_ARGS);
+
+    SuccessOrExit(error = BorderAgent::CreateEphemeralKey(tap));
```
jwhui: If `CreateEphemeralKey` fails, error will be `OTBR_ERROR_INVALID_ARGS`. In the exit: handler, this gets mapped to `StatusCode::BadRequest_400`.

However, failure during internal key generation is a server-side error rather than a client error. We should ensure `CreateEphemeralKey` failure maps to `StatusCode::InternalServerError_500` (e.g. by setting `error = OTBR_ERROR_REST`).

lboue:
All 6 inline comments from @jwhui are now fixed:

1. Strict JSON type validation for lifetime and port (non-numeric values are rejected with HTTP 400),
2. ePSKc handler naming aligned consistently,
3. simplified the two void lambdas (removed redundant return),
4. CreateEphemeralKey failure now maps to server error (HTTP 500) instead of 400,
5. expanded epskc_test() with valid custom lifetime/port and invalid payload coverage (HTTP 400 checks)

You can start CI checks.

src/rest/openapi.yaml
```diff
@@ -1660,6 +1660,114 @@ paths:
...
+            schema:
+              type: object
+              properties:
+                lifetime:
+                  type: integer
+                  description: |-
+                    The key lifetime in milliseconds. Defaults to
+                    OT_BORDER_AGENT_DEFAULT_EPHEMERAL_KEY_TIMEOUT (2 minutes)
+                    when omitted or zero. Capped at
+                    OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT (10 minutes).
```
jwhui: The schema description for `lifetime` states that the timeout is *"Capped at OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT (10 minutes)"*, but `RestWebServer::ActivateEpskcKey` currently rejects values greater than 600,000 ms with an HTTP 400 Bad Request.

If rejecting values > 10 minutes is intended, consider updating this description to:
`Maximum value is OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT (10 minutes).`
Otherwise, if capping is intended, consider updating `ActivateEpskcKey` in `rest_web_server.cpp` to clamp `lifetime` instead of returning `OTBR_ERROR_INVALID_ARGS`.

src/rest/rest_web_server.cpp
```diff
@@ -1066,6 +1084,164 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+    uint32_t    lifetime = OT_BORDER_AGENT_DEFAULT_EPHEMERAL_KEY_TIMEOUT;
+    uint16_t    port     = OTBR_CONFIG_BORDER_AGENT_MESHCOP_E_UDP_PORT;
+    std::string tap;
+    uint16_t    resultPort = 0;
+
+    if (!aRequest.body.empty())
+    {
+        VerifyOrExit(Json::JsonEpskcActivateParams(aRequest.body, lifetime, port), error = OTBR_ERROR_INVALID_ARGS);
+    }
+    VerifyOrExit(lifetime <= OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT, error = OTBR_ERROR_INVALID_ARGS);
```
jwhui, line 1166: OpenThread core automatically caps timeouts to `kMaxTimeout` inside `otBorderAgentEphemeralKeyStart()`. Note that returning `400 Bad Request` here conflicts with the description in openapi.yaml ("Capped at OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT").

src/rest/rest_web_server.cpp
```diff
@@ -1066,6 +1084,164 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+
+    error = BorderAgent::CreateEphemeralKey(tap);
+    VerifyOrExit(error == OTBR_ERROR_NONE, error = OTBR_ERROR_REST);
+    otbrLogInfo("Created Ephemeral Key for REST activation");
```
jwhui: Should move `otbrLogInfo` inside `RunInMainLoop` right after `otBorderAgentEphemeralKeyStart` returns `OT_ERROR_NONE`?

Currently, this log entry is emitted before attempting `otBorderAgentEphemeralKeyStart` in `RunInMainLoop`. If `otBorderAgentEphemeralKeyStart` fails (e.g. returning `OT_ERROR_INVALID_STATE` when a session is already active), an info log will still be produced.

src/rest/json.cpp
lboue: Fixed in ab54bc78: `JsonEpskcActivateParams` now checks `cJSON_IsNumber(value)` whenever the field is present and fails validation (returns `false` → 400) if the type doesn't match, instead of silently skipping it.

src/rest/rest_web_server.cpp
lboue: Fixed in ab54bc78: the lambda now returns `void` directly.

src/rest/rest_web_server.cpp
lboue: Fixed in ab54bc78: the lambda now returns `void` directly.

tests/rest/test_rest.py
lboue: Fixed in ab54bc78: added test cases for custom `lifetime`/`port` parameters, plus invalid payloads (wrong types and `lifetime` exceeding the maximum), asserting the expected `400` responses.

src/rest/rest_web_server.cpp
lboue: Fixed in ab54bc78: a `CreateEphemeralKey` failure now maps to `OTBR_ERROR_REST` → `500 Internal Server Error` instead of `OTBR_ERROR_INVALID_ARGS` → `400`.

src/rest/openapi.yaml
lboue: Fixed in 35fdb106: updated the description to say the maximum value is rejected with `400` instead of "Capped at", matching `ActivateEpskcKey`'s actual behavior. Kept the explicit-rejection behavior (see reply on the `rest_web_server.cpp` thread) rather than relying on OpenThread core's internal clamp.

src/rest/rest_web_server.cpp
lboue: Kept the explicit `400` rejection for lifetimes above `OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT`, rather than relying on `otBorderAgentEphemeralKeyStart()`'s internal clamp, for consistency with the strict type/bounds validation already done for `lifetime`/`port` elsewhere in this handler (silently clamping an out-of-range client-supplied value felt more surprising than rejecting it). Updated the openapi.yaml description accordingly in 35fdb106 to say the value is rejected rather than capped.

src/rest/rest_web_server.cpp
lboue: Fixed in 35fdb106: moved the log line inside the `RunInMainLoop` lambda, right after `otBorderAgentEphemeralKeyStart()` returns `OT_ERROR_NONE`, so it's only emitted on an actual successful activation.

lboue:
@jwhui Addressed both rounds of review comments:

- ab54bc78: strict `cJSON_IsNumber` validation for `lifetime`/`port`, lambdas returning `void` directly, `CreateEphemeralKey` failure now maps to 500 instead of 400, and added test coverage for custom params + invalid payloads.
- 35fdb106: clarified the openapi.yaml `lifetime` description to say values above `OT_BORDER_AGENT_MAX_EPHEMERAL_KEY_TIMEOUT` are rejected with 400 (kept the explicit rejection rather than relying on OpenThread core's internal clamp, for consistency with the rest of the payload validation), and moved the "Created Ephemeral Key" log line to after `otBorderAgentEphemeralKeyStart()` actually succeeds.

All threads replied to and marked resolved. Ready for another look whenever you have time. Thanks!

tests/rest/test_rest.py
```diff
@@ -485,6 +485,76 @@ def error_test(thread_num):
...
+
+    # Invalid payloads must be rejected.
+    expect_http_error(400, lambda: post_key({"lifetime": "3600"}))
+    expect_http_error(400, lambda: post_key({"port": "49191"}))
+    expect_http_error(400, lambda: post_key({"lifetime": 4294967295}))
+
+    # Custom valid parameters should be accepted.
+    req = urllib.request.Request(
+        key_url,
+        data=json.dumps({"lifetime": 300, "port": status["port"]}).encode(),
```
jwhui: `lifetime` is specified in milliseconds (so `300` is 0.3s). To avoid any potential race condition where the ephemeral key session expires before `get_key_status()` is asserted on slow or heavily loaded CI environments, maybe use a larger duration like `30000` (30s) or `60000` (60s).

lboue: Good catch, fixed in 365bc744 — bumped to 30000 (30s).

src/rest/rest_web_server.cpp
```diff
@@ -1066,6 +1084,164 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+    }
+}
+
+void RestWebServer::DeactivateEpskcKey(const Request &aRequest, Response &aResponse) const
+{
+    OT_UNUSED_VARIABLE(aRequest);
+
+    RunInMainLoop([this]() {
+        otBorderAgentEphemeralKeyStop(GetInstance());
+    });
```
jwhui: To fix pretty check:
```suggestion
    RunInMainLoop([this]() { otBorderAgentEphemeralKeyStop(GetInstance()); });
```

src/rest/rest_web_server.cpp
```diff
@@ -1066,6 +1084,164 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+{
+    otbrError   error = OTBR_ERROR_NONE;
+    std::string body;
+
+    VerifyOrExit(Json::JsonString2String(aRequest.body, body), error = OTBR_ERROR_INVALID_ARGS);
+    VerifyOrExit(body == "enable" || body == "disable", error = OTBR_ERROR_INVALID_ARGS);
+
+    RunInMainLoop([this, &body]() {
+        otBorderAgentEphemeralKeySetEnabled(GetInstance(), body == "enable");
+    });
```
jwhui: To fix pretty check:
```suggestion
    RunInMainLoop([this, &body]() { otBorderAgentEphemeralKeySetEnabled(GetInstance(), body == "enable"); });
```

lboue:
@jwhui Could you restart the "NcpMode" CI job?

jwhui (approved):
Thanks! 👍🏼

PR #3486, opened 2026-07-30
[tests] fix ED_1 state after reset in dind_srp_tc_11 script

jwhui:
Fix an integration test failure in dind_srp_tc_11.exp caused by two issues:

1. Software reset in Step 8 clears non-persistent runtime configurations on Node 4 (ED_1), resetting routereligible back to enabled. In Step 9, when starting Thread, Node 4 promotes itself to a router, causing wait_for "state" "child" to time out and fail. Re-apply routereligible disable before starting Thread in Step 9.

2. ot-ctl dataset active -x appends \nDone to the output. Parsing the whole output caused dataset set active to pass Done on a new line to the OpenThread CLI, causing Error 35: InvalidCommand and desynchronizing the expect response matching queue. Clean dataset output to extract only the hex string.


abtink (approved):
👍

PR #3487, opened 2026-07-30
[ubus] fix two memory-safety issues

LorbusChris:
Two independent fixes in the ubus integration:

1. `UbusSetInformation` read `otLinkModeConfig` fields from an uninitialized struct when the caller supplied a partial mode string; value-initialize it.
2. `UbusMgmtset` accepted a short hex string for the extended PAN ID and read past the parsed bytes; require the full length.

Both are reachable from unprivileged ubus callers on default OpenWrt ACLs. Found while integrating the ubus surface with a Matter network manager; tested on a Turris Omnia.

### DIFF
--- a/src/openwrt/ubus/otubus.cpp
+++ b/src/openwrt/ubus/otubus.cpp
@@ -895,7 +895,7 @@ int UbusServer::UbusMgmtset(struct ubus_context      *aContext,
     {
         dataset.mComponents.mIsExtendedPanIdPresent = true;
         VerifyOrExit(Hex2Bin(blobmsg_get_string(tb[EXTPANID]), dataset.mExtendedPanId.m8,
-                             sizeof(dataset.mExtendedPanId.m8)) >= 0,
+                             sizeof(dataset.mExtendedPanId.m8)) == OT_EXT_PAN_ID_SIZE,
                      error = OT_ERROR_PARSE);
     }
     if (tb[PANID] != nullptr)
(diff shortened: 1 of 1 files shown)


src/openwrt/ubus/otubus.cpp
```diff
@@ -895,7 +895,8 @@ int UbusServer::UbusMgmtset(struct ubus_context      *aContext,
     {
         dataset.mComponents.mIsExtendedPanIdPresent = true;
         VerifyOrExit(Hex2Bin(blobmsg_get_string(tb[EXTPANID]), dataset.mExtendedPanId.m8,
-                             sizeof(dataset.mExtendedPanId.m8)) >= 0,
+                             sizeof(dataset.mExtendedPanId.m8)) ==
+                         static_cast<int>(sizeof(dataset.mExtendedPanId.m8)),
```
jwhui: We can compare directly against `OT_EXT_PAN_ID_SIZE` to match the convention used for other `Hex2Bin` checks in this file (e.g., `OT_NETWORK_KEY_SIZE`, `OT_PSKC_MAX_SIZE`, `OT_EXT_ADDRESS_SIZE`) and avoid the `static_cast<int>`.

src/openwrt/ubus/otubus.cpp
```diff
@@ -1512,13 +1513,15 @@ int UbusServer::UbusSetInformation(struct ubus_context      *aContext,
         {
             otExtendedPanId extPanId;
             char           *input = blobmsg_get_string(tb[SETNETWORK]);
-            VerifyOrExit(Hex2Bin(input, extPanId.m8, sizeof(extPanId)) >= 0, error = OT_ERROR_PARSE);
+            VerifyOrExit(Hex2Bin(input, extPanId.m8, sizeof(extPanId.m8)) ==
+                             static_cast<int>(sizeof(extPanId.m8)),
```
jwhui: Similar to above, using `OT_EXT_PAN_ID_SIZE` directly keeps this concise and consistent with the surrounding handlers.

LorbusChris:
@jwhui thank you for taking a look! I've addressed both points.

jwhui (approved):
Thanks! 👍🏼

PR #3488, opened 2026-07-30
[host] advertise mDNS under the system host name

LorbusChris:
With the OpenThread-core mDNS publisher the border router announces itself as `ot<extaddr>.local` instead of the name its administrator gave it. Set the system host name via `otMdnsSetLocalHostName()` right after instance initialisation (the only window where it works, since the mDNS module follows the infra link state). Best effort: on any failure the default stays.

RCP mode only: under NCP the name is generated by the same core code on the co-processor and reaches the host readymade through the dnssd platform, so overriding it there needs spinel support first.

### DIFF
--- a/src/host/rcp_host.cpp
+++ b/src/host/rcp_host.cpp
@@ -34,6 +34,7 @@
 #include <limits.h>
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 
 #include <openthread/backbone_router_ftd.h>
 #include <openthread/border_agent.h>
@@ -44,11 +45,15 @@
 #include <openthread/logging.h>
 #include <openthread/nat64.h>
 #include <openthread/netdiag.h>
+#include <openthread/openthread-system.h>
 #include <openthread/srp_server.h>
 #include <openthread/tasklet.h>
 #include <openthread/thread.h>
 #include <openthread/thread_ftd.h>
 #include <openthread/trel.h>
+#if OTBR_ENABLE_MDNS_OPENTHREAD
+#include <openthread/mdns.h>
+#endif
 #include <openthread/platform/logging.h>
 #include <openthread/platform/misc.h>
 #include <openthread/platform/radio.h>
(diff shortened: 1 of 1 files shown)


src/host/rcp_host.cpp
```diff
@@ -265,6 +270,63 @@ void RcpHost::Init(void)
...
+        // "ot<extaddr>": the border router should be reachable under the name
+        // its administrator gave the device. This must happen before the mDNS
+        // module is enabled, which follows the infra link, so this is the
+        // last easy moment. Best effort: keep the default on any failure.
+        // Larger than any name the kernel allows: glibc reports
+        // ENAMETOOLONG instead of truncating when the name does not fit.
+        char hostname[256];
+
+        if (gethostname(hostname, sizeof(hostname)) == 0)
+        {
```
abtink, line 285: Thank you for this PR.

While I understand the goal, I don’t think the current approach is sufficient, and I would generally advise against it.

### Key Considerations

1. **Uniqueness and Conflict Resolution**
Any host name announced over mDNS **must be unique** among all devices on the local network. If a name collision occurs, mDNS requires the host name to be changed or registration withdrawn.
 
2. **How OpenThread Native mDNS Handles Conflicts**OpenThread’s native mDNS module does **not** automatically rename entities on collision. Instead, it fires callback notifications delegating conflict handling back to the caller (in this case, the platform layer). <br><br> _Note on Border Router Context:_ The primary role of native mDNS on a Border Router is serving as an Advertising Proxy for SRP services registered by Thread devices. For SRP services, the Border Router *cannot* unilaterally change names on conflict because it does not own them—instead, it must report/reject the conflict back to the SRP client.
 
4. **Incompleteness of the Proposed Approach**: Calling `otMdnsSetLocalHostName()` at startup without registering a conflict callback leaves the system in a fragile state. If the administrator-assigned name collides with another device on the link, the local host name announcement will fail or conflict silently without recovery.

---


If you want to support administrator-defined host names, here are some options:

* **Option A: Handle Conflicts in Platform / App Layer**
Implement the conflict callback in the platform code. Upon receiving a conflict notification for the local host name, the platform logic generates an alternative unique name (may require to stop/start mDNS).
* **Option B: Native Auto-Renaming in OT Core (Conditional)**
Extend OT native mDNS to support optional auto-renaming *specifically* for local host names (while strictly keeping it disabled for SRP-delegated records). This adds complexity to the core mDNS module.
* **Option C: Append a Unique Identifier (Simplest)**
If the goal is simply to make the device recognizable while guaranteeing uniqueness, can append a unique hardware identifier to the administrator name (e.g., `<custom-name>-<extaddr>`), avoiding collision resolution logic altogether.

Both Option A and Option B require developing new conflict-handling behavior and are non-trivial tasks with many edge cases to consider.

LorbusChris: Thanks for taking the time to review this! I'd rather not take any shortcuts and would instead try to implement this in the "most correct, RFC-adhering" way. Option C sounds relatively straightforward to do, but I'd be happy to investigate A or B too if there's a preference for either. Please advise.

abtink: Thanks! I would personally lean toward Option B. While it’s likely the most complex one to implement, it provides a unified solution in the OpenThread core that benefits all platforms, not just POSIX.

LorbusChris: I'll take a look but it might be a few days until I get around to it. Converting to draft for the time being.

PR #3489, opened 2026-07-30
[host] fix Leave and migration handling on the RCP path

LorbusChris:
Fixes on the RCP host path, found while driving provision, set_pending and deprovision from the Matter network manager on OpenWrt. The NCP-mode half of the original series was split out into https://github.com/openthread/ot-br-posix/pull/3514.

- **Leave reports its result twice.** It leaves `receiveResultHere` set when handing off to `ThreadDetachGracefully()`, so the exit label posts a result in addition to the one the detach callback delivers.
- **`ScheduleMigration` passes an uninitialised dataset.** `otDatasetSendMgmtPendingSet()` serialises whatever `mComponents` says is present, so stack garbage went on the wire as extra TLVs and the leader answered Reject.
- **Leave picks its erase shortcut from the wrong state.** `mThreadEnabledState` only moves through `SetThreadEnabled()`; a stack started via ubus or ot-ctl still reads disabled while attached, so the shortcut's erase failed silently and the caller was told success. It branches on the device role now.
- **A dataset erase is not reported to state-change subscribers.** They kept serving the erased dataset from their caches.
- **Leave disables Thread before erasing.** A graceful detach ends in `Mle::Stop()`, so the role is already disabled in the callback and the erase succeeds, but the Thread interface is still up and MLE's socket open; Leave erased in that half-way state, left `mThreadEnabledState` at enabled and ignored the erase result. It now brings the stack down the way `SetThreadEnabled(false)` does, walks the state machine to disabled, and propagates a failed erase.
- **Result posts tolerate a null receiver** (`SafeInvoke` in Join, Leave and ScheduleMigration).

`TEST(RcpHostApi, StateChangesCorrectlyAfterLeave)` covers leaving with erase while attached, after an out-of-band start, and enabled-without-a-dataset with a null receiver; `StateChangesCorrectlyAfterJoin` gains a Join while attached as leader with a different dataset.

**Testing:** running on a Turris Omnia (armv7, TurrisOS 9.1) with an EFR32 RCP, as part of an openthread-br 2026.08.0 build, across repeated provision / set_pending / deprovision cycles driven from Matter. Unit tests pass; the new Leave coverage fails without the fixes and passes with them. Also built standalone against current main.

### DIFF
--- a/src/host/rcp_host.cpp
+++ b/src/host/rcp_host.cpp
@@ -561,7 +561,12 @@ void RcpHost::Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const A
     if (GetDeviceRole() != OT_DEVICE_ROLE_DISABLED)
     {
         ThreadDetachGracefully([aActiveOpDatasetTlvs, aReceiver, this] {
-            ConditionalErasePersistentInfo(true);
+            // A graceful detach ends in Mle::Stop(), so the device role is
+            // already disabled here and the erase succeeds; the recursive
+            // Join() below then takes the direct path. Best-effort, as before
+            // this helper reported its result: the join overwrites the
+            // dataset right after, so a failed erase changes nothing.
+            IgnoreError(ConditionalErasePersistentInfo(true));
             Join(aActiveOpDatasetTlvs, aReceiver);
         });
         receiveResultHere = false;
@@ -586,7 +591,7 @@ void RcpHost::Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const A
 exit:
     if (receiveResultHere)
     {
-        mTaskRunner.Post([aReceiver, error, errorMsg](void) { aReceiver(error, errorMsg); });
+        mTaskRunner.Post([aReceiver, error, errorMsg](void) { SafeInvoke(aReceiver, error, errorMsg); });
     }
 }
 
@@ -600,33 +605,47 @@ void RcpHost::Leave(bool aEraseDataset, const AsyncResultReceiver &aReceiver)
     VerifyOrExit(mThreadEnabledState != ThreadEnabledState::kStateDisabling, error = OT_ERROR_BUSY,
                  errorMsg = "Thread is disabling");
 
-    if (mThreadEnabledState == ThreadEnabledState::kStateDisabled)
+    // Branch on the actual device role, not mThreadEnabledState: when the
+    // stack was started outside SetThreadEnabled() (ubus threadstart, ot-ctl)
+    // the state machine still says disabled, and taking the shortcut then
+    // makes otInstanceErasePersistentInfo() fail silently -- the caller is
+    // told the dataset was erased when nothing happened. The role reflects
+    // what the stack is really doing.
+    if (GetDeviceRole() == OT_DEVICE_ROLE_DISABLED)
     {
-        ConditionalErasePersistentInfo(aEraseDataset);
+        // Keep the state machine in step with the stack: SetThreadEnabled(true)
+        // without an active dataset leaves kStateEnabled while the role stays
+        // disabled, and a Leave through this shortcut would preserve the lie.
+        if (mThreadEnabledState != ThreadEnabledState::kStateDisabled)
+        {
+            UpdateThreadEnabledState(ThreadEnabledState::kStateDisabled);
+        }
+        SuccessOrExit(error    = ConditionalErasePersistentInfo(aEraseDataset),
+                      errorMsg = "Failed to erase persistent info");
         ExitNow();
     }
 
-    ThreadDetachGracefully([aEraseDataset, aReceiver, this] {
-        ConditionalErasePersistentInfo(aEraseDataset);
-        if (aReceiver)
-        {
-            aReceiver(OT_ERROR_NONE, "");
-        }
-    });
+    UpdateThreadEnabledState(ThreadEnabledState::kStateDisabling);
+
+    ThreadDetachGracefully([aEraseDataset, aReceiver, this] { LeaveAfterDetach(aEraseDataset, aReceiver); });
+    receiveResultHere = false;
 
 exit:
     if (receiveResultHere)
     {
-        mTaskRunner.Post([aReceiver, error, errorMsg](void) { aReceiver(error, errorMsg); });
+        mTaskRunner.Post([aReceiver, error, errorMsg](void) { SafeInvoke(aReceiver, error, errorMsg); });
     }
 }
 
 void RcpHost::ScheduleMigration(const otOperationalDatasetTlvs &aPendingOpDatasetTlvs,
                                 const AsyncResultReceiver       aReceiver)
 {
-    otError              error = OT_ERROR_NONE;
-    std::string          errorMsg;
-    otOperationalDataset emptyDataset;
+    otError     error = OT_ERROR_NONE;
+    std::string errorMsg;
+    // Value-initialise: otDatasetSendMgmtPendingSet() serialises whichever
+    // fields mComponents says are present, so leaving this indeterminate puts
+    // garbage TLVs on the wire and the leader answers Reject.
+    otOperationalDataset emptyDataset{};
 
     VerifyOrExit(mInstance != nullptr, error = OT_ERROR_INVALID_STATE, errorMsg = "OT is not initialized");
 
@@ -647,7 +666,7 @@ void RcpHost::ScheduleMigration(const otOperationalDatasetTlvs &aPendingOpDatase
 exit:
     if (error != OT_ERROR_NONE)
     {
-        mTaskRunner.Post([aReceiver, error, errorMsg](void) { aReceiver(error, errorMsg); });
+        mTaskRunner.Post([aReceiver, error, errorMsg](void) { SafeInvoke(aReceiver, error, errorMsg); });
     }
     else
     {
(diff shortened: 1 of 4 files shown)


src/host/rcp_host.cpp
```diff
@@ -798,10 +808,19 @@ void RcpHost::ThreadDetachGracefullyCallback(void)
 
 void RcpHost::ConditionalErasePersistentInfo(bool aErase)
```
jwhui: Should `ConditionalErasePersistentInfo(bool aErase)` return `otError`? This allows callers to propagate the result back to the `AsyncResultReceiver` rather than silently ignoring errors.

src/host/rcp_host.cpp
jwhui: `ThreadDetachGracefully()` gracefully detaches the device from the network (role becomes `OT_DEVICE_ROLE_DETACHED`), but it does not disable the Thread stack (`otThreadSetEnabled(false)` is not called).

Because `otInstanceErasePersistentInfo()` requires MLE to be disabled, calling `ConditionalErasePersistentInfo()` here will fail with `OT_ERROR_INVALID_STATE`. As a result:
1. `otInstanceErasePersistentInfo()` fails to erase the dataset.
2. The Thread stack remains enabled in `OT_DEVICE_ROLE_DETACHED`, causing OpenThread to immediately attempt to re-attach to the network with the existing dataset.
3. `mThreadEnabledState` is never transitioned to `kStateDisabled`.

To fix this, we should set `mThreadEnabledState` to `kStateDisabling`, disable Thread and IPv6 after detach completes (similar to `DisableThreadAfterDetach()`), transition state to `kStateDisabled`, and then erase persistent storage.

src/host/rcp_host.cpp
```diff
@@ -600,7 +600,13 @@ void RcpHost::Leave(bool aEraseDataset, const AsyncResultReceiver &aReceiver)
...
                  errorMsg = "Thread is disabling");
 
-    if (mThreadEnabledState == ThreadEnabledState::kStateDisabled)
+    // Branch on the actual device role, not mThreadEnabledState: when the
+    // stack was started outside SetThreadEnabled() (ubus threadstart, ot-ctl)
+    // the state machine still says disabled, and taking the shortcut then
+    // makes otInstanceErasePersistentInfo() fail silently -- the caller is
+    // told the dataset was erased when nothing happened. The role reflects
+    // what the stack is really doing.
+    if (otThreadGetDeviceRole(mInstance) == OT_DEVICE_ROLE_DISABLED)
```
jwhui: Could you also add unit test coverage in `tests/gtest/test_rcp_host_api.cpp` for the fixes in this PR?
Specifically in `TEST(RcpHostApi, StateChangesCorrectlyAfterLeave)`:
1. **`Leave(aEraseDataset = true)` while attached:**
   Test that calling `host.Leave(true, receiver)` while in `OT_DEVICE_ROLE_LEADER` (or `ROUTER`):
   - Successfully disables the stack (`host.GetDeviceRole() == OT_DEVICE_ROLE_DISABLED`).
   - Erases persistent storage (`otDatasetGetActive(...) == OT_ERROR_NOT_FOUND`).
   - Triggers the `ActiveDatasetChangeHandler` callback with an empty dataset (`aDatasetTlvs.mLength == 0`).
2. **`Leave()` when started out-of-band:**
   Bring up the stack directly via `otThreadSetEnabled(instance, true)` (without calling `host.SetThreadEnabled(true)`), then verify `host.Leave(true, receiver)` gracefully detaches, disables the stack, and erases the dataset.

src/host/rcp_host.cpp
LorbusChris: Done, it returns otError now and both callers propagate a failed erase to the receiver.

src/host/rcp_host.cpp
LorbusChris: Thanks. The detach callback now disables Thread and IPv6 like DisableThreadAfterDetach(), moves the state machine through kStateDisabling to kStateDisabled, and erases only then. Covered by the new test.

src/host/rcp_host.cpp
LorbusChris: Added both cases to StateChangesCorrectlyAfterLeave: erasing while attached (stack ends up disabled, dataset NOT_FOUND, the dataset handler fires with an empty dataset) and Leave after an out-of-band otThreadSetEnabled(true). The existing Leave-while-enabled case needed its wait fixed too: it only passed because Leave used to answer before the detach finished, which is one of the bugs this series fixes. It now waits for the detach and asserts the result arrived. The extended test fails without the fixes and passes with them.

src/host/rcp_host.cpp
```diff
@@ -600,19 +600,23 @@ void RcpHost::Leave(bool aEraseDataset, const AsyncResultReceiver &aReceiver)
...
     {
-        ConditionalErasePersistentInfo(aEraseDataset);
+        VerifyOrExit(ConditionalErasePersistentInfo(aEraseDataset) == OT_ERROR_NONE, error = OT_ERROR_FAILED,
```
jwhui: nit: use `SuccessOrExit` instead?

LorbusChris: Done

src/host/rcp_host.cpp
jwhui: Should use `SafeInvoke` here (similar to `SetThreadEnabled` and `LeaveAfterDetach`)?

If `aReceiver` is `nullptr` or empty (e.g. `host.Leave(true, nullptr)`), calling `aReceiver(error, errorMsg)` here directly will throw `std::bad_function_call`.

src/host/rcp_host.cpp
```diff
@@ -600,19 +603,23 @@ void RcpHost::Leave(bool aEraseDataset, const AsyncResultReceiver &aReceiver)
...
+    // Branch on the actual device role, not mThreadEnabledState: when the
+    // stack was started outside SetThreadEnabled() (ubus threadstart, ot-ctl)
+    // the state machine still says disabled, and taking the shortcut then
+    // makes otInstanceErasePersistentInfo() fail silently -- the caller is
+    // told the dataset was erased when nothing happened. The role reflects
+    // what the stack is really doing.
+    if (otThreadGetDeviceRole(mInstance) == OT_DEVICE_ROLE_DISABLED)
     {
-        ConditionalErasePersistentInfo(aEraseDataset);
+        SuccessOrExit(error    = ConditionalErasePersistentInfo(aEraseDataset),
```
jwhui, line 623: Should we update `mThreadEnabledState` to `kStateDisabled` here as well so the host state is consistent regardless of which path `Leave()` takes?

If `SetThreadEnabled(true)` was previously called without an active dataset, `mThreadEnabledState` would be `ThreadEnabledState::kStateEnabled` while the OpenThread device role remained `OT_DEVICE_ROLE_DISABLED`. When `Leave()` takes this shortcut, `mThreadEnabledState` remains in `kStateEnabled`.

src/host/rcp_host.cpp
```diff
@@ -561,7 +561,10 @@ void RcpHost::Join(const otOperationalDatasetTlvs &aActiveOpDatasetTlvs, const A
     if (GetDeviceRole() != OT_DEVICE_ROLE_DISABLED)
     {
         ThreadDetachGracefully([aActiveOpDatasetTlvs, aReceiver, this] {
-            ConditionalErasePersistentInfo(true);
+            // Best-effort, as before this helper reported its result: the join
+            // that follows overwrites the dataset anyway, and the helper logs
+            // the failure itself.
+            IgnoreError(ConditionalErasePersistentInfo(true));
```
jwhui, line 569: Here the device is in `OT_DEVICE_ROLE_DETACHED` (Thread stack is still enabled), so `ConditionalErasePersistentInfo(true)` will fail with `OT_ERROR_INVALID_STATE` and log a warning.

While keeping this as `IgnoreError` matches the previous behavior for now, `Join()` will eventually need a sequence similar to `LeaveAfterDetach` (calling `otThreadSetEnabled(mInstance, false)` before erasing and re-joining).

src/host/rcp_host.cpp
LorbusChris: Done, the exit post uses `SafeInvoke` now, and the new test case covers `Leave(true, nullptr)` through the shortcut.

src/host/rcp_host.cpp
LorbusChris: Right. The shortcut now moves the state machine to `kStateDisabled` (guarded so an already-disabled state doesn't re-notify), and the test covers that exact scenario: `SetThreadEnabled(true)` without a dataset, then `Leave`, asserting the state callback lands on disabled.

src/host/rcp_host.cpp
LorbusChris: Agreed. Kept as `IgnoreError` for now to preserve behavior, and the comment now says outright that the erase fails with `INVALID_STATE` there, with a TODO for the `LeaveAfterDetach`-style disable-then-erase sequence. Happy to do that as a follow-up PR if you'd like.

jwhui: I think there is a recursion loop issue here:

`ThreadDetachGracefully()` brings the role to `OT_DEVICE_ROLE_DETACHED` but does not disable the Thread stack. When the callback runs:
1. `ConditionalErasePersistentInfo(true)` fails with `OT_ERROR_INVALID_STATE` (as noted in the comment).
2. The recursive `Join(aActiveOpDatasetTlvs, aReceiver)` call sees `GetDeviceRole() == OT_DEVICE_ROLE_DETACHED`.
3. Because `OT_DEVICE_ROLE_DETACHED != OT_DEVICE_ROLE_DISABLED` is `true`, `Join()` re-enters this branch and invokes `ThreadDetachGracefully` again in an infinite loop without ever proceeding to set the new dataset and bring up Thread.

To fix this and fulfill the TODO, please disable Thread and IPv6 before erasing by calling `otIp6SetEnabled(mInstance, false)`.

src/host/rcp_host.cpp
jwhui: For consistency with `Leave` (line 630) and `SetThreadEnabled` (line 711), please use `SafeInvoke` here so passing a `nullptr` receiver does not throw `std::bad_function_call` when validation fails.

src/host/rcp_host.cpp
```diff
@@ -600,33 +605,47 @@ void RcpHost::Leave(bool aEraseDataset, const AsyncResultReceiver &aReceiver)
...
                  errorMsg = "Thread is disabling");
 
-    if (mThreadEnabledState == ThreadEnabledState::kStateDisabled)
+    // Branch on the actual device role, not mThreadEnabledState: when the
+    // stack was started outside SetThreadEnabled() (ubus threadstart, ot-ctl)
+    // the state machine still says disabled, and taking the shortcut then
+    // makes otInstanceErasePersistentInfo() fail silently -- the caller is
+    // told the dataset was erased when nothing happened. The role reflects
+    // what the stack is really doing.
+    if (otThreadGetDeviceRole(mInstance) == OT_DEVICE_ROLE_DISABLED)
```
jwhui: nit: Use `GetDeviceRole()` for consistency.

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -254,12 +254,77 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterLeave)
                          [&host]() { return host.GetDeviceRole() != OT_DEVICE_ROLE_DETACHED; });
     EXPECT_EQ(host.GetDeviceRole(), OT_DEVICE_ROLE_LEADER);
     host.Leave(/* aEraseDataset */ false, receiver);
-    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 0, [&resultReceived]() { return resultReceived; });
+    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 1, [&resultReceived]() { return resultReceived; });
```
jwhui, line 257: Add a test case in `TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)` that calls `host.Join(newDatasetTlvs, receiver)` while already attached (`OT_DEVICE_ROLE_LEADER`) with a different dataset, verifying that it successfully detaches, disables, erases, sets the new dataset, and re-attaches to the new network.

src/host/rcp_host.cpp
LorbusChris: Done, the exit posts of ScheduleMigration and Join both use SafeInvoke now.

src/host/rcp_host.cpp
LorbusChris: Done

tests/gtest/test_rcp_host_api.cpp
LorbusChris: Added

src/host/rcp_host.cpp
LorbusChris: My earlier reply saying the erase fails with `INVALID_STATE` was wrong, sorry for the confusion. A graceful detach ends in `Mle::Stop()`, so the detach callback already runs with the role disabled, the erase succeeds, and the recursive `Join()` takes the direct path; step 7 of `StateChangesCorrectlyAfterJoin` now covers exactly this, and the comment says what actually happens. What a graceful detach does leave behind is the interface: `otIp6IsEnabled()` is still true in the detach callback. That is why `LeaveAfterDetach()` disables Thread and IPv6 before erasing, so Leave ends in the same state `SetThreadEnabled(false)` does.

jwhui (approved):
Thanks! 👍🏼

jwhui:
@LorbusChris , can you help squash commits and update commit message?

PR #3490, opened 2026-07-31
[tests] fix expect matching order in dind_trel_tc_5.exp

jwhui:
Fix an integration test failure in dind_trel_tc_5.exp where Step 3 failed with Error: Node 3 did not discover Node 2!.

In Step 3, Node 3 executes discover. The CLI output table lists otbr-net2 on the first row and otbr-net1 on the second row. Because -re {otbr-net1} was listed before -re {otbr-net2} in the expect block, Expect matched otbr-net1 first on the second row. Matching otbr-net1 advanced the buffer pointer past the second row and discarded the preceding text containing otbr-net2. Consequently, found_node2 remained false when exp_continue executed.

Fix this issue in Step 2 and Step 3 by capturing the complete discovery scan output via expect_line "Done" and verifying presence of network names in the full output string.


abtink (approved):
👍

PR #3497, opened 2026-08-04
[ci] package and upload source tarball in monthly release

### LINKED ISSUE #3496: GitHub-generated source archives do not include Git submodules
@stintel noticed this while working on openwrt/packages#29784 and mentioned it in the commit description, but unfortunately did not have time to investigate/report the issue upstream.

Unfortunately, automatically generated GitHub source archives do not include the contents of Git submodules. This is a long-standing GitHub issue (dear-github/dear-github#214), and it does not affect only this project, but also e.g. heiher/natmap#6 and many others.

The issue persists even with the version released yesterday. Just download either of the automatically generated source archives (.tar.gz or .zip) from the release and extract/open it. You can see that the submodule directory located at: ``third_party/openthread/repo`` is empty.

Therefore, it is not possible to build ot-br-posix from the source archives provided by GitHub, as the required submodule sources are missing.

**I think it would be appropriate to provide a proper release tarball as a release asset that includes the required Git submodules, so that the released version can be built directly from the provided source archive.**

jwhui:
GitHub's automatically generated source archives (.tar.gz and .zip) do not include Git submodules, leaving third_party/ submodules empty and making it impossible to build ot-br-posix directly from release archives.

This commit updates monthly-release.yml to checkout submodules recursively and package the source tree into a release tarball asset, which is attached to the GitHub release via gh release create.

Fixes #3496

### DIFF
--- a/.github/workflows/monthly-release.yml
+++ b/.github/workflows/monthly-release.yml
@@ -71,6 +71,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
+          submodules: recursive
           persist-credentials: false
 
       - name: Generate CalVer Tag Name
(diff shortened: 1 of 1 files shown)


abtink (approved):
👍

.github/workflows/monthly-release.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
BKPepe, line 72: Is there a particular reason for pinning actions/checkout to a specific commit instead of following the v7 tag? Is this intended as a supply-chain hardening measure (SHA pinning), or just a leftover?

LorbusChris: This is the best practice afaik, and dependabot handles updating these.

jwhui: Yes, pinning GitHub Actions to full commit SHAs (with human-readable version comments like # v7.0.1) is done intentionally as a supply-chain hardening best practice to ensure immutability and protect against upstream tag mutation.

Dependabot is configured for this repository to monitor and submit PRs for action updates automatically, keeping the pinned SHAs up to date across all workflows.

PR #3508, opened 2026-08-10
[tests] fix flaky timeout in RcpHostApi.StateChangesCorrectlyAfterJoin

### LINKED ISSUE #3507: RcpHostApi.StateChangesCorrectlyAfterJoin fails intermittently due to non-deterministic timeout in MainloopProcessUntil()
## Describe the bug

`RcpHostApi.StateChangesCorrectlyAfterJoin` fails in CI. The test remains in `OT_DEVICE_ROLE_DETACHED` after the second `Join()` instead of reaching `OT_DEVICE_ROLE_LEADER`.

The failure was observed in the `rest-check` job of `ot-br-posix`.

## To Reproduce

1. **Git commit id:** `7c136fc`
2. **IEEE 802.15.4 hardware platform:** N/A — this is a host-side GTest using the fake platform.
3. **Build steps:** Run the `rest-check` test suite / CTest.
4. **Network topology:** N/A — the failing test uses the fake platform and does not require a physical Thread network.

## Expected behavior

`RcpHostApi.StateChangesCorrectlyAfterJoin` should successfully complete the second `Join()` and transition the device to `OT_DEVICE_ROLE_LEADER`.

## Console/log output

The CI run reports:

```text
Test #77: RcpHostApi.StateChangesCorrectlyAfterJoin ***Failed

Expected:
  errorMsg = "Join succeeded"
  host.GetDeviceRole() = OT_DEVICE_ROLE_LEADER (4)

Actual:
  errorMsg = ""
  host.GetDeviceRole() = OT_DEVICE_ROLE_DETACHED (1)

Subsequent Join():
  error = OT_ERROR_ABORT (11)
  errorMsg = "Aborted by leave/disable operation"

98% tests passed, 1 tests failed out of 57

### PR DESCRIPTION
**Pull request by lboue**
## Context

Fixes #3507.

`RcpHostApi.StateChangesCorrectlyAfterJoin` occasionally fails in CI (observed on the `rest-check` job, which builds with `OTBR_COVERAGE=1`) with results that look "shifted by one step": the device role is still `DETACHED` instead of `LEADER`, and the previous Join's abort message ends up on the wrong assertion.

## Root cause

`MainloopProcessUntil()` in `tests/gtest/test_rcp_host_api.cpp` bounds its wait loop by comparing only the whole-second component of `gettimeofday()`:

```cpp
if (now.tv_sec - startTime.tv_sec > aTimeoutSec)
```

For `aTimeoutSec = 0`, this gives a **non-deterministic real-time budget of anywhere between ~0ms and ~999ms**, depending on how close the wait happens to start to a wall-clock second boundary.

Step 3 of `StateChangesCorrectlyAfterJoin` uses this `0s` wait for two consecutive `Join()` calls that go through the full Detach → re-Join → Leader-formation sequence — unlike the other `0s` waits in the same file, which only wait on an immediate/synchronous rejection. On a loaded or coverage-instrumented (gcov) runner, an unlucky start time can leave too little real budget for that sequence to finish, so the test observes stale/shifted state.

I also checked whether the OpenThread submodule bump around the same PR (`211267d` → `cff0c6a`) could be responsible: that diff only touches `src/core/mac/*` (TX operation ordering, key management) and adds Clang lifetime annotations — nothing in MLE/Thread-role/attach code — so it looks unrelated to this test's failure mode.

## Fix

- Switch `MainloopProcessUntil()` to a monotonic, sub-second-precision deadline (`std::chrono::steady_clock`) instead of the truncated `tv_sec` comparison.
- Bump the timeout for the Join-to-Leader wait from `0s` to `1s`, matching the pattern already used for equivalent leader-formation waits elsewhere in the same file (`StateChangesCorrectlyAfterLeave`, `StateChangesCorrectlyAfterScheduleMigration`).

## Testing

I wasn't able to run `ctest` locally (no system GTest CMake package installed in this environment), so this is marked as **draft** pending a green run of `RcpHostApi.StateChangesCorrectlyAfterJoin` in CI. Happy to iterate based on CI results.

### DIFF
--- a/tests/gtest/test_rcp_host_api.cpp
+++ b/tests/gtest/test_rcp_host_api.cpp
@@ -377,15 +377,9 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
     host.Join(datasetTlvs, receiver_);
     host.Join(datasetTlvs, receiver);
 
-    // Unlike the other `Join`/`Leave` calls above, this one goes through the full
-    // Detach -> re-Join -> Leader-formation sequence, so it needs more than an
-    // instant budget to complete (see the other leader-formation waits in this
-    // file, e.g. `StateChangesCorrectlyAfterLeave` and
-    // `StateChangesCorrectlyAfterScheduleMigration`, which use 1s for the same reason).
-    // This can't be reverted to 0s even now that MainloopProcessUntil() runs
-    // Update()/Process() before checking the deadline: with aTimeoutSec == 0
-    // the loop still only gets a single processing pass, which isn't enough
-    // for this multi-step async sequence to reach LEADER.
+    // Unlike the single-pass 0s waits above, this step undergoes a full
+    // Detach -> re-Join -> Leader-formation sequence which requires multiple
+    // mainloop iterations to complete.
     MainloopProcessUntil(mainloop, /* aTimeoutSec */ 1,
                          [&resultReceived, &resultReceived_]() { return resultReceived && resultReceived_; });
     EXPECT_EQ(error_, OT_ERROR_ABORT);
--- a/tests/gtest/test_rcp_host_api.cpp
+++ b/tests/gtest/test_rcp_host_api.cpp
@@ -408,7 +408,9 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
     host.Join(datasetTlvs, receiver_);
     host.SetThreadEnabled(false, receiver);
 
-    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 0,
+    // Like the step-3 wait above, this Join races a Disable that triggers
+    // its own Detach sequence, so it needs multiple mainloop iterations.
+    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 1,
                          [&resultReceived, &resultReceived_]() { return resultReceived && resultReceived_; });
     EXPECT_EQ(error_, OT_ERROR_BUSY);
     EXPECT_STREQ(errorMsg_.c_str(), "Thread is disabling");
@@ -430,7 +432,9 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
     host.Join(datasetTlvs, receiver_);
     host.SetThreadEnabled(false, receiver);
 
-    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 0,
+    // Same Detach -> re-Join -> Leader-formation sequence as step 3, so it
+    // needs the same 1s budget rather than a single mainloop pass.
+    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 1,
                          [&resultReceived, &resultReceived_]() { return resultReceived && resultReceived_; });
     EXPECT_EQ(error_, OT_ERROR_ABORT);
     EXPECT_STREQ(errorMsg_.c_str(), "Aborted by leave/disable operation");
--- a/tests/gtest/test_rcp_host_api.cpp
+++ b/tests/gtest/test_rcp_host_api.cpp
@@ -29,6 +29,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include <algorithm>
 #include <chrono>
 
 #include <openthread/dataset.h>
@@ -49,7 +50,13 @@ static void MainloopProcessUntil(otbr::MainloopContext    &aMainloop,
     // `aTimeoutSec` of 0 could get anywhere between ~0ms and ~999ms of real
     // budget depending on where `startTime` lands relative to the next
     // second boundary, making the wait non-deterministic.
-    const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(aTimeoutSec);
+    //
+    // Clamp to at least 1s: with aTimeoutSec == 0, `deadline` would be set to
+    // `now()`, and since some time always elapses before the first check
+    // inside the loop, the loop would break before ever calling Update()/
+    // Process() even once. The loop still exits as soon as `aCondition()` is
+    // met, so this doesn't add delay to tests that already pass.
+    const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(std::max(aTimeoutSec, 1u));
 
     while (!aCondition())
     {
--- a/tests/gtest/test_rcp_host_api.cpp
+++ b/tests/gtest/test_rcp_host_api.cpp
@@ -29,7 +29,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
-#include <sys/time.h>
+#include <chrono>
 
 #include <openthread/dataset.h>
 #include <openthread/dataset_ftd.h>
@@ -44,15 +44,16 @@ static void MainloopProcessUntil(otbr::MainloopContext    &aMainloop,
                                  uint32_t                  aTimeoutSec,
                                  std::function<bool(void)> aCondition)
 {
-    timeval startTime;
-    timeval now;
-    gettimeofday(&startTime, nullptr);
+    // Use a monotonic clock with sub-second precision. Comparing only whole
+    // seconds (e.g. via `timeval::tv_sec`) truncates the elapsed time, so an
+    // `aTimeoutSec` of 0 could get anywhere between ~0ms and ~999ms of real
+    // budget depending on where `startTime` lands relative to the next
+    // second boundary, making the wait non-deterministic.
+    const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(aTimeoutSec);
 
     while (!aCondition())
     {
-        gettimeofday(&now, nullptr);
-        // Simply compare the second. We don't need high precision here.
-        if (now.tv_sec - startTime.tv_sec > aTimeoutSec)
+        if (std::chrono::steady_clock::now() > deadline)
         {
             break;
         }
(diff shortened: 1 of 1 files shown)


tests/gtest/test_rcp_host_api.cpp
```diff
@@ -44,15 +44,16 @@ static void MainloopProcessUntil(otbr::MainloopContext    &aMainloop,
...
 {
-    timeval startTime;
-    timeval now;
-    gettimeofday(&startTime, nullptr);
+    // Use a monotonic clock with sub-second precision. Comparing only whole
+    // seconds (e.g. via `timeval::tv_sec`) truncates the elapsed time, so an
+    // `aTimeoutSec` of 0 could get anywhere between ~0ms and ~999ms of real
+    // budget depending on where `startTime` lands relative to the next
+    // second boundary, making the wait non-deterministic.
+    const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(aTimeoutSec);
```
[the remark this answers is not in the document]
lboue: Good catch, thanks — fixed in 1bc166ba by clamping to `std::max(aTimeoutSec, 1u)` as suggested.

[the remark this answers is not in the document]
lboue: [1bc166b](https://github.com/openthread/ot-br-posix/pull/3508/commits/1bc166baa184c7ceaa0d871a8151db2c346143d3)

lboue:
## Local verification

CI's main workflows (Build, Docker, Border Router, ...) are currently pending maintainer approval to run (`action_required`), so in the meantime I verified locally:

- Built GoogleTest/GMock from source and reconfigured with `-DBUILD_TESTING=ON`, then built only the `otbr-gtest-host-api` target (contains `test_rcp_host_api.cpp`).
- Full suite (8 tests in `RcpHostApi`): **PASS**, ~17ms.
- `RcpHostApi.StateChangesCorrectlyAfterJoin` alone, `--gtest_repeat=500`: **500/500 PASS**, 0 failures.
- Full suite, `--gtest_repeat=100`: **100/100 PASS**, 0 failures — confirms the `std::max(aTimeoutSec, 1u)` clamp doesn't regress the other tests in this file that rely on `aTimeoutSec = 0`.

(The separate `zizmor-output` failure on this PR is unrelated to this change — all 17 findings are in `.github/workflows/*.yml` files this PR doesn't touch.)

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -44,15 +45,22 @@ static void MainloopProcessUntil(otbr::MainloopContext    &aMainloop,
...
+    // `aTimeoutSec` of 0 could get anywhere between ~0ms and ~999ms of real
+    // budget depending on where `startTime` lands relative to the next
+    // second boundary, making the wait non-deterministic.
+    //
+    // Clamp to at least 1s: with aTimeoutSec == 0, `deadline` would be set to
+    // `now()`, and since some time always elapses before the first check
+    // inside the loop, the loop would break before ever calling Update()/
+    // Process() even once. The loop still exits as soon as `aCondition()` is
+    // met, so this doesn't add delay to tests that already pass.
+    const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(std::max(aTimeoutSec, 1u));
```
jwhui: The need for the clamp seems to come from checking `now > deadline` before executing `Update()` and `Process()`. What if `MainloopProcessUntil()` runs `Update()` and `Process()` first before checking the elapsed timeout? A 0s timeout will process pending tasks and then terminate if the condition is not met.

lboue: Thanks for the suggestion, this makes sense.

I updated the helper so it processes events before checking the timeout deadline, while keeping steady_clock for monotonic and sub-second timing. This preserves the expected 0s behavior (one processing pass, then stop if the condition is still unmet) without forcing a 1s clamp.

I pushed the change in commit df2ebe39.

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -44,21 +44,22 @@ static void MainloopProcessUntil(otbr::MainloopContext    &aMainloop,
...
                                  std::function<bool(void)> aCondition)
 {
-    timeval startTime;
-    timeval now;
-    gettimeofday(&startTime, nullptr);
+    // Use a monotonic clock with sub-second precision. Comparing only whole
+    // seconds (e.g. via `timeval::tv_sec`) truncates the elapsed time, so an
+    // `aTimeoutSec` of 0 could get anywhere between ~0ms and ~999ms of real
+    // budget depending on where `startTime` lands relative to the next
+    // second boundary, making the wait non-deterministic.
```
jwhui: Prefer to keep comments scoped to existing code, rather than pre-existing code.
```suggestion
    // Use a monotonic clock with sub-second precision.
```

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -377,7 +378,12 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
     host.Join(datasetTlvs, receiver_);
     host.Join(datasetTlvs, receiver);
 
-    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 0,
+    // Unlike the other `Join`/`Leave` calls above, this one goes through the full
+    // Detach -> re-Join -> Leader-formation sequence, so it needs more than an
+    // instant budget to complete (see the other leader-formation waits in this
+    // file, e.g. `StateChangesCorrectlyAfterLeave` and
+    // `StateChangesCorrectlyAfterScheduleMigration`, which use 1s for the same reason).
+    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 1,
```
jwhui, line 383: With the change to moving `Update` and `Process` before the timeout check, can we revert this change?

tests/gtest/test_rcp_host_api.cpp
lboue: Trimmed to only describe the new code, as suggested. Fixed in b1feb151.

tests/gtest/test_rcp_host_api.cpp
lboue: Good question. It can't be fully reverted: even with `Update()`/`Process()` now running before the deadline check, an `aTimeoutSec` of 0 still only allows a single processing pass through the loop (the deadline is `now()` at entry, so the check after the first pass breaks out). The Detach -> re-Join -> Leader-formation sequence needs several iterations of the event loop to complete, so it still needs the 1s budget like the other leader-formation waits in this file.

Added a comment explaining this at the call site in b1feb151 so it doesn't look like an oversight.

tests/gtest/test_rcp_host_api.cpp
```diff
@@ -377,7 +377,16 @@ TEST(RcpHostApi, StateChangesCorrectlyAfterJoin)
...
-    MainloopProcessUntil(mainloop, /* aTimeoutSec */ 0,
+    // Unlike the other `Join`/`Leave` calls above, this one goes through the full
+    // Detach -> re-Join -> Leader-formation sequence, so it needs more than an
+    // instant budget to complete (see the other leader-formation waits in this
+    // file, e.g. `StateChangesCorrectlyAfterLeave` and
+    // `StateChangesCorrectlyAfterScheduleMigration`, which use 1s for the same reason).
+    // This can't be reverted to 0s even now that MainloopProcessUntil() runs
+    // Update()/Process() before checking the deadline: with aTimeoutSec == 0
+    // the loop still only gets a single processing pass, which isn't enough
+    // for this multi-step async sequence to reach LEADER.
```
jwhui: Suggest streamlining the comment to focus on the technical reason for future readers without referencing PR review history.
```suggestion
    // Unlike the single-pass 0s waits above, this step undergoes a full
    // Detach -> re-Join -> Leader-formation sequence which requires multiple
    // mainloop iterations to complete.
```

lboue: Trimmed to only describe the new code, as suggested. Fixed by [7f8f182](https://github.com/openthread/ot-br-posix/pull/3508/commits/7f8f1821ecd9e407e933d07c48e3de9368b7fbd1)

lboue:
## Follow-up: same fix applied to steps 5 and 6

**What broke:** the `check`, `rest-check`, and `docker-check (--mdnsresponder)` jobs failed on `RcpHostApi.StateChangesCorrectlyAfterJoin`, at [lines 413–416](https://github.com/openthread/ot-br-posix/blob/9a5828ae9f59bbf1edd7250e46f5c060520739fe/tests/gtest/test_rcp_host_api.cpp#L413-L416) and [435–438](https://github.com/openthread/ot-br-posix/blob/9a5828ae9f59bbf1edd7250e46f5c060520739fe/tests/gtest/test_rcp_host_api.cpp#L435-L438).

**Why:** steps 5 and 6 race a `Join` against a `Disable`, which triggers its own Detach sequence — the same pattern as step 3 above them (step 6's second block is a full Detach → re-Join → Leader-formation sequence, identical to step 3). Both were left at `aTimeoutSec = 0` when step 3 was bumped to `1` earlier in this PR. With the new single-pass-for-0s semantics, neither `receiver`/`receiver_` callback fired within that one pass, so the `EXPECT_*` checks ended up comparing against values left over from the *previous* step:
- `error_` read back as the untouched `OT_ERROR_NONE` reset value
- `errorMsg_` read back as the previous step's leftover string

**Fix:** `9a5828a` bumps `aTimeoutSec` from `0` to `1` at both call sites, each with a comment pointing back to the step-3 rationale.

**Verified locally:**

| Run | Result |
|---|---|
| Full suite (8 tests) | ✅ PASS |
| Full suite × 100 repeat | ✅ 100/100 |
| `StateChangesCorrectlyAfterJoin` × 500 repeat | ✅ 500/500 |

jwhui (approved):
Thanks! 👍🏼

PR #3511, opened 2026-08-12
[tests] fix flaky ncp DeviceRole checks by polling instead of fixed sleep

### LINKED ISSUE #3510: Flaky test: ncp_netif_address_update.exp fails intermittently with D-Bus NoReply on DeviceRole
**Describe the bug**

`tests/scripts/expect/ncp_netif_address_update.exp` intermittently fails in the `NcpMode` CI workflow with a D-Bus `NoReply` error while checking `DeviceRole` right after joining.

Observed on #3480, but unrelated to that PR's changes (see "Additional context" below).

The script joins the network via D-Bus, sleeps a fixed 10s to let the node become leader, then reads `DeviceRole` with a 1s D-Bus reply timeout:

```tcl
spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.Join "array:byte:${dataset_dbus}"
expect eof

# Step 2. Wait 10 seconds for it becomes a leader.
sleep 10
spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply --reply-timeout=1000 /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:DeviceRole
```

When leader formation (or otbr-agent's D-Bus dispatch) takes slightly longer than the fixed 10s window, or the 1000ms reply-timeout is too tight under CI load, the `Properties.Get` call fails with a `NoReply` error, which then makes the following `expect eof` fail with `spawn id exp7 not open`, failing the whole script.

**To Reproduce**

1. Git commit id: `644d6db03c9b4ba1911ef5196a59dcc693706f4e` (PR #3480), also reproduced earlier at `29b1bff92300bf3e288ca2c5ae9494514273d909` (same PR, 2026-07-27)
2. Platform: simulation (`OT_PLATFORM=simulation`), NCP mode, `mDNSResponder` backend
3. Build steps: standard `NcpMode` CI workflow (`.github/workflows/ncp_mode.yml`), `BUILD_TARGET=ncp_mode`
4. Network topology: single simulated NCP node joining via `BorderRouter.Join` over D-Bus (no additional nodes)
5. Run: `tests/scripts/expect/ncp_netif_address_update.exp`

This is intermittent — it does not reproduce on every run, only under CI timing pressure.

**Expected behavior**

`DeviceRole` should be readable and reflect `leader` reliably once the node has had enough time to attach, without a hard failure due to a short fixed D-Bus reply timeout racing against leader formation.

**Console/log output**

```
spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply --reply-timeout=1000 /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:DeviceRole
Error org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.

 *** ERROR: ncp expect script failed!
    while executing
"expect eof"
    (file "tests/scripts/expect/ncp_netif_address_update.exp" line 50)
```

Full logs:
- https://github.com/openthread/ot-br-posix/actions/runs/31569502892/job/94131183385
- https://github.com/openthread/ot-br-posix/actions/runs/30212587620

**Additional context**

Evidence this is a pre-existing, intermittent flake rather than caused by PR #3480:

- Same script, same error, observed on two separate runs weeks apart, on otherwise-unrelated commits of the same PR:
  - https://github.com/openthread/ot-br-posix/actions/runs/31569502892/job/94131183385 (2026-08-12, commit `644d6db0`, which only touches `src/rest/*` / REST ePSKc endpoints — unrelated to D-Bus/mainloop/netif code)
  - https://github.com/openthread/ot-br-posix/actions/runs/30212587620 (2026-07-27, commit `29b1bff9`, also unrelated to D-Bus/mainloop/netif code)
- In both cases the diff between the passing prior commit and the failing commit does not touch anything related to D-Bus, the mainloop, or netif address handling, ruling out those changes as the cause.
- This looks like the same class of "device not yet leader within a fixed wait window" timing race that was recently addressed for `RcpHostApi.StateChangesCorrectlyAfterJoin` (see the `fix-rcphost-join-test-flake` work, e.g. commits `836b6546`, `68b3fbdb`, `df2ebe39`, `b1feb151`, `7f8f1821`, `9a5828ae`) — a fixed `sleep`/short D-Bus reply-timeout instead of a bounded polling loop.

Suggested fix: replace the fixed `sleep 10` + `--reply-timeout=1000` combo in `ncp_netif_address_update.exp` with a bounded polling loop on `DeviceRole` (retry with a longer overall timeout) before asserting `leader`, similar to the approach used for the `RcpHostApi.StateChangesCorrectlyAfterJoin` flake fix.

Related: #3480

lboue:
### Description

Fixes the flaky `NcpMode` CI failure reported in #3510: several `expect` scripts wait a fixed `sleep 10`, then read the `DeviceRole` D-Bus property **once** with a 1000ms reply-timeout. When attach/leader formation takes slightly longer than the fixed window (or otbr-agent's D-Bus dispatch is briefly busy), the read fails with:

```
Error org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
```

...failing the whole script (`expect eof` then errors with `spawn id ... not open`).

Same class of "fixed wait window" timing race already addressed for `RcpHostApi.StateChangesCorrectlyAfterJoin`.

### Change

- Add `wait_for_device_role {pattern} {retries} {bus}` to `tests/scripts/expect/_common.exp`: polls the `DeviceRole` property (bounded, ~30 retries spaced ~1s apart), keeping the same short per-call `reply-timeout`, but tolerating transient `NoReply`/wrong-role reads instead of asserting on a single racy one.
- Apply it to the four scripts sharing the fixed-sleep + single-read pattern:
  - `tests/scripts/expect/ncp_netif_address_update.exp`
  - `tests/scripts/expect/ncp_join_leave.exp`
  - `tests/scripts/expect/ncp_schedule_migration.exp`
  - `tests/scripts/expect/ncp_netif_tx_rx.exp`

### Testing

**Update:** since opening this PR, the fix has also been validated locally against the *real* stack (not just the fake D-Bus service described below), in a privileged Docker container running systemd/dbus/avahi and mirroring the `NcpMode` CI job (`tests/scripts/bootstrap.sh` + `script/cmake-build` + `tests/scripts/ncp_mode build_ot_sim expect ...`) — real `otbr-agent` built from this branch, talking over D-Bus to a real simulated NCP (`ot-ncp-ftd`) via `spinel+hdlc+forkpty`, no mocks:

```
PASS tests/scripts/expect/ncp_netif_address_update.exp
PASS tests/scripts/expect/ncp_join_leave.exp
PASS tests/scripts/expect/ncp_schedule_migration.exp
PASS tests/scripts/expect/ncp_netif_tx_rx.exp
```

In the `ncp_netif_address_update.exp` run, the log shows `wait_for_device_role` polling through several consecutive `DeviceRole = "detached"` reads before the node reaches `leader` — the exact race window this PR's fix is meant to absorb.

Original validation (from before sudo/docker were available in the sandbox used to prepare this change):

CI here doesn't have sudo/docker available in the sandbox used to prepare this change, so instead of the full otbr-agent + simulated NCP stack, the fix was validated against a minimal fake `io.openthread.BorderRouter` D-Bus service that reproduces the exact race:

- Replies to `Properties.Get(DeviceRole)` take 2s (> the 1000ms `reply-timeout`) until a configurable point in time, then reply immediately with the target role (`leader` / `router|child` / `disabled`, matching each script's assertion).
- **Old logic** (fixed `sleep 10` + single read) against this fake service: fails with the **identical** `NoReply` error seen in CI.
- **New logic** (`wait_for_device_role`): retries through the `NoReply`s and succeeds once the role becomes available.
- **Negative case**: when the role never matches, `wait_for_device_role` still fails deterministically after exhausting its bounded retries (no infinite loop).

Also confirmed the modified `_common.exp` sources without syntax errors.

Fixes #3510

### DIFF
--- a/tests/scripts/expect/_common.exp
+++ b/tests/scripts/expect/_common.exp
@@ -84,6 +84,63 @@ proc expect_line {line} {
     return $expect_out(1,string)
 }
 
+# Poll the otbr-agent DeviceRole D-Bus property until it matches `pattern`,
+# retrying up to `retries` times (roughly one attempt per second). Each
+# individual dbus-send call keeps a short reply-timeout, but polling absorbs
+# the CI timing variance in leader/router formation instead of relying on a
+# single read after a fixed sleep, which races with attach and intermittently
+# fails with "Error org.freedesktop.DBus.Error.NoReply".
+proc wait_for_device_role {pattern {retries 30} {bus "--system"}} {
+    # spawn sets the global spawn_id, so the dbus-send calls below would
+    # otherwise silently redirect any send/expect the caller does afterwards
+    # on its own node. Save and restore it so this utility is transparent to
+    # the caller's spawn state.
+    global spawn_id
+    set has_spawn_id [info exists spawn_id]
+    if {$has_spawn_id} {
+        set backup_spawn_id $spawn_id
+    }
+
+    # Localize the timeout so a hung dbus-send can't inherit whatever
+    # timeout was active in the calling script, and so a genuine timeout
+    # (as opposed to a clean eof) still closes the spawned process below.
+    set timeout 2
+
+    set success 0
+    for {set i 0} {$i < $retries} {incr i} {
+        spawn dbus-send $bus --print-reply --reply-timeout=1000 --dest=io.openthread.BorderRouter.wpan0 \
+            /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:DeviceRole
+        expect {
+            -re $pattern {
+                expect eof
+                set success 1
+                break
+            }
+            eof {
+                # NoReply, wrong role, or another transient D-Bus error: retry.
+            }
+            timeout {
+                # dbus-send hung instead of replying or closing; kill it off
+                # before retrying.
+                catch {close}
+                catch {wait}
+            }
+        }
+        sleep 1
+    }
+
+    if {$has_spawn_id} {
+        set spawn_id $backup_spawn_id
+    } else {
+        unset -nocomplain spawn_id
+    }
+
+    if {!$success} {
+        fail "Timed out waiting for DeviceRole matching '$pattern'"
+    }
+    return 0
+}
+
 # type: The type of the node.
 #   Possible values:
 #   1. cli: The cli app. ot-cli-ftd or ot-cli-mtd
--- a/tests/scripts/expect/ncp_get_device_role.exp
+++ b/tests/scripts/expect/ncp_get_device_role.exp
@@ -32,13 +32,7 @@ spawn_node 1 otbr $::env(EXP_OT_NCP_PATH)
 
 sleep 1
 
-spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply --reply-timeout=1000 /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:DeviceRole
-expect -re {disabled} {
-} timeout {
-    puts "timeout!"
-    exit 1
-}
-expect eof
+wait_for_device_role {disabled}
 
 # Shut down otbr-agent
 switch_node 1
--- a/tests/scripts/expect/ncp_join_leave.exp
+++ b/tests/scripts/expect/ncp_join_leave.exp
@@ -52,25 +52,14 @@ sleep 1
 spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.Join "array:byte:${dataset_dbus}"
 expect eof
 
-# Step 3. Wait 10 seconds, check if the otbr-agent has attached successfully
-sleep 10
-spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply --reply-timeout=1000 /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:DeviceRole
-expect -re {router|child} {
-} timeout {
-    puts "timeout!"
-    exit 1
-}
-expect eof
+# Step 3. Check if the otbr-agent has attached successfully. Poll instead of
+# a fixed sleep plus a single short-timeout read: attach timing varies in CI
+# and the fixed wait intermittently races with it (see #3510).
+wait_for_device_role {router|child}
 
 # Step 4. Use dbus leave method to let otbr-agent leave the network
 spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.LeaveNetwork
 expect eof
 
 # Step 5. Verify the state of otbr-agent is 'disabled'
-spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --print-reply --reply-timeout=1000 /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:DeviceRole
-expect -re {disabled} {
-} timeout {
-    puts "timeout!"
-    exit 1
-}
-expect eof
+wait_for_device_role {disabled}
--- a/tests/scripts/expect/ncp_netif_address_update.exp
+++ b/tests/scripts/expect/ncp_netif_address_update.exp
@@ -39,15 +39,10 @@ sleep 1
 spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.Join "array:byte:${dataset_dbus}"
 expect eof
 
-# Step 2. Wait 10 seconds for it becomes a leader.
-sleep 10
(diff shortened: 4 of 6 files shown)


tests/scripts/expect/_common.exp
```diff
@@ -84,6 +84,30 @@ proc expect_line {line} {
...
+                return 0
+            }
+            eof {
+                # NoReply, wrong role, or another transient D-Bus error: retry.
+            }
+        }
+        sleep 1
+    }
+    fail "Timed out waiting for DeviceRole matching '$pattern'"
+}
```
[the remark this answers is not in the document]
lboue: Good catch, fixed in 69aa2785 — `wait_for_device_role` now saves/restores the caller's `spawn_id` around the polling loop, matching the suggested approach. Validated locally that a caller's pre-existing spawn (a long-lived dummy process) is still reachable via `send`/`expect` after the call returns.

jwhui:
@lboue , the `ncp_mode` check still seems to be failing. Or is it a different failure?

tests/scripts/expect/_common.exp
```diff
@@ -84,6 +84,50 @@ proc expect_line {line} {
...
+proc wait_for_device_role {pattern {retries 30} {bus "--system"}} {
+    # spawn sets the global spawn_id, so the dbus-send calls below would
+    # otherwise silently redirect any send/expect the caller does afterwards
+    # on its own node. Save and restore it so this utility is transparent to
+    # the caller's spawn state.
+    global spawn_id
+    set has_spawn_id [info exists spawn_id]
+    if {$has_spawn_id} {
+        set backup_spawn_id $spawn_id
+    }
```
jwhui, line 102: If `spawn_id` was not set before calling `wait_for_device_role` (`has_spawn_id` is false), `spawn dbus-send` will leave the global `spawn_id` variable defined with the closed `dbus-send` process.
Consider unsetting `spawn_id` when `has_spawn_id` is false to leave the global state clean:
```suggestion
    if {$has_spawn_id} {
        set spawn_id $backup_spawn_id
    } else {
        unset -nocomplain spawn_id
    }
```

tests/scripts/expect/_common.exp
```diff
@@ -84,6 +84,50 @@ proc expect_line {line} {
...
+                expect eof
+                set success 1
+                break
+            }
+            eof {
+                # NoReply, wrong role, or another transient D-Bus error: retry.
+            }
+        }
+        sleep 1
+    }
```
jwhui, line 130: If `dbus-send` hangs or doesn't output immediately, the `expect` block uses whatever global `timeout` was previously active in the calling script. Furthermore, if `expect` times out, neither `-re $pattern` nor `eof` will match, and the loop moves on without closing the spawned process.

Consider setting a short localized timeout (e.g., `set timeout 2`) and handling `timeout` explicitly in the `expect` block:
```suggestion
    set timeout 2
    set success 0
    for {set i 0} {$i < $retries} {incr i} {
        spawn dbus-send $bus --print-reply --reply-timeout=1000 --dest=io.openthread.BorderRouter.wpan0 \
            /io/openthread/BorderRouter/wpan0 org.freedesktop.DBus.Properties.Get string:io.openthread.BorderRouter string:DeviceRole
        expect {
            -re $pattern {
                expect eof
                set success 1
                break
            }
            eof {
                # NoReply, wrong role, or another transient D-Bus error: retry.
            }
            timeout {
                catch {close}
                catch {wait}
            }
        }
        sleep 1
    }
```

tests/scripts/expect/_common.exp
```diff
@@ -84,6 +84,50 @@ proc expect_line {line} {
     return $expect_out(1,string)
 }
 
+# Poll the otbr-agent DeviceRole D-Bus property until it matches `pattern`,
+# retrying up to `retries` times (roughly one attempt per second). Each
+# individual dbus-send call keeps a short reply-timeout, but polling absorbs
+# the CI timing variance in leader/router formation instead of relying on a
+# single read after a fixed sleep, which races with attach and intermittently
+# fails with "Error org.freedesktop.DBus.Error.NoReply".
+proc wait_for_device_role {pattern {retries 30} {bus "--system"}} {
```
jwhui, line 93: `ncp_get_device_role.exp` uses the same pattern:
https://github.com/openthread/ot-br-posix/blob/e8a94e516427295eb93f1bd3c20a608b535697c5/tests/scripts/expect/ncp_get_device_role.exp#L35

Should we replace this with `wait_for_device_role` as well?

tests/scripts/expect/_common.exp
lboue: Fixed. wait_for_device_role now does unset -nocomplain spawn_id in the else branch when there was no caller spawn_id to restore, so it doesn't leave a dangling reference to the closed dbus-send.

tests/scripts/expect/_common.exp
lboue: Fixed, matching your suggestion. Added set timeout 2 before the loop and an explicit timeout { catch {close}; catch {wait} } branch that kills the stuck dbus-send before retrying, instead of inheriting the caller's global timeout.

tests/scripts/expect/_common.exp
lboue: Good catch. Migrated it to wait_for_device_role {disabled} as well, so all five scripts now share the same polling helper.

jwhui (approved):
Thanks! 👍🏼

jwhui:
@lboue , can you squash the commits?

lboue:
> @lboue , can you squash the commits?

Done

PR #3514, opened 2026-08-13
[ncp] support the pending dataset under NCP mode

LorbusChris:
Split out of #3489, which is now the RCP-side half of the original series.

- `NcpSpinel` does not refresh its cached active dataset when the co-processor reports an unsolicited change, so callers read a stale dataset after the network changes underneath them.
- NCP mode has no pending-dataset path at all: `SetDatasetPendingTlvs` is added to the `NcpSpinel` interface and implemented in `NcpHost`, so `ScheduleMigration` works in NCP mode the way it already does under RCP.

**Testing:** compile-tested only (builds standalone against current main, unit tests pass). I do not have NCP-mode hardware — the co-processor type is chosen by `otSysInitCoprocessor()` from the firmware on the dongle, and mine reports RCP — so this half has not been exercised at runtime. Reviewers with an NCP setup are very welcome; the RCP counterpart of the same behaviour is in #3489.

### DIFF
--- a/src/host/ncp_host.cpp
+++ b/src/host/ncp_host.cpp
@@ -53,6 +53,7 @@ NcpNetworkProperties::NcpNetworkProperties(void)
     : mDeviceRole(OT_DEVICE_ROLE_DISABLED)
 {
     memset(&mDatasetActiveTlvs, 0, sizeof(mDatasetActiveTlvs));
+    memset(&mDatasetPendingTlvs, 0, sizeof(mDatasetPendingTlvs));
     SetMeshLocalPrefix(kMeshLocalPrefixInit);
 }
 
@@ -90,10 +91,16 @@ void NcpNetworkProperties::GetDatasetActiveTlvs(otOperationalDatasetTlvs &aDatas
     memcpy(aDatasetTlvs.mTlvs, mDatasetActiveTlvs.mTlvs, mDatasetActiveTlvs.mLength);
 }
 
+void NcpNetworkProperties::SetDatasetPendingTlvs(const otOperationalDatasetTlvs &aPendingOpDatasetTlvs)
+{
+    mDatasetPendingTlvs.mLength = aPendingOpDatasetTlvs.mLength;
+    memcpy(mDatasetPendingTlvs.mTlvs, aPendingOpDatasetTlvs.mTlvs, aPendingOpDatasetTlvs.mLength);
+}
+
 void NcpNetworkProperties::GetDatasetPendingTlvs(otOperationalDatasetTlvs &aDatasetTlvs) const
 {
-    // TODO: Implement the method under NCP mode.
-    OTBR_UNUSED_VARIABLE(aDatasetTlvs);
+    aDatasetTlvs.mLength = mDatasetPendingTlvs.mLength;
+    memcpy(aDatasetTlvs.mTlvs, mDatasetPendingTlvs.mTlvs, mDatasetPendingTlvs.mLength);
 }
 
 void NcpNetworkProperties::SetMeshLocalPrefix(const otMeshLocalPrefix &aMeshLocalPrefix)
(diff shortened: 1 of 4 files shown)


src/host/ncp_spinel.cpp
```diff
@@ -866,6 +900,16 @@ otbrError NcpSpinel::HandleResponseForPropGet(spinel_tid_t      aTid,
         break;
     }
 
+    case SPINEL_PROP_THREAD_PENDING_DATASET_TLVS:
+    {
+        otOperationalDatasetTlvs datasetTlvs = {};
+
+        VerifyOrExit(ParseOperationalDatasetTlvs(aData, aLength, datasetTlvs) == OT_ERROR_NONE,
```
jwhui, line 922: If the property get fails or is rejected by the NCP, the NCP returns `aKey == SPINEL_PROP_LAST_STATUS` with the error status payload in `aData`. Because this switch is on `mWaitingKeyTable[aTid]`, it will pass the status payload to `ParseOperationalDatasetTlvs`, leading to a parse failure and a critical log `otbrLogCrit("Error parsing response with tid:%u", aTid)`.

Add a check for `SPINEL_PROP_LAST_STATUS` and verify `aKey == SPINEL_PROP_THREAD_PENDING_DATASET_TLVS` similar to how `HandleResponseForPropSet` handles it.

src/host/ncp_spinel.cpp
```diff
@@ -79,6 +79,18 @@ void NcpSpinel::Init(ot::Spinel::SpinelDriver &aSpinelDriver, PropsObserver &aOb
     mPropsObserver = &aObserver;
     mIid           = mSpinelDriver->GetIid();
     mSpinelDriver->SetFrameHandler(&HandleReceivedFrame, &HandleSavedFrame, this);
+
+    // Get the pending dataset to have an initial value. Unlike the active
+    // dataset, otbr never sets it itself, so without this the cache would stay
+    // empty until the NCP next reports a change.
```
jwhui: Should we also fetch `SPINEL_PROP_THREAD_ACTIVE_DATASET_TLVS` at init time and handle it in `HandleResponseForPropGet` so that both active and pending datasets are initialized from the NCP on startup?

src/host/ncp_host.cpp
```diff
@@ -90,10 +92,19 @@ void NcpNetworkProperties::GetDatasetActiveTlvs(otOperationalDatasetTlvs &aDatas
     memcpy(aDatasetTlvs.mTlvs, mDatasetActiveTlvs.mTlvs, mDatasetActiveTlvs.mLength);
 }
 
+void NcpNetworkProperties::SetDatasetPendingTlvs(const otOperationalDatasetTlvs &aPendingOpDatasetTlvs)
+{
+    // mLength is validated at parse time, but it is a uint8_t and the TLV
+    // buffer holds 254 bytes; clamp rather than trust the invariant.
+    mDatasetPendingTlvs.mLength =
+        static_cast<uint8_t>(std::min<size_t>(aPendingOpDatasetTlvs.mLength, sizeof(mDatasetPendingTlvs.mTlvs)));
```
jwhui: Since `mLength` is already validated against `sizeof(mTlvs)` at parse time in `ParseOperationalDatasetTlvs`, we can drop the clamp and match the existing `SetDatasetActiveTlvs` implementation right above.

src/host/ncp_spinel.cpp
LorbusChris: Done, the get response now checks for `SPINEL_PROP_LAST_STATUS` first and verifies `aKey` against the waiting key before parsing, following the `HandleResponseForPropSet` pattern. A failed get logs a warning with the decoded status instead of a parse error.

src/host/ncp_spinel.cpp
LorbusChris: Done, Init now fetches both datasets, and the get handler covers `ACTIVE_DATASET_TLVS` and `PENDING_DATASET_TLVS` with the same guarded body. Without it the active cache would also start empty against an already-provisioned NCP.

src/host/ncp_host.cpp
LorbusChris: Done, dropped. `SetDatasetPendingTlvs` now matches `SetDatasetActiveTlvs`.

jwhui (approved):

jwhui:
@LorbusChris , can you squash commits?

LorbusChris:
> @LorbusChris , can you squash commits?

Done

PR #1287, opened 2022-03-08
[firewall] clear firewall rules before starting a new firewall 

superwhd:
This PR fixes https://github.com/openthread/ot-br-posix/issues/1286.

Also, this PR clears the firewall rules when the firewall service is uninstalled. Previously firewall rules aren't cleared immediately when the service is uninstalled, which may not be a desired behavior.

Issue #1286, opened 2022-03-07
running otbr fails due to otbr_firewall at the second attempt


doublemis1:
**Describe the bug**
The OTBR docker fails at the second attempt due to firewall-otbr when docker is connected to host network.
Docker creates ip6tables rules at the first attempt and does not clear them before removing the docker container which impacts the second attempt.

**To Reproduce** 

1. ot-br-posix: f0bd2166467464c3fac23b28f7669179ee9ba332
2. docker build -t otbr:test . -f etc/docker/Dockerfile
3. docker run -it --rm --privileged --name otbr --network host --volume /dev/ttyACM0:/dev/radio otbr:test --radio-url spinel+hdlc+uart:///dev/radio?uart-baudrate=115200 -B wlan0
4. Stop and remove docker container
5. docker run -it --rm --privileged --name otbr --network host --volume /dev/ttyACM0:/dev/radio otbr:test --radio-url spinel+hdlc+uart:///dev/radio?uart-baudrate=115200 -B wlan0

**Expected behavior**

The docker container should handle that the ip6tables are already set.
Or Remove them and add them once again.

**Console/log output**
```
+ ipset create -exist otbr-ingress-deny-src hash:net family inet6
+ ipset create -exist otbr-ingress-deny-src-swap hash:net family inet6
+ ipset create -exist otbr-ingress-allow-dst hash:net family inet6
+ ipset create -exist otbr-ingress-allow-dst-swap hash:net family inet6
+ ip6tables -N OTBR_FORWARD_INGRESS
ip6tables: Chain already exists.
+ die 'Failed to start firewall service'
+ echo ' *** ERROR:  Failed to start firewall service'
 *** ERROR:  Failed to start firewall service
+ exit 1
```
**Additional context**

superwhd:
Thanks for raising this issue! I sent a PR to fix it: https://github.com/openthread/ot-br-posix/pull/1287

PR #3517, opened 2026-08-14
[docker] make IPv4 firewall rules idempotent and remove them on exit

### LINKED ISSUE #3516: Docker image: IPv4 NAT64 rules are added without a duplicate check and never removed, accumulating on every start
**Describe the bug**

The `otbr-agent` s6 run script adds four IPv4 rules unconditionally with `-A`,
without checking whether they already exist:

https://github.com/openthread/ot-br-posix/blob/main/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run

```sh
echo "Configuring OpenThread NAT64..."

iptables -t mangle -A PREROUTING -i "${OT_THREAD_IF}" -j MARK --set-mark 0x1001
iptables -t nat -A POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
iptables -t filter -A FORWARD -o "${OT_INFRA_IF}" -j ACCEPT
iptables -t filter -A FORWARD -i "${OT_INFRA_IF}" -j ACCEPT
```

The corresponding `finish` script cleans up **only** the ip6tables chain and the
ipsets — the IPv4 rules are not touched:

https://github.com/openthread/ot-br-posix/blob/main/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish

```sh
while ip6tables -C FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; do
    ip6tables -D FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}"
done
if ip6tables -L "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; then
    ip6tables -w -F "${OT_FORWARD_INGRESS_CHAIN}"
    ip6tables -w -X "${OT_FORWARD_INGRESS_CHAIN}"
fi
ipset_destroy_if_exist otbr-ingress-deny-src
...
```

That cleanup appears to have been added for #1286 (PR #1287). The IPv4 rules
added in the same `run` script were not included.

With `network_mode: host` these rules live in the host's netfilter namespace and
survive the container, so every start of the service appends another copy —
including in-container restarts by s6, not just container restarts.

**Impact**

Functionally benign (the duplicates are identical `ACCEPT`/`MASQUERADE` rules, so
the first match wins), but unbounded: the chains grow without limit, every
forwarded IPv4 packet traverses them, and the chains become unreadable for anyone
trying to diagnose an unrelated firewall problem.

**Evidence from our deployment**

```
iptables -t filter -S FORWARD    | grep -c 'enp2s0 -j ACCEPT'   ->  250
iptables -t nat    -S POSTROUTING| grep -c '0x1001'             ->  125
iptables -t mangle -S PREROUTING | grep -c '0x1001'             ->  125
```

125 accumulated starts. All copies have zero packet counters here, because
nothing on this Thread network uses NAT64:

```
# ot-ctl nat64 counters
|         Total |        0 Pkts |        0 Bytes |   (4 to 6)
|         Total |        0 Pkts |        0 Bytes |   (6 to 4)
# ot-ctl nat64 mappings
(empty)
```

Separately, the IPv6 ingress chain also ends up duplicated if the container is
killed rather than stopped gracefully — `docker kill` sends SIGKILL, so s6 never
runs `finish`:

```
# ip6tables -S OT_FORWARD_INGRESS
-N OT_FORWARD_INGRESS
-A OT_FORWARD_INGRESS -i wpan0 -m pkttype --pkt-type unicast -j DROP
... (5 rules)
-A OT_FORWARD_INGRESS -i wpan0 -m pkttype --pkt-type unicast -j DROP
... (the same 5 rules again)
```

so relying on `finish` alone is not sufficient — `run` should be idempotent too.

**Suggested fix**

Guard each rule in `run`:

```sh
iptables -t nat -C POSTROUTING -m mark --mark 0x1001 -j MASQUERADE 2>/dev/null ||
    iptables -t nat -A POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
```

(and likewise for the other three), and mirror the existing ip6tables removal
loop in `finish` for the four IPv4 rules.

The same argument applies to `ip6tables -N` / `-I FORWARD 1 ... -j OT_FORWARD_INGRESS`
in `run`: making them idempotent would make the container tolerant of an unclean
previous shutdown, which is what #1286 originally asked for.

**Environment**

| | |
|---|---|
| Host | Debian 12 (bookworm), kernel 6.1, x86-64 |
| Docker | 27.4.1 |
| Compose | 2.32.1, `network_mode: host`, `privileged: true` |

skirmess:
Fixes #3516.

### Problem

`etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run` appends four IPv4 rules for NAT64 with `-A` and no `-C` check:

```sh
iptables -t mangle -A PREROUTING -i "${OT_THREAD_IF}" -j MARK --set-mark 0x1001
iptables -t nat -A POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
iptables -t filter -A FORWARD -o "${OT_INFRA_IF}" -j ACCEPT
iptables -t filter -A FORWARD -i "${OT_INFRA_IF}" -j ACCEPT
```

The matching `finish` cleans up only the ip6tables chain and the ipsets — that cleanup was added for #1286 (PR #1287) and the IPv4 rules were not included.

With `--network host` these rules live in the host's netfilter namespace and outlive the container, so every start of the service appends another copy — including in-container restarts by s6, not just container restarts.

On the deployment where this surfaced:

```
iptables -t filter -S FORWARD     | grep -c 'enp2s0 -j ACCEPT'   ->  250
iptables -t nat    -S POSTROUTING | grep -c '0x1001'             ->  125
iptables -t mangle -S PREROUTING  | grep -c '0x1001'             ->  125
```

Functionally harmless — the first match wins — but unbounded, traversed by every forwarded IPv4 packet, and the chains become unreadable when diagnosing something else.

### Change

- `run`: add `iptables_add_if_missing`, guarding each rule with `-C`.
- `finish`: add `iptables_delete_all`, mirroring the `while ip6tables -C … do -D done` loop already used for the ingress chain jump, and remove the four rules.
- `finish` now also defines `OT_INFRA_IF` (same default as `run`), which it needs for the two `filter FORWARD` rules.

The removal loop deletes *every* copy, so a host carrying duplicates from an earlier image is cleaned up on the first graceful stop.

The `-C` guard matters independently of the cleanup: `finish` does not run when the container is killed rather than stopped (`docker kill` sends SIGKILL), and a plain `-A` would duplicate the rules on the next start. This is also what #1286 originally asked for — tolerating a previous unclean shutdown.

### Testing

`shellcheck -s bash run finish` reports nothing for the changed lines. (It flags one pre-existing `SC2015` on the unrelated `mkdir … && … || die` line, left untouched.)

Functional check of both helpers against real `iptables` in an isolated network namespace:

```
--- old behaviour: plain -A three times ---
filter FORWARD : 3     (expected 3)
nat POSTROUTING: 3     (expected 3)
--- iptables_delete_all ---
filter FORWARD : 0     (expected 0)
nat POSTROUTING: 0     (expected 0)
--- new behaviour: iptables_add_if_missing three times ---
mangle PREROUTING: 1   (expected 1)
nat POSTROUTING  : 1   (expected 1)
filter FORWARD   : 2   (expected 2, one per direction)
--- finish path ---
all chains        : 0  (expected 0)
```

The second block covers the upgrade path: the loop removes duplicates left behind by an older image.

### Not in this PR

`run` also creates the ip6tables chain with a bare `ip6tables -N` / `-I FORWARD 1 … -j OT_FORWARD_INGRESS`, which duplicates the same way after an unclean shutdown. Happy to extend this PR to cover that if you would like it in the same change.

### DIFF
--- a/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
+++ b/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
@@ -30,6 +30,9 @@
 OT_THREAD_IF="${OT_THREAD_IF:-wpan0}"
 readonly OT_THREAD_IF
 
+OT_INFRA_IF="${OT_INFRA_IF:-wlan0}"
+readonly OT_INFRA_IF
+
 OT_FORWARD_INGRESS_CHAIN="OT_FORWARD_INGRESS"
 readonly OT_FORWARD_INGRESS_CHAIN
 
@@ -48,8 +51,25 @@ ipset_destroy_if_exist()
     done
 }
 
-while ip6tables -C FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; do
-    ip6tables -D FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}"
+# Remove every copy of an iptables rule. Older images appended these rules
+# without checking for duplicates, so a host may carry several of them.
+#
+# `-w` keeps a busy xtables lock from ending the loop early, and `|| break`
+# catches everything else: without it a failing `-D` leaves the rule in place,
+# `-C` keeps succeeding, and this spins forever -- hanging container shutdown,
+# since finish runs on the way out.
+iptables_delete_all()
+{
+    local table="$1"
+    shift
+
+    while iptables -t "${table}" -w -C "$@" 2> /dev/null; do
+        iptables -t "${table}" -w -D "$@" || break
+    done
+}
+
+while ip6tables -w -C FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; do
+    ip6tables -w -D FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" || break
 done
 
 if ip6tables -L "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; then
@@ -61,6 +81,11 @@ ipset_destroy_if_exist otbr-ingress-deny-src
 ipset_destroy_if_exist otbr-ingress-deny-src-swap
 ipset_destroy_if_exist otbr-ingress-allow-dst
 ipset_destroy_if_exist otbr-ingress-allow-dst-swap
+
+iptables_delete_all mangle PREROUTING -i "${OT_THREAD_IF}" -j MARK --set-mark 0x1001
+iptables_delete_all nat POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
+iptables_delete_all filter FORWARD -o "${OT_INFRA_IF}" -j ACCEPT
+iptables_delete_all filter FORWARD -i "${OT_INFRA_IF}" -j ACCEPT
 
 echo "OpenThread firewall rules removed."
 
--- a/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
+++ b/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
@@ -60,6 +60,24 @@ die()
     exit 1
 }
 
+# Add an iptables rule unless an identical one is already present.
+#
+# With `--network host` these rules are created in the host's netfilter
+# namespace and outlive the container, so a plain `-A` appends another copy on
+# every start of this service. `finish` removes them again, but it does not run
+# when the container is killed rather than stopped.
+#
+# `-w` is required, not cosmetic: without it a busy xtables lock makes `-C`
+# fail, which is indistinguishable from "rule not present" and falls through to
+# `-A` -- the guard would then create the very duplicate it exists to prevent.
+iptables_add_if_missing()
+{
+    local table="$1"
+    shift
+
+    iptables -t "${table}" -w -C "$@" 2> /dev/null || iptables -t "${table}" -w -A "$@"
+}
+
 mkdir -p /data/thread && ln -sft /var/lib /data/thread || die "Could not create directory /var/lib/thread to store Thread data."
 
 echo "Configuring OpenThread firewall..."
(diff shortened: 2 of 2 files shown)


etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
```diff
@@ -48,6 +51,18 @@ ipset_destroy_if_exist()
...
+# Remove every copy of an iptables rule. Older images appended these rules
+# without checking for duplicates, so a host may carry several of them.
+iptables_delete_all()
+{
+    local table="$1"
+    shift
+
+    while iptables -t "${table}" -C "$@" 2> /dev/null; do
+        iptables -t "${table}" -D "$@"
+    done
```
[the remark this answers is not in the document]
skirmess: Good catch, applied in 5fc59a7 — thanks.

I verified it is reproducible rather than just theoretical, using a stub that lets `-C` through and makes `-D` fail, in an isolated network namespace:

```
--- stub sanity ---
  -C: ok        (passes through)
  -D: exit=4    (fails, as intended)

--- with || break ---
  iterations: 1          <- breaks immediately
  rule still present: 1  <- no hang

--- without || break ---
  INFINITE LOOP DETECTED (>20 iterations)
```

One note for the maintainers: the existing loop a few lines below has the same shape and the same exposure —

```sh
while ip6tables -C FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; do
    ip6tables -D FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}"
done
```

I have left it untouched to keep this PR focused, but it is the same one-line change and I am happy to include it here if you would prefer the file to be consistent.

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -60,6 +60,20 @@ die()
...
+# With `--network host` these rules are created in the host's netfilter
+# namespace and outlive the container, so a plain `-A` appends another copy on
+# every start of this service. `finish` removes them again, but it does not run
+# when the container is killed rather than stopped.
+iptables_add_if_missing()
+{
+    local table="$1"
+    shift
+
+    iptables -t "${table}" -C "$@" 2> /dev/null || iptables -t "${table}" -A "$@"
```
jwhui: Should we add the `-w` (`--wait`) flag?

`iptables` calls without `-w` can fail due to xtables lock contention. If `-C` fails due to lock contention rather than the rule being absent, it would fall through to `-A` and potentially append a duplicate rule.

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
```diff
@@ -48,6 +51,22 @@ ipset_destroy_if_exist()
...
+# `|| break` is required: if the deletion fails -- lock contention with another
+# process, for instance -- the rule stays in place and `-C` keeps succeeding,
+# which would spin here forever and hang container shutdown.
+iptables_delete_all()
+{
+    local table="$1"
+    shift
+
+    while iptables -t "${table}" -C "$@" 2> /dev/null; do
+        iptables -t "${table}" -D "$@" || break
```
jwhui: Should we add `-w` here to avoid early break on transient xtables lock contention?

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
```diff
@@ -48,6 +51,22 @@ ipset_destroy_if_exist()
...
+    done
+}
+
 while ip6tables -C FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; do
     ip6tables -D FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}"
```
jwhui: Should we also add `|| break` (and `-w`) to guard agains infinite loop risk?

jwhui:
> `run` also creates the ip6tables chain with a bare `ip6tables -N` / `-I FORWARD 1 … -j OT_FORWARD_INGRESS`, which duplicates the same way after an unclean shutdown. Happy to extend this PR to cover that if you would like it in the same change.

@skirmess , I'd welcome extending this PR to cover that as well if you'd like to include it!

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
skirmess: Yes — good catch, and it is worse than a missed check: the guard would have created the very duplicate it exists to prevent. Added in fcd6d95, on every call in both scripts.

Verified the option order is accepted and the guard still holds, in an isolated network namespace:

```
### option order: -t <table> -w -C ###
  accepted
### iptables_add_if_missing with -w, three times ###
  mangle PREROUTING: 1  (expected 1)
  filter FORWARD:    1  (expected 1)
```

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
skirmess: Added in fcd6d95. With `-w` the `|| break` is no longer reachable through lock contention and stays as the guard for genuine failures — which is what it was meant to be.

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
skirmess: Both added in fcd6d95.

skirmess:
@jwhui thanks — the ip6tables side is now included in fcd6d95, together with the `-w` flags.

The chain is created, or emptied when a previous run left it behind: the five rules below it are appended unconditionally, so a surviving chain ends up holding two copies of each. The jump into it is guarded with `-C`, and the pre-existing removal loop in `finish` got the same `|| break`.

Regression test in an isolated network namespace, running the whole ip6tables block twice:

```
### new block, run twice ###
  after 1st run:  jumps=1  chain rules=5
  after 2nd run:  jumps=1  chain rules=5   (expected 1 and 5)

### control: previous code, run twice ###
  after 2 runs:   jumps=2  chain rules=4   (2x2, shows the problem)
```

The control matches what the host this came from actually had: two `OT_FORWARD_INGRESS` jumps in `FORWARD` and two copies of every rule inside the chain, after container kills where `finish` never ran.

`shellcheck -s bash run finish` is clean for the changed lines (it flags one pre-existing `SC2015` on the unrelated `mkdir … && … || die` line, left untouched).

One judgement call worth flagging: I used a bare `-w` rather than `-w <seconds>`, matching the existing `ip6tables -w -F` / `-w -X` in `finish`. That waits indefinitely for the lock, which in `finish` runs on the shutdown path. Happy to switch to a bounded `-w 5` if you would rather cap it there.

jwhui:
Thanks @skirmess! The IPv6 idempotency handling and `-w` lock guards look great and make the container startup and shutdown clean and resilient.

Regarding bare `-w` vs `-w <seconds>`: keeping bare `-w` is fine here. Container runtimes (such as Docker and s6) enforce their own termination timeouts if a lock were ever held indefinitely by a deadlocked process.

Everything looks good. Please squash the commits into a single commit, and we can get this merged.

skirmess:
Squashed into a single commit (49575915) — thanks for the review, and for the note on bare `-w`; that was the one open judgement call and it is good to have it settled.

jwhui (approved):
Thanks! 👍🏼

PR #3521, opened 2026-08-15
[docker] accept IPv6 traffic leaving the Thread network

### LINKED ISSUE #3515: Docker image: IPv6 egress from the Thread interface is silently dropped when the host's ip6tables FORWARD policy is DROP
**Describe the bug**

The `otbr-agent` s6 run script configures explicit `FORWARD ACCEPT` rules for the
infrastructure interface for **IPv4**, but adds nothing equivalent for **IPv6**.
For IPv6 it only installs the ingress chain, which is jumped to on `-o wpan0`:

https://github.com/openthread/ot-br-posix/blob/main/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run

```sh
ip6tables -N "${OT_FORWARD_INGRESS_CHAIN}"
ip6tables -I FORWARD 1 -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}"   # into Thread only
...
iptables -t filter -A FORWARD -o "${OT_INFRA_IF}" -j ACCEPT                     # IPv4, both directions
iptables -t filter -A FORWARD -i "${OT_INFRA_IF}" -j ACCEPT                     #   explicitly handled
```

Traffic **out of** the Thread network (`-i wpan0 -o <infra>`) is therefore only
forwarded if the ip6tables `FORWARD` policy happens to be `ACCEPT`.

That assumption does not hold on a Docker host. With `network_mode: host`, the
container shares the host's netfilter namespace, and Docker manages ip6tables
there:

```
# ip6tables -S FORWARD
-P FORWARD DROP
-A FORWARD -j DOCKER-USER
```

**Impact — and why it is hard to diagnose**

The failure is one-directional and completely silent:

* Traffic *into* the Thread network works, because the ingress chain accepts it.
* Every reply *out of* the Thread network is dropped by the policy.

The border router looks perfectly healthy throughout: `ot-ctl state` reports
`leader`, the REST API answers, the mDNS announcement is visible, `ot-ctl ping`
to its own OMR address succeeds, devices attach and appear in the child table
with good link quality. Only the application layer fails — in our case every
Matter node went `unavailable` while Thread itself was flawless.

It cost us most of a day before we thought to look at the FORWARD counters.

**To Reproduce**

1. Docker host with IPv6 enabled and Docker managing ip6tables (`ip6tables -S FORWARD`
   shows `-P FORWARD DROP` and a `DOCKER-USER` jump).
2. Run `openthread/border-router` with `network_mode: host`, `privileged: true`.
3. Attach any Thread device and have it talk to something on the LAN.
4. Requests reach the device; replies never arrive. `ip6tables -L FORWARD -n -v`
   shows the policy drop counter rising.

**Expected behavior**

The image should install the IPv6 counterpart of the IPv4 rules it already adds,
rather than relying on a permissive default policy.

**Evidence from our deployment**

```
# ip6tables -L FORWARD -n -v --line-numbers
Chain FORWARD (policy DROP 418 packets, 70851 bytes)
num   pkts bytes target              in      out
  1   2678  302K DOCKER-USER          *       *
  2   2529  330K OT_FORWARD_INGRESS   *       wpan0
  3   3967  463K ACCEPT               wpan0   enp2s0     <-- rule added by hand
```

Rule 3 is not from the image; we add it ourselves as a workaround. Without it
those 3967 packets are counted by the policy instead. The policy counter stopped
moving the moment the rule was added (417 at the time, 418 two days later).

**Suggested fix**

In `run`, alongside the existing IPv4 rules:

```sh
ip6tables -C FORWARD -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT 2>/dev/null ||
    ip6tables -I FORWARD 1 -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT
```

and the matching removal loop in `finish`, mirroring the one that already exists
for the ingress chain jump.

**Environment**

| | |
|---|---|
| Host | Debian 12 (bookworm), kernel 6.1, x86-64 |
| Docker | 27.4.1 |
| Compose | 2.32.1, `network_mode: host`, `privileged: true` |
| RCP | SMLIGHT SLZB-06M (EFR32MG21) over USB, 460800 baud |

skirmess:
Fixes #3515, following @jwhui's direction in https://github.com/openthread/ot-br-posix/issues/3515#issuecomment-5298669158 — explicit forward rules for the Thread and infrastructure interfaces, `DOCKER-USER` with a `FORWARD` fallback.

> ⚠️ **Stacked on #3517.** Both change the same two scripts, so this branch is based on that PR to keep the diff applicable. It carries #3517's commit as its parent; please merge that one first, or say the word and I will rebase onto main once it lands.

### Problem

`run` installs the ingress chain, which filters traffic going *into* the Thread network. There is nothing for the other direction — egress has relied on the ip6tables FORWARD policy being ACCEPT. That holds on a bare host. It does not hold under Docker, which manages ip6tables and sets the policy to DROP:

```
# ip6tables -S FORWARD
-P FORWARD DROP
-A FORWARD -j DOCKER-USER
```

The failure is one-directional and completely silent. Requests into the mesh work, because the ingress chain accepts them; every reply out of it is dropped. Meanwhile `ot-ctl state` reports `leader`, the REST API answers, mDNS is fine, devices attach and show good link quality. Only the application layer fails — in our case every Matter node went `unavailable` while Thread itself was flawless. It cost most of a day before we thought to look at the FORWARD counters.

The IPv4 path does not have this problem because it adds its own `filter FORWARD` ACCEPT rules explicitly. This is the missing IPv6 counterpart.

### Change

`run`:

```sh
if ip6tables -w -L "${OT_DOCKER_USER_CHAIN}" -n > /dev/null 2>&1; then
    egress_chain="${OT_DOCKER_USER_CHAIN}"
else
    egress_chain="FORWARD"
fi

ip6tables -w -C "${egress_chain}" -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT 2> /dev/null ||
    ip6tables -w -I "${egress_chain}" 1 -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT
```

`finish` removes it from **both** chains, because which one `run` picked depends on whether `DOCKER-USER` existed at that moment, and that can differ between start and stop.

The rule is narrow on purpose — `-i <thread> -o <infra>` rather than the broader `-o <infra>` used on the IPv4 side. Aligning IPv4 with that would be a separate discussion; this PR only adds what is missing.

### Testing

`shellcheck -s bash run finish` is clean for the changed lines (it flags one pre-existing `SC2015` on the unrelated `mkdir … && … || die` line, left untouched).

Functional test against real `ip6tables` in an isolated network namespace, exercising both chain scenarios and the transition between them:

```
### case A: DOCKER-USER exists ###
  DOCKER-USER: 1   FORWARD: 0     (add run three times — idempotent)
  rule sits ahead of the chain's RETURN
  after finish: DOCKER-USER=0  FORWARD=0

### case B: no DOCKER-USER ###
  FORWARD: 1                      (add run twice — idempotent)
  after finish: FORWARD=0

### case C: added to FORWARD, DOCKER-USER appears afterwards ###
  before finish: DOCKER-USER=0  FORWARD=1
  after finish:  DOCKER-USER=0  FORWARD=0
```

Case C is why `finish` tries both.

### Environment this was found on

| | |
|---|---|
| Host | Debian 12, kernel 6.1, x86-64, Docker 27.4.1 |
| Compose | `network_mode: host`, `privileged: true` |
| RCP | SMLIGHT SLZB-06M (EFR32MG21) over USB |

### DIFF
--- a/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
+++ b/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
@@ -36,6 +36,9 @@ readonly OT_INFRA_IF
 OT_FORWARD_INGRESS_CHAIN="OT_FORWARD_INGRESS"
 readonly OT_FORWARD_INGRESS_CHAIN
 
+OT_DOCKER_USER_CHAIN="DOCKER-USER"
+readonly OT_DOCKER_USER_CHAIN
+
 if test "$1" -eq 256 ; then
   e=$((128 + $2))
 else
@@ -44,35 +47,44 @@ fi
 
 echo "otbr-agent exited with code ${1} (by signal ${2})."
 
+# `|| break` rather than `|| true`: a destroy that keeps failing leaves the set
+# in place, `ipset list -n` keeps succeeding, and the loop never ends -- hanging
+# container shutdown, since finish runs on the way out.
 ipset_destroy_if_exist()
 {
     while ipset list -n "$1" 2> /dev/null; do
-        ipset destroy "$1" || true
+        ipset destroy "$1" || break
     done
 }
 
-# Remove every copy of an iptables rule. Older images appended these rules
-# without checking for duplicates, so a host may carry several of them.
+# Remove every copy of a rule, for either address family. Older images appended
+# some of these without checking for duplicates, so a host may carry several.
 #
 # `-w` keeps a busy xtables lock from ending the loop early, and `|| break`
 # catches everything else: without it a failing `-D` leaves the rule in place,
 # `-C` keeps succeeding, and this spins forever -- hanging container shutdown,
 # since finish runs on the way out.
-iptables_delete_all()
+xtables_delete_all()
 {
-    local table="$1"
-    shift
+    local cmd="$1"
+    local table="$2"
+    shift 2
 
-    while iptables -t "${table}" -w -C "$@" 2> /dev/null; do
-        iptables -t "${table}" -w -D "$@" || break
+    while "${cmd}" -t "${table}" -w -C "$@" 2> /dev/null; do
+        "${cmd}" -t "${table}" -w -D "$@" || break
     done
 }
 
-while ip6tables -w -C FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; do
-    ip6tables -w -D FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}" || break
+xtables_delete_all ip6tables filter FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}"
+
+# The egress rule. Both chains are tried because which one run picked depends on
+# whether DOCKER-USER existed at the time, and that can differ between start and
+# stop.
+for chain in "${OT_DOCKER_USER_CHAIN}" FORWARD; do
+    xtables_delete_all ip6tables filter "${chain}" -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT
 done
 
-if ip6tables -L "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; then
+if ip6tables -w -L "${OT_FORWARD_INGRESS_CHAIN}" -n > /dev/null 2>&1; then
     ip6tables -w -F "${OT_FORWARD_INGRESS_CHAIN}"
     ip6tables -w -X "${OT_FORWARD_INGRESS_CHAIN}"
 fi
@@ -82,10 +94,10 @@ ipset_destroy_if_exist otbr-ingress-deny-src-swap
 ipset_destroy_if_exist otbr-ingress-allow-dst
 ipset_destroy_if_exist otbr-ingress-allow-dst-swap
 
-iptables_delete_all mangle PREROUTING -i "${OT_THREAD_IF}" -j MARK --set-mark 0x1001
-iptables_delete_all nat POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
-iptables_delete_all filter FORWARD -o "${OT_INFRA_IF}" -j ACCEPT
-iptables_delete_all filter FORWARD -i "${OT_INFRA_IF}" -j ACCEPT
+xtables_delete_all iptables mangle PREROUTING -i "${OT_THREAD_IF}" -j MARK --set-mark 0x1001
+xtables_delete_all iptables nat POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
+xtables_delete_all iptables filter FORWARD -o "${OT_INFRA_IF}" -j ACCEPT
+xtables_delete_all iptables filter FORWARD -i "${OT_INFRA_IF}" -j ACCEPT
 
 echo "OpenThread firewall rules removed."
 
--- a/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
+++ b/etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
@@ -54,6 +54,9 @@ readonly OT_REST_LISTEN_PORT
 OT_FORWARD_INGRESS_CHAIN="OT_FORWARD_INGRESS"
 readonly OT_FORWARD_INGRESS_CHAIN
 
+OT_DOCKER_USER_CHAIN="DOCKER-USER"
+readonly OT_DOCKER_USER_CHAIN
+
 die()
 {
     echo >&2 "ERROR: $*"
@@ -102,6 +105,26 @@ ip6tables -w -A "${OT_FORWARD_INGRESS_CHAIN}" -m set --match-set otbr-ingress-de
 ip6tables -w -A "${OT_FORWARD_INGRESS_CHAIN}" -m set --match-set otbr-ingress-allow-dst dst -j ACCEPT
 ip6tables -w -A "${OT_FORWARD_INGRESS_CHAIN}" -m pkttype --pkt-type unicast -j DROP
 ip6tables -w -A "${OT_FORWARD_INGRESS_CHAIN}" -j ACCEPT
+
+# Traffic leaving the Thread network needs an explicit accept. The ingress chain
+# above only filters what the host already forwards; it does nothing for the
+# other direction, which until now relied on the FORWARD policy being ACCEPT.
+# That does not hold under Docker, which manages ip6tables and sets the policy
+# to DROP. The result is a one-directional, silent failure: requests into the
+# mesh work, every reply out of it is dropped, and the border router looks
+# perfectly healthy throughout.
+#
+# DOCKER-USER is the chain Docker reserves for user rules and does not rewrite
+# when it rebuilds its own; fall back to FORWARD where it does not exist.
+if ip6tables -w -L "${OT_DOCKER_USER_CHAIN}" -n > /dev/null 2>&1; then
+    egress_chain="${OT_DOCKER_USER_CHAIN}"
+else
+    egress_chain="FORWARD"
+fi
+readonly egress_chain
+
+ip6tables -w -C "${egress_chain}" -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT 2> /dev/null ||
+    ip6tables -w -I "${egress_chain}" 1 -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT
 
 echo "Configuring OpenThread NAT64..."
 


etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
```diff
@@ -48,8 +54,34 @@ ipset_destroy_if_exist()
...
+done
+
+# The egress rule. Both chains are tried because which one run picked depends on
+# whether DOCKER-USER existed at the time, and that can differ between start and
+# stop.
+for chain in "${OT_DOCKER_USER_CHAIN}" FORWARD; do
+    while ip6tables -w -C "${chain}" -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT 2> /dev/null; do
+        ip6tables -w -D "${chain}" -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT || break
+    done
 done
```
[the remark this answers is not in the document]
skirmess: Adopted in 35882d4 — deletion is position-independent, so this generalises cleanly and replaces the helper plus both hand-written loops.

Verified against real tables in an isolated network namespace, with duplicates on both families:

```
before:  v6 jump: 2   v6 egress: 1   v4 nat: 3
after :  v6 jump: 0   v6 egress: 0   v4 nat: 0
         DOCKER-USER still has its RETURN: 1
```

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -54,12 +54,33 @@ readonly OT_REST_LISTEN_PORT
...
+# `-w` is required, not cosmetic: without it a busy xtables lock makes `-C`
+# fail, which is indistinguishable from "rule not present" and falls through to
+# `-A` -- the guard would then create the very duplicate it exists to prevent.
+iptables_add_if_missing()
+{
+    local table="$1"
+    shift
+
+    iptables -t "${table}" -w -C "$@" 2> /dev/null || iptables -t "${table}" -w -A "$@"
+}
```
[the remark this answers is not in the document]
skirmess: Not taking this one, and the reason is a functional one rather than a style preference: the helper appends with `-A`, but the two IPv6 rules are inserted with `-I <chain> 1`, and for the egress rule that difference decides whether it works at all.

`DOCKER-USER` ends with a terminating `RETURN`, which Docker puts there. Appending places the rule behind it, where it is never reached:

```
### -I <chain> 1  (this PR)          ### -A <chain>  (as suggested)
  -N DOCKER-USER                       -N DOCKER-USER
  -A ... -i wpan0 -o eth0 -j ACCEPT    -A DOCKER-USER -j RETURN
  -A DOCKER-USER -j RETURN             -A ... -i wpan0 -o eth0 -j ACCEPT   <- dead
```

A generic add helper would have to carry the insert position as well, at which point it is no longer the simplification this is aiming for. The IPv6 chain creation next to it is a third shape again (`-N` or flush), so `iptables_add_if_missing` covers exactly the rules it can.

The deletion side has no such constraint and is generalised in 35882d4.

skirmess:
Took half of the bot's suggestion in 35882d4.

**Generalised the deletion helper** to `xtables_delete_all <cmd> <table> …`, which replaces `iptables_delete_all` plus the two hand-written `ip6tables` loops. Deletion is position-independent, so both address families genuinely share one implementation now.

**Left the add helper alone**, because that one is not equivalent: it appends with `-A`, while the IPv6 jump and the egress rule are inserted with `-I <chain> 1`. `DOCKER-USER` ends with a terminating `RETURN`, so an appended egress rule lands behind it and is never reached — silently. Demonstrated inline on the review comment. A generic version would need the insert position as a parameter, and the IPv6 chain creation beside it is a third shape again, so it would not end up simpler.

`shellcheck -s bash run finish` is clean for the changed lines, and the deletion path is re-verified against real tables in an isolated netns with duplicates on both families.

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
```diff
@@ -51,25 +54,31 @@ ipset_destroy_if_exist()
...
+xtables_delete_all ip6tables filter FORWARD -o "${OT_THREAD_IF}" -j "${OT_FORWARD_INGRESS_CHAIN}"
+
+# The egress rule. Both chains are tried because which one run picked depends on
+# whether DOCKER-USER existed at the time, and that can differ between start and
+# stop.
+for chain in "${OT_DOCKER_USER_CHAIN}" FORWARD; do
+    xtables_delete_all ip6tables filter "${chain}" -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT
 done
 
 if ip6tables -L "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; then
```
jwhui: Add `-w` here too?

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
jwhui: While this is pre-existing code, note that if ipset destroy fails, `|| true` will cause while ipset list " to loop indefinitely on exit.

Using `|| break` here would prevent any potential hang on container shutdown.

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
skirmess: Added in f12dd153. That was the last call without it — I re-checked every `iptables`/`ip6tables` invocation in both `run` and `finish`, and this test was the only one left; the flush and delete right below it already had `-w`.

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
skirmess: Good catch — that is the same shape the rule removal loop was fixed for, and I did not look past the rules. Changed to `|| break` in f12dd153.

Verified with a stub where the set exists and `destroy` always fails, which is exactly the state you describe:

```
|| break (new):  returns
|| true  (old):  had to be killed after 5 s
```

The counter-test matters here: without it the test would pass without distinguishing the two.

Also rebased onto main, and `shellcheck -s bash run finish` reports nothing on the changed lines (the one SC2015 it does report is pre-existing, in the `mkdir -p /data/thread` line, and untouched by this PR).

jwhui:
@skirmess , #3517 is merged. Please rebase this PR.

jwhui:
@skirmess , are you intending to keep the commits separate? If not, can you squash them and update the commit message? Otherwise, can you put the PR number each of the commit titles? Thanks!

etc/docker/border-router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/finish
```diff
@@ -44,35 +47,44 @@ fi
...
+
+# The egress rule. Both chains are tried because which one run picked depends on
+# whether DOCKER-USER existed at the time, and that can differ between start and
+# stop.
+for chain in "${OT_DOCKER_USER_CHAIN}" FORWARD; do
+    xtables_delete_all ip6tables filter "${chain}" -i "${OT_THREAD_IF}" -o "${OT_INFRA_IF}" -j ACCEPT
 done
 
-if ip6tables -L "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; then
+if ip6tables -w -L "${OT_FORWARD_INGRESS_CHAIN}" 2> /dev/null; then
```
jwhui: `ip6tables -L` outputs the chain listing to `stdout` when the chain exists. Should we redirect `stdout` to `/dev/null` and add `-n` (numeric output to skip DNS resolution), consistent with `run`?

```suggestion
if ip6tables -w -L "${OT_FORWARD_INGRESS_CHAIN}" -n > /dev/null 2>&1; then
```

skirmess: Taken in 3a8e2005 — the two chain tests now read identically. `-L` prints the listing on success, so `2> /dev/null` alone left that on stdout during shutdown, and `-n` keeps it from resolving names in a path that only needs an existence check.

skirmess:
Squashed to a single commit (3a8e2005), with the message rewritten to cover the whole change rather than just the original one.

Also took the `-n > /dev/null 2>&1` suggestion, so the ingress chain test in `finish` now matches the DOCKER-USER test in `run`.

Diffing the squashed tree against the previous head shows that one line as the only difference, so nothing else moved in the process. `shellcheck -s bash run finish` is clean apart from the pre-existing SC2015 on the `mkdir -p /data/thread` line.

jwhui (approved):
Thanks! 👍🏼

Issue #3463, opened 2026-07-10
otbr-agent 0.3.0 (2026-07-09 :latest) crashes at startup: InitMulticastRouterSock "Cannot assign requested address" (EADDRNOTAVAIL) with Backbone Router enabled


MarlinFisher:
## Describe the bug

`otbr-agent` crashes ~15 s after every start with a fatal `[C]` critical in the Backbone
Router multicast-routing socket init, and is not restarted:

```
otbr-agent[...]: 00:00:14.798 [C] Platform------: InitMulticastRouterSock() at multicast_routing.cpp:244: Cannot assign requested address
```

`Cannot assign requested address` = `EADDRNOTAVAIL`, which the Linux kernel returns from
`MRT6_ADD_MIF` (`ip6mr` `mif6_add()` → `dev_get_by_index()` returns NULL). Observed process
exit code: 5.

This is a **regression**: the previous `openthread/otbr:latest` ran indefinitely with the
identical compose config, host, kernel, and RCP. The only change was the image update to the
build dated **2026-07-09**. Because the container keeps `otbr-web` running, the container stays
"up" and the failure is silent (Thread simply goes down).

## Environment

- otbr-agent version: **0.3.0**, from `openthread/otbr:latest` @ `sha256:7e8bf38844a64947c761fd852b9f4ce7c96a8f546abe795c200506bd56343060` (built 2026-07-09)
- Thread version: 1.4.0
- RCP firmware: `SL-OPENTHREAD/2.4.4.0_GitHub-7074a43e4; EFR32` (Sonoff Zigbee 3.0 Dongle Plus V2 flashed with OpenThread RCP)
- Host: Ubuntu, kernel **6.8.0-124-generic** (x86_64)
- Docker, `network_mode: host`, `privileged: true`
- otbr-agent opts: `-I wpan0 -B enp7s0 -d7 --rest-listen-address 0.0.0.0 spinel+hdlc+uart:///dev/ttyUSB1?uart-baudrate=460800`
- Backbone interface `enp7s0`: global + link-local IPv6 present; IPv6 not disabled; `net.ipv6.conf.all.forwarding = 1`
- Kernel config: `CONFIG_IPV6_MROUTE=y`, `CONFIG_IPV6_MROUTE_MULTIPLE_TABLES=y`

## To Reproduce

1. Run the 2026-07-09 `openthread/otbr:latest` build with a backbone interface (`-B <iface>`) on a host-network container.
[...]

PR #3523, opened 2026-08-15
[docker] add stable tag to border-router Docker image

drudchenko:
The `latest` tag on [`openthread/border-router`](https://hub.docker.com/r/openthread/border-router) currently tracks every push to `main`, so it can point at unreleased commits in between monthly CalVer releases. Users who want a released image must therefore track the exact CalVer tag (e.g. `v2026.08.0`), and have no fixed rollback target once `latest` regresses — a gap called out by a user in #3463:

> Docker Hub publishes only `openthread/otbr:latest` (no version tags), so downstream users have no clean rollback target once `:latest` regresses.

This PR adds a `stable` tag to the border-router image, applied under the same condition as the CalVer `release_tag` input, so that it only moves when `monthly-release.yml` cuts a release.

## Changes

- Add a `type=raw,value=stable` entry to `docker/metadata-action` in the `merge` job, enabled only when the `release_tag` input is provided.

## Resulting tag behaviour

| Trigger | Tags pushed |
| --- | --- |
| Push to `main` | `main`, `latest`, `sha-<short>` |
| Push to other branch | `<branch>`, `sha-<short>` |
| Manual `v*` tag push | `v2026.08.1`, `sha-<short>` |
| Monthly release (cron or `workflow_dispatch` + `tag_override`) | `main`, `latest`, `sha-<short>`, `v2026.09.0`, **`stable`** |

On release day `latest` and `stable` briefly point at the same digest; `latest` then moves ahead with the next merge to `main` while `stable` stays pinned until the next monthly cut.

## Note for reviewers

A manually pushed `v*` tag deliberately does **not** move `stable`, since the documented release path — including ad-hoc releases via `monthly-release.yml`'s `tag_override` dispatch — always flows through the `release_tag` input. If you would prefer hand-pushed tags to advance `stable` as well, the condition can be widened to:

```yaml
type=raw,value=stable,enable=${{ github.ref_type == 'tag' || (inputs.release_tag != '' && inputs.release_tag != null) }}
```

(the `push:` filter already restricts this workflow to `v*` tags, so `ref_type` alone is sufficient). Happy to make that change if preferred.

### DIFF
--- a/.github/workflows/docker-border-router.yml
+++ b/.github/workflows/docker-border-router.yml
@@ -180,6 +180,7 @@ jobs:
             type=raw,value=latest,enable={{is_default_branch}}
             type=sha,prefix=sha-,format=short
             type=raw,value=${{ inputs.release_tag || 'none' }},enable=${{ inputs.release_tag != '' && inputs.release_tag != null }}
+            type=raw,value=stable,enable=${{ inputs.release_tag != '' && inputs.release_tag != null }}
 
       - name: Create manifest list and push
         working-directory: ${{ runner.temp }}/digests


drudchenko:
@googlebot I signed it!

jwhui (approved):

PR #3524, opened 2026-08-15
[rest] fix PUT errors masked as 405 by RoutingErrorHandler

### LINKED ISSUE #3522: [rest] Errors in PUT method calls masked by RoutingErrorHandler and turned into 405
**Describe the bug**

Any PUT method call on the REST API that creates an error (status >= 400) will be turned to 405 by RoutingErrorHandler because of the missing 
```
    case HttpMethod::kPut:
        // fallthrough
```

**To Reproduce**

call e.g.
```
$ curl -i -XPUT -d'enable' http://192.168.1.22:8081/node/ba-epskc/state
HTTP/1.1 405 Method Not Allowed
Content-Type: application/json
Allow: GET, POST, DELETE, OPTIONS
Content-Length: 85
Access-Control-Allow-Origin: *
Keep-Alive: timeout=5, max=100
Access-Control-Allow-Methods: DELETE, GET, OPTIONS, POST
Access-Control-Allow-Headers: Origin, Accept, X-Requested-With, Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers

{
	"title":	"Method Not Allowed",
	"status":	405,
	"detail":	"method not supported"
}
```
instead of
```
curl -i -XPUT -d'"enable"' http://192.168.1.22:8081/node/ba-epskc/state
HTTP/1.1 200 OK
Content-Length: 0
Access-Control-Allow-Origin: *
Keep-Alive: timeout=5, max=100
Access-Control-Allow-Methods: DELETE, GET, OPTIONS, POST
Access-Control-Allow-Headers: Origin, Accept, X-Requested-With, Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers
```

**Expected behavior**

The causing error/status code should be returned instead of 405 - Method not allowed

**Further Context**

This is especially confusing as the PUT Method is also missing in
https://github.com/openthread/ot-br-posix/blob/main/src/rest/rest_web_server.cpp#L75
```
 #ifndef OTBR_REST_ACCESS_CONTROL_ALLOW_METHODS
-#define OTBR_REST_ACCESS_CONTROL_ALLOW_METHODS "DELETE, GET, OPTIONS, POST"
+#define OTBR_REST_ACCESS_CONTROL_ALLOW_METHODS "DELETE, GET, OPTIONS, POST, PUT"
 #endif
```
and https://github.com/openthread/ot-br-posix/blob/main/src/rest/rest_web_server.cpp#L1260
```
-        aResponse.set_header("Allow", "GET, POST, DELETE, OPTIONS");
+        aResponse.set_header("Allow", "GET, POST, PUT, DELETE, OPTIONS");
```
and hence additionally obfuscates the output of the 405 output (see example above).

lboue:
## Description

`RoutingErrorHandler` treats any `PUT` request as an unsupported method, forcing its status to `405`. Since `405` is only overridden when the handler's real status is *higher* than `405`, a genuine error below `405` (e.g. `400 Bad Request`) returned by a `PUT` handler such as `SetDataState` gets silently replaced by `405`, hiding the actual failure reason from the client.

```
$ curl -i -XPUT -d'enable' http://192.168.1.22:8081/node/state
HTTP/1.1 405 Method Not Allowed
...
{"title": "Method Not Allowed", "status": 405, "detail": "method not supported"}
```

`enable` here is invalid (must be the JSON string `"enable"`), so the correct response is `400 Bad Request`, not `405`.

## Changes

- Fold `kPut` into the same fallthrough group as `kGet`/`kPost`/`kDelete`/`kOptions` in `RoutingErrorHandler`, so it takes a plain `break` instead of falling into `default`. This lets a `PUT` with a real handler behind it (e.g. `/node/state`, `/node/dataset/active`, `/node/dataset/pending`, `/node/commissioner/state`, or `/node/ba-epskc/state` when `OTBR_ENABLE_EPSKC` is set) return whatever status that handler set, while a `PUT` to a path with no `PUT` route at all still falls through cpp-httplib's own routing-miss `404`, consistent with `GET`/`POST`/`DELETE` on the same kind of path.
- Add `PUT` to `OTBR_REST_ACCESS_CONTROL_ALLOW_METHODS` and to the `Allow` header returned alongside a genuine `405`, since it's a legitimately supported method on the REST API.

Per-resource `Allow` headers and the broader `404`/`405`/`OPTIONS` semantics cleanup are tracked separately by #3529.

## Testing

- Verified with `curl` against a live `otbr-agent` (EFR32 RCP, real Thread network):
  - `PUT /node/state` with an invalid body -> `400` (was `405` before the fix)
  - `PUT /node/state` with a valid body -> `200`
  - `PATCH /node/state` (genuinely unsupported method) -> `405`
- Added regression tests in `tests/rest/test_rest.py` covering the above plus `PUT /node/ext-address` (GET-only, no `PUT` route registered) -> `404` (matches `POST`/`DELETE` on the same path); full suite passes against the same live setup (`python3 tests/rest/test_rest.py`).
- `script/clang-format-check src/rest/rest_web_server.cpp` passes.

Fixes #3522

### DIFF
--- a/src/rest/rest_web_server.cpp
+++ b/src/rest/rest_web_server.cpp
@@ -72,7 +72,7 @@
 #endif
 
 #ifndef OTBR_REST_ACCESS_CONTROL_ALLOW_METHODS
-#define OTBR_REST_ACCESS_CONTROL_ALLOW_METHODS "DELETE, GET, OPTIONS, POST"
+#define OTBR_REST_ACCESS_CONTROL_ALLOW_METHODS "DELETE, GET, OPTIONS, POST, PUT"
 #endif
 
 #define OT_REST_RESOURCE_PATH_NODE "/node"
@@ -1250,6 +1250,8 @@ void RestWebServer::RoutingErrorHandler(const Request &aRequest, Response &aResp
         // fallthrough
     case HttpMethod::kGet:
         // fallthrough
+    case HttpMethod::kPut:
+        // fallthrough
     case HttpMethod::kDelete:
         // fallthrough
     case HttpMethod::kOptions:
@@ -1257,7 +1259,7 @@ void RestWebServer::RoutingErrorHandler(const Request &aRequest, Response &aResp
     default:
         errorDetails = "method not supported";
         error        = StatusCode::MethodNotAllowed_405;
-        aResponse.set_header("Allow", "GET, POST, DELETE, OPTIONS");
+        aResponse.set_header("Allow", "GET, POST, PUT, DELETE, OPTIONS");
         break;
     }
 
--- a/tests/rest/test_rest.py
+++ b/tests/rest/test_rest.py
@@ -389,6 +389,65 @@ def node_state_test_attached():
             time.sleep(1)
 
 
+def node_state_put_invalid_body_test():
+    """Regression test for https://github.com/openthread/ot-br-posix/issues/3522
+
+    A PUT request that fails to process (e.g. an invalid body) must return
+    its real error status (400 Bad Request here) instead of being masked as
+    405 Method Not Allowed by RoutingErrorHandler.
+    """
+    url = rest_api_addr + "/node/state"
+
+    # Not valid JSON (missing quotes around the string value), so the
+    # handler rejects it with 400 before touching any device state.
+    request = urllib.request.Request(url, data=b'enable', method='PUT')
+
+    try:
+        urllib.request.urlopen(request)
+        assert False
+    except urllib.error.HTTPError as e:
+        assert (e.code == 400)
+
+    print(" PUT /node/state (invalid body) : status {}, expected 400".format(400))
+
+
+def node_state_put_unsupported_method_test():
+    """A method that genuinely isn't supported (e.g. PATCH) on a PUT-capable
+    endpoint must still return 405.
+    """
+    url = rest_api_addr + "/node/state"
+
+    request = urllib.request.Request(url, method='PATCH')
+
+    try:
+        urllib.request.urlopen(request)
+        assert False
+    except urllib.error.HTTPError as e:
+        assert (e.code == 405)
+
+    print(" PATCH /node/state : status {}, expected 405".format(405))
+
+
+def node_ext_address_put_no_route_test():
+    """A PUT to an endpoint that never registers a PUT handler (e.g.
+    /node/ext-address, GET-only) returns 404, the same routing-miss status
+    POST and DELETE already return on that path. Full 404/405 semantics
+    (a per-resource Allow header, etc.) are tracked by
+    https://github.com/openthread/ot-br-posix/issues/3529.
+    """
+    url = rest_api_addr + "/node/ext-address"
+
+    request = urllib.request.Request(url, data=b'"0000000000000000"', method='PUT')
+
+    try:
+        urllib.request.urlopen(request)
+        assert False
+    except urllib.error.HTTPError as e:
+        assert (e.code == 404)
+
+    print(" PUT /node/ext-address (no route) : status {}, expected 404".format(404))
+
+
 def node_network_name_test(thread_num):
     url = rest_api_addr + "/node/network-name"
 
@@ -512,6 +571,10 @@ def epskc_test():
         except urllib.error.HTTPError as err:
             assert err.code == code
 
+    # Regression test for https://github.com/openthread/ot-br-posix/issues/3522: an invalid
+    # PUT body must return its real error status (400) rather than being masked as 405.
+    expect_http_error(400, lambda: put_state("toggle"))
+
     # Feature disabled: activation must be refused.
     put_state("disable")
     assert get_state() == "disabled"
@@ -561,6 +624,9 @@ def main():
     node_rloc16_test(200)
     node_ext_address_test(200)
     node_state_test(200)
+    node_state_put_invalid_body_test()
+    node_state_put_unsupported_method_test()
+    node_ext_address_put_no_route_test()
     node_network_name_test(200)
     node_state_test_attached()  # wait for attached state
     node_leader_data_test(200)


src/rest/rest_web_server.cpp
```diff
@@ -1082,6 +1093,14 @@ void RestWebServer::RoutingErrorHandler(const Request &aRequest, Response &aResp
...
     case HttpMethod::kOptions:
         break;
+    case HttpMethod::kPut:
+        if (kPutSupportedPaths.find(aRequest.path) == kPutSupportedPaths.end())
+        {
+            errorDetails = "method not supported";
+            error        = StatusCode::MethodNotAllowed_405;
+            aResponse.set_header("Allow", "GET, POST, DELETE, OPTIONS");
+        }
+        break;
```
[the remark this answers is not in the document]
3oris: I was just about to suggest the same @lboue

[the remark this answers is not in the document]
lboue: Fixed [e45d2f3](https://github.com/openthread/ot-br-posix/pull/3524/commits/e45d2f3c59835eee8873109b89376aee8124760c)

3oris:
@lboue :

`kPutSupportedPaths` is now missing the epskc case that you just added a couple of days ago. So I thought it might be handy to introduce a `RegisterPut`put warpper that will also maintain the allow list:

```
diff --git a/src/rest/rest_web_server.cpp b/src/rest/rest_web_server.cpp
index fc98f3dd..eae2c879 100644
--- a/src/rest/rest_web_server.cpp
+++ b/src/rest/rest_web_server.cpp
@@ -146,6 +146,12 @@ HttpMethod GetMethod(const Request &aRequest)
     return HttpMethod::kInvalidMethod;
 }
 
+void RestWebServer::RegisterPut(const std::string &aPath, httplib::Server::Handler aHandler)
+{
+    mPutSupportedPaths.insert(aPath);
+    mServer.Put(aPattern, std::move(aHandler));
+}
+
 RestWebServer::RestWebServer(Host::RcpHost &aHost)
     : mHost(aHost)
 {
@@ -153,7 +159,7 @@ RestWebServer::RestWebServer(Host::RcpHost &aHost)
     mServer.Delete(OT_REST_RESOURCE_PATH_NODE, MakeHandler(&RestWebServer::NodeInfo));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_BAID, MakeHandler(&RestWebServer::BaId));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_STATE, MakeHandler(&RestWebServer::State));
-    mServer.Put(OT_REST_RESOURCE_PATH_NODE_STATE, MakeHandler(&RestWebServer::State));
+    RegisterPut(OT_REST_RESOURCE_PATH_NODE_STATE, MakeHandler(&RestWebServer::State));
     mServer.Options(OT_REST_RESOURCE_PATH_NODE_STATE, MakeHandler(&RestWebServer::State));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_EXTADDRESS, MakeHandler(&RestWebServer::ExtendedAddr));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_NETWORKNAME, MakeHandler(&RestWebServer::NetworkName));
@@ -163,13 +169,13 @@ RestWebServer::RestWebServer(Host::RcpHost &aHost)
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_EXTPANID, MakeHandler(&RestWebServer::ExtendedPanId));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_RLOC, MakeHandler(&RestWebServer::Rloc));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_DATASET_ACTIVE, MakeHandler(&RestWebServer::DatasetActive));
-    mServer.Put(OT_REST_RESOURCE_PATH_NODE_DATASET_ACTIVE, MakeHandler(&RestWebServer::DatasetActive));
+    RegisterPut(OT_REST_RESOURCE_PATH_NODE_DATASET_ACTIVE, MakeHandler(&RestWebServer::DatasetActive));
     mServer.Options(OT_REST_RESOURCE_PATH_NODE_DATASET_ACTIVE, MakeHandler(&RestWebServer::DatasetActive));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING, MakeHandler(&RestWebServer::DatasetPending));
-    mServer.Put(OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING, MakeHandler(&RestWebServer::DatasetPending));
+    RegisterPut(OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING, MakeHandler(&RestWebServer::DatasetPending));
     mServer.Options(OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING, MakeHandler(&RestWebServer::DatasetPending));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_STATE, MakeHandler(&RestWebServer::CommissionerState));
-    mServer.Put(OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_STATE, MakeHandler(&RestWebServer::CommissionerState));
+    RegisterPut(OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_STATE, MakeHandler(&RestWebServer::CommissionerState));
     mServer.Options(OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_STATE, MakeHandler(&RestWebServer::CommissionerState));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_JOINER, MakeHandler(&RestWebServer::CommissionerJoiner));
     mServer.Post(OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_JOINER, MakeHandler(&RestWebServer::CommissionerJoiner));
@@ -179,7 +185,7 @@ RestWebServer::RestWebServer(Host::RcpHost &aHost)
 
 #if OTBR_ENABLE_EPSKC
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_BA_EPSKC_STATE, MakeHandler(&RestWebServer::EpskcState));
-    mServer.Put(OT_REST_RESOURCE_PATH_NODE_BA_EPSKC_STATE, MakeHandler(&RestWebServer::EpskcState));
+    RegisterPut(OT_REST_RESOURCE_PATH_NODE_BA_EPSKC_STATE, MakeHandler(&RestWebServer::EpskcState));
     mServer.Options(OT_REST_RESOURCE_PATH_NODE_BA_EPSKC_STATE, MakeHandler(&RestWebServer::EpskcState));
     mServer.Get(OT_REST_RESOURCE_PATH_NODE_BA_EPSKC_KEY, MakeHandler(&RestWebServer::EpskcKey));
     mServer.Post(OT_REST_RESOURCE_PATH_NODE_BA_EPSKC_KEY, MakeHandler(&RestWebServer::EpskcKey));
@@ -1254,6 +1260,9 @@ void RestWebServer::RoutingErrorHandler(const Request &aRequest, Response &aResp
         // fallthrough
     case HttpMethod::kOptions:
         break;
+    case HttpMethod::kPut:
+        if (mPutSupportedPaths.find(aRequest.path) != mPutSupportedPaths.end()) break;
+        // fallthrough
     default:
         errorDetails = "method not supported";
         error        = StatusCode::MethodNotAllowed_405;
diff --git a/src/rest/rest_web_server.hpp b/src/rest/rest_web_server.hpp
index 9239ae08..b854a471 100644
--- a/src/rest/rest_web_server.hpp
+++ b/src/rest/rest_web_server.hpp
@@ -42,6 +42,7 @@
 #include <netinet/ip.h>
 #include <sys/socket.h>
 
+#include <set>
 #include <unordered_map>
 #include <vector>
 
@@ -207,12 +208,16 @@ private:
         };
     }
 
+    void RegisterPut(const std::string &aPath, httplib::Server::Handler aHandler);
+
     Host::RcpHost &mHost;
 
     httplib::Server mServer;
     std::thread     mServerThread;
 
     Services mServices;
+
+    std::set<std::string> mPutSupportedPaths;
 };
 
 } // namespace rest
```

src/rest/rest_web_server.cpp
```diff
@@ -1084,162 +1066,19 @@ void RestWebServer::CoprocessorVersion(const Request &aRequest, Response &aRespo
...
+    // Paths that register a PUT handler. A PUT to any other path never reaches an application
+    // handler, so aResponse.status here is cpp-httplib's own routing-miss default (404) rather
+    // than a status set by a handler - in that case it should be reported as 405, not passed
+    // through untouched like a genuine handler error would be.
+    static const std::set<std::string> kPutSupportedPaths = {
+        OT_REST_RESOURCE_PATH_NODE_STATE,
+        OT_REST_RESOURCE_PATH_NODE_DATASET_ACTIVE,
+        OT_REST_RESOURCE_PATH_NODE_DATASET_PENDING,
+        OT_REST_RESOURCE_PATH_NODE_COMMISSIONER_STATE,
+    };
```
3oris: If you want to keep it like this, maybe add
```
#if OTBR_ENABLE_EPSKC
        OT_REST_RESOURCE_PATH_NODE_BA_EPSKC_STATE,
#endif
```
to the allow list. But this is why I would suggest introducing a wrapper function

```
   void RegisterPut(const std::string &aPattern, httplib::Server::Handler aHandler)
   {
       mPutSupportedPaths.insert(aPattern);
       mServer.Put(aPattern, std::move(aHandler));
   }
```

See https://github.com/openthread/ot-br-posix/pull/3524#issuecomment-5303473435

3oris (commented):
looks good to me

src/rest/rest_web_server.cpp
```diff
@@ -1254,6 +1260,17 @@ void RestWebServer::RoutingErrorHandler(const Request &aRequest, Response &aResp
         // fallthrough
     case HttpMethod::kOptions:
         break;
+    case HttpMethod::kPut:
```
jwhui: The root cause of #3522 is simply that `HttpMethod::kPut` was missing from switch statement here, causing all PUT requests that hit the error handler to fall into default.

Can we simplify RoutingErrorHandler by treating `kPut` just like `kPost`, `kGet`, `kDelete`, and `kOptions`?

src/rest/rest_web_server.cpp
jwhui: PUT should also be included in the Allow header when returning 405 Method Not Allowed.

src/rest/rest_web_server.cpp
3oris: @jwhui:

Probably yes. What you suggest would fix the originally reported masking issue.

But looking at `RoutingErrorHandler`'s implementation, it became clear that the current implementation wrt Allow headers and the OPTIONS verb is semantically "incomplete".  

- 404 should be issued if the queried resource (resolved route) doesn't exist on the server at all.
- 405 should be issued if the queried resource exists but not supported for the queried method.

In the 405 case an Allow header has to be returned listing all the supported methods for the queried resource, meaning a static list as in `aResponse.set_header("Allow", "GET, POST, DELETE, OPTIONS, PUT");` is in the majority of cases the wrong answer.

The same goes for OPTIONS which is partially implemented, and then only sometimes implements the Allow header.

Would it make sense to open another issue to fix the 404/405/OPTIONS semantics or is it maybe just not important enough?

lboue: Good question — I did consider that, but a plain fallthrough would lose the distinction between two different situations that land in this handler as `kPut`:
- a genuine handler error on a PUT-supported path (e.g. `SetDataState` returning `400`), where `aResponse.status` was actually set by the handler and must pass through untouched;
- a PUT to a path that never registers a PUT handler at all (e.g. `/node/ext-address`), where the request never reaches an application handler and `aResponse.status` is just cpp-httplib's own routing-miss default (`404`) — not a real error status, and it shouldn't leak through as `404` instead of `405`.

Treating `kPut` exactly like `kGet`/`kPost`/`kDelete` (plain `break`) would fix the first case but regress the second back to `404`, which is what the added `node_ext_address_put_not_supported_test` regression test checks for. The `mPutSupportedPaths` allowlist (populated automatically via `RegisterPut`, per @3oris's suggestion) is what lets us tell the two cases apart. Happy to simplify further if you feel returning `404` for that second case is acceptable. Let me know.

Also fixed the missing `PUT` in the `Allow` header on 405 responses in 92b1e4f3, thanks for catching that.

lboue: @3oris agreed, that's a real and separate concern from what #3522/this PR set out to fix. A correct per-resource `Allow` header (and consistent 404 vs 405 semantics, plus a complete `OPTIONS` implementation) would need each route's actual method set to be introspectable from the router, which is a bigger refactor than this PR's scope.

I'll open a follow-up issue to track it here: https://github.com/openthread/ot-br-posix/issues/3529

src/rest/rest_web_server.cpp
lboue: Thanks. Fixed now.

src/rest/rest_web_server.cpp
lboue: Filed #3529 to track this.

3oris: > Good question — I did consider that, but a plain fallthrough would lose the distinction between two different situations that land in this handler as `kPut`:
> 
...
> * a PUT to a path that never registers a PUT handler at all (e.g. `/node/ext-address`), where the request never reaches an application handler and `aResponse.status` is just cpp-httplib's own routing-miss default (`404`) — not a real error status, and it shouldn't leak through as `404` instead of `405`.

Exactly, the assumption here was that the `RoutingErrorHandler` would also be used as some sort of "catch-all rewrite 404 to 405" in the PUT Case.

This should probably be done more explicitly anyway by asking
```
if (StatusCode(aResponse.status) == StatusCode::NotFound_404)
```
which would scope it to the cases yielding from httplibs inability to make the correct distinction, and allow for addressing the issue not only for PUT.

jwhui: Thanks for submitting #3529. With that, should we simplify this PR and remove `mPutSupportedPaths` and `RegisterPut`?

lboue: Done in fb3f7c00 .Replaced mPutSupportedPaths/RegisterPut() with a direct check of aResponse.status != StatusCode::NotFound_404 in the kPut case.

Traced through cpp-httplib's Server::routing()/process_request in httplib.h: when no handler matches for the method+path, res.status is left at its default (-1) by dispatch_request, and the framework then explicitly sets it to 404 right before invoking error_handler_. None of the PUT handlers (State, DatasetActive, DatasetPending, CommissionerState, EpskcState) ever set 404 themselves, so aResponse.status == 404 reliably identifies "PUT hit no registered handler" the same way the allowlist did without needing every PUT route registered through a dedicated wrapper.

Rebuilt the otbr-rest target locally to confirm it still compiles cleanly after the removal.

src/rest/rest_web_server.hpp
```diff
@@ -42,6 +42,7 @@
 #include <netinet/ip.h>
 #include <sys/socket.h>
 
+#include <set>
```
jwhui: Can remove?

src/rest/rest_web_server.cpp
jwhui: Thanks for the update and continuing to work on this.

Checking `aResponse.status != StatusCode::NotFound_404` causes `PUT /nonexistent` to return 405 instead of 404, and would mask any genuine 404 handler errors as 405. It is also inconsistent with other methods (`POST` and `DELETE` on `/node/ext-address` return 404).

With #3529 tracking full 404/405 semantics, should we simply handle `kPut` with a plain `break` like the other methods?

src/rest/rest_web_server.hpp
lboue: Good catch — that one's genuinely dead now that mPutSupportedPaths is gone. Removed in ed12b037.

src/rest/rest_web_server.cpp
lboue: Agreed, done in ed12b037. Folded kPut into the same fallthrough group as kPost/kGet/kDelete/kOptions (plain break), so it just preserves whatever status the handler — or cpp-httplib's routing-miss default — already set, same as the other methods, no more special-casing on aResponse.status.

That still fixes the original #3522 masking bug (kPut was simply missing from the switch, so PUT errors fell into default and got forced to 405), while PUT /node/ext-address now correctly returns 404 like POST/DELETE on that path instead of being coerced to 405. Updated node_ext_address_put_not_supported_test to match (and renamed it to node_ext_address_put_no_route_test since it's now testing 404, not 405). The remaining 405/Allow-header question is exactly what #3529 will address.

tests/rest/test_rest.py
```diff
@@ -389,6 +389,65 @@ def node_state_test_attached():
...
+
+    request = urllib.request.Request(url, data=b'"0000000000000000"', method='PUT')
+
+    try:
+        urllib.request.urlopen(request)
+        assert False
+    except urllib.error.HTTPError as e:
+        assert (e.code == 404)
+
+    print(" PUT /node/ext-address (unsupported) : status {}, expected 404".format(404))
```
jwhui: nit: maybe more accurate?
```suggestion
    print(" PUT /node/ext-address (no route) : status {}, expected 404".format(404))
```

lboue: Done in 49f1d7a8.

3oris (commented):
LGTM

lboue:
Do I have to squash commits?

jwhui:
@lboue , yes please squash commits.

jwhui (approved):
Thanks! 👍🏼

PR #3526, opened 2026-08-15
[tests] fix epskc_test() port assertion against a real ephemeral port

### LINKED ISSUE #3525: [tests] epskc_test() port assertion fails against a real ephemeral port assignment
**Describe the bug**

`epskc_test()` in `tests/rest/test_rest.py` (added in #3480) asserts that the port returned by activating the ephemeral key matches the port read *before* activation:

https://github.com/openthread/ot-br-posix/blob/main/tests/rest/test_rest.py#L541

```python
status = get_key_status()
assert status["state"] == "stopped"
...
data = json.loads(urllib.request.urlopen(req).read())
assert len(data["tap"]) == 9 and data["tap"].isdigit() and data["port"] == status["port"]
```

While the feature is `stopped`, `GET /node/ba-epskc/key` reports `port: 0` (not yet assigned). The test then activates with that same `port: 0` in the request body:

https://github.com/openthread/ot-br-posix/blob/main/tests/rest/test_rest.py#L533-L539

`0` is treated as "auto-assign a port" by `otBorderAgentEphemeralKeyStart()`, so the server picks a real ephemeral port and returns it in `data["port"]`. Comparing that freshly-assigned port to the `0` placeholder captured while stopped will only ever match by coincidence.

**To Reproduce**

Run the REST test suite against a real border agent + RCP (not the simulated `ot-rcp` used in CI) and call `epskc_test()`:

```
$ curl -s http://127.0.0.1:8081/node/ba-epskc/key
{"state": "stopped", "port": 0}

$ curl -s -XPOST -d'{"lifetime": 30000, "port": 0}' -H "Content-Type: application/json" http://127.0.0.1:8081/node/ba-epskc/key
{"tap": "174863805", "port": 49157}
```

`49157 != 0`, so `test_rest.py`'s `epskc_test()` fails:

```
AssertionError
  File "tests/rest/test_rest.py", line 541, in epskc_test
    assert len(data["tap"]) == 9 and data["tap"].isdigit() and data["port"] == status["port"]
```

This reproduced consistently for me on real hardware (EFR32-based RCP), while apparently passing in CI against the simulated `ot-rcp`, suggesting an environment-dependent (possibly coincidental) ephemeral port assignment in the simulated path.

**Expected behavior**

The assertion should not require the newly-assigned port to equal the pre-activation placeholder `port: 0`. It should instead assert the post-activation port is a valid, non-zero UDP port (and, if desired, that the *status* endpoint reflects the same port once started), e.g.:

```python
status = get_key_status()
assert status["state"] == "stopped"
...
data = json.loads(urllib.request.urlopen(req).read())
assert len(data["tap"]) == 9 and data["tap"].isdigit()
assert 0 < data["port"] <= 65535

status = get_key_status()
assert status["state"] == "started" and status["port"] == data["port"]
```

**Further Context**

Found while validating https://github.com/openthread/ot-br-posix/pull/3524 against real hardware — unrelated to that PR's change, but surfaced by running the full REST test suite end-to-end instead of just the new regression checks.

lboue:
## Description

`epskc_test()` asserts that the port returned by activating the ephemeral key equals the port read *before* activation:

```python
status = get_key_status()
assert status["state"] == "stopped"
...
data = json.loads(urllib.request.urlopen(req).read())
assert len(data["tap"]) == 9 and data["tap"].isdigit() and data["port"] == status["port"]
```

While the feature is `stopped`, `GET /node/ba-epskc/key` reports `port: 0` (not yet assigned). The test then activates with that same `port: 0` in the request body, which `otBorderAgentEphemeralKeyStart()` treats as "auto-assign a port" - so the server picks a real ephemeral port and returns it. Comparing that freshly-assigned port to the `0` placeholder captured while stopped only matches by coincidence.

This reproduced consistently against a real EFR32-based RCP (`otbr-agent` returned port `49157` while the pre-activation placeholder was `0`), while apparently passing in CI against the simulated `ot-rcp`.

## Changes

- Replace the `data["port"] == status["port"]` comparison with a check that the returned port is a valid, non-zero UDP port.
- The existing post-activation check (`status["port"] == data["port"]`, right after) already covers the comparison that actually matters: that the *started* status reports the same port the activation response returned.
- Per review feedback, reuse the existing `post_key()` helper and pass `port: 0` directly instead of routing through the pre-activation `status` dict - this makes the "auto-assign a port" intent explicit and drops the data dependency on `status`:

```python
# Custom valid parameters (custom lifetime with auto-assigned port).
data = json.loads(post_key({"lifetime": 30000, "port": 0}).read())
assert len(data["tap"]) == 9 and data["tap"].isdigit()
assert 0 < data["port"] <= 65535
```

## Testing

Ran the full REST test suite twice in a row against a real border agent + EFR32 RCP (Home Assistant Connect ZBT-2):

```
$ python3 tests/rest/test_rest.py
...
 /node/ba-epskc : OK
```

Both runs exited 0. Without this fix, the suite consistently failed with:
```
AssertionError
  File "tests/rest/test_rest.py", line 541, in epskc_test
    assert len(data["tap"]) == 9 and data["tap"].isdigit() and data["port"] == status["port"]
```

Fixes #3525

### DIFF
--- a/tests/rest/test_rest.py
+++ b/tests/rest/test_rest.py
@@ -530,15 +530,10 @@ def epskc_test():
     expect_http_error(400, lambda: post_key({"port": "49191"}))
     expect_http_error(400, lambda: post_key({"lifetime": 4294967295}))
 
-    # Custom valid parameters should be accepted.
-    req = urllib.request.Request(
-        key_url,
-        data=json.dumps({"lifetime": 30000, "port": status["port"]}).encode(),
-        method='POST',
-    )
-    req.add_header('Content-Type', 'application/json')
-    data = json.loads(urllib.request.urlopen(req).read())
-    assert len(data["tap"]) == 9 and data["tap"].isdigit() and data["port"] == status["port"]
+    # Custom valid parameters (custom lifetime with auto-assigned port).
+    data = json.loads(post_key({"lifetime": 30000, "port": 0}).read())
+    assert len(data["tap"]) == 9 and data["tap"].isdigit()
+    assert 0 < data["port"] <= 65535
 
     status = get_key_status()
     assert status["state"] == "started" and status["port"] == data["port"]


tests/rest/test_rest.py
```diff
@@ -530,15 +530,19 @@ def expect_http_error(code, action):
...
+    # still "stopped"), which otBorderAgentEphemeralKeyStart() treats as "auto-assign a port" -
+    # so the response carries a freshly assigned port that has no reason to equal that 0
+    # placeholder. Only the post-activation status is expected to agree with it.
     req = urllib.request.Request(
         key_url,
         data=json.dumps({"lifetime": 30000, "port": status["port"]}).encode(),
         method='POST',
     )
     req.add_header('Content-Type', 'application/json')
     data = json.loads(urllib.request.urlopen(req).read())
```
jwhui: Passing `status["port"]` (which is `0` from the prior stopped state) creates a confusing data dependency on the pre-activation `status` dict and requires extra explanatory comments.

Can we reuse the `post_key()` helper defined on line 503 instead of manually setting up `urllib.request.Request`?

We could simplify to:
```python
    # Custom valid parameters (custom lifetime with auto-assigned port).
    data = json.loads(post_key({"lifetime": 30000, "port": 0}).read())
```

lboue: Good point, thanks — done in 484dee0d: now reusing the `post_key()` helper with `port: 0` passed explicitly.

jwhui (approved):
Thanks! 👍🏼

jwhui:
@lboue , can you squash the commits? Thanks!

lboue:
> @lboue , can you squash the commits? Thanks!

Done

PR #3532, opened 2026-08-18
[agent] enable max power table support for POSIX

### LINKED ISSUE #3520: Docker image: no supported way to set the radio's max transmit power, and the 0 dBm default is too low for some deployments
**Describe the bug**

`otbr-agent` starts the radio at OpenThread's default transmit power of 0 dBm. In the Docker image there is no supported way to change that: no command-line option, no environment variable, and the one mechanism OpenThread provides for it is not compiled in.

The only way we found is `ot-ctl txpower <n>` after startup, which lives in the running agent's memory and is lost on every restart. We currently re-apply it from a Compose `post_start` hook, which is a workaround, not a configuration.

**Why 0 dBm is not a safe default in practice**

End devices typically transmit at +8 to +20 dBm, so at 0 dBm the border router is by far the weaker half of every link. The failure mode is asymmetric and hard to read: the BR hears the device fine, the device barely hears the BR.

On our deployment (SMLIGHT SLZB-06M over USB, EFR32MG21 RCP):

* An IKEA GRILLPLATS plug at 8.5 m line-of-sight could not attach at all. Its Child ID Requests arrived, the replies did not:

  ```
  [W] Mle-----------: Failed to process Child ID Request - Security     (17x)
  [N] MeshForwarder-: Failed to send IPv6 UDP msg ... error:NoAck ...
  ```

* A second device sat with `QMsgCnt 6` — six queued messages the BR could not deliver.

* In Home Assistant the plug was `unavailable` for six hours, having flapped every 30–90 minutes before that.

After a single `ot-ctl txpower 20`, the plug was back within a minute — and not as a child but as a **Router**, at −76 dBm with 0.00 % frame error, giving the network its first second hop. Two other nodes then attached through it. We have since settled at +14 dBm, which keeps ~20 dB of margin over the sensitivity we measured (a device is decoded reliably at −92 dBm here).

Nothing about that deployment is exotic: one border router, seven Matter-over-Thread nodes, ordinary domestic distances.

**The mechanism already exists — it is just not enabled here**

`max-power-table` was added to the POSIX platform in openthread/openthread#4878 (merged 2020-04-30) precisely to "customize the max allowed transmit power of each channel". It is a radio URL parameter, and the radio URL is *already* an environment variable in this image (`OT_RCP_DEVICE`), so no new variable would be needed:

```yaml
OT_RCP_DEVICE: "spinel+hdlc+uart:///dev/ttyOTBR?uart-baudrate=460800&max-power-table=14"
```

But it is gated behind `OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE`, which defaults to `0` in `src/posix/platform/openthread-posix-config.h`, and the image does not turn it on. The container's own help output confirms it — the sibling parameters from the same block are all present, `max-power-table` is not:

```
$ docker exec otbr otbr-agent --help
...
Parameters:
    uart-parity[=even|odd]         ...
    uart-baudrate[=baudrate]       ...
    region[=region-code]           Set the radio's region code...
    cca-threshold[=dbm]            Set the radio's CCA ED threshold in dBm...
    enable-coex[=1|0]              ...
    fem-lnagain[=dbm]              Set the Rx LNA gain in dBm of the external FEM.
    no-reset                       ...
    product-config-file[=path]     ...
    factory-config-file[=path]     ...
```

Defaulting the flag to `0` makes sense for constrained embedded targets, where the table costs footprint. A container image is not that.

**Request**

Build the Docker image with `OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE=1`, so that operators can set the radio's maximum transmit power through the existing `OT_RCP_DEVICE` variable.

**On the regulatory question**

This is a *maximum*, not an override: `max-power-table` caps what the radio may use per channel, so it is as much a compliance tool as a tuning one. An operator who knows their antenna gain and region can set a correct ceiling; today they cannot express that at all in this image, and are left with an undocumented runtime command.

The alternative path — `product-config-file` with `target_power` / `calibrated_power` and a region domain mapping — is the right mechanism for vendors shipping calibrated hardware, but it needs RCP support that is not universally present. On the RCP above:

```
$ docker exec otbr ot-ctl region
Error 12: NotImplemented
```

so that route is not available to us.

**Environment**

| | |
|---|---|
| Host | Debian 12 (bookworm), kernel 6.1, x86-64, Docker 27.4.1 |
| RCP | SMLIGHT SLZB-06M (EFR32MG21) over USB, 460800 baud, channel 15 |

skirmess:
Fixes #3520, as suggested there.

OpenThread's POSIX platform already parses `max-power-table` from the radio URL, but it is behind `OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE`, which defaults to `0`. This enables it, so the setting becomes reachable through `OT_RCP_DEVICE` with no extra moving parts at runtime:

```
OT_RCP_DEVICE=spinel+hdlc+uart:///dev/ttyUSB0?uart-baudrate=460800&max-power-table=14
```

### Why not `openthread-otbr-posix-config.h.in`

That was the suggestion in #3520 and it is where I started, but it only covers half of the feature.

The macro guards code in two libraries:

| file | library | sees the config header? |
|---|---|---|
| `src/posix/platform/radio.cpp` — parses the URL value | `openthread-posix` | yes, via `platform-posix.h` |
| `src/lib/spinel/radio_spinel.cpp` — restores the table after an RCP reset | `openthread-radio-spinel` | **no** |

`radio_spinel.cpp`'s only project config hook is `OPENTHREAD_PROJECT_LIB_CONFIG_FILE`, from `OT_LIB_CONFIG`, which is not set anywhere in either repository. So a definition in the POSIX config header enables the parsing but silently leaves the restore path compiled out — the max power would be applied at startup and then lost on the next RCP reset, which is the failure mode hardest to notice.

Both libraries link `ot-config`, so defining it there covers both. There is precedent for exactly this in `src/posix/platform/CMakeLists.txt`, where `OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE` is lifted into `OT_PLATFORM_DEFINES` with a comment explaining that the core and POSIX libraries would otherwise disagree about the feature.

### This affects `OT_POSIX_MAX_POWER_TABLE` upstream too

Worth flagging separately: `-DOT_POSIX_MAX_POWER_TABLE=ON` puts the definition on `ot-posix-config`, which only `openthread-posix` links. `openthread-radio-spinel` links `ot-config` and `ot-lib-config`. As far as I can tell the restore-after-reset loop is therefore compiled out in openthread's own POSIX builds as well, including `script/cmake-build posix`. I have not filed anything in openthread/openthread yet — happy to, if you agree with the reading.

### Why it is safe for everyone else

- The parser is `VerifyOrExit(aRadioUrl.GetValue("max-power-table") != nullptr)` — it returns before touching the radio when the URL carries no such value.
- `MaxPowerTable` initialises every channel to `kPowerDefault` (30 dBm), so `GetSupportedChannelMask()` still returns every channel and the mask in `radio_spinel.cpp` is unaffected.
- An RCP that does not implement the property is already handled: `SetChannelMaxTransmitPower` returning `OT_ERROR_NOT_IMPLEMENTED` is tolerated and logged as a warning.

### Verification

Checked against the pinned submodule (`7b6ed71`) that the value actually arrives, rather than assuming the include order works out:

```
with this change:    OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE = 1
without:             OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE = 0
```

Behaviour was confirmed on real hardware before this change, by setting the same value at runtime through `ot-ctl txpower`: a mains-powered plug 8.5 m from the border router could not attach at 0 dBm and attached within a minute at 14 dBm. Details and the neighbour-table measurements are in #3520.

### Note on scope

`max-power-table` is a per-channel list, and a single value applies to every channel from the first onwards ("use the last power if omitted"). That is OpenThread's existing semantics; this PR only makes it reachable.

### DIFF
--- a/third_party/openthread/CMakeLists.txt
+++ b/third_party/openthread/CMakeLists.txt
@@ -144,10 +144,18 @@ set(OT_MBEDTLS_CONFIG_FILE "\"${PROJECT_SOURCE_DIR}/third_party/openthread/mbedt
 
 add_subdirectory(repo EXCLUDE_FROM_ALL)
 
+# OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE is set here rather than in
+# openthread-otbr-posix-config.h, because it guards code in two libraries.
+# `src/posix/platform/radio.cpp` parses `max-power-table` from the radio URL and
+# reaches the config header through platform-posix.h, but the loop that restores
+# the table after an RCP reset lives in `src/lib/spinel/radio_spinel.cpp`, whose
+# only project config hook is OPENTHREAD_PROJECT_LIB_CONFIG_FILE. Defining it on
+# `ot-config` puts it on the command line of both.
 target_compile_definitions(ot-config INTERFACE
     "-DOPENTHREAD_CONFIG_LOG_CLI=1"
     "-DOPENTHREAD_CONFIG_MAX_STATECHANGE_HANDLERS=3"
     "-DOPENTHREAD_CONFIG_MLE_STEERING_DATA_SET_OOB_ENABLE=1"
+    "-DOPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE=1"
     "-DOPENTHREAD_POSIX_CONFIG_FILE=\"${PROJECT_BINARY_DIR}/src/agent/openthread-otbr-posix-config.h\""
 )
 


skirmess:
Rewrote this after finding that the config-header approach from #3520 only covers half the feature.

`OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE` guards code in two libraries. `src/posix/platform/radio.cpp` parses the URL value and reaches `openthread-otbr-posix-config.h` through `platform-posix.h`, but the loop that restores the table after an RCP reset is in `src/lib/spinel/radio_spinel.cpp`, which never includes it — its only project config hook is `OPENTHREAD_PROJECT_LIB_CONFIG_FILE`, and `OT_LIB_CONFIG` is not set in either repository. Defining the macro in the header would have enabled the parsing while leaving the restore path compiled out, so the power setting would survive startup but not an RCP reset.

Defining it on `ot-config` instead covers both libraries, following what `OPENTHREAD_POSIX_CONFIG_DAEMON_ENABLE` does via `OT_PLATFORM_DEFINES` for the same reason.

The same reading suggests `-DOT_POSIX_MAX_POWER_TABLE=ON` upstream has the gap too, since it only reaches `ot-posix-config`. Details in the description — I can open an issue there if you agree.

jwhui:
Thank you for the thorough investigation and for identifying why defining `OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE` in `openthread-otbr-posix-config.h.in` was insufficient!

Your analysis of the macro visibility between `openthread-posix` and `openthread-radio-spinel` is correct. Rather than working around this in OTBR by injecting compile definitions onto `ot-config`, we should resolve the underlying issues upstream in `openthread/openthread` and then bump the submodule here.

### Upstream Issues in `openthread/openthread`:

1. **`radio_spinel.cpp` Layering & Crash on RCP Reset:**
   - `src/lib/spinel/` is a platform-agnostic library, yet `radio_spinel.cpp` uses the POSIX-specific macro `OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE`.
   - In `RadioSpinel::RestoreProperties()`:
     ```cpp
     // Some old RCPs doesn't support max transmit power
     otError error = SetChannelMaxTransmitPower(channel, power);

     if (error != OT_ERROR_NONE && error != OT_ERROR_NOT_FOUND)
     {
         DieNow(OT_EXIT_FAILURE);
     }
     ```
     When an RCP lacks `SPINEL_PROP_PHY_CHAN_MAX_POWER` support, `spinel_status_to_ot_error(SPINEL_STATUS_PROP_NOT_FOUND)` returns `OT_ERROR_NOT_IMPLEMENTED` (not `OT_ERROR_NOT_FOUND`). Because `MaxPowerTable` defaults all channels to 30 dBm, `RestoreProperties()` executes this loop on reset even when `max-power-table` is omitted from the radio URL, causing `DieNow()` on unsupported RCPs.
   - **Fix:** Update `RestoreProperties()` to handle `OT_ERROR_NOT_IMPLEMENTED`, and remove the `#if OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE` macro guard from `radio_spinel.cpp` (`MaxPowerTable` is already an unconditional member of `RadioSpinel`).

2. **OpenThread POSIX CMake Configuration:**
   - In `src/posix/platform/CMakeLists.txt`, `OT_POSIX_MAX_POWER_TABLE` currently only sets `ot-posix-config` rather than appending to `OT_PLATFORM_DEFINES` (like `OT_DAEMON` does).
   - Additionally, similar to other radio URL parameters (`cca-threshold`, `fem-lnagain`), `max-power-table` support can simply be enabled by default on POSIX.

### Proposed Plan:
1. Submit an upstream PR to `openthread/openthread` resolving the `radio_spinel.cpp` restore error handling and CMake configuration.
2. Bump the `third_party/openthread/repo` submodule in OTBR (setting `set(OT_POSIX_MAX_POWER_TABLE ON ...)` in `third_party/openthread/CMakeLists.txt` if needed).

Would you be interested in submitting the upstream PR to `openthread/openthread`? We would be happy to review it or help prepare the patch.

skirmess:
Closing in favour of openthread/openthread#13524, which fixes this upstream as @jwhui suggested.

This PR defined `OPENTHREAD_POSIX_CONFIG_MAX_POWER_TABLE_ENABLE` on `ot-config` so that it would reach `src/lib/spinel/radio_spinel.cpp` as well as the POSIX platform. That works, but it is a workaround in OTBR for a layering problem in OpenThread: the guard in `radio_spinel.cpp` is a POSIX macro inside a platform-agnostic library. #13524 removes the guard instead, so nothing has to be injected here.

It also fixes the fatal error path that removing the guard exposes — `RestoreProperties()` tolerates `OT_ERROR_NOT_FOUND` while an RCP lacking `SPINEL_PROP_PHY_CHAN_MAX_POWER` answers `OT_ERROR_NOT_IMPLEMENTED`, which this PR would have made reachable.

Once that lands and `third_party/openthread/repo` is bumped, ot-br-posix needs no change of its own. Happy to send the submodule bump when the time comes.

PR #3534, opened 2026-08-19
[rest] bump REST API version to 0.5.0 for ePSKc endpoints

lboue:
## Description

`OTBR_REST_API_VERSION` (`version.hpp`) and the OpenAPI `info.version` were left at `0.4.0` after #3480 added the Border Agent ephemeral key (ePSKc) REST endpoints (`/node/ba-epskc/state`, `/node/ba-epskc/key`).

The version discovery mechanism introduced in #3330 exists precisely so clients can detect server capabilities via `/.well-known/thread/br-rest` instead of probing an endpoint and handling failure. Adding ePSKc support is new, backward-compatible REST API surface, so per the semver rules documented in `version.hpp` this is a MINOR bump, not a PATCH.

This PR bumps both `version.hpp` and `openapi.yaml` from `0.4.0` to `0.5.0` to keep them in sync.

## Question

Is there a changelog somewhere in this project that should also be updated for this kind of change? I couldn't find one (no `CHANGELOG*` file in the repo), just checking before assuming there isn't one.

### DIFF
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -2751,7 +2751,7 @@ components:
             $ref: "#/components/schemas/WellKnownThread"
           example:
             api:
-              version: 0.3.0
+              version: 0.5.0
               base: /api/
             links:
               - href: /.well-known/thread/br-rest
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -11,7 +11,7 @@ info:
   license:
     name: BSD 3-Clause
     url: https://github.com/openthread/ot-br-posix/blob/main/LICENSE
-  version: 0.4.0
+  version: 0.5.0
 servers:
   - url: "{protocol}://{host}:{port}"
     description: Configurable Border Router destination.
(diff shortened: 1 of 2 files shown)


src/rest/openapi.yaml
jwhui: Should we update this to 0.5.0?

src/rest/openapi.yaml
```diff
@@ -11,7 +11,7 @@ info:
   license:
     name: BSD 3-Clause
     url: https://github.com/openthread/ot-br-posix/blob/main/LICENSE
-  version: 0.4.0
+  version: 0.5.0
```
jwhui, line 14: We currently don't have a test covering `GET /.well-known/thread/br-rest` in `tests/rest/test_rest.py`. If you'd like, we could add a simple check in `test_rest.py` (either in this PR or in a follow-up) to verify that `/.well-known/thread/br-rest` returns HTTP 200, valid JSON structure, and matches the expected API version.

src/rest/openapi.yaml
lboue: Yes

src/rest/openapi.yaml
lboue: Yes, to be done in another PR

jwhui (approved):
Thanks! 👍🏼

Issue #3512, opened 2026-08-12
Flaky test: ncp_*.exp scripts can time out on the initial dbus Join call under CI load


lboue:
**Describe the bug**

Several `ncp_*.exp` scripts join the network via D-Bus like this:

```tcl
spawn dbus-send --system --dest=io.openthread.BorderRouter.wpan0 --type=method_call --print-reply /io/openthread/BorderRouter/wpan0 io.openthread.BorderRouter.Join "array:byte:${dataset_dbus}"
expect eof
```

`expect eof` here inherits whatever `timeout` was last set (either Tcl Expect's implicit default of 10s, or an explicit `set timeout 10` left over from a prior `expect_line` call in the same script) — there is no generous, explicit timeout for this specific call.

The D-Bus `Join` method is not fire-and-forget: `NcpHost::Join` (`src/host/ncp_host.cpp`) chains three sequential spinel round-trips to the (simulated) NCP — `DatasetSetActiveTlvs` → `Ip6SetEnabled` → `ThreadSetEnabled` — and only replies to the D-Bus caller once all three complete. Under CI load, this chain can apparently take longer than the inherited ~10s `expect` timeout (well under `dbus-send`'s own default reply-timeout, ~25s), so `expect eof` fires its `timeout { fail "Timed out" }` handler before `dbus-send` itself would time out — failing the whole script even though nothing is actually broken.

This is the same class of "fixed short wait window racing a variable-latency CI operation" issue as #3510 / #3511, but at a different call site that #3511 did not touch (the initial `Join` RPC, rather than the subsequent `DeviceRole` poll).

The same `spawn dbus-send ... Join ...` + `expect eof` pattern, with no explicit generous timeout, appears identically in:
- `tests/scripts/expect/ncp_netif_address_update.exp`
- `tests/scripts/expect/ncp_join_leave.exp`
- `tests/scripts/expect/ncp_schedule_migration.exp`
- `tests/scripts/expect/ncp_netif_tx_rx.exp`

**To Reproduce**

1. Git commit id: `e57692f1` (PR #3511 branch `fix-ncp-netif-address-update-flake`)
2. Platform: simulation (`OT_PLATFORM=simulation`), NCP mode, `mDNSResponder` backend
3. Build steps: standard `NcpMode` CI workflow (`.github/workflows/ncp_mode.yml`), `BUILD_TARGET=ncp_mode`
[...]

PR #3543, opened 2026-08-21
[tests] pin schemathesis to ~=4.25.0 to avoid 4.24.x crash

### LINKED ISSUE #3539: CI: meshcop HTTP REST API job fails - schemathesis crashes with 'CanonicalSchema' object has no attribute 'is_satisfiable'
## Description

The `meshcop (openthread)` CI job's "HTTP REST API" step fails during the schemathesis `examples` phase with:

```
AttributeError: 'builtins.CanonicalSchema' object has no attribute 'is_satisfiable'
```

This is unrelated to any application code change — it reproduces on PRs that only touch Python test files (e.g. #3538), and is an internal crash inside the `schemathesis` package itself while generating example test cases from `src/rest/openapi.yaml`.

## Root cause

[tests/scripts/meshcop](https://github.com/openthread/ot-br-posix/blob/main/tests/scripts/meshcop#L622-L623) installs schemathesis without an upper version bound:

```sh
python3 -m pip install --user "schemathesis>=4.0.5" || die "Failed to install schemathesis"
schemathesis run src/rest/openapi.yaml --url=http://localhost:8081 --include-path-regex="/api/.*" --phases=examples
```

Since the constraint is unpinned, CI always installs the latest release from PyPI. The currently-latest release (4.24.3) raises the `AttributeError` above internally (`schemathesis/specs/openapi/_hypothesis.py`, `_canonical_strategy_or_none`) before any actual request is made against the REST API, so every run of this job currently fails regardless of the PR's diff.

## Impact

- The `meshcop (openthread)` job is red on every PR/branch until this is fixed, masking real regressions and blocking merges that rely on required checks.

## Suggested fix

Pin `schemathesis` (and/or its `hypothesis`/`hypothesis-jsonschema` dependencies) to a known-good version range in `tests/scripts/meshcop`, e.g.:

```sh
python3 -m pip install --user "schemathesis>=4.0.5,<4.24.0"
```

or whichever version range is confirmed compatible, and consider adding a periodic/manual bump process instead of always tracking `latest`.

## How to reproduce

Run the `meshcop (openthread)` GitHub Actions job on any branch (e.g. https://github.com/openthread/ot-br-posix/actions/runs/32412762482/job/96567052591) — it fails in the "HTTP REST API" step before exercising any endpoint.

lboue:
### Problem

The `meshcop (openthread)` CI job's "HTTP REST API" step fails during the schemathesis `examples` phase with:

```
AttributeError: 'builtins.CanonicalSchema' object has no attribute 'is_satisfiable'
```

[tests/scripts/meshcop](https://github.com/openthread/ot-br-posix/blob/main/tests/scripts/meshcop#L622-L623) installs schemathesis without an upper version bound, so CI always installs the latest release from PyPI. The 4.24.x releases raise the `AttributeError` above internally before any request is made against the REST API, so the job fails regardless of the PR's diff.

The schemathesis maintainer confirmed in [openthread/ot-br-posix#3539](https://github.com/openthread/ot-br-posix/issues/3539#issuecomment) that this is fixed in 4.25.0.

### Change

Pin schemathesis in `tests/scripts/meshcop` to `schemathesis~=4.25.0` (was `schemathesis>=4.0.5`), which contains the upstream fix. Per [review feedback](https://github.com/openthread/ot-br-posix/pull/3543#discussion_r3847326992), this uses the PEP 440 compatible-release operator instead of an open-ended `>=4.25.0` lower bound: it still picks up 4.25.x patch releases automatically, but a future breaking release (4.26, 5.0, ...) can't silently reintroduce a CI failure like this one.

Also update the matching manual testing instructions in [src/rest/README.md](https://github.com/openthread/ot-br-posix/blob/main/src/rest/README.md?plain=1#L361), which still referenced the outdated `v4.0.5` example.

Fixes #3539


lboue:
The failing `ncp_mode (mDNSResponder)` check is unrelated to this PR's diff (which only touches `tests/scripts/meshcop`).

The job fails in `ncp_schedule_migration.exp` with a D-Bus timeout while waiting for the node to become `leader`:

```
Error org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
 *** ERROR: ncp expect script failed!
```

This matches the known, already-tracked flake in #3512, which explicitly lists `ncp_schedule_migration.exp` as one of the affected scripts (fixed short `expect` timeout racing the D-Bus `Join` RPC chain under CI load).

The same `ncp_mode` workflow has passed on the last several runs of `main`, confirming this is CI flakiness rather than a regression from this change.

tests/scripts/meshcop
```diff
@@ -619,7 +619,11 @@ test_restapi()
     cd ../..
 
     # Validate conformance of openapi.yaml specification with the REST API server implementation.
-    python3 -m pip install --user "schemathesis>=4.0.5" || die "Failed to install schemathesis"
+    # schemathesis 4.24.x crashes during the `examples` phase with
+    # `AttributeError: 'builtins.CanonicalSchema' object has no attribute 'is_satisfiable'`
+    # (https://github.com/openthread/ot-br-posix/issues/3539). Require >=4.25.0, which
+    # contains the upstream fix.
+    python3 -m pip install --user "schemathesis>=4.25.0" || die "Failed to install schemathesis"
```
jwhui: Should we pin this to a specific version (`schemathesis==4.25.0` or `schemathesis~=4.25.0`)?

jwhui: Also update:
https://github.com/openthread/ot-br-posix/blob/26f9c41fbdebee8049113a956dabedb843c5947a/src/rest/README.md?plain=1#L361

lboue: Fixed

lboue: Fixed

src/rest/README.md
```diff
@@ -358,10 +358,11 @@ Detailed request examples are provided in the test folder as a "Bruno Request Co
 
 1. Validating the openapi.yaml specification against the implementation.
 
-   Make sure a recent Schemathesis version is installed, for example, v4.0.5:
+   Make sure Schemathesis v4.25.0 or later (4.25.x series) is installed; earlier 4.24.x releases
+   crash during the `examples` phase:
```
jwhui: To fix pretty check:
```suggestion
   Make sure Schemathesis v4.25.0 or later (4.25.x series) is installed; earlier 4.24.x releases crash during the `examples` phase:
```

jwhui (approved):
Thanks! 👍🏼

Issue #3529, opened 2026-08-17
[rest] RoutingErrorHandler: incorrect/incomplete 404 vs 405 semantics and Allow header


lboue:
## Description

Follow-up from the discussion on #3524 (which fixed #3522, PUT errors being masked as `405`).

`RestWebServer::RoutingErrorHandler` doesn't correctly implement REST semantics around `404`/`405`/`OPTIONS`:

- **404 vs 405**: A `404 Not Found` should be returned only when the requested resource (route) doesn't exist on the server at all. A `405 Method Not Allowed` should be returned when the resource exists but doesn't support the requested method. `RoutingErrorHandler` doesn't reliably distinguish these two cases across all methods — #3524 only fixed this for `PUT`, via a hardcoded `mPutSupportedPaths` allowlist populated through a `RegisterPut()` wrapper. The same routing-miss-vs-real-error ambiguity likely exists for other methods (e.g. `DELETE`/`POST` against a path that doesn't support them) since cpp-httplib's own routing-miss default status leaks through unmodified for those methods.
- **Allow header on 405**: When returning `405`, the `Allow` header must list the methods actually supported *by the specific resource being requested*, not a single static, server-wide list. The current implementation (`aResponse.set_header("Allow", "GET, POST, PUT, DELETE, OPTIONS");`) is wrong for the majority of resources, since most endpoints don't actually support all of those methods.
- **OPTIONS**: The `OPTIONS` handling is only partially implemented across handlers, and only some of them set an `Allow` header in their response.

## Suggested direction

Doing this properly likely requires each route's supported-method set to be introspectable from the router (e.g. extending the `RegisterPut`-style wrapper pattern to `RegisterGet`/`RegisterPost`/`RegisterDelete`/`RegisterOptions`, or building a small per-path method registry), so that:
- a request to an unregistered path returns `404`,
- a request with an unsupported method on a registered path returns `405` with a correct, per-resource `Allow` header,
- `OPTIONS` responses consistently report the same per-resource `Allow` header.

## References

- #3522 (original masking bug)
- #3524 (fix for PUT; see review discussion for the semantics gap noted above)

cc @3oris @jwhui

3oris:
@lboue, I have proposed something similar as you suggest [here](https://github.com/openthread/ot-br-posix/issues/3522#issuecomment-5308183340), replacing the `RegisterMethod(pattern,handler)` idea by `Register(method, pattern, handler)` 

[...]

PR #3546, opened 2026-08-22
[rest] add per-route method registry to fix 404/405/OPTIONS handling

3oris:
Tracks, for each registered route, which HTTP methods it supports. This lets the routing error handler and a new pre-routing OPTIONS handler tell an unknown path (404) apart from a known path used with an unsupported method (405), and report the Allow header for the specific resource instead of a static server-wide method list.

Replaces the ad hoc per-handler OPTIONS cases and the PUT-only 404/405 workaround from https://github.com/openthread/ot-br-posix/pull/3524 with a single generic mechanism covering GET/POST/PUT/DELETE on every route.

Addresses https://github.com/openthread/ot-br-posix/issues/3529

Since OTBR REST endpoints are strictly defined by OpenAPI specifications the registry features a regex-free path matcher.

Co-authored-by: Jonathan Hui <jonhui@google.com>

### DIFF
--- a/src/rest/rest_web_server.cpp
+++ b/src/rest/rest_web_server.cpp
@@ -31,7 +31,9 @@
 
 #include "rest/rest_web_server.hpp"
 
+#include <algorithm>
 #include <chrono>
+#include <utility>
 
 #include <arpa/inet.h>
 #include <errno.h>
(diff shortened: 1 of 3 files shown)


src/rest/rest_web_server.cpp
```diff
@@ -225,6 +219,119 @@ RestWebServer::~RestWebServer(void)
...
+    if (aPattern.find("/:") == std::string::npos)
+    {
+        return std::regex(aPattern);
+    }
+
+    static const std::regex kRegexMetaChars(R"([.^$|()\[\]{}*+?\\])");
+    std::string             escaped = std::regex_replace(aPattern, kRegexMetaChars, R"(\$&)");
+
+    return std::regex(std::regex_replace(escaped, std::regex(R"(/:[^/]+)"), R"(/[^/]+)"));
+}
```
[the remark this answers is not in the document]
3oris: No, this is intentional, because it is exactly how httplib behaves, and the registry must match the exact same patterns.

Meaning that `rest_web_server` has always served content under all these wrong URLs.

We could open an issue to fix this by correcting the definition of `OT_REST_ROUTE_WELLKNOWN_THREAD` to escape the dot.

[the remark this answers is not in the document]
3oris: Opened https://github.com/openthread/ot-br-posix/issues/3547

3oris:
@lboue , please take a look.

src/rest/rest_web_server.cpp
```diff
@@ -1240,37 +1328,31 @@ void RestWebServer::EpskcKey(const Request &aRequest, Response &aResponse) const
...
-        otbrLogWarning("%s:%d Error (%d) - %s", __FILE__, __LINE__, error, errorDetails.c_str());
-        ErrorHandler(aResponse, error, errorDetails);
+    }
+
+    if (status == StatusCode::MethodNotAllowed_405)
+    {
+        aResponse.set_header(OT_REST_ALLOW_HEADER, methodsString);
+        ErrorHandler(aResponse, status, "method not supported");
+    }
+    else
```
jwhui: Should we only generate the error body if `aResponse.body` is currently empty (or if the status was rewritten from 404 to 405)?

`cpp-httplib` calls `error_handler_` whenever `res.status >= 400`. When a route handler sets an error response with specific error details via `ErrorHandler(aResponse, status, detail)`, `RoutingErrorHandler` executes the `else` branch and calls `ErrorHandler(aResponse, status)` unconditionally. This creates a new `Json::Error2JsonString` body and overwrites the error detail already generated by the handler.

```suggestion
    else if (aResponse.body.empty())
```

src/rest/rest_web_server.cpp
```diff
@@ -1240,37 +1328,31 @@ void RestWebServer::EpskcKey(const Request &aRequest, Response &aResponse) const
...
+        // both status require the methodsString to be loaded
+        methodsString = mRouteRegistry.GetMethodsString(aRequest.path);
 
-    if (error != StatusCode::OK_200 || aResponse.status >= StatusCode::MultipleChoices_300)
-    {
-        if (error < aResponse.status)
+        // Check if status has to be rewritten to 405:
+        // if methodsString.empty(): no handler registered for this route -> real 404, no rewrite
+        // method not in methods: method not registered -> rewrite to 405
+        if (!methodsString.empty() && methodsString.find(aRequest.method) == std::string::npos)
```
jwhui: Could we instead expose a helper method on `RouteRegistry` (or return the `mMethodMask`) and check `(mask & MethodBit(GetMethod(aRequest))) == 0`? That would make the check exact and avoid string searches.

Checking method support via substring search inside `methodsString` couples validation with the formatted `Allow` header representation.

src/rest/rest_web_server.hpp
```diff
@@ -193,10 +237,13 @@ class RestWebServer : public std::enable_shared_from_this<RestWebServer>
...
+    auto MakeHandler(HandlerType aHandler) -> std::function<
+        decltype((std::declval<RestWebServer *>()->*aHandler)(std::declval<const Request &>(),
+                                                              std::declval<Response &>()))(const Request &, Response &)>
     {
-        return [this, aHandler](const Request &aRequest, Response &aResponse) -> void {
-            (this->*aHandler)(aRequest, aResponse);
+        return [this, aHandler](const Request &aRequest, Response &aResponse) {
+            return (this->*aHandler)(aRequest, aResponse);
         };
     }
```
jwhui: The trailing return type with `std::declval` expressions is fairly complex to read. Since `MakeHandler` is only used for two signatures, consider using function overloading or a distinct `MakePreRoutingHandler` helper:

```suggestion
    httplib::Server::Handler MakeHandler(void (RestWebServer::*aHandler)(const Request &, Response &))
    {
        return [this, aHandler](<const Request &aRequest, Response &aResponse>) { (this->*aHandler)(aRequest, aResponse); };
    }

    httplib::Server::HandlerWithResponse MakePreRoutingHandler(
        httplib::Server::HandlerResponse (RestWebServer::*aHandler)(const Request &, Response &))
    {
        return [this, aHandler](<const Request &aRequest, Response &aResponse>) { return (this->*aHandler)(aRequest, aResponse); };
    }
```

src/rest/rest_web_server.hpp
```diff
@@ -99,6 +102,40 @@ class RestWebServer : public std::enable_shared_from_this<RestWebServer>
...
+
+    private:
+        struct Route
+        {
+            std::string mPattern;
+            std::regex  mRegex;
+            uint8_t     mMethodMask = 0;
+        };
+
+        // Reflects httplibs's path matching by creating the path pattern's regex
```
jwhui:
```suggestion
        // Reflects httplib's path matching by creating the path pattern's regex
```

src/rest/rest_web_server.cpp
3oris: Sure!

One observation though: The old RoutingErrorHandler implementation would always run into
```
    if (error != StatusCode::OK_200 || aResponse.status >= StatusCode::MultipleChoices_300)
    {
        if (error < aResponse.status)
        {
            error = StatusCode(aResponse.status);
        }
        otbrLogWarning("%s:%d Error (%d) - %s", __FILE__, __LINE__, error, errorDetails.c_str());
        ErrorHandler(aResponse, error, errorDetails);
    }
```
because, as you quoted, `aResponse.status >= 400` is always true). Consequently the old implementation would also "unconditionally" overwrite `errorDetails` which is either set to empty or "method not supported" (in case of 405).

I assumed this was maybe left as a catch-all protection against anyone forgetting to call `ErrorHandler()` in some places or to enforce no HTML body being returned inadvertently in case of an error. So I tried to stick to the old behavior here.

Still thinking your suggestion makes sense. What do you think?

src/rest/rest_web_server.cpp
3oris: You are right.

I tried to keep the registry's class interface simple and not expose the implementation details (like how the method assignment is organized as a bit mask) to the RoutingErrorHandler. I also tried to avoid doing the path matching twice (first in the suggested decoupled validation in a helper method, and later to compose the methodsString).

Let me see how to address this better.

src/rest/rest_web_server.hpp
3oris: Addressing this in [Reduce template complexity, introduce MakePreRoutingHandler](https://github.com/openthread/ot-br-posix/pull/3546/commits/3bcba89db2d2669d252dbbf4d37fee907d1f50e4)

src/rest/rest_web_server.cpp
3oris: Addressing this in "[decouple validation from string representation of method list](https://github.com/openthread/ot-br-posix/pull/3546/commits/cdb711e6c4f35bcf542b94e62c0048e26f3cdbbd)"

Also, I have restructured the 404->405 rewrite decision logic to make it better comprehensible.

3oris:
Added/fixed some [Tests](https://github.com/openthread/ot-br-posix/pull/3546/changes/922602c4c447cfb9cb64e1c3904dadcfd81ba37a)

src/rest/rest_web_server.cpp
jwhui: Checking `aResponse.body.empty()` provides the best of both worlds:

1. **Catch-all for routing misses & unformatted errors:** If `aResponse.body` is empty (such as on a genuine 404 route miss, or if a handler sets `status >= 400` without generating a body), `ErrorHandler(aResponse, status)` will be called to generate the standard RFC 7807 JSON error payload (`{"title": "Not Found", "status": 404}`).
2. **Preserves handler details:** If an endpoint handler has already populated `aResponse.body` (e.g. with custom error details or a JSON:API error collection), it will not be overwritten.

In the current revision, genuine 404s (where `methods == 0`) currently skip calling `ErrorHandler` entirely because `status == 405` evaluates to `false` inside the `if (status == 404 || status == 405)` block. This causes genuine 404 responses to return with an empty/unformatted body instead of `Content-Type: application/json`.

Structuring `RoutingErrorHandler` as follows resolves both issues:

```cpp
void RestWebServer::RoutingErrorHandler(const Request &aRequest, Response &aResponse)
{
    httplib::StatusCode status = StatusCode(aResponse.status);

    if (status == StatusCode::NotFound_404 || status == StatusCode::MethodNotAllowed_405)
    {
        RouteRegistry::RouteMethods methods = mRouteRegistry.GetMethods(aRequest.path);

        if (RouteRegistry::AnyMethod(methods) && !RouteRegistry::MatchMethod(methods, GetMethod(aRequest)))
        {
            status = StatusCode::MethodNotAllowed_405;
        }

        if (status == StatusCode::MethodNotAllowed_405)
        {
            aResponse.set_header(OT_REST_ALLOW_HEADER, RouteRegistry::BuildMethodsString(methods));
            ErrorHandler(aResponse, status, "method not supported");
            return;
        }
    }

    if (aResponse.body.empty())
    {
        ErrorHandler(aResponse, status);
    }
}
```

src/rest/rest_web_server.hpp
```diff
@@ -99,6 +102,53 @@ class RestWebServer : public std::enable_shared_from_this<RestWebServer>
...
+        // Records that aPattern was registered for aMethod.
+        void Add(const std::string &aPattern, HttpMethod aMethod);
+
+        // Returns the methods registered for the route matching aPath, or 0 if no route matches.
+        RouteMethods GetMethods(const std::string &aPath) const;
+
+        // Returns whether aMethod is supported by aMethods.
+        static bool MatchMethod(RouteMethods aMethods, HttpMethod aMethod);
+
+        // Returns whether aMethod has any methods at all.
```
jwhui:
```suggestion
        // Returns whether aMethods has any methods at all.
```

tests/rest/test_rest.py
```diff
@@ -444,9 +445,70 @@ def node_ext_address_put_no_route_test():
...
+    rejected with 405, and the Allow header must list every method actually
+    registered on that route, not just the one that was tried.
+    """
+    request = urllib.request.Request(rest_api_addr + "/node/state", method='DELETE')
+
+    try:
+        urllib.request.urlopen(request)
+        assert False
+    except urllib.error.HTTPError as e:
+        assert (e.code == 405)
```
jwhui, line 470: Consider checking the `Content-Type` header and response body in these tests to verify that the error response is formatted as valid JSON matching the RFC 7807 schema:
```python
assert e.headers.get("Content-Type") == "application/json"
body = json.loads(e.read().decode('utf-8'))
assert body["status"] == 405
assert body["title"] == "Method Not Allowed"
assert body.get("detail") == "method not supported"
```

tests/rest/test_rest.py
```diff
@@ -444,9 +445,70 @@ def node_ext_address_put_no_route_test():
...
+    """
+    url = rest_api_addr + "/node/this-path-does-not-exist"
+
+    for method in ("GET", "OPTIONS", "POST", "PUT", "DELETE"):
+        request = urllib.request.Request(url, method=method)
+        try:
+            urllib.request.urlopen(request)
+            assert False, "expected HTTP 404 for {}, but request succeeded".format(method)
+        except urllib.error.HTTPError as e:
+            assert e.code == 404, "expected HTTP 404 for {}, got {}".format(method, e.code)
```
jwhui, line 555: Consider checking the `Content-Type` header and response body in these tests to verify that the error response is formatted as valid JSON matching the RFC 7807 schema:
```python
assert e.headers.get("Content-Type") == "application/json"
body = json.loads(e.read().decode('utf-8'))
assert body["status"] == 404
assert body["title"] == "Not Found"
```

src/rest/rest_web_server.cpp
3oris: Good catch! I introduced a bug while restructuring the code. Also completely agree on 1. and 2.

src/rest/rest_web_server.cpp
3oris: Since you already proposed a fix for the RoutingErrorHandler, I suppose you are OK with the general approach of decoupling validation and the formatted Allow header representation. So resolving this one.

src/rest/rest_web_server.hpp
3oris: Can we resolve? I followed your considerations.

src/rest/rest_web_server.cpp
3oris: Addessed here https://github.com/openthread/ot-br-posix/pull/3546/commits/0ef96cba0dc06bac208d6557ee8491251c40cb87

tests/rest/test_rest.py
3oris: https://github.com/openthread/ot-br-posix/pull/3546/changes/8c7be95ffff130746a5707d78b820350b5d03e34

tests/rest/test_rest.py
3oris: https://github.com/openthread/ot-br-posix/pull/3546/changes/8c7be95ffff130746a5707d78b820350b5d03e34

3oris:
@jwhui , let me know when to rebase/squash.

src/rest/rest_web_server.cpp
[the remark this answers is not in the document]
3oris: Being addressed here https://github.com/openthread/ot-br-posix/pull/3548

src/rest/rest_web_server.hpp
```diff
@@ -99,6 +102,53 @@ class RestWebServer : public std::enable_shared_from_this<RestWebServer>
...
+     *
+     * cpp-httplib routes each method (GET/POST/PUT/DELETE) independently and has no notion of
+     * "this path exists but not for that method" - an unmatched method simply falls through to
+     * a generic 404. This class lets RestWebServer tell an unknown path (status 404) apart from
+     * a known path used with an unsupported method (status 405).
+     */
+    class RouteRegistry
+    {
+    public:
+        // Bitmask of the HTTP methods (see HttpMethod) registered for a route.
```
jwhui: Please use standard OpenThread Doxygen comments (`/** ... */` or `///<`) for `RouteRegistry` members to match the rest of the header.

tests/rest/test_rest.py
```diff
@@ -444,9 +445,116 @@ def node_ext_address_put_no_route_test():
         urllib.request.urlopen(request)
         assert False
     except urllib.error.HTTPError as e:
-        assert (e.code == 404)
+        assert (e.code == 405)
+        assert (e.headers.get("Allow") == "GET, OPTIONS")
```
jwhui, line 449: For consistency with `node_state_delete_method_not_allowed_test`, consider adding the JSON response body and `Content-Type` checks here as well:

```suggestion
        assert (e.code == 405)
        assert (e.headers.get("Allow") == "GET, OPTIONS")
        assert e.headers.get("Content-Type") == "application/json"
        body = json.loads(e.read().decode('utf-8'))
        assert body["status"] == 405
        assert body["title"] == "Method Not Allowed"
        assert body.get("detail") == "method not supported"
```

jwhui (commented):
Looks good overall. Feel free to rebase and squash. Thanks!

src/rest/rest_web_server.hpp
3oris: addressed here: https://github.com/openthread/ot-br-posix/pull/3546/commits/71c1fb8a5a9aa08928c161ddb59233a9d91d5081

3oris:
> Looks good overall. Feel free to rebase and squash. Thanks!

https://github.com/openthread/ot-br-posix/pull/3546/commits/aa15887d6b06e0c132b96e008a6262cad2e93530

src/rest/rest_web_server.cpp
```diff
@@ -225,6 +219,131 @@ RestWebServer::~RestWebServer(void)
...
+{
+    mServer.Delete(aPattern, aHandler);
+    mRouteRegistry.Add(aPattern, HttpMethod::kDelete);
+}
+
+std::regex RestWebServer::RouteRegistry::PathPatternToRegex(const std::string &aPattern)
+{
+    if (aPattern.find("/:") == std::string::npos)
+    {
+        return std::regex(aPattern);
```
jwhui: Based on our discussion in https://github.com/openthread/ot-br-posix/pull/3548/changes#r3839660680, thoughts on trying to avoid regex here?

3oris: That actually depends on if you want to support httplib style regex routes at all. The registry currently mimics the httplib features, because I assume a necessity of matching request paths the same way, hence supporting regex matching. If you consider restricting routes to `:id` patterns, regex could possibly be avoided. It would also make the workaround in [https://github.com/openthread/ot-br-posix/pull/3548](https://github.com/openthread/ot-br-posix/pull/3548) obsolete. It is maybe not expected by people implementing features for the rest web server though.

httplib tries to limit the impact of regex defining  `CPPHTTPLIB_REGEX_ROUTE_PATH_MAX_LENGTH 256`

Would you consider restricting routes to only supporting `:id` patterns?

3oris: Even more so, would you consider restricting routes to `/some/prefix/path` and `/some/prefix/path/:id`, possibly `:id` being the only allowed keyword?

jwhui: OTBR REST endpoints are strictly defined by OpenAPI specifications and do not require arbitrary regex matching. Supporting literal paths and parameterized segments (such as `:id` or any `:param`) should be sufficient.

Rather than restricting strictly to the keyword `:id`, a segment-by-segment matcher can handle both literal paths and any `:param` segment without `<regex>` overhead or metacharacter escaping issues. Something like:
```cpp
bool RestWebServer::RouteRegistry::MatchPath(const std::string &aPattern, const std::string &aPath)
{
    bool   matched    = false;
    size_t patternPos = 0;
    size_t pathPos    = 0;
    while (patternPos < aPattern.size() && pathPos < aPath.size())
    {
        size_t nextPatternSlash = aPattern.find('/', patternPos);
        size_t nextPathSlash    = aPath.find('/', pathPos);
        std::string patternSeg = aPattern.substr(patternPos, nextPatternSlash - patternPos);
        std::string pathSeg    = aPath.substr(pathPos, nextPathSlash - pathPos);
        if (!patternSeg.empty() && patternSeg[0] == ':')
        {
            VerifyOrExit(!pathSeg.empty());
        }
        else
        {
            VerifyOrExit(patternSeg == pathSeg);
        }
        VerifyOrExit((nextPatternSlash == std::string::npos) == (nextPathSlash == std::string::npos));
        if (nextPatternSlash == std::string::npos)
        {
            patternPos = aPattern.size();
            pathPos    = aPath.size();
            break;
        }
        patternPos = nextPatternSlash + 1;
        pathPos    = nextPathSlash + 1;
    }
    matched = (patternPos == aPattern.size() && pathPos == aPath.size());
exit:
    return matched;
}
```
Thoughts?

3oris: OK, I understand. Given that OTBR REST endpoints are strictly defined by OpenAPI specifications your proposal generally looks good to me. We would avoid regex usage and also would't have to save a second version in the registry.
Let me check how httplib and your proposal handle `/pathes//with/double//slashes`, just to make sure. Then I would use the above, if you allow.

jwhui: Feel free to use the code above in any way you see fit.

3oris: Thank you! I put it in here https://github.com/openthread/ot-br-posix/pull/3546/commits/73fc5812945addaf108ccf7500b79c7667c82154

3oris: forgot to remove the regex include obviously..

3oris: forgot to remove the regex include obviously.. https://github.com/openthread/ot-br-posix/pull/3546/commits/f89ce3babefceeb4bb482da5e825fb9a374bed01

3oris:
@jwhui , please have another look.

jwhui:
@3oris , thanks for the updates. This looks good to me.

@lboue , did you want to take a look as well?

3oris:
Thank you for your support!

lboue (approved):
Great job. LGTM

jwhui (approved):
Thanks! 👍🏼

3oris:
> @3oris , thanks for the updates. This looks good to me.
> 
Does it need another squash?

jwhui:
> Does it need another squash?

@3oris , I squashed this one for you :D

3oris:
> > Does it need another squash?
> 
> @3oris , I squashed this one for you :D

Thank you!

PR #3552, opened 2026-08-25
[rest] honor If-None-Match on the dataset PUT endpoints

LorbusChris:
A pending dataset that is not newer than the one the mesh already holds is silently ignored while this handler accepts the write, so a client that did not know a dataset was in flight reports success for a change that never happens. The client can check first, but not atomically with the write.

Honor the RFC 9110 If-None-Match header with the value `*` on PUT /node/dataset/active and /node/dataset/pending: the write then only happens when the addressed endpoint has no dataset in place yet (no pending dataset on the pending endpoint, no active dataset on the active one), and answers 412 Precondition Failed otherwise, checked inside the main-loop task atomically with the write. Header values other than `*` answer 400 rather than running the write unconditionally.

The checks run in the order RFC 9110 section 13.2.1 prescribes: first the existing request checks (the active dataset is only writable while Thread is stopped, so on a running node the answer stays 409 and the header cannot downgrade it to a 412), then the precondition (412), and only then is the request content processed (400 for a malformed body) and the dataset written. Requests without the header keep the replace semantics unchanged, so existing clients are unaffected.

cc @agners re https://github.com/home-assistant-libs/python-otbr-api/pull/269#pullrequestreview-5007594914

### DIFF
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -1464,6 +1464,8 @@ paths:
         Creates or updates the the active operational dataset on the current node. Only allowed if the Thread node
         is inactive. If the Thread node is active, a pending dataset should be used to update the current active
         operational dataset.
+      parameters:
+        - $ref: "#/components/parameters/IfNoneMatchStar"
       requestBody:
         description: |-
           Operational dataset that will be stored as active operational dataset. Supports request body Content-Type
@@ -1481,8 +1483,12 @@ paths:
           description: Successfully updated the active operational dataset.
         "201":
           description: Successfully created the active operational dataset.
+        "412":
+          $ref: "#/components/responses/PreconditionFailedError"
         "400":
-          description: Invalid request body.
+          description: |-
+            Invalid request body, or an `If-None-Match` header value other
+            than `*`.
         "409":
           description: Writing active operational dataset rejected because Thread network is active.
         "500":
@@ -1511,6 +1517,8 @@ paths:
       description: |-
         Creates or updates the the pending operational dataset on the current node. Delay needs to be set to let
         the pending dataset apply as active dataset in the near future.
+      parameters:
+        - $ref: "#/components/parameters/IfNoneMatchStar"
       requestBody:
         description: |-
           Operational dataset that will be stored as pending operational dataset. Supports request body Content-Type
@@ -1528,8 +1536,12 @@ paths:
           description: Successfully updated the pending operational dataset.
         "201":
           description: Successfully created the pending operational dataset.
+        "412":
+          $ref: "#/components/responses/PreconditionFailedError"
         "400":
-          description: Invalid request body.
+          description: |-
+            Invalid request body, or an `If-None-Match` header value other
+            than `*`.
         "500":
           $ref: "#/components/responses/InternalServerError"
   /node/commissioner/state:
@@ -2488,6 +2500,21 @@ components:
   ### Parameters ##############################################################
   #############################################################################
   parameters:
+    IfNoneMatchStar:
+      name: If-None-Match
+      in: header
+      required: false
+      schema:
+        type: string
+        enum: ["*"]
+      description: |-
+        With the value `*`, the dataset is only written when none is in place
+        yet; if one exists the request fails with 412 and nothing is written.
+        Any other value is rejected with 400: the dataset resources carry no
+        entity tags to match against. Without the header an existing dataset
+        is replaced. The active dataset is only writable while Thread is
+        stopped, and that check runs first: on a running node the response
+        is 409 whether or not this header is present.
 
     actionId:
       name: actionId
@@ -2706,6 +2733,18 @@ components:
             title: Request Timeout
             status: 408
 
+    PreconditionFailedError:
+      description: |
+        A dataset is already in place and the request carried
+        `If-None-Match: *`. Nothing was written; retry without the header to
+        replace the dataset deliberately, or wait until none is in place.
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/SimpleError"
+          example:
+            title: Precondition Failed
+            status: 412
     ConflictError:
       description: |
         The request conflicts with the current state of the target resource;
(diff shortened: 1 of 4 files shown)


src/rest/rest_web_server.cpp
```diff
@@ -697,8 +697,25 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
+    // match against here and are rejected rather than mistaken for an
+    // unconditional write.
+    const std::string ifNoneMatch  = aRequest.get_header_value(OT_REST_IF_NONE_MATCH_HEADER);
+    const bool        onlyIfAbsent = ifNoneMatch == "*";
+
+    if (!ifNoneMatch.empty() && !onlyIfAbsent)
+    {
+        ErrorHandler(aResponse, StatusCode::BadRequest_400);
+        ExitNow();
+    }
```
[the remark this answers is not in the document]
LorbusChris: Folded into the amended commit: the manual block is now `VerifyOrExit(ifNoneMatch.empty() || onlyIfAbsent, error = OTBR_ERROR_INVALID_ARGS)`. The exit block already maps `OTBR_ERROR_INVALID_ARGS` to 400, so the behavior is unchanged.

src/rest/openapi.yaml
```diff
@@ -2706,6 +2729,11 @@ components:
             title: Request Timeout
             status: 408
 
+    PreconditionFailedError:
+      description: |
+        A dataset is already in place and the request carried
+        `If-None-Match: *`. Nothing was written; retry without the header to
+        replace the dataset deliberately, or wait until none is in place.
```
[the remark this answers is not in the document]
LorbusChris: Added in the amended commit: `PreconditionFailedError` now carries the same `application/vnd.api+json`/`application/json` content blocks as the other error responses, with 412 examples.

src/rest/openapi.yaml
```diff
@@ -2706,6 +2729,26 @@ components:
...
+      content:
+        application/vnd.api+json:
+          schema:
+            $ref: "#/components/schemas/Errors"
+          example:
+            errors:
+              - title: Precondition Failed
+                status: 412
+                detail: A dataset is already in place.
```
agners: This seems not entirely accurate, `ErrorHandler` only emits title/status.

agners (approved):
Other than the small nit, LGTM! Nice RFC 9110 13.2.1 compliant implementation!

src/rest/openapi.yaml
LorbusChris: Seems like the vnd.api+json variants are aspirationally included here for all errors but unused?

src/rest/rest_web_server.cpp
```diff
@@ -776,6 +796,10 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
     {
         ErrorHandler(aResponse, StatusCode::Conflict_409);
     }
+    else if (error == OTBR_ERROR_DUPLICATED && onlyIfAbsent)
```
jwhui: The `&& onlyIfAbsent` check is redundant here because `OTBR_ERROR_DUPLICATED` is only returned by `RunInMainLoop` when `onlyIfAbsent` is `true`.

src/rest/openapi.yaml
```diff
@@ -1481,6 +1483,8 @@ paths:
           description: Successfully updated the active operational dataset.
         "201":
           description: Successfully created the active operational dataset.
+        "412":
+          $ref: "#/components/responses/PreconditionFailedError"
         "400":
           description: Invalid request body.
```
jwhui: With this change, a 400 status code can also be returned if an invalid If-None-Match header value is provided (e.g., entity tags other than *). Consider updating the description or referencing `$ref: "#/components/responses/BadRequestError"` so the 400 response reflects both invalid headers and invalid bodies.

src/rest/openapi.yaml
```diff
@@ -1528,6 +1534,8 @@ paths:
           description: Successfully updated the pending operational dataset.
         "201":
           description: Successfully created the pending operational dataset.
+        "412":
+          $ref: "#/components/responses/PreconditionFailedError"
         "400":
           description: Invalid request body.
```
jwhui: See https://github.com/openthread/ot-br-posix/pull/3552/changes#r3854563576

tests/rest/test_rest.py
```diff
@@ -49,6 +49,14 @@ def get_data_from_url(url, result, index):
     result[index] = data
 
 
+def expect_http_error(code, action):
+    try:
+        action()
+        raise AssertionError("expected HTTP {}".format(code))
+    except urllib.error.HTTPError as err:
+        assert err.code == code
```
jwhui: Add diagnostic message to the assertion
```suggestion
       assert err.code == code, f"expected HTTP {code}, got {err.code}"
```

src/rest/openapi.yaml
LorbusChris: removed.

tests/rest/test_rest.py
LorbusChris: Applied, thanks.

src/rest/rest_web_server.cpp
LorbusChris: Dropped here.

src/rest/openapi.yaml
LorbusChris: Description changed to

```code
Invalid request body, or an `If-None-Match` header value other than `*`.
```

src/rest/openapi.yaml
LorbusChris: Description changed to

```code
Invalid request body, or an `If-None-Match` header value other than `*`.
```

tests/rest/test_rest.py
```diff
@@ -713,6 +714,49 @@ def expect_http_error(code, action):
...
+    # run left behind (there is no DELETE to clean it up with); the
+    # conditional create is only expected on a node that has none.
+    try:
+        with urllib.request.urlopen(url) as response:
+            has_pending = response.status == 200
+    except urllib.error.HTTPError:
+        has_pending = False
+    if not has_pending:
+        with put(body, {'Content-Type': 'application/json', 'If-None-Match': '*'}) as response:
+            assert (response.status == 201)
```
jwhui: style nit:
```suggestion
            assert response.status == 201
```

LorbusChris: Applied, thanks

jwhui (approved):
Thanks 👍🏼

PR #3553, opened 2026-08-25
[rest] add entity tags and If-Match to the dataset endpoints

LorbusChris:
Follow-up to https://github.com/openthread/ot-br-posix/pull/3552, which has merged; rebased onto main, so this is now a single commit.

If-None-Match gives a client create-only semantics; what it cannot
express is "replace, but only the dataset I read". A controller that
reads a dataset, stamps a successor and writes it back has no way to
detect another writer slipping in between the read and the write.

Give the dataset resources entity tags: GET answers with an ETag
derived from the stored TLVs (hashed, so the tag does not echo the
credentials it stands for; the pending dataset's delay timer is left
out of the hash since the running countdown does not make it a
different dataset), and a successful PUT reports the tag of what it
stored. The JSON representation is now rendered from the same stored
TLV bytes the tag covers, rather than the struct getters.

PUT honors If-Match per RFC 9110 section 13.1.1: the write only
happens when one of the given entity tags matches the stored dataset,
or, with `*`, when any dataset is stored. Tags are compared with the
strong comparison function (section 8.8.3.2), and the precondition is
evaluated before If-None-Match (section 13.2.2), atomically with the
write in the main-loop task. A failed precondition answers 412, a
syntactically invalid header value 400, and the role check on the
active dataset still answers 409 first (section 13.2.1). This turns a
read-stamp-write sequence into a compare-and-swap.

cc @agners

### DIFF
--- a/src/rest/openapi.yaml
+++ b/src/rest/openapi.yaml
@@ -11,7 +11,7 @@ info:
   license:
     name: BSD 3-Clause
     url: https://github.com/openthread/ot-br-posix/blob/main/LICENSE
-  version: 0.5.0
+  version: 0.6.0
 servers:
   - url: "{protocol}://{host}:{port}"
     description: Configurable Border Router destination.
@@ -1447,6 +1447,11 @@ paths:
       responses:
         "200":
           description: Returns currently active operational dataset
+          headers:
+            ETag:
+              $ref: "#/components/headers/DatasetEtag"
+            Cache-Control:
+              $ref: "#/components/headers/DatasetCacheControl"
           content:
             application/json:
               schema:
@@ -1465,6 +1470,7 @@ paths:
         is inactive. If the Thread node is active, a pending dataset should be used to update the current active
         operational dataset.
       parameters:
+        - $ref: "#/components/parameters/IfMatchEtag"
         - $ref: "#/components/parameters/IfNoneMatchStar"
       requestBody:
         description: |-
@@ -1481,14 +1487,20 @@ paths:
       responses:
         "200":
           description: Successfully updated the active operational dataset.
+          headers:
+            ETag:
+              $ref: "#/components/headers/DatasetEtag"
         "201":
           description: Successfully created the active operational dataset.
+          headers:
+            ETag:
+              $ref: "#/components/headers/DatasetEtag"
         "412":
           $ref: "#/components/responses/PreconditionFailedError"
         "400":
           description: |-
-            Invalid request body, or an `If-None-Match` header value other
-            than `*`.
+            Invalid request body, a syntactically invalid `If-Match` header
+            value, or an `If-None-Match` header value other than `*`.
         "409":
           description: Writing active operational dataset rejected because Thread network is active.
         "500":
@@ -1501,6 +1513,11 @@ paths:
       responses:
         "200":
           description: Returns currently pending operational dataset
+          headers:
+            ETag:
+              $ref: "#/components/headers/DatasetEtag"
+            Cache-Control:
+              $ref: "#/components/headers/DatasetCacheControl"
           content:
             application/json:
               schema:
@@ -1518,6 +1535,7 @@ paths:
         Creates or updates the pending operational dataset on the current node. Delay needs to be set to let
         the pending dataset apply as active dataset in the near future.
       parameters:
+        - $ref: "#/components/parameters/IfMatchEtag"
         - $ref: "#/components/parameters/IfNoneMatchStar"
       requestBody:
         description: |-
@@ -1534,14 +1552,20 @@ paths:
       responses:
         "200":
           description: Successfully updated the pending operational dataset.
+          headers:
+            ETag:
+              $ref: "#/components/headers/DatasetEtag"
         "201":
           description: Successfully created the pending operational dataset.
+          headers:
+            ETag:
+              $ref: "#/components/headers/DatasetEtag"
         "412":
           $ref: "#/components/responses/PreconditionFailedError"
         "400":
           description: |-
-            Invalid request body, or an `If-None-Match` header value other
-            than `*`.
+            Invalid request body, a syntactically invalid `If-Match` header
+            value, or an `If-None-Match` header value other than `*`.
         "500":
           $ref: "#/components/responses/InternalServerError"
   /node/commissioner/state:
@@ -2499,7 +2523,47 @@ components:
 
   ### Parameters ##############################################################
   #############################################################################
+  headers:
+    DatasetCacheControl:
+      description: |-
+        Always `no-store`. A dataset carries the network key and the PSKc,
+        and a pending one also carries a delay timer that is stale as soon
+        as it is stored, so no cache on the way should keep a copy.
+      schema:
+        type: string
+
+    DatasetEtag:
+      description: |-
+        The entity tag of the stored dataset, for use in `If-Match`. Derived
+        from the dataset TLVs, so it changes exactly when the stored dataset
+        does and is shared by the JSON and TLV representations.
+      schema:
+        type: string
+
   parameters:
+    IfMatchEtag:
+      name: If-Match
+      in: header
+      required: false
+      schema:
+        type: string
+      description: |-
+        One or more entity tags as returned in `ETag` by a GET of this
+        resource, or `*`. The dataset is only written when one of the tags
+        matches the stored dataset (`*`: when any dataset is stored),
+        turning a read-modify-write sequence into a compare-and-swap;
+        otherwise the request fails with 412 and nothing is written.
+        Evaluated before `If-None-Match` (RFC 9110 section 13.2.2),
+        atomically with the write. Tags are compared with the weak function
+        (section 8.8.3.2), so a tag matches whether or not it is sent in the
+        `W/` form: the tags this server hands out are weak, and the strong
+        comparison section 13.1.1 asks for never matches a weak tag, which
+        would leave `If-Match` unusable here. A syntactically
+        invalid value is rejected with 400. The active dataset is only
+        writable while Thread is stopped, and that check runs first: on a
+        running node the response is 409 whether or not this header is
+        present.
+
     IfNoneMatchStar:
       name: If-None-Match
       in: header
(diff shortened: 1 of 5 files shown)


src/rest/rest_web_server.cpp
```diff
@@ -628,51 +631,137 @@ void RestWebServer::Rloc(const Request &aRequest, Response &aResponse) const
...
+        }
+        if (end == std::string::npos)
+        {
+            break;
+        }
+        begin = end + 1;
+    }
+
+    return true;
+}
```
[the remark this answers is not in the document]
LorbusChris: Fixed in the amended commit: ParseIfMatch now requires at least one entity tag -- empty list elements are still ignored per RFC 9110 section 5.6.1, but a value containing only empty elements answers 400; test added.

src/rest/rest_web_server.cpp
```diff
@@ -715,6 +836,27 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
+
+                if (matched && !ifMatchStar)
+                {
+                    const std::string current = DatasetEtag(datasetTlvs);
+
+                    matched = std::find(ifMatchTags.begin(), ifMatchTags.end(), current) != ifMatchTags.end();
+                }
+                VerifyOrReturn(matched, OTBR_ERROR_DUPLICATED);
+            }
+            VerifyOrReturn(!onlyIfAbsent || errorOt != OT_ERROR_NONE, OTBR_ERROR_DUPLICATED);
```
jwhui, line 1025: Should update the check so `If-None-Match` is only evaluated when `If-Match` is not present?

Per RFC 9110 Section 13.2.2, for state-changing methods like PUT, if `If-Match` is present and evaluates to `true`, the server must perform the requested method without evaluating `If-None-Match`.

Currently, if a client provides both `If-Match: "<matching_etag>"` and `If-None-Match: *` when a dataset exists, `VerifyOrReturn(!onlyIfAbsent || errorOt != OT_ERROR_NONE, OTBR_ERROR_DUPLICATED)` will evaluate to false and fail with 412.

src/rest/rest_web_server.cpp
jwhui, line 82: Since `If-Match` and `If-None-Match` are not CORS-safelisted request headers, browser-based clients (such as the web GUI or external dashboards) making conditional PUT requests will trigger a CORS preflight OPTIONS check.

To support browser clients:

1. Add `If-Match`, `If-None-Match` to `OTBR_REST_ACCESS_CONTROL_ALLOW_HEADERS`.
2. Add `Access-Control-Expose-Headers: ETag` to the default CORS response headers in `Init` so cross-origin browser scripts can read `response.headers.get("ETag")`.

src/rest/rest_web_server.cpp
```diff
@@ -697,8 +793,33 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
 
+    // RFC 9110 section 13.1.2: If-None-Match with "*" makes the write
+    // conditional on no dataset being in place. A pending dataset that is
+    // not newer than the one already held is silently ignored by the mesh
+    // while this handler accepts the write, so a client that does not intend
+    // to supersede an in-flight dataset can turn that collision into an
+    // error it sees. Other forms of the header (entity tags) have nothing to
+    // match against here and are rejected rather than mistaken for an
+    // unconditional write.
+    const std::string ifNoneMatch  = aRequest.get_header_value(OT_REST_IF_NONE_MATCH_HEADER);
```
jwhui: `ParseIfMatch` trims OWS via `TrimOws()`, but `ifNoneMatch` does not. A valid header carrying optional whitespace (e.g., `If-None-Match:  *`) will fail the check on line 817 and return 400 Bad Request.

Consider applying `TrimOws` to `ifNoneMatch` as well.

src/rest/rest_web_server.cpp
```diff
@@ -628,51 +631,144 @@ void RestWebServer::Rloc(const Request &aRequest, Response &aResponse) const
...
-                }
-            }
+                      if (aRequest.get_header_value(OT_REST_ACCEPT_HEADER) == OT_REST_CONTENT_TYPE_PLAIN)
+                      {
+                          contentType = OT_REST_CONTENT_TYPE_PLAIN;
+                          body        = Utils::Bytes2Hex(datasetTlvs.mTlvs, datasetTlvs.mLength);
+                      }
+                      else
+                      {
+                          otOperationalDataset dataset;
```
jwhui:
```suggestion
                          otOperationalDataset dataset = {}
```

src/rest/openapi.yaml
jwhui: Since this introduces new feature capabilities (`ETag`, `If-Match`, and `If-None-Match` on dataset endpoints), consider bumping the REST API version from 0.5.0 to 0.6.0 in both `src/rest/version.hpp` and `src/rest/openapi.yaml`.

src/rest/rest_web_server.cpp
LorbusChris: I read 13.2.2 the other way: Step 1 says a true If-Match continues to step 3, and step 3 evaluates If-None-Match

src/rest/rest_web_server.cpp
LorbusChris: Done, thanks.

src/rest/openapi.yaml
LorbusChris: Also done.

src/rest/rest_web_server.cpp
LorbusChris: Done.

jwhui:
@LorbusChris , can you help rebase?

src/rest/openapi.yaml
```diff
@@ -2499,7 +2519,36 @@ components:
 
   ### Parameters ##############################################################
   #############################################################################
+  headers:
+    DatasetEtag:
+      description: |-
+        The entity tag of the stored dataset, for use in `If-Match`. Derived
+        from the dataset TLVs, so it changes exactly when the stored dataset
+        does and is shared by the JSON and TLV representations.
```
agners, line 2539: This is not 100% honest: The delay timer TLV is left out, and I think it's worth pointing this out here too.

Especially since this actually violates [RFC 9110 §8.8.1.](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.8.1)

> A "strong validator" is representation metadata that changes value whenever a change occurs to the representation data that would be observable in the content of a [200 (OK)](https://www.rfc-editor.org/rfc/rfc9110.html#status.200) response to GET.

So this actually makes a case emitting a weak tag. However, this then doesn't allow to use the tag in the If-Match. Unless we break [RFC 9110 §13.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-13.1.1)

> An origin server MUST use the strong comparison function when comparing entity tags for If-Match ([Section 8.8.3.2](https://www.rfc-editor.org/rfc/rfc9110.html#entity.tag.comparison)), since the client intends this precondition to prevent the method from being applied if there have been any changes to the representation data.

We are a bit between a rock and a hard place here: But I think I'd go with weak E-Tag by default, and accept weak comparison in the If-Match. This violates the spec too, but it feels more honest to me, since we don't need to lie on the read side and it is the write side where we consider the delay timer not relevant for dataset comparison.

In a way, this is what we are aiming for:

> This weakness might be due to limitations in how the value is calculated (e.g., clock resolution), an inability to ensure uniqueness for all possible representations of the resource, **or a desire of the resource owner to group representations by some self-determined set of equivalency rather than unique sequences of data.**

Besides, Claude also claims that reverse proxies (which we also considered adding to the HA OTBR app) potentially convert E-Tag to weak ones when using gzip. I haven't verified that, but using weak E-Tags from the origin neatly sides steps that too.

We also share the tag between JSON and plain text/TLV representation, which technically is not spec compliant:

> A strong validator is unique across all versions of all representations associated with a particular resource over time.

src/rest/openapi.yaml
```diff
@@ -1534,14 +1548,20 @@ paths:
       responses:
         "200":
           description: Successfully updated the pending operational dataset.
+          headers:
+            ETag:
+              $ref: "#/components/headers/DatasetEtag"
```
agners, line 1557: Can we also add `Cache-Control: no-store` to the `GET` method to avoid caching? Especially with ETags present, this potentially return outdated delay timers if the response gets cached along the way. Also we include keys which we don't want to be cached 😅 . The latter is technically pre-existing, but since we add ETags which are also cache relevant, it still fits with this PR IMHO.

src/rest/rest_web_server.cpp
```diff
@@ -764,51 +770,144 @@ void RestWebServer::Rloc(const Request &aRequest, Response &aResponse) const
...
+    while (offset + 2u <= aDatasetTlvs.mLength)
+    {
+        const uint8_t *tlv    = &aDatasetTlvs.mTlvs[offset];
+        uint16_t       length = static_cast<uint16_t>(2 + tlv[1]);
+
+        if (offset + length > aDatasetTlvs.mLength)
+        {
+            break;
+        }
+        if (tlv[0] != OT_MESHCOP_TLV_DELAYTIMER)
```
agners, line 806: We should include `openthread/dataset.h` probably. It appears to compile, so I guess it is included through other headers, but this is potentially fragile.

src/rest/rest_web_server.cpp
```diff
@@ -764,51 +770,144 @@ void RestWebServer::Rloc(const Request &aRequest, Response &aResponse) const
...
+        {
+            sha256.Update(tlv, length);
+        }
+        offset += length;
+    }
+    sha256.Finish(hash);
+
+    return '"' + Utils::Bytes2Hex(hash.GetBytes(), 8) + '"';
```
agners: Same here, we should include `utils/hex.hpp` for this.

agners:
> A controller that reads a dataset, stamps a successor and writes it back has no way to detect another writer slipping in between the read and the write.

Maybe we should also ask ourselfs if this is really a use case we want to/need to handle :thinking:

LorbusChris:
Moved to draft while I give this some more thought.

PR #3555, opened 2026-08-26
[agent] use sigaction for signal handlers at start of Application::Run

jwhui:
This commit updates Application::Run to install signal handlers for SIGTERM and SIGPIPE using sigaction() instead of signal() at the very top of the function, before any service readiness notifications (such as systemd, Upstart, or generic file-descriptor supervisors) are dispatched.

Using sigaction() with sa_flags = 0 provides portable, reliable semantics across UNIX-like systems and ensures blocking system calls (such as select()) are interrupted upon receiving a signal so that termination (sShouldTerminate) is handled immediately without waiting for the poll timeout.

### DIFF
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -35,6 +35,7 @@
 
 #include <errno.h>
 #include <limits.h>
+#include <signal.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


abtink (approved):
👍

PR #3564, opened 2026-09-01
[misc] fix cmake install manifest spelling

dsr-vafokhon-safokhujaev:
Change from install_manifests.txt to install_manifest.txt
This fixes uninstall of otbr

### DIFF
--- a/script/_otbr
+++ b/script/_otbr
@@ -95,9 +95,9 @@ otbr_uninstall()
     sudo killall otbr-web otbr-agent || true
 
     (
-        if cd "${OTBR_TOP_BUILDDIR}"; then
-            # shellcheck disable=SC2024
-            sudo xargs rm <install_manifests.txt || true
+        if cd "${OTBR_TOP_BUILDDIR}" 2>/dev/null && [ -f install_manifest.txt ]; then
+            sudo xargs -a install_manifest.txt -r rm -f || true
+            rm -f install_manifest.txt
         fi
     )
     if have systemctl; then


script/_otbr
```diff
@@ -97,7 +97,7 @@ otbr_uninstall()
     (
         if cd "${OTBR_TOP_BUILDDIR}"; then
             # shellcheck disable=SC2024
-            sudo xargs rm <install_manifests.txt || true
+            [ -f install_manifest.txt ] && sudo xargs rm -f <install_manifest.txt || true
```
jwhui: How about something like:

```
    (
        if cd "${OTBR_TOP_BUILDDIR}" 2>/dev/null && [ -f install_manifest.txt ]; then
            sudo xargs -a install_manifest.txt -r rm -f || true
        fi
    )
```

1. Rather than chaining `[ -f ... ] && ... || true`, a standard `if` block is clearer and more idiomatic in bash.
2. By using `xargs -a install_manifest.txt -r rm -f`, `sudo` reads the file directly (avoiding potential permission issues if the manifest was created with a restrictive umask) and eliminates the need for `# shellcheck disable=SC2024`.
3. Redirecting `cd 2>/dev/null` avoids printing noisy errors to stderr if the build directory doesn't exist yet.

dsr-vafokhon-safokhujaev: Applied your suggestion!
Thank you for explanation!

script/_otbr
```diff
@@ -95,9 +95,8 @@ otbr_uninstall()
     sudo killall otbr-web otbr-agent || true
 
     (
-        if cd "${OTBR_TOP_BUILDDIR}"; then
-            # shellcheck disable=SC2024
-            sudo xargs rm <install_manifests.txt || true
+        if cd "${OTBR_TOP_BUILDDIR}" 2>/dev/null && [ -f install_manifest.txt ]; then
+            sudo xargs -a install_manifest.txt -r rm -f || true
```
jwhui, line 99: Would it make sense to remove the manifest file itself once the uninstallation completes?

```suggestion
            sudo xargs -a install_manifest.txt -r rm -f || true
            rm -f install_manifest.txt
```

jwhui (commented):
Also, please rebase against the latest main branch and squash the 3 commits into a single commit. Thanks.

script/_otbr
dsr-vafokhon-safokhujaev: Indeed! Thanks!

jwhui (approved):
Thanks! 👍🏼

Issue #3469, opened 2026-07-15
macOS host support for otbr-agent (RCP mode) — working port, seeking guidance before PRs


clara-mostly-human:
---

## Summary

I have otbr-agent running as a full border router on macOS (Apple Silicon
Mac mini, bare metal, RCP mode with a SONOFF EFR32MG24 dongle). Real
Matter-over-Thread devices (two Nuki smart locks) are commissioned through
it into Home Assistant and work reliably, including across host reboots.

The surprising part: the tree is *nearly there already*. A stock checkout
gets 96% through compilation with `-DOTBR_MDNS=mDNSResponder`. It took two
small changes in this repo plus one in openthread to reach a fully working
border router:

1. `src/rest/json.cpp`: a C++ VLA fails Apple clang with `-Werror`
   (`-Wvla-cxx-extension`); fixed with a fixed-size buffer.
2. `src/mdns/CMakeLists.txt`: don't link a standalone `dns_sd` library on
   Apple platforms (the API is in libSystem).
3. openthread `src/posix/platform/infra_if.cpp`: `IPV6_BOUND_IF` takes an
   interface index on macOS, not a name; the shared BSD branch passed the
   name and died with EINVAL at startup. (Separate PR to
   openthread/openthread.)

## What works (verified with real hardware)

- RCP over HDLC/UART (460800 baud, EFR32MG24)
- utun-based Thread netif; network forms, state=leader, dataset persistence
- mDNS via native mDNSResponder: `_meshcop._udp` advertised; discovered by
  Home Assistant on the LAN
- SRP server + advertising proxy: Thread devices' `_matter._tcp` records
  published to the LAN with AAAA resolution
- Border routing: RA emission with PIO + RIO on the infra interface works
  out of the box (tcpdump-verified); LAN hosts learn routes and reach mesh
  devices; Matter CASE sessions established host->mesh
- Full Matter commissioning of Thread devices via HA + phone flow

## Platform requirements found

- `sysctl net.inet6.ip6.forwarding=1` is required (macOS defaults to 0).
  Without it the host neither forwards infra<->utun traffic nor delivers
  packets arriving on the infra interface for utun-owned addresses.
  Question for maintainers: should otbr-agent set this at startup on macOS
  (matching what the Linux setup scripts do), or should it be a documented
  deployment requirement?

## Not attempted / out of scope for a first PR

- Backbone router: Darwin has no usable IPv6 multicast routing (MRT6
[...]

PR #3575, opened 2026-09-05
[mdns] fix the mDNSResponder build on Apple platforms

clara-mostly-human:
Two things keep `-DOTBR_MDNS=mDNSResponder` from building on macOS:

- On Apple platforms the dns_sd API is part of libSystem; the standalone library only exists elsewhere (e.g. mDNSResponder compatibility packages on Linux). Linking `dns_sd` unconditionally fails the link with `ld: library 'dns_sd' not found`. Link it only on non-Apple platforms.
- `kDNSServiceErr_StaleData` entered `dns_sd.h` with mDNSResponder 2559 (the macOS 15 SDK). Against an older Apple SDK — such as the macOS 14 SDK the macos-14 CI runner ships — the build fails with `use of undeclared identifier 'kDNSServiceErr_StaleData'` (since #2971). Apple's header stamps its build into `_DNS_SD_H`, so the two uses are gated on it; on an older SDK the error falls through to the default cases.

Part of #3469, split into separate PRs as requested there. #3577 adds a macOS CI leg that builds this configuration.

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -52,6 +52,16 @@
 #include "utils/dns_utils.hpp"
 #include "utils/string_utils.hpp"
 
+// kDNSServiceErr_StaleData entered dns_sd.h with mDNSResponder 2559 (the
+// macOS 15 SDK); older Apple SDKs predate it, and Apple's header stamps its
+// build into _DNS_SD_H. The mDNSResponder builds used on other platforms are
+// recent enough to have it.
+#if defined(__APPLE__) && defined(_DNS_SD_H) && ((_DNS_SD_H + 0) < 2559000000LL)
+#define OTBR_MDNSSD_HAVE_STALE_DATA 0
+#else
+#define OTBR_MDNSSD_HAVE_STALE_DATA 1
+#endif
+
 namespace otbr {
 
 namespace Mdns {
@@ -214,8 +224,10 @@ static const char *DNSErrorToString(DNSServiceErrorType aError)
     case kDNSServiceErr_DefunctConnection:
         return "Defunct Connection";
 
+#if OTBR_MDNSSD_HAVE_STALE_DATA
     case kDNSServiceErr_StaleData:
         return "Stale Data";
+#endif
 
     default:
         return "Unhandled Error";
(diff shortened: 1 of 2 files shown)


clara-mostly-human:
Added a second commit to this PR: the new macOS CI leg in #3577 (mDNSResponder + border agent on the macos-14 runner) surfaced that `src/mdns/mdns_mdnssd.cpp` uses `kDNSServiceErr_StaleData` unguarded. That code entered `dns_sd.h` with mDNSResponder 2559 (the macOS 15 SDK), so the mDNSResponder configuration has not built against the macOS 14 SDK since #2971. Apple stamps the build into `_DNS_SD_H`, so the two uses are now gated on it — an older SDK simply falls through to the default cases. Kept here rather than in a separate PR since it is the same concern (the mdns module building on Apple platforms); happy to split it out if you prefer.

src/mdns/mdns_mdnssd.cpp
```diff
@@ -52,6 +52,16 @@
 #include "utils/dns_utils.hpp"
 #include "utils/string_utils.hpp"
 
+// kDNSServiceErr_StaleData entered dns_sd.h with mDNSResponder 2559 (the
+// macOS 15 SDK); older Apple SDKs predate it, and Apple's header stamps its
+// build into _DNS_SD_H. The mDNSResponder builds used on other platforms are
+// recent enough to have it.
+#if defined(__APPLE__) && (_DNS_SD_H < 2559000000)
```
jwhui: Consider refining this preprocessor check to be slightly more defensive:

```suggestion
#if defined(__APPLE__) && defined(_DNS_SD_H) && ((_DNS_SD_H + 0) < 2559000000LL)
```

1. `defined(_DNS_SD_H)` and `+ 0`: Checking `defined(_DNS_SD_H)` and using `(_DNS_SD_H + 0)` is the idiomatic way to query `_DNS_SD_H` on Apple platforms and avoids warnings if `-Wundef` is enabled.
2. `2559000000LL`: Since `2559000000` exceeds `INT32_MAX (2,147,483,647)`, appending the `LL` suffix avoids integer constant truncation/overflow warnings on 32-bit targets or under strict compiler warning modes.

clara-mostly-human: Done — `defined(_DNS_SD_H) && ((_DNS_SD_H + 0) < 2559000000LL)`.

jwhui (approved):
Thanks! 👍🏼

jwhui:
@clara-mostly-human , can you squash commits and update commit message?

clara-mostly-human:
Done — squashed into one commit, message and PR description updated to cover both fixes.

PR #3576, opened 2026-09-05
[script] enable IP forwarding on macOS in _ipforward

clara-mostly-human:
macOS defaults `net.inet6.ip6.forwarding` to 0. Without it the host neither forwards traffic between the infrastructure and Thread interfaces nor delivers packets arriving on the infrastructure interface for addresses owned by the Thread utun, so a border router on a macOS host does not work until the sysctl is set.

This mirrors the Linux implementation, as suggested in #3469: enable forwarding immediately with `sysctl` and persist it across reboots. macOS has no `sysctl.d` and `/etc/sysctl.conf` is not reliably applied on recent releases, so the setting is persisted as a launchd daemon (`io.openthread.otbr-ipforward`) that re-applies it at boot; `ipforward_uninstall` boots the daemon out and removes the plist. Linux behavior is unchanged.

Checked with `shfmt -i 4 -bn -ci -fn -s` and `shellcheck -x`.

Part of #3469.

### DIFF
--- a/script/_ipforward
+++ b/script/_ipforward
@@ -32,8 +32,24 @@
 
 SYSCTL_IP_FORWARD=/etc/sysctl.d/60-otbr-ip-forward.conf
 
+# macOS has no sysctl.d, and /etc/sysctl.conf is not reliably applied on
+# recent releases, so the setting is persisted as a launchd daemon that
+# re-applies it at boot (the counterpart of the sysctl.d drop-in on Linux).
+LAUNCHD_IP_FORWARD_LABEL=io.openthread.otbr-ipforward
+LAUNCHD_IP_FORWARD_PLIST=/Library/LaunchDaemons/${LAUNCHD_IP_FORWARD_LABEL}.plist
+
+ipforward_is_macos()
+{
+    [[ ${PLATFORM:-} == "macOS" || ${OSTYPE:-} == darwin* ]]
+}
+
 ipforward_install()
 {
+    if ipforward_is_macos; then
+        ipforward_install_macos
+        return
+    fi
+
     sudo tee $SYSCTL_IP_FORWARD <<EOF
 net.ipv6.conf.all.forwarding = 1
 net.ipv4.ip_forward = 1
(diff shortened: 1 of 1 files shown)


script/_ipforward
```diff
@@ -46,13 +62,51 @@ EOF
...
+        <string>/usr/sbin/sysctl</string>
+        <string>net.inet6.ip6.forwarding=1</string>
+        <string>net.inet.ip.forwarding=1</string>
+    </array>
+    <key>RunAtLoad</key>
+    <true/>
+</dict>
+</plist>
+EOF
+    sudo chmod 644 "$LAUNCHD_IP_FORWARD_PLIST"
```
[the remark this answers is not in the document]
clara-mostly-human: Done.

script/_ipforward
```diff
@@ -46,13 +62,51 @@ EOF
...
+    sudo launchctl bootstrap system "$LAUNCHD_IP_FORWARD_PLIST" || die 'Failed to load the IP forwarding launchd daemon!'
```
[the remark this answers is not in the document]
clara-mostly-human: Done — `bootout` (ignoring failure) before `bootstrap`.

script/_ipforward
```diff
@@ -46,13 +62,55 @@ EOF
     fi
 }
 
+ipforward_install_macos()
+{
+    sudo tee "$LAUNCHD_IP_FORWARD_PLIST" <<EOF
```
jwhui: `sudo tee` will echo the entire 17-line XML plist to standard output during setup. Consider redirecting stdout to `/dev/null` to keep installation logs clean:

```suggestion
    sudo tee "$LAUNCHD_IP_FORWARD_PLIST" >/dev/null <<EOF
```

script/_ipforward
```diff
@@ -46,13 +62,55 @@ EOF
...
+    <key>RunAtLoad</key>
+    <true/>
+</dict>
+</plist>
+EOF
+    # launchd refuses daemons whose plist is not owned by root:wheel with 644.
+    sudo chown root:wheel "$LAUNCHD_IP_FORWARD_PLIST"
+    sudo chmod 644 "$LAUNCHD_IP_FORWARD_PLIST"
+    # Re-running setup must not trip over an instance that is already loaded.
+    sudo launchctl bootout system "$LAUNCHD_IP_FORWARD_PLIST" 2>/dev/null || true
```
jwhui: For `bootout`, targeting the service label directly (`system/${LAUNCHD_IP_FORWARD_LABEL}`) rather than the plist path is standard in modern `launchctl` and avoids re-reading the file from disk:
```suggestion
    sudo launchctl bootout system/${LAUNCHD_IP_FORWARD_LABEL} 2>/dev/null || true
```

script/_ipforward
```diff
@@ -46,13 +62,55 @@ EOF
...
+    sudo launchctl bootstrap system "$LAUNCHD_IP_FORWARD_PLIST" || die 'Failed to load the IP forwarding launchd daemon!'
```
jwhui, line 92: If the service was previously disabled on the host, `launchctl bootstrap` can fail with `Bootstrap failed: 119: Service is disabled`. It can be helpful to run `enable` right before `bootstrap`:
```suggestion
    sudo launchctl enable system/${LAUNCHD_IP_FORWARD_LABEL} 2>/dev/null || true
    sudo launchctl bootstrap system "$LAUNCHD_IP_FORWARD_PLIST" || die 'Failed to load the IP forwarding launchd daemon!'
```

script/_ipforward
```diff
@@ -46,13 +62,55 @@ EOF
...
+    if ipforward_is_macos; then
+        if [[ -f $LAUNCHD_IP_FORWARD_PLIST ]]; then
+            sudo launchctl bootout system "$LAUNCHD_IP_FORWARD_PLIST" || true
+            sudo rm -v "$LAUNCHD_IP_FORWARD_PLIST"
+        fi
+        return
+    fi
+
     test ! -f $SYSCTL_IP_FORWARD || sudo rm -v $SYSCTL_IP_FORWARD
 }
```
jwhui, line 105: 1. `2>/dev/null` is missing on the `launchctl bootout` invocation (unlike line 90). If the service is not active (e.g. during a clean run or repeated uninstall), `launchctl` will output `Boot-out failed: 5: Input/output error` to stderr.
2. If the `.plist` file was deleted or missing, the `if [[ -f ... ]]` check skips `bootout` altogether, leaving the daemon registered in `launchd` memory.

Booting out by service label and separating the file removal avoids both issues and matches the Linux `test ! -f ...` pattern:

```suggestion
ipforward_uninstall()
{
    if ipforward_is_macos; then
        sudo launchctl bootout system/${LAUNCHD_IP_FORWARD_LABEL} 2>/dev/null || true
        test ! -f "$LAUNCHD_IP_FORWARD_PLIST" || sudo rm -v "$LAUNCHD_IP_FORWARD_PLIST"
        return
    fi

    test ! -f $SYSCTL_IP_FORWARD || sudo rm -v $SYSCTL_IP_FORWARD
}
```

script/_ipforward
```diff
@@ -32,8 +32,24 @@
...
 
+# macOS has no sysctl.d, and /etc/sysctl.conf is not reliably applied on
+# recent releases, so the setting is persisted as a launchd daemon that
+# re-applies it at boot (the counterpart of the sysctl.d drop-in on Linux).
+LAUNCHD_IP_FORWARD_LABEL=io.openthread.otbr-ipforward
+LAUNCHD_IP_FORWARD_PLIST=/Library/LaunchDaemons/${LAUNCHD_IP_FORWARD_LABEL}.plist
+
+ipforward_is_macos()
+{
+    [[ ${OSTYPE} == darwin* ]]
```
jwhui: To stay consistent with platform detection in `script/_initrc` (which sets `PLATFORM=macOS` when `${OSTYPE}` is `darwin*`), and to guard against potential unbound variable issues if `set -u` is enabled in calling shells, consider checking `PLATFORM` as well:

```suggestion
    [[ ${PLATFORM:-} == "macOS" || ${OSTYPE:-} == darwin* ]]
```

script/_ipforward
clara-mostly-human: Done.

script/_ipforward
clara-mostly-human: Done — bootout by `system/<label>` in both install and uninstall.

script/_ipforward
clara-mostly-human: Done — `enable` right before `bootstrap`.

script/_ipforward
clara-mostly-human: Done — bootout by label with stderr quieted, and the file removal is a separate `test ! -f … || rm -v` like the Linux path, so a missing plist no longer skips the bootout.

script/_ipforward
clara-mostly-human: Done — `[[ ${PLATFORM:-} == "macOS" || ${OSTYPE:-} == darwin* ]]`. Applied the same to `firewall_is_macos` in #3578 for consistency.

jwhui (approved):
Thanks! 👍🏼

PR #3577, opened 2026-09-05
[github-actions] build the mDNSResponder configuration on macOS

clara-mostly-human:
The macOS job only builds with `OTBR_MDNS=OFF` and the border agent off, so it does not cover the configuration a macOS host actually runs: native mDNSResponder with the border agent (and border routing) on. This turns the job into a two-leg matrix so the Apple-specific mDNS link path stays covered.

Builds on #3575 (merged), which the new leg needs to link.

Verified locally on macOS 26 / Apple Silicon with the exact `script/test build` invocation used by the new leg.

Part of #3469.

### DIFF
--- a/.github/workflows/macOS.yml
+++ b/.github/workflows/macOS.yml
@@ -45,7 +45,20 @@ permissions:
 
 jobs:
   build-check:
+    name: build-check (${{ matrix.name }})
     runs-on: macos-14
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: "Minimal (mDNS OFF, Border Agent OFF)"
+            mdns: "OFF"
+            border_agent: "OFF"
+          # The border router configuration used on macOS hosts: native
+          # mDNSResponder with the border agent (and border routing) on.
+          - name: "Native (mDNSResponder, Border Agent ON)"
+            mdns: "mDNSResponder"
+            border_agent: "ON"
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/macOS.yml
```diff
@@ -46,6 +46,18 @@ permissions:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - mdns: "OFF"
+            border_agent: "OFF"
+          # The border router configuration used on macOS hosts: native
+          # mDNSResponder with the border agent (and border routing) on.
+          - mdns: "mDNSResponder"
+            border_agent: "ON"
```
jwhui, line 61: GitHub Actions names matrix jobs using the values of matrix variables, which currently results in check names like `build-check (OFF, OFF)`. Adding a descriptive `name` to each matrix item makes the check list clearer at a glance:

```suggestion
    strategy:
      fail-fast: false
      matrix:
        include:
          - name: "Minimal (mDNS OFF, Border Agent OFF)"
            mdns: "OFF"
            border_agent: "OFF"
          - name: "Native (mDNSResponder, Border Agent ON)"
            mdns: "mDNSResponder"
            border_agent: "ON"
```

And in the job definition:
```
name: build-check (${{ matrix.name }})
```

.github/workflows/macOS.yml
```diff
@@ -46,6 +46,18 @@ permissions:
...
+    env:
+      OTBR_MDNS: ${{ matrix.mdns }}
```
jwhui: `matrix.mdns` is currently passed as an ambient environment variable `OTBR_MDNS` at the job level, while `border_agent` is passed directly in `OTBR_OPTIONS` in the `Build` step.

For consistency, consider passing `-DOTBR_MDNS=${{ matrix.mdns }}` directly within `OTBR_OPTIONS` alongside `-DOTBR_BORDER_AGENT=${{ matrix.border_agent }}` and omitting the top-level `env:` block. This keeps all CMake options visible in one place.

.github/workflows/macOS.yml
clara-mostly-human: Done — `build-check (Minimal (mDNS OFF, Border Agent OFF))` and `build-check (Native (mDNSResponder, Border Agent ON))`.

.github/workflows/macOS.yml
clara-mostly-human: Done — `-DOTBR_MDNS=${{ matrix.mdns }}` in `OTBR_OPTIONS`, `env:` block gone.

jwhui (approved):
Thanks! 👍🏼

PR #3578, opened 2026-09-05
[firewall] add an opt-in in-process pf backend for macOS

clara-mostly-human:
macOS has no netfilter; its firewall is pf. This adds a pf counterpart of the in-process nftables backend from #3325, as the follow-up promised in #3469.

With `-DOTBR_PF=ON` (opt-in, macOS only, mutually exclusive with `OTBR_NFTABLES`), otbr-agent owns the Thread ingress filter and the NAT44 masquerade in the pf anchor `otbr`, and produces the ingress prefix tables from Thread network data on `OT_CHANGED_THREAD_NETDATA` / `OT_CHANGED_ACTIVE_DATASET`, replacing the OpenThread posix platform firewall (which has no macOS port; `OT_FIREWALL` is forced off, as for nftables).

### Shape

- `PfFirewall` implements the same surface the agent already uses from `FirewallManager` (`Init`/`Deinit`, `EnableIngressFilter`, `EnableNat44Masquerade`, `ReplaceIngressPrefixes`) rather than `INftables`, whose vocabulary (tables/chains/hooks) does not map onto pf's anchors and tables. `Application` picks the backend at build time; `UpdateIngressPrefixes()` is shared.
- pf is driven through `pfctl(8)`, spawned directly with `posix_spawn` (no shell): Apple does not ship the pf ioctl interface (`net/pfvar.h`) in the public SDK, and pfctl is part of the base system, so there is no build-time dependency. The invocation sits behind `IPfctl` so the policy is unit-tested against a fake (`tests/gtest/test_pf_firewall.cpp`, 12 tests).
- The ruleset is regenerated and reloaded atomically (`pfctl -a otbr -f -`) when a feature is enabled; the two prefix tables are `persist` (a reload keeps their contents) and are each replaced atomically with `-T replace` as network data changes.

```
table <ingress_deny_src> persist
table <ingress_allow_dst> persist
nat pass on <infra> inet tagged otbr_thread -> (<infra>) # NAT44
pass in quick on <thread> inet tag otbr_thread            # NAT44
pass out quick on <thread> inet6 from (<thread>) to any
pass out quick on <thread> inet6 from (self) to any
block out quick on <thread> inet6 from <ingress_deny_src> to any
pass out quick on <thread> inet6 from any to <ingress_allow_dst>
pass out quick on <thread> inet6 from any to ff00::/8
block out quick on <thread> inet6 all
```

The ingress rules match the ip6tables FORWARD chain 1:1, with one difference: pf also sees the host's own output on the Thread interface, so it is exempted first. `(<thread>)` tracks the interface's addresses dynamically — the OMR address only arrives after the ruleset is loaded — and `(self)` covers the host's other addresses. A pf nat rule cannot match the inbound interface, so IPv4 from the Thread interface is tagged on the way in and the nat rule keys on the tag.

### Setup

pf only evaluates an anchor the main ruleset references, so `script/_firewall` gains a macOS branch: install adds `nat-anchor "otbr"` (next to the system's nat-anchor — pf requires translation before filtering) and `anchor "otbr"` to `/etc/pf.conf` and reloads it; uninstall removes them. `Init()` verifies the reference (`pfctl -s rules` / `-s nat`) and fails otherwise, so a build with the firewall never runs unfiltered — the same policy as the nftables backend. The rules are scoped to the interface name the platform reports (`otSysGetThreadNetifName()`), since macOS assigns `utunN` itself and it differs from the requested name.

### Tested

Apple Silicon Mac mini (macOS 26), otbr-agent in RCP mode as a Matter-over-Thread border router, in a mesh with three routers (the border router attaches as a router, the Matter locks parent elsewhere). After startup the anchor held the rules above on `utun0` — the kernel picked a different utun number than on the previous boot, which the backend followed — with `ingress_deny_src` = {OMR prefix, mesh-local prefix} and `ingress_allow_dst` = {OMR prefix}, all produced from network data. Host→mesh traffic sourced from the OMR address reached the locks through the filter, and Matter availability was unchanged across the switch from the previous (template/script) firewall. Running as the household's production border router since. Formatted with the `script/make-pretty` tooling (clang-format 19, shfmt, shellcheck).

Part of #3469. Adds a third leg to the macOS CI matrix from #3577 that builds this configuration with `-DOTBR_PF=ON`.

### DIFF
--- a/.github/workflows/macOS.yml
+++ b/.github/workflows/macOS.yml
@@ -54,11 +54,18 @@ jobs:
           - name: "Minimal (mDNS OFF, Border Agent OFF)"
             mdns: "OFF"
             border_agent: "OFF"
+            pf: "OFF"
           # The border router configuration used on macOS hosts: native
           # mDNSResponder with the border agent (and border routing) on.
           - name: "Native (mDNSResponder, Border Agent ON)"
             mdns: "mDNSResponder"
             border_agent: "ON"
+            pf: "OFF"
+          # The same, with the in-process pf firewall backend.
+          - name: "Native + pf firewall (mDNSResponder, Border Agent ON, OTBR_PF ON)"
+            mdns: "mDNSResponder"
+            border_agent: "ON"
+            pf: "ON"
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
@@ -72,9 +79,16 @@ jobs:
       run: |
         OTBR_OPTIONS="-DOTBR_MDNS=${{ matrix.mdns }} \
                       -DOTBR_BORDER_AGENT=${{ matrix.border_agent }} \
+                      -DOTBR_PF=${{ matrix.pf }} \
                       -DOTBR_BACKBONE_ROUTER=OFF \
                       -DOTBR_TREL=OFF \
                       -DOTBR_NAT64=OFF \
                       -DOTBR_DNS_UPSTREAM_QUERY=OFF \
                       -DOT_FIREWALL=OFF \
                       -DOTBR_DBUS=OFF" ./script/test build
+    # `script/test check` installs into /usr and runs the Linux-only suites;
+    # run the firewall unit tests directly (the pf tests only exist with
+    # OTBR_PF=ON, the FirewallManager tests in every configuration).
+    - name: Test
+      run: |
+        cd build/otbr && CTEST_OUTPUT_ON_FAILURE=1 ctest -R 'FirewallManagerTest|PfFirewall'
--- a/script/_firewall
+++ b/script/_firewall
@@ -29,7 +29,9 @@
 
 FIREWALL_SERVICE=/etc/init.d/otbr-firewall
 
-sudo modprobe ip6table_filter || true
+if have modprobe; then
+    sudo modprobe ip6table_filter || true
+fi
 
 FIREWALL="${FIREWALL:-1}"
 
@@ -51,9 +53,85 @@ if [ -z "${OTBR_NFTABLES:-}" ]; then
     fi
 fi
 
+# The in-process pf backend (-DOTBR_PF=ON, macOS) installs its rules into the
+# pf anchor "otbr" itself, and pf only evaluates an anchor the main ruleset
+# (/etc/pf.conf) references. Installing on macOS therefore means adding that
+# reference; there is no legacy firewall service there. The lines are tagged
+# so uninstall can find them again. pf wants translation rules before filter
+# rules, so the nat-anchor goes next to the system's own nat-anchor and only
+# the filter anchor is appended.
+# /usr/share sits on the SIP-protected system volume on macOS, so installs
+# there use another prefix; look in the usual ones.
+OTBR_PF_MARKER="/usr/share/otbr/pf-backend"
+[ -r "${OTBR_PF_MARKER}" ] || OTBR_PF_MARKER="/usr/local/share/otbr/pf-backend"
+[ -r "${OTBR_PF_MARKER}" ] || OTBR_PF_MARKER="/opt/homebrew/share/otbr/pf-backend"
+if [ -z "${OTBR_PF:-}" ]; then
+    OTBR_PF=0
+    if [ -r "${OTBR_PF_MARKER}" ] && grep -qx '1' "${OTBR_PF_MARKER}"; then
+        OTBR_PF=1
+    fi
+fi
+PF_CONF=/etc/pf.conf
+PF_ANCHOR_TAG="# otbr-agent (managed by script/_firewall)"
+
+firewall_is_macos()
+{
+    [[ ${PLATFORM:-} == "macOS" || ${OSTYPE:-} == darwin* ]]
+}
+
+# Replaces the main ruleset with the rewritten copy at "${PF_CONF}.otbr",
+# but only once pfctl has parsed it: a broken /etc/pf.conf would otherwise
+# be left in place.
+firewall_replace_pf_conf()
+{
+    # An empty file parses, so check for content first: a failed rewrite must
+    # not become an empty main ruleset.
+    [ -s "${PF_CONF}.otbr" ] || die 'Rewritten pf.conf is empty; leaving the original in place!'
+    sudo pfctl -n -f "${PF_CONF}.otbr" || die 'Rewritten pf.conf does not parse; leaving the original in place!'
+    sudo mv "${PF_CONF}.otbr" "${PF_CONF}"
+    sudo pfctl -f "${PF_CONF}" || die 'Failed to reload pf.conf!'
+}
+
+firewall_uninstall_macos()
+{
+    if grep -qF "${PF_ANCHOR_TAG}" "${PF_CONF}" 2>/dev/null; then
+        # awk matches the tag literally; it contains characters sed would
+        # read as a delimiter or metacharacters.
+        sudo awk -v tag="${PF_ANCHOR_TAG}" 'index($0, tag) { next } { print }' "${PF_CONF}" \
+            | sudo tee "${PF_CONF}.otbr" >/dev/null
+        sudo pfctl -a otbr -F all 2>/dev/null || true
+        firewall_replace_pf_conf
+    fi
+}
+
+firewall_install_macos()
+{
+    if [ "${OTBR_PF}" != "1" ]; then
+        echo "OTBR_PF=0: otbr-agent was built without the pf backend; no firewall to install on macOS."
+        return 0
+    fi
+
+    # Idempotent: earlier tagged lines are dropped before the new ones go in,
+    # so re-running setup never duplicates them. This deliberately does not
+    # go through firewall_uninstall_macos, which flushes the anchor: a running
+    # agent would lose its live rules until restarted.
+    sudo awk -v tag="${PF_ANCHOR_TAG}" '
+        index($0, tag) { next }
+        { print }
+        /^nat-anchor / && !done { print "nat-anchor \"otbr\" " tag; done = 1 }
+        END { if (!done) print "nat-anchor \"otbr\" " tag; print "anchor \"otbr\" " tag }
+    ' "${PF_CONF}" | sudo tee "${PF_CONF}.otbr" >/dev/null
+    firewall_replace_pf_conf
+}
+
 firewall_uninstall()
 {
     with FIREWALL || return 0
+
+    if firewall_is_macos; then
+        firewall_uninstall_macos
+        return 0
+    fi
 
     firewall_stop
     if have systemctl; then
@@ -70,6 +148,11 @@ firewall_install()
 {
     with FIREWALL || return 0
 
+    if firewall_is_macos; then
+        firewall_install_macos
+        return 0
+    fi
+
     if [ "${OTBR_NFTABLES}" = "1" ]; then
         echo "OTBR_NFTABLES=1: OTBR ingress filter is managed in-process (nftables); skipping legacy firewall service."
         return 0
@@ -85,6 +168,11 @@ firewall_install()
 
 firewall_start()
 {
+    # On macOS the agent installs the firewall itself; nothing to start.
+    if firewall_is_macos; then
+        return 0
+    fi
+
     # Same gate as the install: with the in-process backend there is no
     # otbr-firewall service to start, and starting the legacy one would put
     # its chain alongside the agent's.
@@ -104,6 +192,10 @@ firewall_start()
 firewall_stop()
 {
     with FIREWALL || return 0
+
+    if firewall_is_macos; then
+        return 0
+    fi
 
     if with DOCKER; then
         service otbr-firewall stop || true
--- a/src/agent/application.cpp
+++ b/src/agent/application.cpp
@@ -51,13 +51,16 @@
 #include "host/posix/dnssd.hpp"
 #include "utils/infra_link_selector.hpp"
 
-#if OTBR_ENABLE_NFTABLES
+#if OTBR_ENABLE_NFTABLES || OTBR_ENABLE_PF
 #include <algorithm>
 #include <string.h>
 #include <vector>
 
 #include <openthread/netdata.h>
 #include <openthread/thread.h>
+#endif
+#if OTBR_ENABLE_PF
+#include <openthread/openthread-system.h>
 #endif
 
 namespace otbr {
@@ -400,18 +403,26 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
     });
 #endif
 #endif // OTBR_ENABLE_BORDER_AGENT
-#if OTBR_ENABLE_NFTABLES
-    // The firewall is a security control: if it was built in (OTBR_NFTABLES) but
-    // cannot be installed, abort rather than silently forward traffic unfiltered.
-    // This mirrors the legacy script path, which `die`d when the otbr-firewall /
-    // otbr-nat44 services failed to start.
+#if OTBR_ENABLE_NFTABLES || OTBR_ENABLE_PF
+    // The firewall is a security control: if it was built in (OTBR_NFTABLES or
+    // OTBR_PF) but cannot be installed, abort rather than silently forward
+    // traffic unfiltered. This mirrors the legacy script path, which `die`d when
+    // the otbr-firewall / otbr-nat44 services failed to start.
     // Each result is taken once: SuccessOrDie() names its argument twice, so a
     // call passed to it directly runs a second time on the failing path and the
     // fatal log reports what that second attempt returned.
     otbrError firewallError;
 
+#if OTBR_ENABLE_NFTABLES
     firewallError = mNftables->Init();
     SuccessOrDie(firewallError, "Failed to initialize the nftables firewall!");
+#else
+    // pf rules are scoped to the Thread interface by name, and on macOS the
+    // kernel picks that name (utunN) when mHost.Init() creates the interface,
+    // so the backend is created here rather than in CreateRcpMode().
+    mPfctl    = MakeUnique<Firewall::PfctlProcess>();
+    mFirewall = MakeUnique<Firewall::PfFirewall>(*mPfctl, otSysGetThreadNetifName());
+#endif
     firewallError = mFirewall->Init();
     SuccessOrDie(firewallError, "Failed to initialize the firewall manager!");
     firewallError = mFirewall->EnableIngressFilter();
@@ -457,7 +468,7 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
 #endif
 }
 
-#if OTBR_ENABLE_NFTABLES
+#if OTBR_ENABLE_NFTABLES || OTBR_ENABLE_PF
 void Application::UpdateIngressPrefixes(void)
 {
     otNetworkDataIterator    iterator = OT_NETWORK_DATA_ITERATOR_INIT;
@@ -505,7 +516,7 @@ void Application::UpdateIngressPrefixes(void)
 
     if (mFirewall->ReplaceIngressPrefixes(denySrc, allowDst) != OTBR_ERROR_NONE)
     {
-        otbrLogWarning("FirewallManager: failed to update ingress prefixes");
+        otbrLogWarning("Firewall: failed to update ingress prefixes");
     }
 
 exit:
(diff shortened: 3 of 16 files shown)


script/_firewall
```diff
@@ -51,10 +53,62 @@ if [ -z "${OTBR_NFTABLES:-}" ]; then
...
+}
+
+firewall_install_macos()
+{
+    if [ "${OTBR_PF}" != "1" ]; then
+        echo "OTBR_PF=0: otbr-agent was built without the pf backend; no firewall to install on macOS."
+        return 0
+    fi
+
+    sudo awk -v tag="${PF_ANCHOR_TAG}" '
```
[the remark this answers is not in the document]
clara-mostly-human: Done, with one difference from the suggestion: the install now drops previously tagged lines itself (the awk skips lines carrying the tag) rather than calling `firewall_uninstall_macos` first, because uninstall flushes the `otbr` anchor — with an agent running, that would drop the live ingress rules until its next restart. Same idempotency, no window without a filter.

src/firewall/pfctl.cpp
```diff
@@ -0,0 +1,233 @@
...
+    {
+        ssize_t count = write(inPipe[1], aInput.data() + written, aInput.size() - written);
+
+        if (count < 0)
+        {
+            VerifyOrExit(errno == EINTR, error = OTBR_ERROR_ERRNO);
+            continue;
+        }
+        written += static_cast<size_t>(count);
+    }
```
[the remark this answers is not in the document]
clara-mostly-human: Done — and it needed one more piece to actually work: otbr-agent does not ignore SIGPIPE process-wide, so a write to a pipe pfctl had already closed would have killed the agent before EPIPE was ever returned. The input pipe is now marked `F_SETNOSIGPIPE` (a macOS fcntl; this backend is macOS-only), the write loop breaks on EPIPE, and the drain loop then captures pfctl's stderr so the ruleset error is logged.

src/firewall/pf_firewall.cpp
```diff
@@ -0,0 +1,351 @@
...
+        mEnableToken += all[pos];
+    }
+    VerifyOrExit(!mEnableToken.empty(), error = OTBR_ERROR_PARSE);
+
+exit:
+    if (error == OTBR_ERROR_PARSE)
+    {
+        otbrLogWarning("`pfctl -E` did not report a reference token: %s", all.c_str());
+    }
+    return error;
```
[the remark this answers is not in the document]
clara-mostly-human: Done — a missing token now logs a warning and Init proceeds; Deinit skips `pfctl -X` when there is nothing to release. Added a unit test for that path (`InitSucceedsWithoutReferenceTokenAndDeinitSkipsRelease`).

src/firewall/pfctl.cpp
```diff
@@ -0,0 +1,246 @@
...
+    pid = -1;
+
+    VerifyOrExit(WIFEXITED(status) && WEXITSTATUS(status) == 0, error = OTBR_ERROR_ERRNO);
+
+exit:
+    if (pid > 0)
+    {
+        // Spawned but not reaped because something failed part-way: don't
+        // leave a zombie behind.
+        while (waitpid(pid, &status, 0) < 0 && errno == EINTR)
```
jwhui, line 246: `waitpid()` is called before `ClosePipeEnd()` closes the remaining pipes.

f an error occurs during `write()` or during `poll()` before `ClosePipeEnd(inPipe[1])` is reached, `pfctl` will still be running and waiting for input / EOF on `stdin`. The parent then blocks indefinitely in `waitpid`, waiting for `pfctl` to exit, causing a deadlock.

All pipe ends should be closed before attempting to reap `pid`. In addition, if aborting early with a running child, sending `SIGKILL` (or `SIGTERM`) before waiting prevents blocking indefinitely if the child hangs.

script/_firewall
```diff
@@ -51,10 +53,67 @@ if [ -z "${OTBR_NFTABLES:-}" ]; then
...
+
+firewall_is_macos()
+{
+    [[ ${PLATFORM:-} == "macOS" || ${OSTYPE:-} == darwin* ]]
+}
+
+firewall_uninstall_macos()
+{
+    if grep -q "${PF_ANCHOR_TAG}" "${PF_CONF}" 2>/dev/null; then
+        sudo sed -i '' "/${PF_ANCHOR_TAG}/d" "${PF_CONF}"
```
jwhui: `PF_ANCHOR_TAG` contains a slash (`# otbr-agent (managed by script/_firewall)`). In `sed -i '' "/${PF_ANCHOR_TAG}/d" "${PF_CONF}"`, BSD `sed` on macOS treats the slash before `_firewall` as the terminating regex delimiter, causing `firewall_uninstall_macos` to fail.

Consider using an alternate delimiter with `sed` or using `awk` similar to `firewall_install_macos`.

src/firewall/pf_firewall.cpp
```diff
@@ -0,0 +1,352 @@
...
+    rules += "table " + denySrc + " persist\n";
+    rules += "table " + allowDst + " persist\n";
+
+    if (mNat44Enabled)
+    {
+        // Masquerade IPv4 that arrived on the Thread interface (the NAT64 egress
+        // path) out the upstream interface. A nat rule cannot match the inbound
+        // interface, so the packets are tagged on the way in and the nat rule
+        // keys on the tag.
+        rules += "nat on " + mUpstreamIfName + " inet tagged " + kThreadTag + " -> (" + mUpstreamIfName + ")\n";
```
jwhui: `nat on <interface>` performs address translation, but does not automatically permit packets through filter rules. If `/etc/pf.conf` or an anchor applies outbound filtering on the upstream interface, the translated IPv4 packets will be blocked unless explicitly matched by a `pass out` rule.

Using `nat pass on ...` tells PF to skip subsequent filter evaluation for translated packets and automatically create state for return traffic:

```suggestion
        rules += "nat pass on " + mUpstreamIfName + " inet tagged " + kThreadTag + " -> (" + mUpstreamIfName + ")\n";
```

script/_firewall
```diff
@@ -51,10 +53,67 @@ if [ -z "${OTBR_NFTABLES:-}" ]; then
...
+    # so re-running setup never duplicates them. This deliberately does not
+    # go through firewall_uninstall_macos, which flushes the anchor: a running
+    # agent would lose its live rules until restarted.
+    sudo awk -v tag="${PF_ANCHOR_TAG}" '
+        index($0, tag) { next }
+        { print }
+        /^nat-anchor / && !done { print "nat-anchor \"otbr\" " tag; done = 1 }
+        END { if (!done) print "nat-anchor \"otbr\" " tag; print "anchor \"otbr\" " tag }
+    ' "${PF_CONF}" | sudo tee "${PF_CONF}.otbr" >/dev/null
+    sudo mv "${PF_CONF}.otbr" "${PF_CONF}"
```
jwhui: `${PF_CONF}.otbr` is moved over `/etc/pf.conf` before validating it with `pfctl`. If `pfctl -f` fails (e.g. due to syntax issues in `/etc/pf.conf`), the system configuration file is already overwritten and left in an invalid state.

Consider test-parsing the file first using `pfctl -n -f` before replacing `/etc/pf.conf`.

src/firewall/pfctl.cpp
```diff
@@ -0,0 +1,246 @@
...
+                continue;
+            }
+            if (fds[i].fd == outPipe[0])
+            {
+                if (!DrainPipe(outPipe[0], aOutput))
+                {
+                    ClosePipeEnd(outPipe[0]);
+                }
+            }
+            else if (!DrainPipe(errPipe[0], aError))
```
jwhui: The `else if` branch drains `errPipe[0]` without checking `fds[i].fd == errPipe[0]`. While `fds` only contains up to two descriptors, adding the explicit check `else if (fds[i].fd == errPipe[0])` is more symmetric and prevents subtle bugs if descriptor handling is ever extended.

script/_firewall
```diff
@@ -51,10 +53,67 @@ if [ -z "${OTBR_NFTABLES:-}" ]; then
...
 fi
 
+# The in-process pf backend (-DOTBR_PF=ON, macOS) installs its rules into the
+# pf anchor "otbr" itself, and pf only evaluates an anchor the main ruleset
+# (/etc/pf.conf) references. Installing on macOS therefore means adding that
+# reference; there is no legacy firewall service there. The lines are tagged
+# so uninstall can find them again. pf wants translation rules before filter
+# rules, so the nat-anchor goes next to the system's own nat-anchor and only
+# the filter anchor is appended.
+OTBR_PF_MARKER="/usr/share/otbr/pf-backend"
```
jwhui, line 65: On macOS, `/usr/share` is part of the read-only root system volume protected by System Integrity Protection (SIP). Standard CMake installations on macOS use a prefix like `/usr/local` or `/opt/homebrew`, placing the marker in `/usr/local/share/otbr/pf-backend`.

Consider checking `/usr/local/share` as a fallback.

```suggestion
OTBR_PF_MARKER="/usr/share/otbr/pf-backend"
[ -r "${OTBR_PF_MARKER}" ] || OTBR_PF_MARKER="/usr/local/share/otbr/pf-backend"
```

src/agent/application.cpp
```diff
@@ -400,18 +403,26 @@ void Application::InitRcpMode(const std::string &aRestListenAddress, int aRestLi
...
 
+#if OTBR_ENABLE_NFTABLES
     firewallError = mNftables->Init();
     SuccessOrDie(firewallError, "Failed to initialize the nftables firewall!");
+#else
+    // pf rules are scoped to the Thread interface by name, and on macOS the
+    // kernel picks that name (utunN) when mHost.Init() creates the interface,
+    // so the backend is created here rather than in CreateRcpMode().
+    mPfctl    = MakeUnique<Firewall::PfctlProcess>();
+    mFirewall = MakeUnique<Firewall::PfFirewall>(*mPfctl, otSysGetThreadNetifName());
```
jwhui, line 424: `mPfctl` and `mFirewall` are allocated via `MakeUnique`.

In `DeinitRcpMode`, `mFirewall->Deinit()` is called, but neither pointer is reset. If `InitRcpMode` were ever called again within the same process lifecycle, re-assigning `mPfctl` first would destroy the old `PfctlProcess` while `mFirewall` still holds a reference to it.

Consider resetting both pointers in `DeinitRcpMode` (in reverse order of creation).

src/firewall/pfctl.cpp
clara-mostly-human: Done — the exit path now closes all six pipe ends first (so a pfctl still reading stdin sees EOF), then kills a child that was spawned but not reaped before waiting on it.

script/_firewall
clara-mostly-human: Done — uninstall now uses the same literal-match awk as install (`index($0, tag)`), through a shared helper; no sed.

src/firewall/pf_firewall.cpp
clara-mostly-human: Done — `nat pass`, and the comment explains why. Verified it parses on macOS pf.

script/_firewall
clara-mostly-human: Done — a shared `firewall_replace_pf_conf` runs `pfctl -n -f` on the rewritten copy first and only then moves it over `/etc/pf.conf` and reloads; both install and uninstall go through it.

src/firewall/pfctl.cpp
clara-mostly-human: Done.

script/_firewall
clara-mostly-human: Done — falls back to `/usr/local/share` and then `/opt/homebrew/share`.

src/agent/application.cpp
clara-mostly-human: Done — both are reset in `DeinitRcpMode`, `mFirewall` before `mPfctl`.

src/firewall/pfctl.cpp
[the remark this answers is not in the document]
clara-mostly-human: Correction to my earlier reply here: the agent does ignore SIGPIPE, but in `Run()`; the firewall is installed during `Init()`, before that handler exists, so the pipe-level `F_SETNOSIGPIPE` still stands for the startup calls. The behavior described above is unchanged.

script/_firewall
```diff
@@ -51,10 +53,83 @@ if [ -z "${OTBR_NFTABLES:-}" ]; then
...
+{
+    [[ ${PLATFORM:-} == "macOS" || ${OSTYPE:-} == darwin* ]]
+}
+
+# Replaces the main ruleset with the rewritten copy at "${PF_CONF}.otbr",
+# but only once pfctl has parsed it: a broken /etc/pf.conf would otherwise
+# be left in place.
+firewall_replace_pf_conf()
+{
+    sudo pfctl -n -f "${PF_CONF}.otbr" || die 'Rewritten pf.conf does not parse; leaving the original in place!'
```
jwhui, line 90: If `awk` encounters an error or produces empty output, `pfctl -n -f` will still succeed (an empty `pf.conf` is syntactically valid in pf), and `sudo mv` will overwrite `/etc/pf.conf` with a 0-byte file.

Consider ensuring `${PF_CONF}.otbr` is non-empty before validating and replacing.

src/firewall/pf_firewall.cpp
```diff
@@ -0,0 +1,354 @@
...
+    std::istringstream lines;
+    std::string        line;
+    bool               found = false;
+
+    SuccessOrExit(error = RunPfctl({"-s", aListFlag}, "", output));
+
+    lines.str(output);
+    while (std::getline(lines, line))
+    {
+        if (line.compare(0, wanted.size(), wanted) == 0)
```
jwhui: `line.compare(0, wanted.size(), wanted) == 0` requires the rule to start at column 0. If `pfctl -s rules` output is ever indented or contains leading tabs/whitespace, the comparison will fail.

Consider trimming leading whitespace before comparing.

src/firewall/pfctl.cpp
```diff
@@ -0,0 +1,253 @@
...
+                 error = OTBR_ERROR_ERRNO);
+
+    argv.push_back(const_cast<char *>("pfctl"));
+    for (const std::string &arg : aArgs)
+    {
+        argv.push_back(const_cast<char *>(arg.c_str()));
+    }
+    argv.push_back(nullptr);
+
+    VerifyOrExit(posix_spawn(&pid, kPfctlPath, &actions, nullptr, argv.data(), environ) == 0, error = OTBR_ERROR_ERRNO);
```
jwhui: `posix_spawn` returns the error number directly on failure rather than setting `errno`. If `posix_spawn` fails, `errno` will retain a stale value (or 0) when `error = OTBR_ERROR_ERRNO` is set.

Consider capturing the return code.

src/agent/application.cpp
jwhui: To be consistent with the updated log message in `DeinitRcpMode`.

```suggestion
        otbrLogWarning("Firewall: failed to update ingress prefixes");
```

script/_firewall
clara-mostly-human: Done — `firewall_replace_pf_conf` now requires the rewritten file to be non-empty (`test -s`) before the parse check and the move.

src/firewall/pf_firewall.cpp
clara-mostly-human: Done — the comparison starts at the first non-blank character of each line.

src/firewall/pfctl.cpp
clara-mostly-human: Done — the return code of `posix_spawn` and of the `posix_spawn_file_actions_*` calls is captured and stored in `errno` before the error path.

src/agent/application.cpp
clara-mostly-human: Done.

src/firewall/pf_firewall.cpp
```diff
@@ -0,0 +1,356 @@
...
+    error = mPfctl.Run({"-E"}, "", output, stderrText);
+    VerifyOrExit(error == OTBR_ERROR_NONE, otbrLogWarning("`pfctl -E` failed: %s", stderrText.c_str()));
+
+    // pfctl prints the reference as "Token : <n>"; only the digits after the
+    // word matter, whichever stream and spacing a given release uses. pf is
+    // enabled either way, so a token that cannot be found is not fatal: the
+    // reference just cannot be released on Deinit().
+    all = output + stderrText;
+    mEnableToken.clear();
+    pos = all.find("oken");
```
jwhui: Searching for the substring `"oken"` is quite broad and could match unrelated words (e.g. `"broken"`).

Consider searching for `"Token"` or `"token"`.

script/_firewall
```diff
@@ -51,10 +53,86 @@ if [ -z "${OTBR_NFTABLES:-}" ]; then
...
+# so uninstall can find them again. pf wants translation rules before filter
+# rules, so the nat-anchor goes next to the system's own nat-anchor and only
+# the filter anchor is appended.
+# /usr/share sits on the SIP-protected system volume on macOS, so installs
+# there use another prefix; look in the usual ones.
+OTBR_PF_MARKER="/usr/share/otbr/pf-backend"
+[ -r "${OTBR_PF_MARKER}" ] || OTBR_PF_MARKER="/usr/local/share/otbr/pf-backend"
+[ -r "${OTBR_PF_MARKER}" ] || OTBR_PF_MARKER="/opt/homebrew/share/otbr/pf-backend"
+if [ -z "${OTBR_PF:-}" ]; then
+    OTBR_PF=0
```
jwhui, line 69: In `script/_otbr`, `otbr_options` passes `-DOTBR_NFTABLES` to CMake:
https://github.com/openthread/ot-br-posix/blob/9a467e59b0123e6ce302d99e81a948851e36310b/script/_otbr#L124

However, `-DOTBR_PF` is not currently passed in `otbr_options`. When running `script/setup` on macOS with `OTBR_PF=1` in the environment, `script/_firewall` configures the `/etc/pf.conf` anchors, but `script/_otbr` builds otbr-agent with `OTBR_PF=OFF`.

Consider adding `-DOTBR_PF` to `otbr_options` in `script/_otbr`.

src/firewall/pf_firewall.cpp
clara-mostly-human: Done — `Token`, then `token`.

clara-mostly-human:
Also addressed the point about `script/_otbr`: it now passes `-DOTBR_PF=$([ "${OTBR_PF:-0}" = "1" ] && echo ON || echo OFF)` next to the `-DOTBR_NFTABLES` line, so `OTBR_PF=1 script/setup` builds the pf backend and `_firewall` resolves the same shell variable it does (the thread itself does not seem to be visible on my side to reply on).

clara-mostly-human:
Rebased onto main now that #3575–#3577 are in, and added a third leg to the macOS matrix that builds the native configuration with `-DOTBR_PF=ON`, so this backend is covered by CI rather than only by my border router.

.github/workflows/macOS.yml
```diff
@@ -54,11 +54,18 @@ jobs:
...
             border_agent: "OFF"
+            pf: "OFF"
           # The border router configuration used on macOS hosts: native
           # mDNSResponder with the border agent (and border routing) on.
           - name: "Native (mDNSResponder, Border Agent ON)"
             mdns: "mDNSResponder"
             border_agent: "ON"
+            pf: "OFF"
+          # The same, with the in-process pf firewall backend.
+          - name: "Native + pf firewall (mDNSResponder, Border Agent ON, OTBR_PF ON)"
```
jwhui, line 65: `googletest` is installed via `brew` in the bootstrap step and this PR adds thorough unit tests in `test_pf_firewall.cpp`, but the macOS workflow only runs `./script/test build` (which compiles default targets and doesn't run tests via `ctest`).

Should we add a step to run `ctest` for `otbr-gtest-firewall` on macOS (or running `./script/test check`)?

clara-mostly-human: Done — added a `Test` step to all three macOS legs: `cd build/otbr && CTEST_OUTPUT_ON_FAILURE=1 ctest -R 'FirewallManagerTest|PfFirewall'`. A targeted `ctest` rather than `./script/test check`, because `check` also runs `sudo ninja install` into `/usr` (read-only on macOS) and the Linux-only suites (netif/infra-if need netlink). The FirewallManager tests run in every configuration, the pf tests where `OTBR_PF=ON`; on a local macOS build all 36 pass.

script/_firewall
clara-mostly-human: Done — `script/_otbr` passes `-DOTBR_PF=$([ "${OTBR_PF:-0}" = "1" ] && echo ON || echo OFF)` next to `OTBR_NFTABLES`, so `OTBR_PF=1 ./script/setup` builds the backend it configured pf for.

jwhui (approved):
Thanks! 👍🏼

PR #3582, opened 2026-09-08
[rest] register the pending dataset via MGMT_PENDING_SET

agners:
`PUT /node/dataset/pending` commits the dataset with `otDatasetSetPendingTlvs`, a plain local write. That bypasses every check the TMF MGMT_PENDING_SET handler performs when the border router is the leader, and worse when it is not attached:

- A pending timestamp that does not advance is accepted locally and then silently ignored by the mesh — the client reports success for a change that never happens (the collision #3552 lets a client opt out of; here the mandatory arbitration applies again).
- A delay timer below the allowed minimum is applied verbatim: a 1 ms delay migrates the network with effectively zero warning time, bypassing the 30 s minimum (and the default-delay floor for network key changes) that `MGMT_PENDING_SET` enforces.
- On a disabled or detached border router the dataset is applied to that node alone when the delay expires; no other device ever sees it, and in the worst case the border router migrates by itself and leaves the mesh. This is the failure mode home-assistant-libs/python-otbr-api#278 adds client-side detection for; this change removes it at the source.

**What this does**

Send the dataset to the leader with `otDatasetSendMgmtPendingSet` and report the leader's verdict to the client. When the border router is the leader itself the request goes through the same TMF handler (leader ALOC), so validation applies uniformly.

Behavior changes on `PUT /node/dataset/pending`:

- A rejected registration answers **409** instead of a phantom success; no response from the leader answers **504**. An accepted dataset (200/201) has actually been registered with the leader.
- The node must be attached, otherwise **409**. The check runs with the other request checks, so a failing `If-None-Match` precondition still answers in the order RFC 9110 section 13.2.1 prescribes (409 before 412).
- A JSON request without a `pendingTimestamp` gets one generated from the current time: the previous behavior of inheriting the timestamp from the dataset being replaced now amounts to a guaranteed rejection.
- Raw TLV requests are passed through unchanged — expert callers get exactly the leader's verdict.

**Tests**

The new test drives the full path against the test node (a leader): a 1 ms delay is raised to the leader-enforced minimum instead of migrating the node, a stale pending timestamp answers 409, and an omitted pending timestamp is generated so a subsequent write advances. Verified locally: the full `tests/rest/test_rest.py` suite passes with this change (and without it, on current main, as a baseline).

`node_dataset_if_none_match_test` moves before `epskc_test`: pending writes now require an attached node, and after `epskc_test` disables Thread the node only comes back through auto-attach, so its attach state at that point is racy rather than guaranteed.

**Notes**

- **Seeding (review follow-up):** when no pending dataset exists, the request is now applied on top of the current **active** dataset instead of an `otDatasetCreateNewNetwork()` base. Digging through the history, the random seed was never a decision taken for the pending endpoint: it came from folding POST's "create like `dataset init new`" semantics into PUT for the *active* dataset in #1682 (see the `openapi.yaml:197` thread there), and the pending endpoint shared the code path. Now that the dataset reaches the whole mesh, a partial request (e.g. only a new channel) would otherwise rotate the network to a random key and PAN ID. The pending JSON shape (active dataset + pending timestamp + delay) was defined in #1682 precisely to express that a pending dataset is a change to the current network. Consequence: with no key change, the leader requires the request's `activeTimestamp` to advance (documented in the spec; python-otbr-api's `set_channel` already does this).
- A merged pending dataset must carry a delay timer in either request format (the TLV path could previously produce one without, which the leader accepts but never applies).
- The REST API version reported by `/.well-known/thread/br-rest` is bumped to 0.6.0 so clients can feature-detect the new behavior; it also covers the If-None-Match support from #3552, which did not bump it.
- The REST worker blocks on the MGMT response (the callback fires on response or CoAP timeout; a 30 s wait is a safety net only) while the main loop keeps running.

cc @LorbusChris — this is the server-side enforcement half of what #3552 (concurrency guard) and home-assistant-libs/python-otbr-api#278 (client-side role check) address; a client on a new OTBR now gets a 409 instead of a silent local-only apply even if it skips the role check.

### DIFF
--- a/src/rest/rest_web_server.cpp
+++ b/src/rest/rest_web_server.cpp
@@ -33,6 +33,8 @@
 
 #include <algorithm>
 #include <chrono>
+#include <future>
+#include <memory>
 #include <utility>
 
 #include <arpa/inet.h>
@@ -40,6 +42,7 @@
 #include <fcntl.h>
 #include <httplib.h>
 #include <string.h>
+#include <time.h>
 
 #include <openthread/commissioner.h>
 
@@ -834,6 +837,43 @@ static std::string TrimOws(const std::string &aValue)
     return begin == std::string::npos ? "" : aValue.substr(begin, end - begin + 1);
 }
 
+static void HandleMgmtPendingSetResponse(otError aResult, void *aContext)
+{
+    std::unique_ptr<std::promise<otError>> result(static_cast<std::promise<otError> *>(aContext));
+
+    result->set_value(aResult);
+}
+
+// Generates a Pending Timestamp from the current time. The leader rejects a
+// MGMT_PENDING_SET whose Pending Timestamp does not advance the one it holds,
+// and the clock may lag that value (not yet synchronized, or the dataset was
+// registered by a commissioner with a faster clock), so the result is made to
+// strictly advance the Pending Timestamp in @p aBaseTlvs when there is one.
+static otbrError GeneratePendingTimestamp(const otOperationalDatasetTlvs &aBaseTlvs, otTimestamp &aTimestamp)
+{
+    otbrError            error = OTBR_ERROR_NONE;
+    otOperationalDataset base;
+    timespec             now;
+
+    VerifyOrExit(clock_gettime(CLOCK_REALTIME, &now) == 0, error = OTBR_ERROR_REST);
+
+    aTimestamp.mSeconds = static_cast<uint64_t>(now.tv_sec) & 0xffffffffffff;
+    aTimestamp.mTicks   = static_cast<uint16_t>((static_cast<uint64_t>(now.tv_nsec) * 32768 / 1000000000) & 0x7fff);
+    aTimestamp.mAuthoritative = false;
+
+    if (otDatasetParseTlvs(&aBaseTlvs, &base) == OT_ERROR_NONE && base.mComponents.mIsPendingTimestampPresent &&
+        (aTimestamp.mSeconds < base.mPendingTimestamp.mSeconds ||
+         (aTimestamp.mSeconds == base.mPendingTimestamp.mSeconds &&
+          aTimestamp.mTicks <= base.mPendingTimestamp.mTicks)))
+    {
+        aTimestamp.mSeconds = (base.mPendingTimestamp.mSeconds + 1) & 0xffffffffffff;
+        aTimestamp.mTicks   = 0;
+    }
+
+exit:
+    return error;
+}
+
 void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest, Response &aResponse) const
 {
     otbrError       error = OTBR_ERROR_NONE;
@@ -841,13 +881,16 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
     std::string     body;
     StatusCode      errorCode = StatusCode::OK_200;
 
+    std::unique_ptr<std::promise<otError>> mgmtSetResult;
+    std::future<otError>                   mgmtSetFuture;
+    bool                                   mgmtSetSent = false;
+
     // RFC 9110 section 13.1.2: If-None-Match with "*" makes the write
-    // conditional on no dataset being in place. A pending dataset that is
-    // not newer than the one already held is silently ignored by the mesh
-    // while this handler accepts the write, so a client that does not intend
-    // to supersede an in-flight dataset can turn that collision into an
-    // error it sees. Other forms of the header (entity tags) have nothing to
-    // match against here and are rejected rather than mistaken for an
+    // conditional on no dataset being in place, so a client that does not
+    // intend to supersede an in-flight pending dataset can turn that
+    // collision into an error it sees, before anything is sent to the
+    // leader. Other forms of the header (entity tags) have nothing to match
+    // against here and are rejected rather than mistaken for an
     // unconditional write.
     const std::string ifNoneMatch  = TrimOws(aRequest.get_header_value(OT_REST_IF_NONE_MATCH_HEADER));
     const bool        onlyIfAbsent = ifNoneMatch == "*";
@@ -855,7 +898,8 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
     VerifyOrExit(ifNoneMatch.empty() || onlyIfAbsent, error = OTBR_ERROR_INVALID_ARGS);
 
     SuccessOrExit(
-        error = RunInMainLoop([this, aDatasetType, onlyIfAbsent, &errorCode, &aRequest]() {
+        error = RunInMainLoop([this, aDatasetType, onlyIfAbsent, &errorCode, &aRequest, &mgmtSetResult, &mgmtSetFuture,
+                               &mgmtSetSent]() {
             bool                     isTlv;
             otOperationalDataset     dataset = {};
             otOperationalDatasetTlvs datasetTlvs;
@@ -869,21 +913,38 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
             }
             else if (aDatasetType == DatasetType::kPending)
             {
+                // The Pending Dataset is registered with the leader (see below), which
+                // requires the device to be attached.
+                VerifyOrReturn(otThreadGetDeviceRole(GetInstance()) >= OT_DEVICE_ROLE_CHILD, OTBR_ERROR_INVALID_STATE);
                 errorOt = otDatasetGetPendingTlvs(GetInstance(), &datasetTlvs);
             }
 
             // RFC 9110 section 13.2.1: the precondition is evaluated after the
-            // request checks (the role check above still answers 409 first, the
+            // request checks (the role checks above still answer 409 first, the
             // header cannot downgrade that to a 412) and before the content is
-            // processed; evaluated in this main-loop task, it is atomic with
-            // the write below.
+            // processed. Evaluated in this main-loop task, it is atomic with
+            // the write below for the Active Dataset; for the Pending Dataset
+            // it is atomic with the MGMT_PENDING_SET send, and the leader's
+            // Pending Timestamp ordering arbitrates concurrent registrations.
             VerifyOrReturn(!onlyIfAbsent || errorOt != OT_ERROR_NONE, OTBR_ERROR_DUPLICATED);
 
-            // Create a new operational dataset if it doesn't exist.
             if (errorOt == OT_ERROR_NOT_FOUND)
             {
-                VerifyOrReturn(otDatasetCreateNewNetwork(GetInstance(), &dataset) == OT_ERROR_NONE, OTBR_ERROR_REST);
-                otDatasetConvertToTlvs(&dataset, &datasetTlvs);
+                if (aDatasetType == DatasetType::kPending)
+                {
+                    // A Pending Dataset describes a change to the network this node is
+                    // attached to, so a new one starts from the current Active Dataset:
+                    // fields the request leaves out keep their current value.
+                    VerifyOrReturn(otDatasetGetActiveTlvs(GetInstance(), &datasetTlvs) == OT_ERROR_NONE,
+                                   OTBR_ERROR_REST);
+                }
+                else
+                {
+                    // Create a new operational dataset if it doesn't exist.
+                    VerifyOrReturn(otDatasetCreateNewNetwork(GetInstance(), &dataset) == OT_ERROR_NONE,
+                                   OTBR_ERROR_REST);
+                    otDatasetConvertToTlvs(&dataset, &datasetTlvs);
+                }
                 errorCode = StatusCode::Created_201;
             }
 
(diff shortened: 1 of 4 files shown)


src/rest/rest_web_server.cpp
```diff
@@ -914,6 +934,21 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
+
+                    if (!dataset.mComponents.mIsPendingTimestampPresent)
+                    {
+                        // The leader rejects a MGMT_PENDING_SET whose Pending Timestamp does
+                        // not advance, so an omitted timestamp cannot be inherited from an
+                        // existing Pending Dataset.
+                        timespec now;
+
+                        clock_gettime(CLOCK_REALTIME, &now);
+                        dataset.mPendingTimestamp.mSeconds = static_cast<uint64_t>(now.tv_sec) & 0xffffffffffff;
```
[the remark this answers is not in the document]
agners: Done in the amended commit: the return value is checked and the request fails with 500 rather than sending a timestamp built from uninitialized memory.

src/rest/rest_web_server.cpp
```diff
@@ -924,11 +959,52 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
+                // Register the Pending Dataset with the leader via MGMT_PENDING_SET
+                // rather than writing it locally, so that leader-side validation
+                // (Pending Timestamp ordering, minimum delay timer) applies even
+                // when this device is the leader.
+                otOperationalDataset emptyDataset = {};
+
+                errorOt =
+                    otDatasetSendMgmtPendingSet(GetInstance(), &emptyDataset, datasetTlvs.mTlvs, datasetTlvs.mLength,
+                                                HandleMgmtPendingSetResponse, mgmtSetResult.get());
+                VerifyOrReturn(errorOt != OT_ERROR_ALREADY && errorOt != OT_ERROR_BUSY, OTBR_ERROR_INVALID_STATE);
```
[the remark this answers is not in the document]
agners: Done in the amended commit: OT_ERROR_INVALID_STATE from otDatasetSendMgmtPendingSet now maps to OTBR_ERROR_INVALID_STATE and answers 409, covering a detach that races the role check.

LorbusChris (approved):
Thanks, this seems sensible to me. LGTM!

src/rest/rest_web_server.cpp
```diff
@@ -924,11 +959,54 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
             }
             return OTBR_ERROR_NONE;
         }));
 
+    if (mgmtSetSent)
+    {
+        otError mgmtError = OT_ERROR_RESPONSE_TIMEOUT;
+
+        // The callback is invoked on response reception or timeout; the wait
+        // duration here is only a safety net beyond the TMF exchange lifetime.
```
jwhui: The comment notes that 30 seconds is *"beyond the TMF exchange lifetime"*, but OpenThread's default CoAP confirmable retransmission parameters (`kDefaultAckTimeout = 2000ms`, `kDefaultMaxRetransmit = 4`, backoff factor up to 1.5) yield a `MAX_TRANSMIT_WAIT` of **62 to 93 seconds** (RFC 7252 Section 4.8.2).

src/rest/rest_web_server.cpp
```diff
@@ -914,6 +934,21 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
+
+                    if (!dataset.mComponents.mIsPendingTimestampPresent)
+                    {
+                        // The leader rejects a MGMT_PENDING_SET whose Pending Timestamp does
+                        // not advance, so an omitted timestamp cannot be inherited from an
+                        // existing Pending Dataset.
+                        timespec now;
+
+                        VerifyOrReturn(clock_gettime(CLOCK_REALTIME, &now) == 0, OTBR_ERROR_REST);
+                        dataset.mPendingTimestamp.mSeconds = static_cast<uint64_t>(now.tv_sec) & 0xffffffffffff;
```
jwhui: If a Pending Dataset is already present (`errorOt == OT_ERROR_NONE`), its `mPendingTimestamp` may be greater than or equal to `CLOCK_REALTIME`. In that case, the leader will reject this `MGMT_PENDING_SET` even though the caller omitted `pendingTimestamp`.

When `errorOt == OT_ERROR_NONE`, consider comparing the generated `now` timestamp against the existing pending dataset's `mPendingTimestamp` and ensuring the new timestamp strictly advances (e.g., `std::max(nowSeconds, existingPendingTimestamp.mSeconds + 1)` when `now <= existingPendingTimestamp`).

src/rest/rest_web_server.cpp
```diff
@@ -924,11 +959,54 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
...
+        {
+        case OT_ERROR_NONE:
+            break;
+        case OT_ERROR_REJECTED:
+            ErrorHandler(aResponse, StatusCode::Conflict_409, "rejected by leader");
+            ExitNow();
+        case OT_ERROR_RESPONSE_TIMEOUT:
+            ErrorHandler(aResponse, StatusCode::GatewayTimeout_504, "no response from leader");
+            ExitNow();
+        default:
```
jwhui, line 1061: If Thread is disabled or the node detaches while `MGMT_PENDING_SET.req` is in flight, OpenThread aborts the CoAP request and invokes the response callback with `OT_ERROR_ABORT`. Currently, that falls through to `default:` and returns `500 Internal Server Error`.

Consider handling `case OT_ERROR_ABORT:` as `StatusCode::Conflict_409` (consistent with the pre-flight role check).

src/rest/rest_web_server.cpp
```diff
@@ -834,13 +837,24 @@ static std::string TrimOws(const std::string &aValue)
...
+}
+
 void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest, Response &aResponse) const
 {
     otbrError       error = OTBR_ERROR_NONE;
     struct NodeInfo node;
     std::string     body;
     StatusCode      errorCode = StatusCode::OK_200;
 
+    std::unique_ptr<std::promise<otError>> mgmtSetResult(new std::promise<otError>());
```
jwhui: `mgmtSetResult` is heap-allocated on every `SetDataset` call, including `PUT /node/dataset/active` where it is never used. Consider initializing `mgmtSetResult` only when `aDatasetType == DatasetType::kPending` (or inside the `kPending` branch before calling `otDatasetSendMgmtPendingSet`).

src/rest/rest_web_server.cpp
```diff
@@ -869,14 +884,19 @@ void RestWebServer::SetDataset(DatasetType aDatasetType, const Request &aRequest
             }
             else if (aDatasetType == DatasetType::kPending)
             {
+                // The Pending Dataset is registered with the leader (see below), which
+                // requires the device to be attached.
+                VerifyOrReturn(otThreadGetDeviceRole(GetInstance()) >= OT_DEVICE_ROLE_CHILD, OTBR_ERROR_INVALID_STATE);
                 errorOt = otDatasetGetPendingTlvs(GetInstance(), &datasetTlvs);
```
jwhui, line 919: When `aDatasetType == DatasetType::kPending` and no pending dataset currently exists (`otDatasetGetPendingTlvs` returns `OT_ERROR_NOT_FOUND`), `SetDataset` falls back to `otDatasetCreateNewNetwork()`, which generates a random Network Key, PAN ID, Extended PAN ID, and Mesh-Local Prefix.

Now that `PUT /node/dataset/pending` requires the node to be attached and registers the dataset with the leader via `MGMT_PENDING_SET`, seeding a missing Pending Dataset from `otDatasetCreateNewNetwork()` means any partial `PUT /node/dataset/pending` request (e.g., changing only `channel`) will broadcast a pending dataset that rotates the entire Thread network to a random Network Key and PAN ID when the delay timer expires.

Would it make sense (maybe in a follow-up PR) to seed `OT_ERROR_NOT_FOUND` for `kPending` from the node's current Active Dataset (`otDatasetGetActiveTlvs()`) instead of `otDatasetCreateNewNetwork()`?

tests/rest/test_rest.py
```diff
@@ -870,6 +870,46 @@ def put(body, headers, target=None):
...
+        req = urllib.request.Request(url,
+                                     data=json.dumps(body_dict).encode(),
+                                     method='PUT',
+                                     headers={'Content-Type': 'application/json'})
+        return urllib.request.urlopen(req)
+
+    # The pending dataset is registered with the leader via MGMT_PENDING_SET,
+    # so the leader raises the delay timer to the allowed minimum rather than
+    # applying a 1 ms delay verbatim (which, written locally, would migrate
+    # this node alone almost immediately).
```
jwhui: Note on what this test exercises: because there is no pending dataset initially, `SetDataset` seeds the base dataset via `otDatasetCreateNewNetwork()` with a random Network Key before applying `{"activeDataset": {"activeTimestamp": {"seconds": 20}}, "delay": 1}`. Consequently, the leader sees a Network Key change (`mAffectsNetworkKey == true`) and raises `delay` to `DelayTimerTlv::kDefaultDelay` (300,000 ms / 5 min) rather than `DelayTimerTlv::kMinDelay` (30,000 ms).

src/rest/rest_web_server.cpp
agners: You are right, I had the parameters wrong — with kDefaultAckTimeout=2000 ms, kDefaultMaxRetransmit=4 and ACK_RANDOM_FACTOR 1.5, MAX_TRANSMIT_WAIT is 62–93 s. Rewrote the comment to say so. I kept the 30 s wait deliberately: blocking an HTTP worker for up to 93 s seemed worse than answering earlier with 504 (outcome unknown, which the OpenAPI text spells out along with how to find out what happened). A late callback is absorbed by the heap-owned promise, and until the exchange finalizes further registrations answer 409 via OT_ERROR_ALREADY, which is now also stated in the comment. Happy to raise the wait past MAX_TRANSMIT_WAIT if you would rather have the 504 mean "the exchange is over".

src/rest/rest_web_server.cpp
agners: Done. Factored the generation into `GeneratePendingTimestamp()`: wall clock as the base, and when a pending dataset exists and the generated value does not strictly exceed its pending timestamp (seconds and ticks compared), it becomes `existing.seconds + 1` with ticks 0. This is also what the in-stack `DatasetUpdater` does in spirit (it advances the current pending timestamp by random ticks rather than trusting a clock). Test added: an explicit far-future pending timestamp followed by a request without one still succeeds, and the resulting timestamp is verified to have advanced past it.

src/rest/rest_web_server.cpp
agners: Done: `OT_ERROR_ABORT` now answers 409 ("no longer attached"), consistent with the pre-flight role check. With that the callback error space is fully mapped (NONE, REJECTED, ABORT, RESPONSE_TIMEOUT; PARSE falls to 500, which is right for a malformed response).

src/rest/rest_web_server.cpp
agners: Done: the promise and future are now created inside the `kPending` branch right before `otDatasetSendMgmtPendingSet`; the active-dataset path no longer allocates anything.

src/rest/rest_web_server.cpp
agners: Yes, and I did it in this PR rather than a follow-up: the version bump and the changed contract are already here, and with the dataset now reaching the whole mesh the random seed became genuinely dangerous.

I went back through the history to see how the random seed came about. It was never a decision taken for the pending endpoint: #1682 originally had POST = "create like `dataset init new`" and PUT = "update the existing dataset, error if none". During review the two were folded into PUT for idempotence (the `openapi.yaml:197` thread there), with the *active* dataset commissioning case in mind, and the pending endpoint simply shared the code path. In the same PR the pending JSON shape (active dataset + pending timestamp + delay) was introduced precisely to express that a pending dataset is a change to the current network, so seeding from the active dataset restores the original intent rather than adding new behavior.

Consequence worth noting: with no key change the leader requires the request's active timestamp to advance, so a partial request needs a newer `activeTimestamp` (documented in the spec; python-otbr-api's `set_channel` already does this). Auto-advancing it when omitted, like `DatasetUpdater` does, would be a reasonable follow-up. I also added a check that the merged dataset carries a Delay Timer in either request format: the TLV path could previously produce one without, which the leader accepts but which never applies.

tests/rest/test_rest.py
agners: Good catch, and with the seeding change above this now resolves the other way: the base is the active dataset, the network key is unchanged, and the leader raises the 1 ms delay to `kMinDelay`. The test now asserts `1000 < delay <= 30000` (30 s, not the 5 min key-change default) and additionally checks that the pending dataset's network key and PAN ID equal the active dataset's, which pins the seeding behavior.

src/rest/rest_web_server.cpp
```diff
@@ -834,28 +837,69 @@ static std::string TrimOws(const std::string &aValue)
...
+    aTimestamp.mSeconds = static_cast<uint64_t>(now.tv_sec) & 0xffffffffffff;
+    aTimestamp.mTicks   = static_cast<uint16_t>((static_cast<uint64_t>(now.tv_nsec) * 32768 / 1000000000) & 0x7fff);
+    aTimestamp.mAuthoritative = false;
+
+    if (otDatasetParseTlvs(&aBaseTlvs, &base) == OT_ERROR_NONE && base.mComponents.mIsPendingTimestampPresent &&
+        (aTimestamp.mSeconds < base.mPendingTimestamp.mSeconds ||
+         (aTimestamp.mSeconds == base.mPendingTimestamp.mSeconds &&
+          aTimestamp.mTicks <= base.mPendingTimestamp.mTicks)))
+    {
+        aTimestamp.mSeconds = base.mPendingTimestamp.mSeconds + 1;
```
jwhui: nit: Line 860 masks `now.tv_sec` with `& 0xffffffffffff`. Should we mask here as well?

agners: Done, masked for consistency. Since the base value comes out of a 48-bit TLV field, the increment can only exceed 48 bits when the existing timestamp is the all-ones maximum, and then no advancing timestamp exists at all: masked or not the value wraps to 0 and the leader rejects the registration with a 409, which is the honest outcome for that (unreachable in practice) case.

jwhui (approved):
Thanks! 👍🏼

PR #3588, opened 2026-09-12
[mdns] treat mDNSResponder timeout as terminal error

### LINKED ISSUE #3242: [mdns] handle mDNSResponder timeout as explicit terminal failure path
## Problem

In `src/mdns/mdns_mdnssd.cpp`, there is an explicit TODO noting timeout handling should be improved during mDNS registration:

- `TODO: Abort on Timeout error, as it indicates an unresponsive mDNSResponder.`

Today, timeout can be treated as retryable in generic retry paths. This can lead to prolonged retry loops when mDNSResponder is unhealthy, which reduces reliability and slows failure detection.

## Proposed change

Handle `kDNSServiceErr_Timeout` explicitly as a terminal/unhealthy condition for the current registration/resolution attempt, instead of treating it like a generic retryable error.

Initial scope:

1. Update timeout classification/handling in `mdns_mdnssd` paths.
2. Ensure logs clearly indicate timeout as a distinct failure mode.
3. Add/adjust tests for timeout behavior.

## Expected outcome

- Faster and clearer failure signaling when mDNSResponder is unresponsive.
- Less churn from retries that are unlikely to succeed.
- More predictable recovery behavior for OTBR users and downstream integrations.

## Context

I am happy to implement this as a small, focused PR with tests.

jwhui:
In src/mdns/mdns_mdnssd.cpp, kDNSServiceErr_Timeout was previously classified as a retryable error in IsRetryableError(). This caused two severe issues:

1. During service, host, and key registrations, DNSServiceRegister blocks synchronously for up to 60 seconds when mDNSResponder is unresponsive before returning kDNSServiceErr_Timeout. Because Timeout was marked retryable, HandleRegisterResult scheduled a retry in 5 seconds, causing otbr-agent to hang in an indefinite 60-second blocking loop and starving other operations.

2. During service and host resolution, DNSServiceResolve passes kDNSServiceFlagsTimeout so mDNSResponder returns a Timeout error if the query cannot be answered on the network. Treating Timeout as retryable led to endless resolution retry loops instead of reporting failure back to OpenThread / discovery proxy.

This commit fixes the issue by:
- Removing kDNSServiceErr_Timeout from IsRetryableError() so that timeouts during registration or resolution are treated as terminal failures rather than retryable.
- Removing the outdated TODO comment in DnssdServiceRegistration.
- Adding unit tests in tests/gtest/test_mdns_mdnssd.cpp verifying the expected behavior of IsRetryableError().

Resolves #3242

### DIFF
--- a/src/mdns/mdns_mdnssd.cpp
+++ b/src/mdns/mdns_mdnssd.cpp
@@ -247,7 +247,6 @@ bool IsRetryableError(DNSServiceErrorType aError)
     case kDNSServiceErr_NoRouter:
     case kDNSServiceErr_NATTraversal:
     case kDNSServiceErr_DoubleNAT:
-    case kDNSServiceErr_Timeout:
     case kDNSServiceErr_DefunctConnection:
 #if OTBR_MDNSSD_HAVE_STALE_DATA
     case kDNSServiceErr_StaleData:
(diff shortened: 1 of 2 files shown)


abtink (approved):
👍

Issue #3590, opened 2026-09-12
macOS: userspace Backbone Router multicast forwarding for platforms without kernel multicast routing (MRT6)


clara-mostly-human:
With the macOS host support from #3469 merged (#3575, #3576, #3577, #3578, openthread#13587), one gap remains for a *full-functionality* border router on macOS: the Backbone Router's multicast forwarding. The OpenThread posix platform does it through the kernel's multicast routing (MRT6), which only Linux offers — `src/posix/platform/multicast_routing.cpp` is `#error`ed out elsewhere, and on the `ot-br-posix` side the NCP-mode `MulticastRoutingManager` compiles to nothing, so `OTBR_BACKBONE_ROUTER=ON` does not even link on macOS.

This issue tracks a userspace implementation for platforms without MRT6. It is working and in production on my macOS border router; the PRs are linked below.

## Approach

`OTBR_USERSPACE_MCAST_FWD` (default OFF; macOS only for now). When on, the posix MRT6 knob is forced off (`OPENTHREAD_POSIX_CONFIG_BACKBONE_ROUTER_MULTICAST_ROUTING_ENABLE=0`) while the OpenThread core keeps its multicast listener table, and a `McastForwarder` in `src/host/posix/` takes the MRT6 role in RCP mode, applying the same rules as the Linux path:

- **Thread → backbone**: a BPF tap on the Thread tunnel interface (inbound only, i.e. what the stack hands the host) picks up multicast of admin-local scope or larger and writes it to the backbone interface as an Ethernet frame — forwarded regardless of backbone listeners, as Linux does.
- **Backbone → Thread**: a BPF tap on the backbone interface; a packet is injected (written to the tunnel) only for a group with a registered Thread listener (MLR), as the Linux inbound forwarding cache does. Injected packets are MPL-flooded, so this direction has the tighter limit.
- **Both directions**: duplicate suppression (fingerprint window; it also catches the forwarder's own emissions coming back through the backbone tap), per-group and overall token-bucket rate limits, hop limit decremented.
- **Listener tracking**: Thread side from the core's MLR events; backbone side by snooping MLD reports (MLDv1/v2 — note macOS reports the last leave as an empty `BLOCK_OLD_SOURCES` record).
- **Observability**: per-direction counters, logged and exposed as `GET /node/mcast-forwarder` on the REST server.
- The OpenThread core is untouched; the Linux MRT6 path is unchanged; no kernel extension.

[...]

PR #3591, opened 2026-09-12
[host] stub MulticastRoutingManager on non-Linux so a Backbone Router build links

clara-mostly-human:
The NCP-mode `MulticastRoutingManager` (`src/host/posix/multicast_routing_manager.cpp`) is implemented on the Linux MRT6 socket and compiles to nothing elsewhere, so `OTBR_BACKBONE_ROUTER=ON` fails to link on macOS with three undefined symbols. This gives it a non-Linux body: the constructor logs a warning that multicast routing is not available on the platform, the other methods are no-ops. No change on Linux.

Enabling piece for #3590 (userspace Backbone Router multicast forwarding on macOS); the forwarder itself is in a follow-up PR.

### DIFF
--- a/src/host/posix/multicast_routing_manager.cpp
+++ b/src/host/posix/multicast_routing_manager.cpp
@@ -661,4 +661,62 @@ bool MulticastRoutingManager::MatchesMeshLocalPrefix(const Ip6Address        &aA
 } // namespace otbr
 
 #endif // OTBR_ENABLE_BACKBONE_ROUTER
+#else  // __linux__
+#if OTBR_ENABLE_BACKBONE_ROUTER
+
+// Kernel multicast routing (MRT6) is Linux-only. Elsewhere the class exists so
+// that a Backbone Router build links, but it forwards nothing: the Backbone
+// Router will not relay multicast between the Thread network and the backbone
+// link on this platform.
+
+namespace otbr {
+
+MulticastRoutingManager::MulticastRoutingManager(const Netif                   &aNetif,
+                                                 const InfraIf                 &aInfraIf,
+                                                 const Host::NetworkProperties &aNetworkProperties)
+    : mNetif(aNetif)
+    , mInfraIf(aInfraIf)
+    , mNetworkProperties(aNetworkProperties)
+    , mLastExpireTime(otbr::Timepoint::min())
+    , mMulticastRouterSock(-1)
+    , mState(kStateDisabled)
+    , mRetryIntervalUs(kMinRetryIntervalUs)
+    , mNextRetryTime(otbr::Timepoint::min())
+{
+    OTBR_UNUSED_VARIABLE(mMulticastForwardingCacheTable);
+    OTBR_UNUSED_VARIABLE(mMulticastListeners);
+
+    otbrLogWarning(
+        "Multicast routing is not available on this platform; the Backbone Router will not forward multicast");
+}
+
+void MulticastRoutingManager::HandleStateChange(otBackboneRouterState aState)
+{
+    OTBR_UNUSED_VARIABLE(aState);
+}
+
+void MulticastRoutingManager::HandleBackboneMulticastListenerEvent(otBackboneRouterMulticastListenerEvent aEvent,
+                                                                   const Ip6Address                      &aAddress)
+{
+    OTBR_UNUSED_VARIABLE(aEvent);
+    OTBR_UNUSED_VARIABLE(aAddress);
+}
+
+void MulticastRoutingManager::FinalizeMulticastRouterSock(void)
+{
+}
+
+void MulticastRoutingManager::Update(MainloopContext &aContext)
+{
+    OTBR_UNUSED_VARIABLE(aContext);
+}
+
+void MulticastRoutingManager::Process(const MainloopContext &aContext)
+{
+    OTBR_UNUSED_VARIABLE(aContext);
+}
+
+} // namespace otbr
+
+#endif // OTBR_ENABLE_BACKBONE_ROUTER
 #endif // __linux__


src/host/posix/multicast_routing_manager.cpp
```diff
@@ -660,5 +660,63 @@ bool MulticastRoutingManager::MatchesMeshLocalPrefix(const Ip6Address        &aA
...
+    , mInfraIf(aInfraIf)
+    , mNetworkProperties(aNetworkProperties)
+    , mLastExpireTime(otbr::Timepoint::min())
+    , mMulticastRouterSock(-1)
+    , mState(kStateDisabled)
+    , mRetryIntervalUs(kMinRetryIntervalUs)
+    , mNextRetryTime(otbr::Timepoint::min())
+{
+    OTBR_UNUSED_VARIABLE(mMulticastForwardingCacheTable);
+    OTBR_UNUSED_VARIABLE(mMulticastListeners);
```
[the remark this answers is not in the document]
clara-mostly-human: Checked rather than assumed: this exact code builds in the macOS CI legs of this repository, which compile with `-Wall -Wextra -Werror` under clang (see the green macOS workflow on this PR's branch), and it has been running on a macOS border router for days. Clang's `-Wunused-private-field` only fires when it sees every member function of the class defined in the translation unit; here most of `MulticastRoutingManager`'s methods are declared in the header and not defined on non-Linux, so the fields cannot be proven unused and no warning is emitted. The two array members are already marked with `OTBR_UNUSED_VARIABLE`; I'd rather not add nine more no-op statements for a warning that does not occur.

src/host/posix/multicast_routing_manager.cpp
```diff
@@ -660,5 +660,63 @@ bool MulticastRoutingManager::MatchesMeshLocalPrefix(const Ip6Address        &aA
...
+#if OTBR_ENABLE_BACKBONE_ROUTER
+
+// Kernel multicast routing (MRT6) is Linux-only. Elsewhere the class exists so
+// that a Backbone Router build links, but it forwards nothing: the Backbone
+// Router will not relay multicast between the Thread network and the backbone
+// link on this platform.
+
+namespace otbr {
+
+MulticastRoutingManager::MulticastRoutingManager(const Netif                   &aNetif,
```
jwhui, line 674: Could we guard the MRT6-specific private methods and member variables in `src/host/posix/multicast_routing_manager.hpp` with `#ifdef __linux__` (and make `Deinit()` / `IsEnabled()` no-ops returning `false` on `#else`)?

That avoids carrying `mMulticastForwardingCacheTable[750]` (~42 KB) and ~20 declared-but-undefined private methods on non-Linux, eliminates `FinalizeMulticastRouterSock()`, and simplifies this constructor to:
```cpp
MulticastRoutingManager::MulticastRoutingManager(const Netif                   &aNetif,
                                                 const InfraIf                 &aInfraIf,
                                                 const Host::NetworkProperties &aNetworkProperties)
{
    OTBR_UNUSED_VARIABLE(aNetif);
    OTBR_UNUSED_VARIABLE(aInfraIf);
    OTBR_UNUSED_VARIABLE(aNetworkProperties);
}
```

src/host/posix/multicast_routing_manager.cpp
```diff
@@ -660,5 +660,63 @@ bool MulticastRoutingManager::MatchesMeshLocalPrefix(const Ip6Address        &aA
 
 } // namespace otbr
 
+#endif // OTBR_ENABLE_BACKBONE_ROUTER
```
jwhui, line 663: nit: if we place `#if OTBR_ENABLE_BACKBONE_ROUTER` and `namespace otbr` outside `#ifdef __linux__` at the top of the file, we don't need to close and immediately re-open both across `#else // __linux__`.

src/host/posix/multicast_routing_manager.cpp
```diff
@@ -660,5 +660,63 @@ bool MulticastRoutingManager::MatchesMeshLocalPrefix(const Ip6Address        &aA
...
+    , mNextRetryTime(otbr::Timepoint::min())
+{
+    OTBR_UNUSED_VARIABLE(mMulticastForwardingCacheTable);
+    OTBR_UNUSED_VARIABLE(mMulticastListeners);
+
+    otbrLogWarning(
+        "Multicast routing is not available on this platform; the Backbone Router will not forward multicast");
+}
+
+void MulticastRoutingManager::HandleStateChange(otBackboneRouterState aState)
```
jwhui, line 693: `Application::CreateNcpMode()` constructs `MulticastRoutingManager` unconditionally at startup even when Backbone Router is disabled. Consider logging the warning here when `aState == OT_BACKBONE_ROUTER_STATE_PRIMARY` (where Linux calls `Enable()`) instead of in the constructor:
```cpp
void MulticastRoutingManager::HandleStateChange(otBackboneRouterState aState)
{
    if (aState == OT_BACKBONE_ROUTER_STATE_PRIMARY)
    {
        otbrLogWarning(
            "Multicast routing is not available on this platform; the Backbone Router will not forward multicast");
    }
}
```

src/host/posix/multicast_routing_manager.cpp
```diff
@@ -660,5 +660,63 @@ bool MulticastRoutingManager::MatchesMeshLocalPrefix(const Ip6Address        &aA
...
+    , mLastExpireTime(otbr::Timepoint::min())
+    , mMulticastRouterSock(-1)
+    , mState(kStateDisabled)
+    , mRetryIntervalUs(kMinRetryIntervalUs)
+    , mNextRetryTime(otbr::Timepoint::min())
+{
+    OTBR_UNUSED_VARIABLE(mMulticastForwardingCacheTable);
+    OTBR_UNUSED_VARIABLE(mMulticastListeners);
+
+    otbrLogWarning(
```
jwhui, line 689: nit: `multicast_routing_manager.cpp` doesn't define `OTBR_LOG_TAG` before `#include "multicast_routing_manager.hpp"`, so `common/code_utils.hpp` defaults it to `"UTILS"`. Consider adding `#define OTBR_LOG_TAG "MCAST"` at the top of this file.

PR #3594, opened 2026-09-12
[netif] verify unicast netlink ACKs and gate cache updates

### LINKED ISSUE #3244: [netif] keep unicast cache aligned with netlink ACK results

jwhui:
Previously, Netif::UpdateIp6UnicastAddresses updated its internal mIp6UnicastAddresses cache optimistically without confirming whether the Linux kernel accepted or rejected netlink requests (RTM_NEWADDR / RTM_DELADDR). If the kernel rejected an address addition or deletion, OTBR's local cache diverged from the kernel interface state. Because subsequent updates diff against the local cache, failed operations were never retried, leaving interface addresses desynchronized.

This commit resolves the issue:
1. Categorizes address reconciliation into removals, in-place replaces, and additions.
2. Gates mutations of mIp6UnicastAddresses on confirmed netlink ACK outcomes.
3. Uses NLM_F_REPLACE for address metadata modifications (e.g., preferred lifetime changes) to update attributes in-place without a racy delete-then-add cycle.
4. Binds mNetlinkFd with sa.nl_groups = 0 so the socket receives only unicast ACKs/replies and avoids buffer exhaustion (ENOBUFS) from unhandled system-wide multicast link/address events.
5. Removes NLM_F_ACK from SetAddrGenModeToNone to avoid generating unhandled netlink ACKs on initialization.
6. Implements netlink ACK sequence verification bounded by a tight 50 ms timeout, handling EAGAIN/EWOULDBLOCK on non-blocking sends and correctly parsing ACKs when NETLINK_CAP_ACK is enabled.
7. Treats idempotent outcomes (EEXIST for add, ENOENT/EADDRNOTAVAIL for remove) as success.
8. Fixes Ip6AddressInfo constructor and operator== to avoid comparing uninitialized bitfield padding.
9. Adds unit tests in test_netif.cpp verifying address reconciliation caching behavior under success, failure, and batched changes.

Resolves #3244

### DIFF
--- a/src/host/posix/netif.cpp
+++ b/src/host/posix/netif.cpp
@@ -57,6 +57,15 @@
 
 namespace otbr {
 
+namespace {
+
+bool HasSameIp6Address(const Ip6AddressInfo &aFirst, const Ip6AddressInfo &aSecond)
+{
+    return Ip6Address(aFirst.mAddress) == Ip6Address(aSecond.mAddress);
+}
+
+} // namespace
+
 otbrError Netif::Dependencies::Ip6Send(const uint8_t *aData, uint16_t aLength)
 {
     OTBR_UNUSED_VARIABLE(aData);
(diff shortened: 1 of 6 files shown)


src/host/posix/netif.cpp
```diff
@@ -600,6 +771,44 @@ void Netif::Update(MainloopContext &aContext)
...
+
+            for (const auto &pending : mPendingNetlinkRequests)
+            {
+                if (pending.mExpireTime < minExp)
+                {
+                    minExp = pending.mExpireTime;
+                }
+            }
+
+            auto delayUs = std::chrono::duration_cast<std::chrono::microseconds>(minExp - now).count();
```
abtink: minor nit: can we move this var declaration to the top of scope?

src/host/posix/netif.cpp
```diff
@@ -600,6 +771,44 @@ void Netif::Update(MainloopContext &aContext)
...
+            if (delayUs < 0)
+            {
+                delayUs = 0;
+            }
+
+            int64_t currentTimeoutUs =
```
abtink: Same for this maybe?

abtink (approved):
LGTM, thanks! The logic for handling and updating addresses looks solid to me (which is definitely not a simple problem to solve!).

PR #3604, opened 2026-09-19
[firewall] leave IPv4 out of the nftables ingress filter

### LINKED ISSUE #3601: nftables: ingress filter drops IPv4 replies needed for NAT64
**Describe the bug**

With the native nftables backend enabled, IPv4 replies forwarded towards `wpan0` miss the [IPv6-only allow-set](https://github.com/openthread/ot-br-posix/blob/fd872ab9d4afdfaf034b3b1154fed08b8b6d1e43/src/firewall/nftables_impl.cpp#L1047-L1075) and reach the [unicast drop rule](https://github.com/openthread/ot-br-posix/blob/fd872ab9d4afdfaf034b3b1154fed08b8b6d1e43/src/firewall/firewall_manager.cpp#L129-L140). This blocks the return path needed by NAT64.

**To Reproduce**

- Commit: `fd872ab9d4afdfaf034b3b1154fed08b8b6d1e43` (`v2026.09.0`).
- Hardware: none; the firewall components were exercised in isolated Linux network namespaces.
- Build configuration: native nftables backend (`OTBR_NFTABLES=ON`).
- Topology: a Thread-facing test interface and an IPv4 upstream interface.

Enable ingress filtering and NAT44 masquerading. Forward a UDP packet from the Thread-facing side upstream, then send a reply over the established flow.

**Expected behavior**

The IPv4 reply reaches the Thread interface for NAT64 translation.

**Console/log output**

No agent log attached. The isolated test forwarded and masqueraded the outgoing packet, but dropped its reply. Adding an IPv4 return rule restored delivery. This was not a physical-radio NAT64 test.

**Additional context**

An early `meta nfproto != ipv6 return` in `forward_ingress` would keep IPv4 out of this filter without changing the shared `inet` table.

LorbusChris:
Fixes #3601.

The ingress filter is an IPv6 filter -- the ip6tables rules it replaces never saw
IPv4 -- but the nftables backend keeps its chains in one `inet` table, so
`forward_ingress` sees IPv4 as well. An IPv4 packet forwarded to the Thread interface
matches neither address set and falls through to `meta pkttype unicast drop`. That is
the return path NAT64 depends on.

`nat_forward` accepting the reply does not help, which is what the comment there
assumed: an `accept` ends the base chain it is in, and every other base chain on the
hook still runs.

The fix is the one suggested in the issue: `meta nfproto != ipv6 return` as the first
rule of `forward_ingress` (new `INftables::AddRuleNfprotoNeqIp6Return`). The per-rule
`nfproto ipv6` guards on the set lookups stay; they are what makes those rule builders
safe to use in any chain.

**Testing.** Unit test: the rule is installed, and installed first. Beyond the mocks,
checked against a kernel in network namespaces (Thread side / router / upstream, the
router carrying the rules of v2026.09.0 with NAT44 on):

- before: a ping from the Thread side goes out masqueraded, its reply is lost, and the
  unicast drop in `forward_ingress` counts exactly those packets;
- with the rule: the reply is delivered, and IPv6 to an address outside the allow set
  is still dropped.

The real backend renders the chain as
`meta nfproto != ipv6 return; oifname != "wpan0" return; ...` on a 7.1 kernel.

Written with AI assistance (`Assisted-by` trailer); I've reviewed and understand all of
it and will answer questions myself.


src/firewall/nftables.hpp
```diff
@@ -166,6 +166,15 @@ class INftables
...
+     * `meta nfproto != ipv6 return` — the table is of the inet family, so its
+     * chains see IPv4 as well. A chain that filters IPv6 alone opens with
+     * this rule, or its closing verdicts apply to IPv4 traffic too.
+     */
+    virtual otbrError AddRuleNfprotoNeqIp6Return(const std::string &aTable,
+                                                 const std::string &aChain,
+                                                 uint64_t          *aHandleOut) = 0;
+
     /**
      * `meta oifname != <ifname> return` — used as the first rule of an
```
jwhui, line 179: nit: now that `AddRuleNfprotoNeqIp6Return` is placed first in `forward_ingress`, `AddRuleOifnameNeqReturn` is no longer the first rule of the chain. Consider updating this comment from `"used as the first rule of an inet/forward base chain"` to `"used at the start of an inet/forward base chain"` so the two docstrings right next to each other don't conflict.

src/firewall/firewall_manager.cpp
```diff
@@ -175,8 +181,9 @@ otbrError FirewallManager::EnableNat44Masquerade(const std::string &aUpstreamInt
     SuccessOrExit(error = mNftables.AddRuleMarkMasquerade(kTableName, kNatPostroutingChain, kNat44Mark, nullptr));
 
     // Forward: accept traffic in either direction on the upstream interface.
-    // Hooked at FORWARD/filter alongside forward_ingress; the iifname/oifname
-    // matches scope each rule cleanly.
+    // Hooked at FORWARD/filter alongside forward_ingress. Accepting here does
+    // not exempt a packet from that chain -- every base chain on a hook runs
+    // -- so forward_ingress leaves IPv4 alone by itself.
```
jwhui, line 186: Observation (for a potential follow-up): Since `nat_forward` is a base chain in `table inet otbr` with default policy `accept` and contains only `accept` rules (`oifname <upstream> accept` and `iifname <upstream> accept`), and since `accept` in an `inet otbr` base chain neither exempts traffic from `forward_ingress` nor overrides a `DROP` policy in another table (e.g. `ip filter FORWARD` on the host, per #3515), `nat_forward` is effectively a no-op for all packets.

No need to change that in this PR so the bugfix stays minimal, but it may be worth noting here or cleaning up `nat_forward` in a follow-up.

tests/gtest/test_firewall_manager.cpp
```diff
@@ -242,6 +247,11 @@ TEST(FirewallManagerTest, EnableIngressFilterInstallsChainAndRulesInOrder)
         EXPECT_CALL(mock, AddChain(StrEq(FirewallManager::kTableName), StrEq(FirewallManager::kIngressChain),
                                    Hook::kForward, ChainPriority::kFilter, ChainType::kFilter))
             .Times(1);
+        // First, so that nothing below it -- the unicast drop in particular --
+        // ever sees IPv4: NAT64 replies are forwarded to the Thread interface.
+        EXPECT_CALL(mock, AddRuleNfprotoNeqIp6Return(StrEq(FirewallManager::kTableName),
+                                                     StrEq(FirewallManager::kIngressChain), _))
+            .Times(1);
```
jwhui, line 254: Since `FirewallManagerTest` uses `MockNftables`, the new `Nftables::AddRuleNfprotoNeqIp6Return` implementation in `src/firewall/nftables_impl.cpp` isn't executed by the mock test. Could we also add a quick check in `NftablesBackendTest` (under `#if OTBR_ENABLE_NFTABLES`) that calls `mNftables.AddRuleNfprotoNeqIp6Return(...)` inside (and outside) a batch so the `libnftnl` expression-building path is exercised by unit tests?

Also consider adding `EXPECT_CALL(mock, AddRuleNfprotoNeqIp6Return(_, _, _)).Times(0);` alongside `AddRuleOifnameNeqReturn` in `FirewallManagerTest.InitOnlyCreatesTable` (line 181).

Issue #3602, opened 2026-09-18
nftables: overlapping on-mesh prefixes cause ingress-set updates to fail


JamieMagee:
**Describe the bug**

Overlapping on-mesh prefixes, such as `fd11::/48` without SLAAC alongside `fd11::/64`, cause the native firewall's set update to fail. The allow-set remains stale or empty, blocking ingress to newly advertised addresses.

**To Reproduce**

- Commit: `fd872ab9d4afdfaf034b3b1154fed08b8b6d1e43` (`v2026.09.0`).
- Hardware: none; an isolated Linux network namespace.
- Build configuration: native nftables backend (`OTBR_NFTABLES=ON`).
- Network topology: no live Thread network; the firewall components were exercised directly.

Initialize the firewall and enable ingress filtering. Pass both prefixes in each list to [`ReplaceIngressPrefixes()`](https://github.com/openthread/ot-br-posix/blob/fd872ab9d4afdfaf034b3b1154fed08b8b6d1e43/src/firewall/firewall_manager.cpp#L318-L345).

**Expected behavior**

The update succeeds and both sets cover the advertised address ranges.

**Console/log output**

The isolated call returned `OTBR_ERROR_ERRNO` with `errno = EEXIST (17)`. The legacy `ipset hash:net family inet6` backend accepted both prefixes.

**Additional context**

Overlapping intervals were mentioned in [openthread/ot-br-posix#3325](https://github.com/openthread/ot-br-posix/pull/3325#discussion_r3754228932). The [collector removes exact duplicates](https://github.com/openthread/ot-br-posix/blob/fd872ab9d4afdfaf034b3b1154fed08b8b6d1e43/src/agent/application.cpp#L470-L504), but distinct overlapping prefixes still reach the kernel. Removing contained prefixes separately within each set would preserve coverage and avoid the conflict.

PR #3606, opened 2026-09-19
[firewall] remove the unused single-element ingress set operations

LorbusChris:
`FirewallManager` offers `AddIngressSetElement`, `DelIngressSetElement` and `FlushIngressSet` next to `ReplaceIngressPrefixes`, and nothing calls them: the agent rebuilds both sets from the network data with `ReplaceIngressPrefixes`, and the PF backend never had them.

They are not merely unused, which is why this removes them rather than leaving them be:

- The sets are interval sets, so adding a prefix that overlaps one already in the set fails with `EEXIST`, in either order (the single-element face of #3602).
- They do not compose with `ReplaceIngressPrefixes` once that leaves out prefixes another one covers (#3605): deleting such a prefix fails because it was never added, and deleting the covering one silently removes the covered one's addresses with it.

Doing it properly means `FirewallManager` tracking what each set is meant to hold and re-deriving the kernel set on every change -- new state, for callers that do not exist and with no counterpart in the other backend.

Removed with them: the `IngressSet` enum and `SetName()`, which served only them, and `INftables::DelSetElement`, which had no other caller.

**Tests.** The one that checked per-set routing of the removed calls goes; that is covered by `ReplaceIngressPrefixesFlushesThenRefillsBothSets`. The three that guard properties which still matter -- a failed operation aborts the batch, and the sets need `EnableIngressFilter` and `Init` first -- now drive `ReplaceIngressPrefixes`. `otbr-agent` builds with `OTBR_NFTABLES=ON`.

Independent of #3604 and #3605: the three merge cleanly in either order (36 firewall tests green combined).

Written with AI assistance (`Assisted-by` trailer); I've reviewed and understand all of it and will answer questions myself.


src/firewall/nftables.hpp
```diff
@@ -163,7 +163,6 @@ class INftables
     virtual otbrError AddIp6PrefixSet(const std::string &aTable, const std::string &aSet) = 0;
 
     virtual otbrError AddSetElement(const std::string &aTable, const std::string &aSet, const Ip6Prefix &aPrefix) = 0;
-    virtual otbrError DelSetElement(const std::string &aTable, const std::string &aSet, const Ip6Prefix &aPrefix) = 0;
```
jwhui, line 166: nit: above on lines 102–104, the `INftables` class doc comment still says:
```cpp
 * Rule-creation methods return a kernel-assigned handle through @p aHandleOut
 * so the rule can be deleted later with DelRule(). Set elements are addressed
 * by value, not handle.
```

Now that `DelSetElement` is removed and set elements are never deleted individually, consider updating that second sentence (e.g. "Sets are cleared as a whole with FlushSet(), so AddSetElement() does not return a handle.").

tests/gtest/test_firewall_manager.cpp
```diff
@@ -493,27 +469,25 @@ TEST(FirewallManagerTest, EnableNdProxyKeepsRuleHandleWhenCommitFails)
...
     SetSuccessfulDefaults(mock);
 
     FirewallManager fw(mock, "wpan0");
     ASSERT_EQ(fw.Init(), OTBR_ERROR_NONE);
 
-    // EnableIngressFilter NOT called — set ops must reject.
-    Ip6Prefix prefix("2001:db8::", 64);
-    EXPECT_EQ(fw.AddIngressSetElement(FirewallManager::IngressSet::kDenySrc, prefix), OTBR_ERROR_INVALID_STATE);
+    // EnableIngressFilter NOT called — there are no sets to fill yet.
+    EXPECT_EQ(fw.ReplaceIngressPrefixes({Ip6Prefix("2001:db8::", 64)}, {}), OTBR_ERROR_INVALID_STATE);
```
jwhui, line 481: Consider adding `EXPECT_CALL(mock, BeginBatch()).Times(0);` after `fw.Init()` (matching `EnableNdProxyRejectsEmptyBackboneInterface` and `EnableNat44RejectsEmptyUpstreamInterface`) to verify that `ReplaceIngressPrefixes` fails fast before opening a batch or flushing sets:
```suggestion
    // EnableIngressFilter NOT called — there are no sets to fill yet.
    EXPECT_CALL(mock, BeginBatch()).Times(0);
    EXPECT_EQ(fw.ReplaceIngressPrefixes({Ip6Prefix("2001:db8::", 64)}, {}), OTBR_ERROR_INVALID_STATE);
```

